PR #173, opened 2018-09-24
fix: Allow null value to be sent into Avatar component (#172)

adrian1358:
affects: webiny-ui


adrian1358:
I added requested information, please let me know if anything else is needed.

PR #278, opened 2018-11-05
Editor - added "Image" element and ability to set background image on any element

adrian1358:
You can now add images - background image on any element or images via image element.


adrian1358:
@SvenAlHamad please check for visual issues, and I will merge this PR.

Known issues:
- aligns work in frontend render, but in editor, align is not visible due to `Image` component which is always `width: 100%`. Eg. in the following image, right align is set, but the image is still centered:

image

But in frontend render, it works OK:
image

- when hovering over the image component, you will see an overlays-related glitch - image component will show its overlay and immediately hide it. @adrian1358 (will check this one with @Pavel910  )

adrian1358:
4. Style background image section (especially positioning). Please note that components are disabled when image is not selected.

image

adrian1358:
5. Bonus :)
Not necessarily tied to image component... but I noticed that these controls might be a bit bigger. Please take a look and if you think they could be a bit bigger, please increase the size.

image

adrian1358:
@SvenAlHamad Let me know if you need any help with this one.

PR #518, opened 2019-04-19
Inactive tabs are now hidden instead of completely unmounted.

doitadrian:
## Related Issue
Inactive tabs would unmount, which can cause issues. For example, when validating forms, we want to validate elements on all tabs, not just on the current one.

## Your solution
Instead of unmounting inactive tabs, we hide them with `display: none`. Please check the screenshot for the new functionality.

## How Has This Been Tested?
Manual testing.

## Screenshots (if relevant):
image


doitadrian:
@Pavel910 Please let me know if this would work.

PR #594, opened 2019-11-04
Remove JSON response types for `createFiles` & `uploadFiles` fields

doitadrian:
## Related Issue
`uploadFiles` and `createFiles` had `JSON ` type for the `data` response field. `JSON` is an anti-pattern in the GQL world, and also it would cause the `createFiles` field to break when a `data`  field would be listed as a part of the GQL response.

## Your solution
Replace `JSON` with adequate fully-specified GQL types.

## How Has This Been Tested?
Manual testing - I tested the install process completely, just to make sure nothing breaks there as it is a critical piece of Webiny.

## Screenshots (if relevant):
N/A


doitadrian:
Added comment for the `data` field @Pavel910.

PR #616, opened 2019-11-16
changed form label and title

enmesarru:
Changed Forms label to Form Builder and changed page title, too.


packages/app-forms/src/admin/plugins/routes.js
```diff
@@ -24,7 +24,7 @@ export default [
                 render={() => (
                     <SecureRoute roles={["form-editors-editor"]}>
                         <Helmet>
-                            <title>Edit form</title>
+                            <title>Edit form builder</title>
```
doitadrian: Please remove the "builder" word here, we're not really editing form builder, but form. :)

enmesarru: Got it, i have removed

Issue #657, opened 2020-01-07
Minimum password length is different from defined in Cognito


naokia:
## This is:

- Bug

## Specifications

In any environment

## Expected Behavior

Password shorter than 8 characters should be rejected by validation module.
(because minimum password length on cognito is 8 characters)

## Actual Behavior

Password shorter than 8 characters passes validation on front end.
But, Validation on Cognito reject it.

## Steps to Reproduce the Problem

1. Display initial account setting screen. (titled "Install Security")
2. Input password shorter than 8 characters. and click "Install Security"
3. Validation pass, but an error by Cognito is shown

## Possible Solution
Set minimum length on validation module to 8 characters same as cognito

Pavel910:
Hey @naokia, good find! Our default password policy for `cognito` component looks like this:

```js
const defaultPasswordPolicy = {
    minimumLength: 8,
    requireLowercase: false,
    requireNumbers: false,
    requireSymbols: false,
    requireUppercase: false,
    temporaryPasswordValidityDays: 7
};
```

However, the installation form is not in sync with those settings. I'm leaving this issue opened; we need to figure out a good approach to keeping those settings in sync. @doitadrian This is something to discuss after our upcoming release.

Pavel910:
Just to document a potential solution: the idea is to fetch Cognito Password Policy via SDK when a login form is rendered, so we can render an appropriate validation on password field. I'll try to implement this after we get our next release through the door.

PR #658, opened 2020-01-07
fix: set minimum password length to the same as defined in cognito

naokia:
## Related Issue

https://github.com/webiny/webiny-js/issues/657

## Your solution

Set minimum length on validation module to 8 characters same as cognito

## How Has This Been Tested?

Manual and Unit test

### DIFF
--- a/packages/validation/__tests__/password.test.js
+++ b/packages/validation/__tests__/password.test.js
@@ -8,14 +8,14 @@ describe("password test", () => {
 
     it("should fail - values are too short", () => {
         return Promise.all([
-            validation.validate("12312", "password").should.be.rejected,
+            validation.validate("1231231", "password").should.be.rejected,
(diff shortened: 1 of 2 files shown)


Pavel910:
The referenced issue is not really connected to the `validation` package itself. The validation is passing because of this: https://github.com/webiny/webiny-js/blob/master/packages/app-plugin-security-cognito/src/components/states/SetNewPassword.js#L71

Here, the input validation is not defining the length. Our goal is to somehow sync the values of Cognito parameters with the actual login system.

I'm closing this PR as it is not relevant to the actual problem. Next time please do discuss the intended solution with the maintainers before working on a PR (as stated in the [contribution guide](https://github.com/webiny/webiny-js/blob/master/CONTRIBUTING.md#working-on-an-issue)).

Thank you for your time :)

Issue #683, opened 2020-01-29
Fix links included in the initial Page Builder pages


doitadrian:
Initial Page Builder pages, that come out-of-the box with every Webiny installation, contain links to some of the docs.webiny.com pages.

But, a few links on the `For Developer` and `Get Started` pages point to non-existing pages. So we need to fix this (and we probably need to check all of the links on these initial pages).

And additionally, for some reason, all of the links in these pages are prefixed with the current site's domain, so basically all links are broken. 

The fix for that should probably land somewhere in the `packages/app-page-builder/src/render/plugins/slate/link/index.tsx` file.

Fsalker:
I'd like to take this on ☄

doitadrian:
Be my guest @Fsalker :)

doitadrian:
I will provide a few instructions quickly @Fsalker . Hold on :)

doitadrian:
Added a short document that says a few words about the initial data and files that are utilized in the Page Builder process (all located in a single ZIP file).

https://github.com/webiny/webiny-js/blob/master/packages/api-page-builder/README.md

The links we're talking about in this issue are all located in the initial pages that are created in the Page Builder installation process, and these pages can be found in the mentioned ZIP file. 

So you will have to unzip that, correct the broken links in the `data/pagesData.json`, create a new ZIP and commit it.

Once you're done let me know, and I'll test it out.

And of course if you have more questions, feel free to ask, I'm here.

Thanks!

Fsalker:
Sure, I'm very easygoing with the issue.

If this is critical I may always speed it up 😄. I'm coming with questions / feedback soon.

Fsalker:
Thanks!

doitadrian:
@Fsalker, it's not critical, take as much time as you need!

Thank you! 🍻

PR #688, opened 2020-02-01
fix: All external links now work. Two broken links have been replaced

Fsalker:
## Related Issue
#683 

## Your solution
The external links have been fixed by using `<a>` instead of `<Link>` elements. It seems like Link elements haven't been designed to allow "routing" to external URLs (https://github.com/ReactTraining/react-router/issues/1147), so I've just scrapped them. They still look the same to me now on the site.

I've unzipped `webiny\components\serverless-page-builder\installation\page_builder_installation_files.zip` and replaced 

`https://docs.webiny.com/docs/developer-tutorials/local-setup`
`https://docs.webiny.com/docs/cms-guides/get-started-webiny-administration` 

with  

`https://docs.webiny.com/docs/webiny/introduction/`
`https://docs.webiny.com/docs/get-started/quick-start`.

## How Has This Been Tested?
By running `site` app locally and clicking the new `<a>` elements.

### DIFF
--- a/packages/app-page-builder/src/render/plugins/slate/link/index.tsx
+++ b/packages/app-page-builder/src/render/plugins/slate/link/index.tsx
@@ -18,16 +18,16 @@ export default (): PbRenderSlateEditorPlugin => {
                     const noFollow = data.get("noFollow");
                     const newTab = data.get("newTab");
 
-                    return (
-                        <a
-                            href={href}
-                            {...attributes}
-                            rel={noFollow ? "nofollow" : null}
-                            target={newTab ? "_blank" : "_self"}
-                        >
-                            {children}
-                        </a>
-                    );
+                    const isInternal = href.startsWith("/")
+                    const LinkComponent = isInternal ? Link : "a"
+                    const linkProps = {
+                        ...attributes,
+                        rel: noFollow ? "nofollow" : null,
+                        target: newTab ? "_blank" : "_self",
+                        [isInternal ? "to" : "href"]: href
+                    }
+
+                    return <LinkComponent {...linkProps}>{children}</LinkComponent>
                 }
 
                 return next();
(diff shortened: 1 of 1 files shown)


Pavel910:
The `<Link>` component there is really important. You may have noticed that it is imported from `@webiny/react-router` and not the `react-router`. The reason is, we have some special logic for links that allows data prefetching for internal pages.

So we can't just replace the `Link` with `a`. What we need is a check if the link is external - then render `a`, otherwise, render a `<Link>` for internal URLs.

I kindly ask you to fix that, and if you have any questions, just post them here :)

Thanks for your time 🚀

Fsalker:
Sure, I'll do that. Sounds fair enough.

Fsalker:
@Pavel910 How about this? It looks clean enough to me, although I'm open to suggestions about formatting that ternary.

I've tested it and it works fine locally.

Pavel910:
This is how I would do it:

```jsx
const isInternal = href.startsWith("/");
const LinkComponent = isInternal ? Link : "a";

const linkProps = {
  ...attributes,
  rel: noFollow ? "nofollow" : null,
  target: newTab ? "_blank" : "_self",
  [isInternal ? "to" : "href"]: href
};

return <LinkComponent {...linkProps}>{children}</LinkComponent>;
```

Your check for `http(s)` doesn't cover all the possibilities so it would fail if somebody entered an `ftp` or any other protocol. We know that all internal links will always start with `/`, without the domain name, so it's the safest check.

Fsalker:
I see, dunno if there will ever be any other links than http(s) on the first pages, but your solution is more consistent.

I also like your `LinkComponent` way more than my variant, I'll do that.

Fsalker:
Here, I've written it following your advice. I think it's identical, but it works (tested both cases locally) 😅.

Fsalker:
Btw I did initially try putting the component in a variable, but I messed it up when I wrote `const Component = isExternal ? <a> : <Link>`. Glad I know how to write it now properly heh.

Pavel910:
Well, this `Link` component is not just used on the initial pages, it's a component that will be used for all links created via Page Builder. You never know what users might enter, so it's best to write bullet-proof code or at least cover as many use-cases as possible.

Thanks for the changes, approved 👍

Fsalker:
My pleasure

Issue #676, opened 2020-01-27
Check if installation was completed on site render.


Pavel910:
We need to do a check if the installation of the system was completed when a user visits the site app. Currently it explodes with an ugly error, so we need to improve that.

doitadrian:
Excellent, will fix/release this over the following days.

PR #692, opened 2020-02-05
When opening the site, and PB wasn't installed, show an appropriate error message

doitadrian:
## Related Issue
#676 

## Your solution
If a user visits the site, and the Page Builder installation wasn't completed, we will print an error message, telling the user to open the admin and complete the installation wizard.

### Additional notes
In general, if an error occurs or a page wasn't found, Page Builder's component will show error / not-found pages that were set via the admin area. But in case PB wasn't installed, in which these pages do not exist in the database, the fallback is to show generic error / not-found components (located in `apps/site/src/pageBuilder` folder).

These generic pages were not visually nice, so I took the chance and did a small refactor and made them look more appealing. 

The old version:
Snipaste_2020-02-04_18-48-45

The new version, error page:
Snipaste_2020-02-04_11-53-51

Not found page:
Snipaste_2020-02-04_11-53-16

Note that when deployed to the actual cloud, the `Admin UI` will be a link, that leads to the admin area. We couldn't do that in the localhost environment, because we don't know the domain on which the admin will be hosted (we actually don't know the port).

## How Has This Been Tested?
Manual testing.


Pavel910:
@doitadrian I would like to change the `An error occurred` message completely, to something like: `System not ready` or `Installation incomplete`, or anything else, but I want to avoid the word `error`. Just to make the UI more pleasant.

doitadrian:
@Pavel910 I like it.

doitadrian:
I changed the title:

image

I also noticed `emotion` was used in both admin and site apps, but it wasn't listed as a dependency in respective `package.json` files.

Made the change in the examples folder and in the CLI template.

Issue #678, opened 2020-01-27
Update Apollo cache when a new Form is created.


Pavel910:
Forms created in form builder are not visible in the page builder if the `listForms` query is executed BEFORE the form was created.

Say I opened the page builder, tried adding a Form, but realized I have not yet created a form, I went to Form builder, created a new form, but when I go back to the page editor, the forms list is not refreshed.

doitadrian:
Excellent, will fix/release this over the following days.

PR #694, opened 2020-02-05
Make sure new forms are visible in page builder (without refreshing the page)

doitadrian:
## Related Issue
#678

## Your solution
First, I fixed another bug here - if a user created a new form (which redirects him to the form editor) and then returned back to the forms view, the new form would not be visible in the list of forms. That's why we added the following  at `packages/app-form-builder/src/admin/views/Forms/NewFormDialog.tsx:58`:

```
   refetchQueries: ["FormsListForms"],
   awaitRefetchQueries: true
```

Regarding the initial issue...

I added `fetchPolicy="network-only"` in the `packages/app-form-builder/src/admin/views/Forms/NewFormDialog.tsx:57`, because the just added `refetchQueries: ["FormsListForms"]` in the `packages/app-form-builder/src/admin/views/Forms/NewFormDialog.tsx:57` had no effect, or in other words, when re-opening the page builder and then form element's settings, the list of forms would still be old. I believe this solution would suffice for now.

## How Has This Been Tested?
Added 2 Cypress tests that test:
- basic form creation flow and if the forms list was updated correctly
- if the forms list in the page builder's form element contains the new form after it was created

Note: the 2nd one is work in progress, but will finish it soon.

## Screenshots (if relevant):
N/A

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,14 +1,14 @@
 {
   "viewportWidth": 1600,
   "viewportHeight": 1200,
-  "baseUrl": "https://dx7ymomoh1jp3.cloudfront.net/admin",
+  "baseUrl": "http://localhost:3001",
   "defaultCommandTimeout": 10000,
   "env": {
-    "AWS_COGNITO_USER_POOL_ID": "eu-central-1_42ThoxnJI",
-    "AWS_COGNITO_CLIENT_ID": "2u43k5t0a2d2tpnh6drfeofvq4",
-    "API_URL": "https://diux8va81evaz.cloudfront.net",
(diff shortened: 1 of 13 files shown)


Pavel910:
I'm thinking about this `network-only` part. Could you please try adding an `id` to the query at this reference: `packages/app-form-builder/src/page-builder/admin/plugins/components/FormElementAdvancedSettings.tsx:65`. I think the reason Apollo is not refreshing data is because the query does not contain the cache ID. I think if we add an `id` to the query `{ data { id parent ... } }` it will be able to connect the dots.

Issue #682, opened 2020-01-29
Database connection errors - improve error messaging


doitadrian:
In spite of the fact that the steps for setting up the MongoDB Atlas service are written in the official docs, database connection establishment errors still happen frequently.

In order to help our users resolve this issue more quickly, we can improve the error messaging. 

Once a connection establishment error occurs, let's try to give them a few suggestions on how to fix the issue. A link to the docs might also be helpful.

PR #696, opened 2020-02-05
fix: show a more friendly database connection error message

doitadrian:
## Related Issue
#682 

## Your solution
The following message will be returned from the API, if the database connection wasn't successfully established:
> Could not connect to MongoDB server, make sure the connection string is correct and that the database server allows outside connections. Check https://docs.webiny.com/docs/get-started/quick-start#3-setup-database-connection for more information.

This will make debugging connection-related issues easier for new developers.

## How Has This Been Tested?
Manual testing.

## Screenshots (if relevant):
Snipaste_2020-02-05_14-29-47
Snipaste_2020-02-05_14-20-57


doitadrian:
Let's make that DX as awesome as possible! :)

Issue #702, opened 2020-02-08
Menu Edit crashes if Title not provided


tjrexer:
When adding a Page item to the Menu, if the Title is not provided first, the builder can't set the menu.

## This is:
- Bug

## Specifications

  - version: 3.0.1
  - OS: Ubuntu 18 (WSL)
  - Browser: Firefox

## Expected Behavior
Page is mapped, then user can move to title.

## Actual Behavior
Page is not mapped, error window shows up with the error below:
TypeError: selection is undefined

## Steps to Reproduce the Problem
1. Navigate to edit menu.
2. Add new page menu item.
3. Select Page to map.
4. Error window shows.

## Detailed Description
It appears that the menu items require the Title first, however the Page to select is the first value.  This will cause users to fill out the Page first and end up with the crash.

## Possible Solution
Switch Title to be the first field like with Link.  Won't fix the bug entirely, but will make user experience better.

doitadrian:
Found the error, and submitted a PR, will merge it soon.

doitadrian:
Hi @tjrexer, we just released a set of fixes, so this should work for you.

But before updating, can you tell me which version of CLI you have installed (`webiny --version`)? And maybe also `app-page-builder` (`yarn list @webiny/app-page-builder`)?

Just checking so we can see what would be the best way to upgrade.

If the CLI is on major version 3 or higher, then you can just run `yarn upgrade --scope @webiny`.

Thanks!

tjrexer:
webiny@3.0.2
webiny/app-page-builder@3.1.1

tjrexer:
Ran `yarn upgrade --scope @webiny` from the project root, but it didn't update the app-page-builder, it's still @ version 3.1.1.  It pulled the update for @webiny/api-page-builder to 3.0.3, but not app-page-builder.

doitadrian:
Very weird... I just tried it, had a package `3.1.1`, after running the command, it upgraded it to `3.1.2`.

Can you...
1. try running the command once more?
2. try running it from e.g. the `apps/site` folder?
3. check what package version is set in the `apps/site/package.json`? 

Thank you @tjrexer!

tjrexer:
I had had tried all of those combinations. Semverlevel is set correctly in package.json with ^3.1.1. It may be a mirror replication delay since I'm in the US, so I'll try again tomorrow.

tjrexer:
[...]

PR #703, opened 2020-02-09
fix: menu items form (page + pages list items)

doitadrian:
## Related Issue
#702 

## Your solution
The described error was happening because a second onChange argument in the Autocomplete component was removed. I brought it back.

## How Has This Been Tested?
I will add a Cypress test, later in the day, so let's not merge this until I add it.

## Screenshots (if relevant):
N/A


doitadrian:
While I was testing the menus module, I found a bug in the `searchTags` GQL field. This is fixed.

Also, I added some nice tests for the menus module :)
ezgif-6-54b391b1cbb5

doitadrian:
@Pavel910 please take a look on the newly added changes, and I'll merge this.

PR #704, opened 2020-02-09
fix: change the name of the before/after `http-handler` plugins

doitadrian:
## Related Issue
I changed the type of the before/after handler plugins in the `http-handler` function:
- ~before-handle~ `before-handler`
- ~after-handle~ `after-handler`

## How Has This Been Tested?
Manual testing.

## Screenshots (if relevant):
N/A

### DIFF
--- a/packages/http-handler/src/create.ts
+++ b/packages/http-handler/src/create.ts
@@ -1,8 +1,8 @@
 import { PluginsContainer } from "@webiny/plugins";
 import createResponse from "./createResponse";
 import {
-    HttpAfterHandlePlugin,
-    HttpBeforeHandlePlugin,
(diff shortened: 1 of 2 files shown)


packages/http-handler/src/create.ts
```diff
@@ -32,7 +32,7 @@ export default (...plugins) => async (...args) => {
             }
         }
 
-        const afterHandle = context.plugins.byType<HttpAfterHandlePlugin>("after-handle");
+        const afterHandle = context.plugins.byType<HttpAfterHandlerPlugin>("after-handler");
```
Pavel910: Might as well change the variable names ;)

doitadrian:
Done @Pavel910 (code reviews rule).

Issue #686, opened 2020-01-31
Implement ErrorBoundary in PB Editor


Pavel910:
The related issue is #673. We need to implement [React Error Boundaries](https://reactjs.org/docs/error-boundaries.html) on the `Element` components, so they don't blow up the entire editor. This way content editor will be able to remove the broken element from the content.

The related components that need to implement the error boundaries are:
1) `packages/app-page-builder/src/render/components/Element.tsx`
2) `packages/app-page-builder/src/editor/components/Element.tsx`

Fsalker:
I'll snatch this

Pavel910:
Closed by #726

PR #726, opened 2020-02-24
Handle PB Element plugin errors during render

Fsalker:
## Related Issue
#686 

## Your solution
Wrapped the two `Element.tsx` components inside Error Boundaries.

## How Has This Been Tested?
The `Render` `Element ` was tested by opening up the Pages. Here we clearly see an `Element` component which I've tested by intentionally throwing an error inside the Error Boundary.

image

### DIFF
--- a/packages/app-page-builder/src/editor/components/Element.tsx
+++ b/packages/app-page-builder/src/editor/components/Element.tsx
@@ -36,41 +36,6 @@ export type ElementProps = {
     dragging: boolean;
 };
 
-// Using a class element for a change because `componentDidCatch` only works with class elements ;/
-class ErrorBoundary extends React.Component {
-    constructor(props) {
-        super(props)
-        this.state = {
-            error: null,
-            errorInfo: null
-        }
-    }
-
-    componentDidCatch(error, errorInfo) {
-        this.setState({
-            error,
-            errorInfo
-        })
-    }
-
-    render() {
-        if(this.state.error) {
-            return (
-                <div>
-                    <h3>An unexpected error has occurred</h3>
-                    <details>
-                        <div>{this.state.error.toString()}</div>
-                        <div>{this.state.errorInfo.componentStack}</div>
-                    </details>
-                </div>
-            )
-        }
-
-        return this.props.children
-    }
-}
-
-
 const getElementPlugin = (element): PbEditorPageElementPlugin => {
     if (!element) {
         return null;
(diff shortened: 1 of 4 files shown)


packages/app-page-builder/src/editor/components/Element.tsx
```diff
@@ -36,6 +36,41 @@ export type ElementProps = {
     dragging: boolean;
 };
 
+// Using a class element for a change because `componentDidCatch` only works with class elements ;/
+class ErrorBoundary extends React.Component {
```
Pavel910: I would ask you to separate this into another file. The convention is to have 1 component per file. Just create a file right next to `Element` - `ErrorBoundary.tsx` and import that.

packages/app-page-builder/src/render/components/Element.tsx
```diff
@@ -7,6 +7,40 @@ export type ElementProps = {
     element: PbElement;
 };
 
+// Using a class element for a change because `componentDidCatch` only works with class elements ;/
+class ErrorBoundary extends React.Component {
```
Pavel910: Same as above. Separate file please :)

packages/app-page-builder/src/editor/components/Element.tsx
```diff
@@ -101,29 +136,31 @@ const Element = (props: ElementProps) => {
...
+                        highlight={highlight}
+                        active={active}
+                        style={{ ...defaultStyle, ...transitionStyles[state] }}
+                        className={"webiny-pb-page-element-container"}
+                    >
+                        <div className={["innerWrapper", className].filter(c => c).join(" ")}>
+                            <Draggable target={plugin.target} beginDrag={beginDrag} endDrag={endDrag}>
+                                {renderDraggable}
+                            </Draggable>
+                            {plugin.render({ element })}
```
Pavel910: I think we only want to wrap this line with the ErrorBoundary. If an error occurs, we still want to be able to select and delete the element.

Pavel910:
Can you also attach an image to the PR of what an error looks like when rendered? Thanks!

Fsalker:
Here we go, there's an image now

packages/app-page-builder/src/editor/components/Element.tsx
```diff
@@ -100,6 +100,19 @@ const Element = (props: ElementProps) => {
         return null;
     }
 
+    let renderedPlugin = null
+    try {
+        renderedPlugin = plugin.render({ element })
+    } catch(err) {
+        renderedPlugin = (
+            <div>
+                <div>{err.message}</div>
```
Pavel910: Could you spare some time and style this error a little bit better? See how CRA styles their error overlay. Something along those lines would be great. Just to have a clear error message, and maybe skip the stack output, since you're console.logging it, and simply output a message like `To see the error stack, open dev tools ...`. It will make the output less messy in a page full of elements.

packages/app-page-builder/src/editor/components/Element.tsx
```diff
@@ -100,6 +100,19 @@ const Element = (props: ElementProps) => {
         return null;
     }
 
+    let renderedPlugin = null
```
Pavel910: I'd still ask you to extract this into a separate function  at the top of the file, the way `getElementPlugin` on line `39` is done.

packages/app-page-builder/src/editor/components/Element.tsx
```diff
@@ -115,7 +128,7 @@ const Element = (props: ElementProps) => {
                         <Draggable target={plugin.target} beginDrag={beginDrag} endDrag={endDrag}>
                             {renderDraggable}
                         </Draggable>
-                        {plugin.render({ element })}
+                        {renderedPlugin}
```
Pavel910: Then here you can simply call that function:
```js
{renderPlugin({ plugin, element })}
```

It's easier to read code and navigate around function calls this way.

packages/app-page-builder/src/render/components/Element.tsx
```diff
@@ -27,7 +27,20 @@ const Element = (props: ElementProps) => {
         return null;
     }
 
-    return <>{plugin.render({ theme, element })}</>;
+    let renderedPlugin = null
```
Pavel910: Since this is an exact copy, maybe we could extract this whole tool into a separate utility. I'm thinking maybe `app-page-builder/src/utils/renderElementPlugin.tsx`? Then import and use it in both editor and render Element. Let's do that and see how it looks.

packages/app-page-builder/src/editor/components/Element.tsx
Fsalker: yep

packages/app-page-builder/src/render/components/Element.tsx
Fsalker: Here, I've done so

packages/app-page-builder/src/editor/components/Element.tsx
Fsalker: Fingers crossed that this will do 💈

Fsalker:
Also, I'll remove the commented lines currently present in the code if this is ok. Those lines help me test it out 😄

Pavel910:
@Fsalker this look ok, we can tweak minor things later on. One last thing: replace the `An error has occured` text with a simple `Error: ${error.message}`. That long text doesn't bring any meaning but takes up a lot of space. And remove the test errors of course ;)

packages/app-page-builder/src/editor/plugins/elements/row/Row.tsx
```diff
@@ -3,7 +3,6 @@ import ConnectedElement from "@webiny/app-page-builder/editor/components/Connect
 import { ElementRoot } from "@webiny/app-page-builder/render/components/ElementRoot";
 import RowContainer from "./RowContainer";
 import ElementAnimation from "@webiny/app-page-builder/render/components/ElementAnimation";
-
```
Fsalker: dunno what this is tbh

Fsalker: oh btw here are the -s and +s you despise @Pavel910 🤣

Fsalker:
alright, here we go. everything should be k now

Issue #727, opened 2020-02-24
Page Builder - Invalidate SSR cache when user changes sensitive settings


doitadrian:
## This is:

- Bug

## Expected Behavior
When the user changes some of the page builder settings that are located behind these two menu items, we should refresh the SSR cache. 

![image](https://user-images.githubusercontent.com/5121148/75179067-92733e80-5739-11ea-9297-13f86d44ff6e.png)

The SSR caching mechanism is unfortunately not documented yet, so ping me for details if interested, or check out [this article](https://blog.webiny.com/serverless-side-rendering-e1c0924b8da1) (scroll to the "SSR HTML Caching" section).

## Actual Behavior
When a user changes some of these settings, nothing happens, and the user is stuck with the previous version of pages until some other event in the system invalidates SSR HTML.

## Possible Solution
To store Page Builder settings data (or any other data in the system), the `PbSettings` Commodo model is used (`packages/api-page-builder/src/plugins/models/pbSettings.model.ts`).

You need to use `withHooks` higher-order function, to add an `afterSave` hook in which you will trigger cache invalidation. You will accomplish this using the SSR Cache client.

There are already a couple of events that trigger cache invalidation, and they can all be found here: ```packages/api-page-builder/src/plugins/useSsrCacheTags.ts```. You should add your logic in this file too.

## Notes
This task will enable you to get a feeling of how the API and the system in general works. If you find the time, walk through the code a bit. You will notice that we have our own ORM/ODM, which might be different from what you might've seen already in the JS world. Check out the just-written 
 [docs](https://docs.webiny.com/docs/api-development/commodo/introduction) if you want to find out more about it.

You can check out the article I linked above too. No need to read the whole piece if you're already familiar with the concept of SPA and SEO problems they present, but the SSR HTML Caching section would be useful. 

Finally, ping if you get stuck :)

PR #728, opened 2020-02-27
WIP - Cache invalidation when updating PB settings

Fsalker:
## Related Issue
#727 

## Your solution
I've written an `afterSave` hook in `withHook` (as advised: thanks, Adrian!). It doesn't correctly get the plugin at the moment, but perhaps this is a good stab at solving the issue. I'll need a few opinions here.

## How Has This Been Tested?
It will be tested by deploying the app, changing a few settings and seeing if the cache is cleared (ie: changing the site's name) since cache clearing can't be tested locally yet.

### DIFF
--- a/packages/api-page-builder/src/plugins/models/pbSettings.model.ts
+++ b/packages/api-page-builder/src/plugins/models/pbSettings.model.ts
@@ -3,7 +3,6 @@ import {
     withStaticProps,
     withProps,
     withName,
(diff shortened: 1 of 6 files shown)


Fsalker:
Here's the most recent and functional implementation.

I've also refactored a little code.

doitadrian:
Closing this in favor of #746.

Issue #730, opened 2020-02-28
CLI - Scaffolding


doitadrian:
## This is:
- Feature request

## Detailed Description
When it comes to creating initial files for a new GraphQL service or a new React app, currently, our users still have to do a significant amount of manual repetitive work.

That's why we want to add the "scaffolding" function to the CLI. Basically, if a user wants to start creating a new GraphQL service, he would use the CLI to generate all of the initial files. 

This will enable the user to immediately jump to the code, which not only will make the DX better, but will also reduce the amount of docs we'll need to write.

## Possible Solution
The command might look something like the following:

```webiny scaffold```

After that, the CLI will internally search for all `@webiny/cli-scaffold-*` and `webiny-cli-scaffold-*` packages, and output the list with the [Inquirer](https://github.com/SBoudrias/Inquirer.js/).

Once a user selects a scaffold template, the CLI will select and execute the template's "list-options" code. In other words, the CLI will start asking questions that were defined in the selected template (e.g. a folder in which the scaffolding should be executed). Once all of the options have been chosen, the scaffolding starts and the user should end up with all of the necessary files in place.

I suggest that in these scaffold templates, we just use named exports for every part of the scaffolding process. For example:

```js
// packages/cli-scaffold-graphql-service/index.js

module.exports.name = "GraphQL Service";
module.exports.description = "GraphQL Service";

// Generates the interactive options selection . Not sure if this just 
// returns the params for the inquirer, or the inquirer can be directly 
// used here. I believe the first option is better.
module.exports.options = () => { ... }

// Generates (or copy&pastes, depends on the template) all of the necessary files, 
// based on the selected options.
module.exports.scaffold = ({ options }) => { ... } // or module.exports.generate = () => {}
```

Either that, or we can just export a simple object with all of these properties. We can even export the plugin here or array of plugins, so in the future, we could maybe add additional customizations here. If you ask me, I would actually do something like that. For example:

```js
// packages/cli-scaffold-graphql-service/index.js

module.exports = [
	{
		type: "scaffold-template",
		name: "scaffold-template-apollo-graphql-service",
		scaffold: {
			name: "...",
			description: "...",
			options: () => {}
			generate: () => {}
		}
	},
	{ ... } // If needed, more plugins could be exported, but I don't see an example right now.
]
```

The actual CLI code is located in the `packages/cli` folder.

doitadrian:
[...]

PR #733, opened 2020-03-01
CLI Scaffolding

Fsalker:
## Related Issue
#730 and #741 

## Your solution
I've created two new packages `cli-scaffold-graphql-service` and `cli-scaffold-model` which can be called through `cli` with `webiny scaffold <name>`.

## How Has This Been Tested?
**GraphQL service** template - by updating `serverless.yml` correctly and validating the generated code in `./examples/api`
**Model** template - WIP


Pavel910:
@Fsalker I'm wondering what happened to the code formatting 🤔 Are you using `prettier`? Please do configure `prettier` so when you push the code it adheres to our project standard.

And I have a pro tip: when you want to find all packages on local system, use this command:

```
yarn list --pattern="@babel/preset*|babel-preset-*" --depth=0 --json

// .... stripped output
{"type":"activityTick","data":{"id":0,"name":"secure-site@0.1.0"}}
{"type":"activityTick","data":{"id":0,"name":"site@0.1.0"}}
{"type":"activityEnd","data":{"id":0}}
{"type":"tree","data":{"type":"list","trees":[{"name":"@babel/preset-env@7.8.3","children":[],"hint":null,"color":null,"depth":0},{"name":"@babel/preset-react@7.8.3","children":[],"hint":null,"color":null,"depth":0},{"name":"@babel/preset-typescript@7.8.3","children":[],"hint":null,"color":null,"depth":0},{"name":"babel-preset-jest@24.9.0","children":[],"hint":null,"color":null,"depth":0},{"name":"@babel/preset-flow@7.8.3","children":[],"hint":null,"color":null,"depth":0},{"name":"babel-preset-react-app@9.1.0","children":[],"hint":null,"color":null,"depth":0},{"name":"babel-preset-minify@0.5.1","children":[],"hint":null,"color":null,"depth":0}]}}```

The last line of JSON will be the actual result of the command, which you can then parse using `JSON.parse` and get a nice clean object to work with.

**Comment by Fsalker** [comment]
Ah I've just noticed the msg's now.

**Code formatting**: it should be fine now. I've fixed my Webstorm File Watcher 😄
**Finding packages (pro tip)**: a tad too late for me to notice this now. I've just gone vanilla with `fs.readdirSync` and figured out all the correct packages. This works just fine, as `webiny scaffold <name>` will be always called from within ./webiny/<app_name>/ (ie: ./webiny/examples)

This new commit fully implements the GraphQL service scaffolding. If you approve this aswell I'll update the remaining files in `templateFiles` with the code to be generated (atm only `templateFiles/index.ts` is fine).

File: packages/cli/cli.js [hunk 1]
```diff
@@ -144,12 +147,55 @@ yargs.command(
...
+            moduleName => `@webiny/${moduleName}`
+        );
+        const nodeModulesNames = await fs.readdir("../node_modules");
+        const scaffoldModulesNames = webinyModulesNames
+            .concat(nodeModulesNames)
+            .filter(moduleName => moduleName.match(regexIsCliModule));
+        const scaffoldModules = scaffoldModulesNames.reduce(
+            (modules, crtModuleName) => modules.concat(require(crtModuleName)),
+            []
+        );
```
Pavel910: @Fsalker With `fs.readdir("../node_modules/@webiny")` you are limiting the modules to only those located in `{cwd}/../node_modules` which is not very robust. So your code is very dependent on the CWD. Also that's a lot of code for something that can be achieved with a single line and cover all node_modules relevant to your project.

Pavel910:
@Fsalker `webiny scaffold` doesn't have to be executed in the root of the project, as not every scaffolding template has to edit the `serverless.yaml` file. You can scaffold app plugins, api plugins, and include them in your React app manuallu; there are plenty of options, so we need to make the tool as robust as possible.

packages/cli/cli.js
```diff
@@ -144,12 +147,55 @@ yargs.command(
...
+            name: "scaffoldTemplate",
+            message: "Pick the template you'd like to employ",
+            choices
+        });
+        const = scaffoldModules.find(
+            module => moduleToChoice(module) === scaffoldTemplate
+        );
+        selectedScaffoldModule.scaffold.generate({
+            serviceName: argv.name
+        });
```
Pavel910: Once you have the template, you should get options from it and run them through `inquirer`, so each templte can have its own wizard. `argv.name` is not really necessary as we don't know what the user wants to create until he selects a template. Let's do that per-template for now; basically, let every template handle its own wizard.

doitadrian: As mentioned, `scaffold` will be flexible and allow users to write their own plugins (via `webiny-cli-scaffold-*` packages), which means they might write a scaffold that's not related to services at all.

I agree with @Pavel910, I believe `webiny scaffold` will suffice, then users select what they want via the inquirer.

packages/cli/cli.js
doitadrian: Very good point @Pavel910 🚀

doitadrian: I would still use the `PluginsContainer` we've mentioned in our conversations here.

Have in mind when we require the modules on line 175, that these modules can export an array of plugins, or a single plugin.

Using `new PluginsContainer(...)` will enable you to "not care" about that fact, and also will enable you later down the road to query registered plugins by type via `pluginsContainer.byType(...)` method, or by name `pluginsContainer.byName(...)`.

Fsalker:
all good? @Pavel910

packages/plugins/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@webiny/plugins",
   "version": "3.0.1",
-  "main": "dist/index.js",
+  "main": "index.js",
```
Pavel910: For future testing, it's easier to simply import the packages with the `dist` path included, like: 
```import { PluginsContainer } from "@webiny/plugins/dist"```.
This is only necessary when working on the `cli` package.

File: packages/cli-scaffold-graphql-service/index.ts [hunk 4]
```diff
@@ -0,0 +1,61 @@
...
+                    if (serverlessJson[serviceName])
+                        throw new Error(
+                            "This service already exists in serverless.yml! Please pick a different name for it."
+                        );
+
+                    serverlessJson[serviceName] = {
+                        component: "@webiny/serverless-apollo-service",
+                        inputs: {
+                            description: `Apollo service '${serviceName}' scaffoled by the CLI.`,
+                            region: "${vars.region}",
```
doitadrian: What if `vars.x`, in this case, `vars.region` does not exist?

When the template is deployed, errors will be thrown.

I guess the scaffold could try to use these variables, but if it determines they don't exist, then we could use sensible defaults?

For example: 

- region: `us-east-1`
- debug: `false`
- vars.apollo: 
```
    server:
      introspection: true
      playground: true
```

--- 

The second idea I had is - ask users via inquirer. I think that might be also a good idea, and especially if we prefill these values in the answers, so users can just hit enter.

We could allow users to type in ${vars.region}, and we could detect if he's referencing a variable here. If so, check if it exists immediately in the inquirer, and show him an error message if referencing was invalid. I would use these ${vars.xyz} values as defaults for every question.

If you ask me, I would go with the second solution.

@Pavel910 @Fsalker

Fsalker: `vars.region` and `vars.debug` do already exist in serverless.yml though, don't they?

That's where I got them from 😄

Most other components in there (gateway, security, i18n, ...) use it.

doitadrian: Yeah, I understand, but I was thinking it would be a good idea to try to handle this case / remove the possibility of potential errors.

Two more ideas I had in mind...

3) If the vars don't exist in the `serverless.yml`, maybe the scaffold template could print a warning once the scaffolding is done. This way user will know about the missing variables.

4) If vars are not existing, the scaffolding template could just add them at the top of the serverless.yml file, so the user doesn't have to do it by himself. Then, the user can edit these variables. And of course, we can print a message at the end of the scaffolding - telling the user he should edit these to his needs.

All in all, these vars will be present because of other services. But still, we will probably, at one point in time, have to maybe add some other variable, that's not already listed in the sls.yml, and then we'll have to proceed with one of the four solutions (or fifth if you have one in mind).

Now we just need to choose the right solution.

doitadrian: Per our conversations, let's go with the solution 4 - add missing variables directly to the serverless.yml, and warn the user about the addition. Also, put sensible defaults where possible.

Fsalker:
Here's what it looks like at the moment:
image

doitadrian:
> Here's what it looks like at the moment:
> image

My thoughts:

1. `Pick the template you'd like to employ` -> `Choose a template:`

2. `GraphQL Apollo Service - Creates an /api ...` -> I know that we have a description in the plugin, but looking at this, it kinda feels crowded, so I was thinking maybe we could just print `GraphQL Apollo Service` and that's it. Keep it simple. Correct me if I'm wrong of course.

3. I would do `The name of your service` -> `Name:` or -> `Service name:`.

4. Please use more realistic examples when discussing / providing screenshots. It will give us a better feel for the thing we're building.

5. Before warning, I would definitiely show a success message, like "Scaffolding complete!". Play around and see what looks good.

6. Warning message `Variable "swagz" does not exist...` looks good to me. I am just thinking if there are better solutions for the `[Warning]` prefix. Maybe we could try to use [Info] or even emoji ℹ️(not sure how looks on each platform). [Warning] kinda gives me a feeling of an error. Remember when you run something like `yarn`, and while the packages are being installed, all kind of warnings pop up and you're scratching your head and asking yourself "will this work?". So because of that, I would avoid the word "warning".

doitadrian:
We'll need @Pavel910's assistance for this.

```
i18n:
  component: "@webiny/serverless-apollo-service"
  inputs:
    region: ${vars.region}
    memory: 512
    debug: ${vars.debug}
    webpackConfig: ./webpack.config.js
    plugins:
      - factory: "@webiny/api-plugin-create-apollo-handler"
        options: ${vars.apollo}
      - factory: "@webiny/api-plugin-commodo-db-proxy"
        options:
          functionArn: ${dbProxy.arn}
      - factory: "@webiny/api-security/plugins/service"
        options: ${vars.security}
      - "@webiny/api-i18n/plugins"
```


This is a simple `i18n` service, which uses `"@webiny/serverless-apollo-service"` component.

What should be some of the values here when a user creates a new service?

The following is what I'm not sure what to do with:
1. `webpackConfig: ./webpack.config.js` - should we use this as a default?
2. `- "@webiny/api-i18n/plugins"` - OK, this works for `@webiny/...`, but how will the user reference his plugins? Is the relative path possible here? Or should we put the package name here, since, with yarn monorepos, everyhing is symlinked in `node_modules`?

doitadrian:
Another question that's going through my mind is the following...

Basically, every Apollo Service that you create will need this:

```
      - factory: "@webiny/api-security/plugins/service"
        options: ${vars.security}
```

So maybe, we could ask the user which additional plugins he wants to immediately have in his new Apollo Service?

We could add this functionality again via new plugins, which will be read by the "Apollo GraphQL" scaffolding template, and will output a list of options to the user.

@Pavel910 @Fsalker

Fsalker:
It is possible to use `inquirer` to allow the users to select a list of plugins. Of course, as this grows larger so will the amount of text outputted by us. 

At some point, wouldn't there be dozens of plugins to choose from? This basically means we'll have to create a 'textbook' of plugins to choose from (ie: list 10 plugins per page). And who really goes through that kind of thing anyway? I'd rather write them myself in serverless.yml or maybe create a different CLI utility altogether where I may insert the name of the plugin(s) I'd like to add.

Within `webiny scaffold` it could also be possible to ask the user for the plugins' names which they would like to use and we'd quickly search for it and come back with a `Yes, good! It will be added.` or `Sorry, that plugin doesn't exist` message. But there's 1 issue that instantly pops in my mind here - imagine going through a 5-10-15-step process where you configure your Scaffolding template just to have an unexpected error thrown (for whatever reason); heck, maybe the user just closes their terminal or mispresses CTRL+C. So I'd rather avoid such a long process and thus separate scaffolding & plugin selection (if we were to provide it). What do you think?

Maybe we should just let them insert their own plugins though directly in serverless.yml. If the users know of their plugins' existence, they will be able to configure them, right? And putting a few additional variables won't be a big deal I guess...

doitadrian:
I was also worried about the possibility that the user will have to make a lot of choices via the CLI.

I am open for the more manual approach too, but just wanted to add 2 comments:

1. I don't believe we'll ever reach a level in which we'll need pagination. Right now we have `i18n` service plugins and `security` service plugins:

```
- "@webiny/api-i18n/plugins/service"
- factory: "@webiny/api-security/plugins/service"
            options: ${vars.security}
```

2. I think that by asking the user if he needs security/i18n and adding these for them immediately, we would reduce unnecessary DX friction. The user won't have to open the docs and do everyhing manually. 

3. Even more importantly, by offering him this option via the scaffolding, he will instantly be **aware** that he needs security, and it's not built-in. Otherwise, he might deploy services without it, thinking that it's already enabled, or even worse, roll out his own solution, which is completely unnecessary.

Fsalker:
(the newest version of the CLI interface)
image

doitadrian:
Have you tested this when no variables are provided in the yml? Try running the command, without any vars that this template might rely on. How will the screen look like?

Fsalker:
It works well. Before:
image

After:
image

Also don't mind `factory` in `My Service 1` generated in the second picture. I've fixed that now, here's how the services are generated:

```
My Service 2:
  component: '@webiny/serverless-apollo-service'
  inputs:
    description: Apollo service 'My Service 2' scaffoled by the CLI.
    region: '${vars.region}'
    memory: 512
    debug: '${vars.debug}'
    plugins:
      - factory: '@webiny/api-security/plugins'
        options: '${vars.security}'
```

Issue #748, opened 2020-03-10
I18N Service Plugins - Improve Error Messaging


doitadrian:
## This is:

- Feature request

## Detailed Description
Currently, if a user forgets to pass the `I18N_API_URL` into the serverless component (`serverless.yml`), the whole service would not work at all. 

![image](https://user-images.githubusercontent.com/5121148/76301239-3b837280-62be-11ea-8b91-558fe73fa40c.png)

And if you open the CloudWatch log, you can see the following error:
![image](https://user-images.githubusercontent.com/5121148/76301183-24448500-62be-11ea-9516-f802a3d05219.png)

## Possible Solution
This error probably won't mean a lot to the user, so let's make it a bit more friendly. 

Let's check the passed env variable here: `packages/api-i18n/src/plugins/service.ts:32`. 

If it doesn't exist, or it's not a valid URL, we should throw an error message like 
> `I18N_API_URL env variable is not a valid URL. Make sure it's specified in the service configuration.`. 

This is just a quick example. Of course, you are allowed to use your imagination / be creative 😃

Arvisix:
May I start working on this issue?

doitadrian:
Sure, go for it! 

If you need any assistance, just let us know here, or via [gitter](https://gitter.im/Webiny/webiny-js).

PR #750, opened 2020-03-12
I18N Service Plugins - Improved error messaging 

Arvisix:
## Related Issue
https://github.com/webiny/webiny-js/issues/748

## Your solution
I added validation for valid url inside `packages/api-i18n/src/plugins/service.ts`

## How Has This Been Tested?
I don't know how this could be tested

### DIFF
--- a/packages/api-i18n/src/plugins/service.ts
+++ b/packages/api-i18n/src/plugins/service.ts
@@ -31,15 +31,13 @@ export default () => [
             }
 
             if (!process.env.I18N_API_URL) {
-                throw new Error(
-                    `I18N_API_URL env variable is missing. Make sure it's specified in the service 'serverless.yml' configuration.`
-                );
+                throw new Error('I18N_API_URL environment variable is missing. Please check the service configuration, located in the "api/serverless.yml" file.');
             }
 
-            if (validation.validate(process.env.I18N_API_URL, 'url', { throw: false }) instanceof ValidationError) {
-                throw new Error(
-                    `Invalid URL specified for I18N_API_URL env variable.`
-                )
+            try {
+                validation.validateSync(process.env.I18N_API_URL, "url");
+            } catch (e) {
+                throw new Error('The value specified for the I18N_API_URL env variable is not a valid URL. Please check the service configuration, located in the "api/serverless.yml" file.')
             }
 
             const client = new GraphQLClient(process.env.I18N_API_URL);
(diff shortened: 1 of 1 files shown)


packages/api-i18n/src/plugins/service.ts
```diff
@@ -29,6 +29,14 @@ export default () => [
                 return localesCache;
             }
 
+            const urlRegEx = /^(https?:\/\/)((([a-z\d]([a-z\d-]*[a-z\d])*)\.)+[a-z]{2,}|((\d{1,3}\.){3}\d{1,3}))(\:\d+)?(\/[-a-z\d%_.~+]*)*(\?[;&a-z\d%_.~+=-]*)?(\#[-a-z\d_]*)?$/i;
+
+            if (!urlRegEx.test(process.env.I18N_API_URL)) {
```
doitadrian: Please use our own [validation library](https://github.com/webiny/webiny-js/tree/master/packages/validation) for checking URLs.

packages/api-i18n/src/plugins/service.ts
```diff
@@ -29,6 +29,14 @@ export default () => [
...
+                throw Error(
+                    `I18N_API_URL env variable is not a valid URL. Make sure it's specified in the service configuration.`
```
doitadrian: I would just mention the exact file we are talking about here and that is the `serverless.yml`. This way users will know what exactly we meant, "service configuration" to some users might be confusing.

Arvisix:
@doitadrian please check updated PR

packages/api-i18n/src/plugins/service.ts
```diff
@@ -29,6 +30,18 @@ export default () => [
                 return localesCache;
             }
 
+            if (!process.env.I18N_API_URL) {
+                throw new Error(
+                    `I18N_API_URL env variable is missing. Make sure it's specified in the service 'serverless.yml' configuration.`
```
doitadrian:
```js
throw new Error("I18N_API_URL environment variable is missing. Please check the service configuration, located in the "api/serverless.yml" file.`);
```

doitadrian: I changed the message a bit, just to kinda sync it with the lower one.

packages/api-i18n/src/plugins/service.ts
```diff
@@ -29,6 +30,18 @@ export default () => [
...
+                );
+            }
+
+            if (validation.validate(process.env.I18N_API_URL, 'url', { throw: false }) instanceof ValidationError) {
```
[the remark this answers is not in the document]
Arvisix: Oh, sorry, I though I changed it to `validateSync`

PR #752, opened 2020-03-13
Updated React URL links

visshaljagtap:
Updated React URL links


Fsalker:
Hey visshaljagtap, I think this is a cool update. Even though the improvement is marginal, `reactjs.org` is the actual domain we're landing on and the one we should hit when clicking those documentation anchors.

Let's see what the other guys say about it. 😄

visshaljagtap:
Yeah but due to this we can avoid unnecessary redirection.

visshaljagtap:
Thats Nice! @Pavel910 Surely I will have a look on same!

Issue #737, opened 2020-03-03
Clicking outside of the text of a menu button doesn't link to another page


Fsalker:
When creating a `Navigation` component (`packages/app-page-builder-theme/src/components/Navigation.tsx`), it is only possible to access the linked url by clicking on the button's text itself (the **blue** area in the screenshot).

Even though the `li` element intuitively shows us we're focusing the button, clicking it sometimes won't do anything. This happens when we click on the `li`'s padding rather than its anchor's content (the **green** area in the screenshot).

![image](https://user-images.githubusercontent.com/16700631/75762815-f7cfbc80-5d43-11ea-9950-4f2ed41f6e2a.png)

## This is:
- Bug

## Steps to Reproduce the Problem
Run the `site` app (ie: from `./examples/apps/site`) and check the menu.

SvenAlHamad:
We should remove the padding from the list item `li`, and move it to the anchor element...this will fix it.

doitadrian:
Had to revert changes made in the PR, since it introduced another bug with displaying the menu items:
![image](https://user-images.githubusercontent.com/5121148/80815021-37dcdf80-8bcd-11ea-8899-c9e70396b310.png)

TommyJackson85:
I will work on this issue if it is still relevant (after the recent changes to the project I mean).

Pavel910:
@Ashu96 @doitadrian please confirm this is still an issue, I'm totally out of the loop on this one.

doitadrian:
Please reopen this issue once you add more information and updates here. If this is not the case and you need some help, feel free to seek help from our [Slack Community](https://www.webiny.com/slack) or ping one of the reviewers. Thank you for your contributions!

PR #755, opened 2020-03-13
fix: Navigation link padding

Arvisix:
## Related Issue
https://github.com/webiny/webiny-js/issues/737

## Your solution
I wrapped `<li>` item in `<Link>` component. Also refactor `getLink` method

## How Has This Been Tested?
This was tested by clicking outside  of text on navigation item

## Screenshots (if relevant):

### DIFF
--- a/packages/app-page-builder-theme/src/components/Navigation.tsx
+++ b/packages/app-page-builder-theme/src/components/Navigation.tsx
@@ -1,14 +1,6 @@
 import React from "react";
 import { Link } from "@webiny/react-router";
 
-function getLink({ url, title }) {
-    if (url.startsWith("/")) {
-        return <Link to={url}>{title}</Link>;
-    }
-
-    return <a href={url}>{title}</a>;
-}
-
 const Navigation = ({ data }) => {
     if (!data) {
         return null;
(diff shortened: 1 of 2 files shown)


packages/app-page-builder-theme/src/components/Navigation.tsx
```diff
@@ -23,14 +15,14 @@ const Navigation = ({ data }) => {
                             {item.title}
                             <ul>
                                 {item.children.map((item, index) => {
-                                    return <li key={item.id + index}>{getLink(item)}</li>;
+                                    return <Link to={item.url} key={item.id + index}><li>{item.title}</li></Link>;
```
Pavel910: I think this will produce invalid HTML, since you'd get `<ul><a><li>...</li></a></ul>`. Only `li` is allowed inside of `ul`, so we need a different approach here. Maybe create an class using emotion to set `li` padding to 0 (or something along those lines)?

packages/app-page-builder-theme/src/components/Navigation.tsx
```diff
@@ -1,14 +1,6 @@
...
 import { Link } from "@webiny/react-router";
 
-function getLink({ url, title }) {
-    if (url.startsWith("/")) {
-        return <Link to={url}>{title}</Link>;
-    }
-
-    return <a href={url}>{title}</a>;
-}
-
```
Pavel910: Good call! This is totally not needed since we handle this inside of the Link component 👍

PR #775, opened 2020-03-28
fix: 🐛  fixed storybook disabled input and select was hard coded

### LINKED ISSUE #759: Cannot disable form inputs
## This is:

- Bug

## Detailed Description
Passing a `disabled={true}` prop on components like `Input` and `Select` does not have any effect. For example:

https://storybook.webiny.com/?path=/story/components-input--usage
https://storybook.webiny.com/?path=/story/components-select--usage

It works in the RMWC library - https://rmwc.io/text-fields.

Not sure where the bug is but if I hardcoded the `disabled={true}` in `packages/ui/src/Input/Input.tsx:112`, all inputs would become disabled as expected:

image

So, it seems that the flag is not correctly passed to the actual RMWC `TextArea` component.

ahmad-reza619:
## Related Issue

close #759 
## Your solution

Disabled props was hard coded as mentioned in #759 and so i followed the directions by @doitadrian 
## How Has This Been Tested?

not tested yet actually
## Screenshots (if relevant):

### DIFF
--- a/packages/form/src/Bind.ts
+++ b/packages/form/src/Bind.ts
@@ -74,8 +74,12 @@ const createBind = (form: Form) => {
 
         form.inputs[name].props = newProps;
 
+        
         if (React.isValidElement(children)) {
-            return React.cloneElement(children, newProps);
+            if (!newProps.disabled) {
+                newProps.disabled = children.props.disabled;
+            }
+            return React.cloneElement(children, { ...children.props, ...newProps });
         }
 
         return children(newProps);
--- a/packages/form/src/Bind.ts
+++ b/packages/form/src/Bind.ts
@@ -48,6 +48,7 @@ const createBind = (form: Form) => {
 
         // Build new input props
         const newProps = {
+            disabled: false,
             form,
             validate: form.getValidateFn(name),
             validation: form.state.validation[name] || {
@@ -74,7 +75,7 @@ const createBind = (form: Form) => {
         form.inputs[name].props = newProps;
 
         if (React.isValidElement(children)) {
-            return React.cloneElement(children, { ...children.props ,...newProps });
+            return React.cloneElement(children, newProps);
         }
 
         return children(newProps);
--- a/packages/form/src/Bind.ts
+++ b/packages/form/src/Bind.ts
@@ -48,7 +48,6 @@ const createBind = (form: Form) => {
 
         // Build new input props
         const newProps = {
-            disabled: false,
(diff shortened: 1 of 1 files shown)


Pavel910:
Could you test it in storybook locally and confirm that it's actually fixed? Thanks.

ahmad-reza619:
Ok sorry, it was late night and I was sleepy, very sorry

packages/form/src/Bind.ts
```diff
@@ -48,7 +48,6 @@ const createBind = (form: Form) => {
 
         // Build new input props
         const newProps = {
-            disabled: false,
```
doitadrian: This will cause TS error on line 70:
image

I would maybe undo this change,  and add the following piece on line 76:
```js
if (!newProps.disabled) {
   newProps.disabled = children.props.disabled;
}
```

Basically, if the `disabled={true}` prop wasn't present on the whole `Form` component, then we should just take the `disabled` value from the actual child component, which the `Bind` component is wrapping.

ahmad-reza619: yes, those error also happened in my local machine. ok, so am i just have to revert the change?

doitadrian: Yeah, revert it, add the `if` statement mentioned above, and we should be good to go I believe.

ahmad-reza619: unfortunately, i have tried that and it says typescript error, because there's no props property on children 🤔 

Screen Shot 2020-03-29 at 19 44 12

doitadrian: This needs to go inside of the `if (React.isValidElement(children)) { ...` statement.

ahmad-reza619: ok thanks for pointing out @doitadrian

ahmad-reza619:
Ok, i have tested this on my local machine, node v13.7.0 and it's working

packages/form/src/Bind.ts
```diff
@@ -74,7 +74,11 @@ const createBind = (form: Form) => {
 
         form.inputs[name].props = newProps;
 
+        
         if (React.isValidElement(children)) {
+            if (!newProps.disabled) {
+                newProps.disabled = children.props.disabled;
+            }
             return React.cloneElement(children, newProps);
```
doitadrian: We still need to have `return React.cloneElement(children, { ...children.props, ...newProps });` here.

Issue #783, opened 2020-04-06
Support for DynamoDB - data restructuring - Page Builder - querying latest pages


doitadrian:
Continuing from #781, this issue gives more details about solving the "Page Builder - querying latest pages" issue.

## From the introduction issue
Pages can have multiple revisions, where each revision has the `parent` field which just contains a reference to the initial revision of every page. Note that when a brand new page is created, `id` and `parent` fields are the same.

![image](https://user-images.githubusercontent.com/5121148/78422529-2cdf6f80-7660-11ea-808e-17eeeed93117.png)

Pages are queried in various ways, but the one that required us to create a custom query was listing latest pages, which is used in the Page Builder admin section:

![image](https://user-images.githubusercontent.com/5121148/78421661-9dcf5900-7659-11ea-94bc-1d7c1a0e81f5.png)

So if you take a look at the query in the ["pb-resolver-list-pages" plugin](https://github.com/webiny/webiny-js/blob/master/packages/api-plugin-page-builder-resolvers-mongodb/src/index.ts#L36), in order to fetch the latest revision for every page, we had to first sort the pages by version number (descending), then group the pages by the parent.  

## Possible Solution
The solution is to copy the same behavior we already implemented in the [Form model](https://github.com/webiny/webiny-js/blob/master/packages/api-form-builder/src/plugins/models/form.model.ts), in the Form Builder app.

If you open the file Form model file, you might notice the [`withLatestVersion`](https://github.com/webiny/webiny-js/blob/master/packages/api-form-builder/src/plugins/models/Form/withLatestVersion.ts) higher-order function. This is basically all that is needed in order for us to be able to remove the custom MongoDB query. 

In short, it will just add the `latestVersion` prop to the model, which will be used to track which revision of the form is the latest. On the creation of a new form revision, it will mark it as the latest, and unmark the previous revisions.

So, just copy this to the [`packages/api-page-builder/src/plugins/models/pbPage`](https://github.com/webiny/webiny-js/tree/master/packages/api-page-builder/src/plugins/models/pbPage) folder, and apply it to the `PbPage` model ([packages/api-page-builder/src/plugins/models/pbPage.model.ts:29](https://github.com/webiny/webiny-js/blob/master/packages/api-page-builder/src/plugins/models/pbPage.model.ts)). 

[...]

PR #792, opened 2020-04-20
Page Builder: A custom MongoDB data resolver for querying latest pages is no longer needed

Ashu96:
## Related Issue

Support for DynamoDB - data restructuring - Page Builder - querying latest pages
https://github.com/webiny/webiny-js/issues/783
## Your solution

I followed the instructions mentioned in the issue for a possible solution which is basically
 to copy the same behavior we already implemented in the Form model, in the Form Builder app. 
## How Has This Been Tested?

For testing, I also followed the instructions mentioned in the Issue itself.

1. You should open the Page Builder app in your Webiny project (http://localhost:3001/admin/page-builder/pages), and create a new page.
In the page editor that immediately opens (after you've chosen the page category from the dialog), immediately click on the Publish button in the top right corner:
2. Once you've done it, you should be able to see it in the pages list 
3. After that, create a new revision of the same page by clicking on the edit icon. This should again open the page editor. Return back to the pages list, and you should see your new revision at the top of the list because it's the latest revision.
## Screenshots (if relevant):
https://www.loom.com/share/44149f371af342dcb592c258bca28076


packages/api-page-builder/src/plugins/models/pbPage/withLatestVersion.ts
```diff
@@ -0,0 +1,54 @@
...
+            async beforeDelete() {
+                // If parent is being deleted, do not do anything. Both parent and children will be deleted anyways.
+                if (this.id === this.parent) {
+                    return;
+                }
+
+                if (this.version > 1 && this.latestVersion) {
+                    this.latestVersion = false;
+                    const removeCallback = this.hook("afterDelete", async () => {
+                        const previousLatestForm = await this.constructor.findOne({
```
doitadrian, line 38: `previousLatestForm` should be `previousLatestPage`. 

Please check if there are any other similar occurrences.

Ashu96: @doitadrian 
Thank you for the feedback. I've updated the PR

doitadrian:
Cool, BTW can you check if the installation process works as expected? More specifically, the initial pages are created as expected and that you can open them on the actual site?

You can check that by just deploying your project into a new environment. For example:

`webiny deploy-api --env dev && webiny deploy-apps --env dev`

or

`webiny deploy-api --env prod && webiny deploy-apps --env prod`

Once you have that, open the site.com/admin, finish the installation process, click over pages,  see if they work, and then open them on the actual site.

Let me know if you'll need any help.

P.S.: I just wanted to mention that we have the Cypress test that asserts the installation wizard can be finished properly, but I'm still working on finishing our CI GitHub action, which will run these tests automatically, so we don't have to do it manually.

Issue #795, opened 2020-04-20
Remove "graphql-shield" package


doitadrian:
## This is:

- Feature request

## Intro

All Webiny apps (like for example Page Builder, Form Builder, I18N, ...) rely on the GraphQL API in order to do queries and mutations in the system.

> Read more about the API infrastructure and federated GraphQL schemas [here](https://docs.webiny.com/docs/deep-dive/architecture/api).

For all **admin** queries and mutations, users need to be logged in, otherwise, the GraphQL server will respond with the "Not authorised!" error.

![Snipaste_2020-04-15_10-27-17](https://user-images.githubusercontent.com/5121148/79759550-1ce1b400-831f-11ea-9a55-a30309f34e56.png)

How does the backend know when to throw this error?

For starters, every GraphQL service, that needs authentication / authorization functionality, imports a set of plugins from the `"@webiny/api-security/plugins/service"` package/folder, which can be seen in the `sample-project/api/services` folder, in which all of the different GraphQL APIs are located.

For example, if we open `sample-project/api/services/i18n/src/index.ts`, we can see the`securityServicePlugins` added.

This enables us to utilize the [`hasScope`](https://docs.webiny.com/docs/webiny-apps/security/development/api/graphql-helpers/has-scope) and [`hasRole`](https://docs.webiny.com/docs/webiny-apps/security/development/api/graphql-helpers/has-role) helpers, in order to check if the user has the required security scopes and roles, respectively. For examples, check the `graphql.ts` files, located across several `packages/api-*` packages. For example:

- `packages/api-i18n/src/plugins/graphql.ts:75`
- `packages/api-page-builder/src/plugins/graphql.ts:97`

## What's wrong with this?
Nothing wrong with this approach, but if you go over the project files a bit, you'll notice that we rely on the `graphql-shield` package in order for the `hasRole` and `hasScope` functions to work (check the code). And basically over time, we figured we don't need the `graphlq-shield` package at all.

#### Remove `graphql-shield`
So, we want you to implement the `hasScope` without using the `graphql-shield` package. How are we going to do that?

Well, using function composition. The new `hasScope` will just wrap existing resolvers, and make sure to check the user, before continuing with the execution of the wrapped resolver. Pretty simple.

#### Adapt existing code
Once you've ensured the new `hasScope` works, you will need to go through all of the `graphql.ts` files in all API packages and adapt the existing code so that it works with the new solution. The `security` section that can be found in almost every `graphql.ts` file will dissappear.

[...]

PR #799, opened 2020-04-23
Feat: Remove "graphql-shield" package

Ashu96:
## Related Issue

Remove "graphql-shield" package
[#795](https://github.com/webiny/webiny-js/issues/795)
## Your solution

Followed the instructions mentioned in the issue

1. Implement the `hasScope` using function composition and without using the `graphql-shield` package.
2. Update graphql.ts files in all API packages and adapt the existing code to work with the new solution and without security section
3. Remove `hasRole` helper function as well it's [reference](https://docs.webiny.com/docs/webiny-apps/security/development/api/graphql-helpers/has-role) from doc.webiny.com
4. Update `hasScope` usage example in documentation

## How Has This Been Tested?

Manual testing using graphQL playground and admin app


Ashu96:
#### Query: 
Should we remove `graphql-shield` package and it's usage from `packages/api-security/src/plugins/security.ts`

packages/api-page-builder/src/plugins/graphql/Page.ts
```diff
@@ -86,73 +87,86 @@ export default {
...
             _empty: String
         }
-        
+
         input PbCreatePageInput {
             category: ID!
         }
-        
+
+        # Response types
```
doitadrian, line 99: Not sure why this addition? Was this something that was missing or?

doitadrian:
BTW @Ashu96 , please check `packages/app-security`, and remove role-based checking `hasRole` function as well (`packages/app-security/src/hasRoles.ts`).

We don't need it.

And of course, please check if `hasRoles` is being used in other apps as well (I doubt it).

doitadrian:
@Pavel910 , just wanted to confirm the thing we discussed a couple of days ago..

So, we are going with complete removal of "graphql-middleware" plugins?

Pavel910:
@doitadrian I'm pretty sure we don't need them. The only middleware plugin that was using it was the `graphql-shield` middleware. We never had any need for middleware other than that. I think for now we can safely remove it, and remove that `graphql-middleware` plugin processing as well.

If we ever decide we need it again, we'll add it.

doitadrian:
Awesome stuff @Pavel910 :)

We'll shave off ~100kb of code, which is always great :)

https://bundlephobia.com/result?p=graphql-middleware@4.0.2

packages/api-security/src/scopes.ts
```diff
@@ -1,31 +1,28 @@
...
-    return rule()(async (parent, args, ctx) => {
-        if (!ctx.user) {
-            return false;
-        }
-
-        const access = await ctx.user.access;
-        if (access.fullAccess) {
-            return true;
-        }
+export class ScopeError extends Error {
```
doitadrian, line 3: Let's call this `SecurityError`, calling it `ScopeError` is a bit vague.

doitadrian: Please include `code` and `data` properties in the error. Like we have for example in `packages/graphql/src/responses.ts:16`.

This way, in GQL, you'll be able to get the error code.

image

packages/app-security/src/components/SecureRoute.tsx
```diff
@@ -1,24 +1,21 @@
...
+    scopes
 }: {
     children: any;
     scopes?: ResourcesType;
-    roles?: ResourcesType;
 }): React.ReactElement => {
     const checks = {
-        scopes: scopes ? hasScopes(scopes, { forceBoolean: true }) : true,
-        roles: roles ? hasRoles(roles, { forceBoolean: true }) : true
+        scopes: scopes ? hasScopes(scopes, { forceBoolean: true }) : true
```
doitadrian, line 15: Clean this area please. We had the `checks` object here just because there were `scopes` and `roles` props available. Not we only have `scopes`.

packages/app-security/src/components/SecureView.tsx
```diff
@@ -1,24 +1,21 @@
...
+    scopes
 }: {
     children: any;
     scopes?: ResourcesType;
-    roles?: ResourcesType;
 }): React.ReactElement => {
     const checks = {
-        scopes: scopes ? hasScopes(scopes) : true,
-        roles: roles ? hasRoles(roles) : true
+        scopes: scopes ? hasScopes(scopes) : true
```
doitadrian, line 13: Same comment as in `SecureRoute`.

packages/app-headless-cms/src/admin/plugins/menus.tsx
```diff
@@ -13,14 +13,14 @@ export default [
         render({ Menu, Section, Item }) {
             return (
                 <SecureView
-                    roles={{
+                    scopes={{
```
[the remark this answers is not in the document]
Ashu96: Sorry, I missed it completely.
Thanks for catching it :+1:

Issue #809, opened 2020-05-02
Headless CMS - close menu when selecting "Manage environments"


doitadrian:
When you click on the "Manage environments" link in the Environment selector dialog, the page refreshes (because of dynamic imports) and the menu is reopened. Please check the video:

![ezgif-2-f756f89cf066](https://user-images.githubusercontent.com/5121148/80869546-7e9c0980-8ca1-11ea-815e-f267eeeb5b23.gif)

The menu should not reopen, it should close.

The fix should be somewhere at `packages/app-headless-cms/src/admin/plugins/menus/HeadlessCmsMenu.ts`.

PR #816, opened 2020-05-03
Fix: Headless CMS - close menu when selecting "Manage environments"  

Ashu96:
## Related Issue

Issue #809 
Headless CMS - close menu when selecting "Manage environments"
## Your solution

The issue is generally we use navigation `Item` component to navigate between different views which implicitly hides the menu using `hideMenu` from `useNavigation` hook.

My solution: I've used `hideMenu` from `useNavigation` hook viz basically doing the same stuff that Item component is doing just explicitly. This is one of the solutions. I'm happy to discuss how we can achieve the same with a better approach :smile_cat: 
## How Has This Been Tested?

Manually, by following these steps:
1. Visit the admin app
2. Open the navigation drawer by clicking on the hamburger icon
3. Click on `change environment` and then `Manage Environment`
4. The user navigates to `settings/cms/environments` and drawer is closed now

## Screenshots (if relevant):
Check out this video
https://www.loom.com/share/7ca8b03c2e0e4ef7815a00cc9cce43ce


doitadrian:
Cool @Ashu96.

BTW I can see there are more fixes in this PR, and that the PR is in "draft" state.

I'll revisit this when you feel that this might be ready to get merged.

Issue #814, opened 2020-05-02
Headless CMS - Merge "float" and "integer" fields into a single "number" field


doitadrian:
No need to have two separate fields. For now, let's just go with a single "number" field and later down the road we can adjust this if needed.

BONUS:
I totally forgot that we have the `graphql-scalar` plugin.  So while you're at it, please document it as well.

Just create a new entry here: https://docs.webiny.com/docs/api-development/plugins-reference

doitadrian:
@Pavel910 For this one, we're gonna need a separate `Number` type for the GraphQL schema I'd say. It's either that or the user will have to select the number type in field options.

What do you think?

Some good info here - https://www.apollographql.com/docs/apollo-server/schema/scalars-enums/

Also, check this to see how you can add new scalars: `packages/graphql/src/createSchema/prepareSchema.ts:29`

Pavel910:
I think a custom scalar type for GQL is actually a good idea 👍

PR #834, opened 2020-05-08
Feat: Headless CMS - Merge "float" and "integer" fields into a single "number" field

Ashu96:
## Related Issue

Issue #814 
Headless CMS - Merge "float" and "integer" fields into a single "number" field 
## Your solution

As suggested
> For this one, we're gonna need a separate Number type for the GraphQL schema

Created a `Number` scalar field and added it in `packages/graphql/src/createSchema/prepareSchema.ts:29`
## How Has This Been Tested?

Manually, by following these steps:
1. Visit the admin app and create a new _content model_
2. Use the `Number` field in the editor instead of `Integer or Float` and save the content model
3. Successfully created a content entry for the same.

## Screenshots (if relevant):
image


packages/graphql/src/createSchema/NumberScalar.ts
```diff
@@ -0,0 +1,26 @@
+import { GraphQLScalarType } from "graphql";
+import { Kind } from "graphql/language";
+
+// TODO: Need to check whether user can send number as string
```
doitadrian, line 4: Don't allow sending numbers as string.

packages/graphql/src/createSchema/NumberScalar.ts
```diff
@@ -0,0 +1,26 @@
...
+    name: "Number",
+    description: "A custom input type to be used with numbers. Supports Int and Float.",
+    serialize: getNumber,
+    parseValue: getNumber,
+    parseLiteral: ast => {
+        if (ast.kind === Kind.INT || ast.kind === Kind.FLOAT) {
+            return ast.value;
+        }
+
+        throw new Error("Invalid Number value!");
```
doitadrian, line 24: For example, when you send `false` instead of a string to a `String` field, the error message would be the following:

> Expected type String, found false.

To be a bit more uniformed, lets change the error message to:
> `"Expected type Number, found {value}."`.

doitadrian:
Hey @Ashu96, before merging this one, let's check one more thing here. 

Just make sure users can enter numbers with decimals as well. Currently, with the`Float` field, they cannot. For now, I think we can get away with this by adding `step="any"` to the Input component.

I'll see if we'll need to revisit this one during the following days.

PR #896, opened 2020-05-26
feat: ✨  Add newsletter checkbox to installation process

### LINKED ISSUE #892: Webiny install process - add the option to subscribe to the newsletter.
Reference: https://github.com/webiny/website/blob/master/src/components/footer/newsletter.js

EmilK15:
Closes #892


packages/app-security/src/admin/plugins/installation.tsx
```diff
@@ -156,6 +171,15 @@ const Install = ({ onInstalled }) => {
                                     )
                                 }
                             })}
+                            {!authUserMessage && (
+                                <Grid>
+                                    <Cell span={12}>
+                                        <Bind name="subscribed">
+                                            <Checkbox label={t`Subscribe to Newsletter`} />
```
Pavel910, line 178: @SvenAlHamad @EmilK15 should we add some more text here so user knows what this is?

SvenAlHamad: I agree! @EmilK15 please set the message to something like this:
```
I want to receive updates on product improvements and new features. Doing so I accept Webiny's [privacy policy](https://www.webiny.com/privacy-policy/). 
```

EmilK15:
Added the above comments.
Also removed the unneeded text below password input.
Added a try/catch block to fetch code so that if it does fail it will toss and error

EmilK15:
image
image
Current iteration of changes for new Welcome page

PR #897, opened 2020-05-27
fix: 🐛  Fixed bug of #871 pressing Esc closes the edit field dialog

### LINKED ISSUE #871: Headless CMS - pressing Esc closes the edit field dialog
Clicking on this, shows the I18N overlay. Pressing escape, closes both the overlay, and the edit field dialog.

image

sunguru98:
## Related Issue
Closes #871 

The `I18NInputLocalesOverlay` component being present at a higher z-index was not being closed alone, but also closed its parent component `I18NInput` before.

## Your solution
This bug was fixed by isolating the child component via stopping the event propagation and then closing only the dialog of the child component. React Hotkeyz' `useHotkeys` hook was super useful, as the keys object came with properties embedded with the event object.

## How Has This Been Tested?
Manual testing.

## Screenshots (if relevant):
N/A

### DIFF
--- a/packages/app-i18n/package.json
+++ b/packages/app-i18n/package.json
@@ -33,7 +33,8 @@
     "react-apollo": "^3.1.3",
     "react-helmet": "^5.2.0",
     "slate": "^0.42.2",
-    "slate-react": "^0.19.3"
+    "slate-react": "^0.19.3",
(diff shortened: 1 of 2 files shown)


doitadrian:
P.S. I edited the PR body just to make it a bit nicer (sorted your text into relevant sections).

PR #908, opened 2020-05-30
Introduced the brand new "Files" field

### LINKED ISSUE #850: Headless CMS - File field
The last field to implement - the File field.

Before we continue, a note:
 - CM => Content Model
 - FM => File Manager

Had an additional discussion with @Pavel910 regarding this one, and here is the summary. Basically, in the end, we're not sure if the File field should work directly with FM, or over a File CM entry.

### 1. Working directly with FM
Basically, when you add a File field onto the CM, and then continue with creating a new CM entry, by clicking on the File field, the File Manager is opened, where you choose (or upload and choose) a file, and basically, in the database, you end up with a simple string - a unique file key (which is returned by the FM).

This is great because it's super simple, it will work. The only problem with it is the fact that if you, for example, had 10 CM entries, all linking to the same file, and if you wanted to change the file on all 10 CM entries, you would have to make the change one-by-one. Which is not the case with the next approach I'm about to explain.

### 2. Working with File CM entries, which work with the FM
In this approach, we have this additional layer between the CM entry and the FM, and that's the File CM entry.

So, every time you'd want to link an image to a CM entry, we'd show a UI which would, basically, list all File CM entries, where each File CM entry has the file key to which it points to.  So, technically, you're not picking a file directly from the FM, but an existing File CM entry. And if you were to upload a new file, a new File CM entry would be created, and that would be selected and end up in the database, a simple ID that references the File CM entry.

This approach would solve the issue mentioned in 1. So, if you had 10 CM entries, and wanted to change an image for each, ASSUMING that they all link to the same File CM entry, you'd be able to open that specific File CM entry, change the image to which it points to, and you would have instant change on all 10 entries.

### Which to choose?
Well, that is the question. Implementing the 1st solution would be super easy, and less complicated in terms of how the data would be structured in the database. But then, there's that issue with applying a file change to 10 CM entries, which in the Headless CMSes world maybe presents a significant issue.

If it's a MUST that this field is present in the beta version, that we should definitely go with the 1st solution. If it's not a MUST, then we can research the 10-entries-with-the-same-image issue a bit more, and decide which approach we want to take. As mentioned, if this might be a significant issue for our users, then I guess we won't have any other choice, but to go with the 2nd solution. Which is not a problem, but at the moment, we have other things to think about, and we probably won't be able to squeeze the 2nd solution into the release.

@webiny/webinydevteam @SvenAlHamad

Ashu96:
## Related Issue
Closes #850 
Headless CMS - File field 

## Your solution
Basically followed the solution mentioned in the issue

## How Has This Been Tested?
Manually, by performing the following steps:

- Visit the admin app and head over to the `content model` section

- Create a brand new content model with the `file field ` (you can also toggle `multiValues` option in the field settings)

- Head over to the content entry section and create a new content entry with file

- User should be able to select and upload file

## Screenshots :
https://www.loom.com/share/f776c2d44ac54b978fcb1cc1bb993f9c


packages/app-headless-cms/src/admin/plugins/fields/file.tsx
```diff
@@ -0,0 +1,53 @@
...
+                renderer: {
+                    name: ""
+                }
+            };
+        },
+        renderSettings({ form: { Bind } }) {
+            return (
+                <Grid>
+                    <Cell span={12}>
+                        <Bind name={"settings.type"}>
```
doitadrian, line 40: No need for this, this is controlled via the `multipleValues` flag.

packages/app-headless-cms/src/admin/plugins/fields/file.tsx
```diff
@@ -0,0 +1,53 @@
...
+        allowIndexes: {
+            singleValue: true,
+            multipleValues: false
+        },
+        createField() {
+            return {
+                type: this.type,
+                validation: [],
+                settings: {
+                    type: "single",
```
doitadrian, line 29: Remove this.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileField.tsx
```diff
@@ -0,0 +1,38 @@
...
+import { Cell, Grid } from "@webiny/ui/Grid";
+const t = i18n.ns("app-headless-cms/admin/fields/date-time");
+
+const plugin: CmsEditorFieldRendererPlugin = {
+    type: "cms-editor-field-renderer",
+    name: "cms-editor-field-renderer-file",
+    renderer: {
+        rendererName: "file-input",
+        name: t`File Input`,
+        description: t`Renders inputs for file.`,
```
doitadrian, line 15: Reword this to `Enables selecting a single file via File Manager.`

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/SingleFile.tsx
```diff
@@ -0,0 +1,59 @@
...
+                props.bind.onChange(value.key);
+                setPreviewURL(value.src);
+            } else {
+                props.bind.onChange(value);
+                setPreviewURL(value);
+            }
+        }}
+        value={getValue()}
+        imagePreviewProps={imagePreviewProps}
+        accept={[]}
```
doitadrian, line 53: `accept` is `[]` here?

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/SingleFile.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from "react";
+import { makeRenderImagePreview } from "./utils";
+import SingleImageUpload from "@webiny/app-admin/components/SingleImageUpload";
+import fileIcon from "../../fields/icons/round_insert_drive_file-24px.svg";
+
+const imageExtensions = [".jpg", ".jpeg", ".gif", ".png", ".svg"];
```
doitadrian, line 6: BTW, `FileManager` accepts `images` prop, so you don't have to specify file extensions by yourself.

`packages/app-admin/src/components/FileManager.tsx:73`

doitadrian: BTW now I see why you needed this, for the lower `isImage` check.

At least, you got that tip.

Probably you've seen it already anyways.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/SingleFile.tsx
```diff
@@ -0,0 +1,59 @@
+import * as React from "react";
```
doitadrian, line 1: In the future, this might be refactored a bit. 

We should allow easy registration of new file-type renderers via a simple plugin. So, if a user had a special file extension, and wanted to create a different file renderer for it, he should be able to do it.

No time for that now unfortunately. 

I'll make an issue.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/SingleFile.tsx
```diff
@@ -0,0 +1,59 @@
...
+            } else {
+                props.bind.onChange(value);
+                setPreviewURL(value);
+            }
+        }}
+        value={getValue()}
+        imagePreviewProps={imagePreviewProps}
+        accept={[]}
+        placeholder="Select a file"
+        renderImagePreview={!isImage && makeRenderImagePreview({ value: props.bind.value, imagePreviewProps })}
```
doitadrian, line 55: I don't understand this check. It says, if not image, then call `makeRenderImagePreview`. 

🤔

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileFields.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { useState } from "react";
+import { CmsEditorFieldRendererPlugin } from "@webiny/app-headless-cms/types";
+import { I18NValue } from "@webiny/app-i18n/components";
+import { i18n } from "@webiny/app/i18n";
+import { Cell, Grid } from "@webiny/ui/Grid";
+import MultipleFile from "./MultipleFile";
+
+const t = i18n.ns("app-headless-cms/admin/fields/date-time");
```
doitadrian, line 8: `date-time`

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileFields.tsx
```diff
@@ -0,0 +1,91 @@
...
+    );
+}
+
+const plugin: CmsEditorFieldRendererPlugin = {
+    type: "cms-editor-field-renderer",
+    name: "cms-editor-field-renderer-files",
+    renderer: {
+        rendererName: "file-inputs",
+        name: t`File Inputs`,
+        description: t`Renders a list of file inputs.`,
```
doitadrian, line 81: Check description I put for a single image and put it here as well (with minor adjustments ofc).

sample-project/api/cms/content/tsconfig.json
```diff
@@ -4,6 +4,7 @@
     { "path": "../../../../packages/api-plugin-commodo-db-proxy" },
     { "path": "../../../../packages/api-headless-cms" },
     { "path": "../../../../packages/api-i18n" },
+    { "path": "../../../../packages/api-files" },
```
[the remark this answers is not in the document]
Ashu96: No, it's not needed anymore. Removing it

packages/app-headless-cms/src/admin/plugins/fieldRenderers/longText/longTextLongTextAreas.tsx
```diff
@@ -27,66 +17,26 @@ const plugin: CmsEditorFieldRendererPlugin = {
...
-                                >{t`+ Add value`}</ButtonDefault>
-                            </Cell>
-                        </Grid>
+                <DynamicListMultipleValues {...props}>
+                    {({ bind, index }) => (
+                        <Input
+                            {...bind.index}
+                            autoFocus
+                            rows={5}
+                            onEnter={() => bind.field.appendValue("")}
```
doitadrian, line 29: Actually, for text area, don't enter a new value on pressing the "Enter" key.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/utils.tsx
```diff
@@ -0,0 +1,42 @@
...
+    fontSize: "0.8rem",
+    color: "var(--mdc-theme-on-surface)",
+    backgroundColor: "var(--mdc-theme-on-background)"
+});
+
+export const RenderFilePreview = ({ value, imagePreviewProps }) => {
+    const plugin: AdminFileManagerFileTypePlugin = getPlugin("file-manager-file-type-default");
+    const fileName = getFileName(value);
+
+    const renderImagePreview = (renderImageProps) => {
```
doitadrian, line 26: I don't understand this, a React component that returns a function. 🧐

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileFields.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+const plugin: CmsEditorFieldRendererPlugin = {
+    type: "cms-editor-field-renderer",
+    name: "cms-editor-field-renderer-files",
+    renderer: {
+        rendererName: "file-inputs",
+        name: t`File Inputs`,
+        description: t`Enables selecting multiple files via File Manager.`,
+        canUse({ field }) {
+            return field.type === "file" && field.multipleValues && !field.predefinedValues;
```
doitadrian, line 83: File filed doesn't allow `predefinedValues`, so no need for the last `&& !field.predefinedValues;`. Check the same in the `fileField.tsx`.

doitadrian:
Before merging, I would still like @SvenAlHamad to take a look at the video and the UX in general.

This was the video that @Ashu96 shared:
https://www.loom.com/share/f776c2d44ac54b978fcb1cc1bb993f9c

Issue #798, opened 2020-04-23
Don't hide PB editor loader image


Pavel910:
Our PB Editor has a loader which has this animated image, unfortunately data loading sometimes takes longer than the duration of the animation so the image is hidden and the screen looks broken:

![image](https://user-images.githubusercontent.com/3920893/80086831-8dc0e000-855a-11ea-90ba-190de0263716.png)

We need to prevent the image from disappearing to make it look like this:

![image](https://user-images.githubusercontent.com/3920893/80086856-974a4800-855a-11ea-93de-e1ab762e7205.png)

The image location: `packages/app-page-builder/src/admin/views/Pages/Editor.tsx:47`
Styles reference: `packages/app-page-builder/src/admin/views/Pages/EditorStyled.ts`

asashay:
I'd like to handle this one

Pavel910:
@asashay Great, I'll assign you to this issue. Btw. I'm sorry I didn't get back to you yet, we were busy with the new launch. I'll take a look at that deploy issue you were having. Did you, by any chance, manage to deploy the whole system ?

Also, branch off the `develop` branch for this issue :)

asashay:
@Pavel910 No worries, it was for the better. I pulled the latest code from `master` and deploy went smoothly. Think you don't need to bother (unless someone else experienced it as well)

Pavel910:
Glad to hear that! :) Ok then, let us know if you need help with this issue! 🚀

TommyJackson85:
Should this issue not be closed if @asashay 's pull request has been merged?

Pavel910:
@TommyJackson85 in theory yes, but sometimes it's not closed. Closing manually.

PR #941, opened 2020-06-05
feat: ✨  make PB Editor loader image stay at screen after animation

asashay:
Make image that accompanies `Loading Editor...` text stay at the screen after animation finishes

## Related Issue

https://github.com/webiny/webiny-js/issues/798

## Your solution

Removed animation step that made image disappear from the screen. 

## How Has This Been Tested?

Manually at localhost (https://cl.ly/88b9bf308042)


Pavel910:
@asashay This is great, thank you! And thanks for the video recording 🚀

PR #954, opened 2020-06-08
File Manager - file upload size restrictions are now configurable via settings in the Admin app

### LINKED ISSUE #951: Headless CMS - enable max-file-size adjusting
Users should be able to change the max file size that can be uploaded.

This can be done in the `resources.js`, via the `UPLOAD_MAX_FILE_SIZE` param, but this only adjusts it on the API side. The default value set there is 25MB.

The client still enforces the 10MB file size limit.

So, we should do two things:
1. Increase the limit on client side to at 25MB.
2. Enable adjusting this via the user interface, in the Edit Field dialog.

jmrapp1:
Added file manager settings and components to allow configurable file upload sizes

## Related Issue
Closes #951

## Your solution
- Two new variables were added to the `file-manager` settings DB model: `uploadMinFileSize` and `uploadMaxFileSize` which hold the min and max file size constraints in bytes
- `api-files` was updated to add new GraphQL queries and mutations to allow for modifying the variables through the API
- The `getPresignedPostPayload` function was modified to use these settings when getting a presigned post payload
- A new plugin `app-file-manager` was created to add a new admin settings page. This page is accessible through the admin panel, and allows the user to modify the min and max file upload sizes
- The `FileManagerView` component is what is used to upload files to the cloud. It was modified to run a GraphQL query that retrieves the min and max file upload constraints. These constraints were then passed to the lower level components to ensure that the selected files stayed within the constraints

## How Has This Been Tested?
- This was tested component-by-component
- Testing was first done on the API side to ensure that generating a new project set the two new min and max settings variables in the DB correctly
- Testing was then done on the settings component to validate that the form was populating the fields with the current settings from the DB, and saving any changes updated the fields in the DB
- Final testing that was done by attempting to upload files that exceeded the max file size, ensuring that the upload failed, changing the max file size via the new settings page to allow that file to upload, then retrying the upload and making sure that it works

## Screenshots (if relevant):

[Database Model](https://imgur.com/hwkvibC)
[Sidebar Menu](https://imgur.com/Gy8wdWW)

### Test 1 (fail)
[Settings Page With ~10mb Max](https://imgur.com/vAQ9QJp)
[Image Upload Failed With 40mb image](https://imgur.com/QFCVD6Y)

### Test 2 (pass)
[Settings Page With ~50mb Max](https://imgur.com/vAQ9QJp)
[Image Upload Passed With 40mb image](https://imgur.com/dHeKivt)

### DIFF
--- a/packages/api-files/src/plugins/models/filesSettings.model.ts
+++ b/packages/api-files/src/plugins/models/filesSettings.model.ts
@@ -2,6 +2,8 @@ import { flow } from "lodash";
 import {
     withStaticProps,
     withName,
+    number,
+    int,
     string,
     boolean,
     fields,
@@ -34,6 +36,8 @@ export default ({ createBase }) => {
                 value: {},
                 instanceOf: withFields({
                     installed: boolean({ value: false }),
+                    uploadMinFileSize: number({ value: 0 }),
+                    uploadMaxFileSize: number({ value: 26214401 }),
                     srcPrefix: onSet(value => {
                         // Make sure srcPrefix always ends with forward slash.
                         if (typeof value === "string") {
--- a/packages/api-files/src/plugins/models/filesSettings.model.ts
+++ b/packages/api-files/src/plugins/models/filesSettings.model.ts
@@ -36,8 +36,14 @@ export default ({ createBase }) => {
                 value: {},
                 instanceOf: withFields({
                     installed: boolean({ value: false }),
-                    uploadMinFileSize: number({ value: 0 }),
-                    uploadMaxFileSize: number({ value: 26214401 }),
+                    uploadMinFileSize: number({
+                        value: 0,
+                        validation: validation.create("required,gte:0")
+                    }),
+                    uploadMaxFileSize: number({
+                        value: 26214401,
+                        validation: validation.create("required")
+                    }),
                     srcPrefix: onSet(value => {
                         // Make sure srcPrefix always ends with forward slash.
                         if (typeof value === "string") {
--- a/packages/app-admin/src/components/FileManager/FileManagerView.tsx
+++ b/packages/app-admin/src/components/FileManager/FileManagerView.tsx
@@ -3,7 +3,7 @@ import Files from "react-butterfiles";
 import { ButtonPrimary, ButtonIcon } from "@webiny/ui/Button";
 import { Icon } from "@webiny/ui/Icon";
 import File from "./File";
-import { useQuery, useMutation, useApolloClient } from "react-apollo";
+import { useQuery, useMutation, useApolloClient, Query } from "react-apollo";
 import { FilesRules } from "react-butterfiles";
 import { LIST_FILES, CREATE_FILE } from "./graphql";
 import getFileTypePlugin from "./getFileTypePlugin";
@@ -27,6 +27,7 @@ const t = i18n.ns("app-admin/file-manager/file-manager-view");
 
 import { ReactComponent as SearchIcon } from "./icons/round-search-24px.svg";
 import { ReactComponent as UploadIcon } from "./icons/round-cloud_upload-24px.svg";
+import gql from "graphql-tag";
 
 const style = {
     draggingFeedback: css({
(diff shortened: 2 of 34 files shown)


packages/api-files/src/plugins/models/filesSettings.model.ts
```diff
@@ -34,6 +36,8 @@ export default ({ createBase }) => {
                 value: {},
                 instanceOf: withFields({
                     installed: boolean({ value: false }),
+                    uploadMinFileSize: number({ value: 0 }),
```
doitadrian, line 39: Please apply validation on `uploadMinFileSize` and `uploadMaxFileSize` fields.

```
uploadMinFileSize: number({ value: 0, validation.create("required,gte:0") }),
uploadMaxFileSize: number({ value: 0, validation.create("required") }),
```

doitadrian: For `uploadMaxFileSize`, I didn't put a limit. Not sure what would be a good one really. I guess we can leave it for now as is.

packages/app-admin/src/components/FileManager/FileManagerView.tsx
```diff
@@ -106,6 +107,21 @@ type FileManagerViewProps = {
     multipleMaxSize: number | string;
 };
 
+const graphql = {
```
doitadrian, line 110: Please move this into the "./graphql", in the same fashion we already did with `LIST_FILES` and `CREATE_FILE`. Note that const naming (all uppercase) as well.

packages/app-admin/src/components/FileManager/FileManagerView.tsx
```diff
@@ -303,123 +319,129 @@ function FileManagerView(props: FileManagerViewProps) {
...
-                        />
-
-                        <FileListWrapper>
-                            <Scrollbar
-                                onScrollFrame={scrollFrame =>
-                                    refreshOnScroll({
-                                        scrollFrame,
-                                        fetchMore
-                                    })
+        <Query query={graphql.fileSettingsQuery}>
```
doitadrian, line 323: Can we use `useQuery` here maybe, like we already did on line `265` ? :)

packages/app-file-manager/src/admin/graphql.ts
```diff
@@ -0,0 +1,29 @@
+import gql from "graphql-tag";
```
doitadrian, line 1: Functionally everything is good of course. But just for consistency sakes, please rename query/mutation into something like `GET_SETTINGS` / `UPDATE_SETTINGS`. 

We're basically trying to mimic the same approach that's used on [Apollo docs](https://www.apollographql.com/docs/react/data/queries/).

packages/app-file-manager/src/admin/components/FileManagerSettings.tsx
```diff
@@ -0,0 +1,94 @@
...
+                                        <SimpleFormHeader title="File Manager Settings" />
+                                        <SimpleFormContent>
+                                            <Grid>
+                                                <Cell span={12}>
+                                                    <Grid>
+                                                        <Cell span={6}>
+                                                            <Bind name={"uploadMinFileSize"}>
+                                                                <Input
+                                                                    type="number"
+                                                                    label="Minimum File Upload Size (Bytes)"
```
doitadrian, line 58: Please follow the same style we had in other settings for labels. In other words, no need for uppercasing every word in the label. For example:

image

Also, please move the (Bytes) part into the `description` prop, and let's use something like `In bytes.`.

packages/app-file-manager/src/admin/menus/menus.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { AdminLayout } from "@webiny/app-admin/components/AdminLayout";
+import FileManagerSettings from "../views/FileManagerSettings";
+import { SecureRoute } from "@webiny/app-security/components";
+import { RoutePlugin } from "@webiny/app/types";
+import { i18n } from "@webiny/app/i18n";
+import { AdminMenuSettingsPlugin } from '@webiny/app-admin/types';
+
+const t = i18n.ns("app-file-manager/admin");
+
+const ROLE_FM_SETTINGS = ["cms:file-manager:crud"];
```
doitadrian, line 13: Please rename `cms:file-manager:crud`, to just `files:settings`. I checked the API files, and `files:...` scopes are what we've been using.

BTW I know scopes like `files:crud` are super wide, and should be more specific. It's something we're planning to tackle, hopefully, in the near future.

jmrapp1: Yeah a lot of this I've been learning by just comparing plugins to understand how the whole project architecture works together, and the scope implementation is something I still haven't fully, deeply investigated. I'll make sure to update this!

jmrapp1:
@doitadrian Pushed up those changes! Here's what the menu looks like now too https://imgur.com/dgmv12w

Issue #933, opened 2020-06-04
Headless CMS - Rich Text editor - minor padding issue


doitadrian:
Small, but noticeable. :)

![image](https://user-images.githubusercontent.com/5121148/83770105-f71b2f00-a680-11ea-94f1-4916fb0510b0.png)

Ashu96:
This one will be resolved with PR #979

PR #979, opened 2020-06-09
Fix: Headless CMS - Rich Text Editor Max Height

### LINKED ISSUE #970: Headless CMS - Rich Text Editor Max Height
Let's set a max height on the editor, so that the editor doesn't expand too much, and toolbar icons don't get invisible.

With this issue, we could probably also solve this one: https://github.com/webiny/webiny-js/issues/933

Ashu96:
## Related Issue
Closes #970 #933 
Headless CMS - Rich Text Editor Max Height
## Your solution
- Fix content getting too tall that toolbar hide due to scroll, by adding `height` property
- Add resizer so that the user can resize the editor if needed

## How Has This Been Tested?
Manually, via the Admin app

## Screenshots:
image

### DIFF
--- a/packages/app-i18n/src/admin/components/RichTextEditor/RichTextEditor.tsx
+++ b/packages/app-i18n/src/admin/components/RichTextEditor/RichTextEditor.tsx
@@ -15,11 +15,7 @@ const EditorWrapper = styled("div")({
     border: "1px solid var(--mdc-theme-on-background)",
     borderRadius: 2,
     boxSizing: "border-box",
-    padding: 10,
-
-    "& > div:first-of-type": {
-        marginLeft: "-8px"
-    },
+    padding: 10
 });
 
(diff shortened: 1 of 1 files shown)


packages/app-i18n/src/admin/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -10,16 +10,24 @@ import { MenuButton } from "./index";
...
-import { Range, Descendant } from "slate";
 
 const EditorWrapper = styled("div")({
     border: "1px solid var(--mdc-theme-on-background)",
     borderRadius: 2,
     boxSizing: "border-box",
-    padding: 10
+    padding: 10,
+
+    "& > div:first-of-type": {
```
doitadrian, line 20: Let's go with this approach maybe:
image

PR #981, opened 2020-06-09
[Draft] Feature: Unique validator for form fields

### LINKED ISSUE #822: Headless CMS - create "unique" validator
image

~A couple of things to do here.~

~### API~
~1. Make sure the `unique` can only be set for text fields. BTW, because of this, I think the `unique` should be a separate Commodo field (`packages/api-headless-cms/src/handler/plugins/models/ContentModel/createFieldsModel.ts:23`). Since it's a field-specific option, I would put it into `options` field.~

~1.1 Because of the change above, there will probably be some adjustments in the GraphQL schema creation process.~

~2. Make sure that if a user makes a field `unique`, that he cannot submit another entry with the same value for that field.~

~### App
~1. Add description below in the dialog, explaining what `unique` actually means.~

Create a "unique" field validator.

jmrapp1:
Draft PR to mark story as in-progress

## Related Issue
Closes #822 

## Currently Done
- Added new unique validator components to `app-headless-cms` and `app-form-builder`
- Added unique validator logic to `api-form-builder`
- Validated that the unique validator is working; values can not be duplicated for the same form field

## To-Do
- Add unique logic to `api-headless-cms`
- Add error message on client when unique validation fails on server-side
- Look into creating plugins for `api-` logic and `app-` logic that can be shared in the respective `form-builder` and `headless-cms` plugins (de-duplicate code)

## How Has This Been Tested?
- Tests have been done by creating a form with a unique text field and validating that entering the same value more than once for that form and field combination gets rejected

## Screenshots (if relevant):
*Will be adding these as I go along*

### DIFF
--- a/packages/api-form-builder/src/plugins/graphql/formSubmission.ts
+++ b/packages/api-form-builder/src/plugins/graphql/formSubmission.ts
@@ -2,6 +2,7 @@ import { resolveGet, resolveList } from "@webiny/commodo-graphql";
 import { hasScope } from "@webiny/api-security";
 import exportFormSubmissions from "./formSubmissionResolvers/exportFormSubmissions";
 import createFormSubmission from "./formSubmissionResolvers/createFormSubmission";
+import getFormSubmissionByFieldAndValue from './formSubmissionResolvers/getFormSubmission';
 
 const getFormSubmission = ctx => ctx.models.FormSubmission;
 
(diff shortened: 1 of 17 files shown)


doitadrian:
Hi @jmrapp1,

thanks for this contribution. Unfortunately, I will have to postpone the merge of this for a little while, because we have a couple of other things to wrap up first in the whole "Validators" section.

Once we're done with those, I'll revisit this, and hopefully merge it.

Thanks again! 🍻

PR #985, opened 2020-06-10
feat(app-headless-cms): add link to actual content in content model list

### LINKED ISSUE #965: Headless CMS - in content models view, add link to the actual content
We already added `VIEW CONTENT` button in the content model editor, which leads the users immediately to the content view.

We should also have the same in content models view.

From the feedback sessions:
> When I'm on the content model list, there should be an easy way to access the records for that content model -> icon or a link that would send me to the list of records

Ashu96:
## Related Issue
Closes #965 
Headless CMS - in content models view, add a link to the actual content

## Your solution
Add a link in content model list, which will take the user to the actual content entry list.

## How Has This Been Tested?
Manually, using the Admin app

## Screenshots:
https://www.loom.com/share/2284d274633b4322925266299d648555


packages/app-headless-cms/src/admin/views/ContentModels/ContentModelsDataList.tsx
```diff
@@ -19,6 +20,8 @@ import {
     ListItemMeta,
     ListActions
 } from "@webiny/ui/List";
+import {IconButton} from "@webiny/ui/Button";
```
doitadrian, line 23: Prettier.

packages/app-headless-cms/src/admin/views/ContentModels/ContentModelsDataList.tsx
```diff
@@ -133,16 +144,27 @@ const ContentModelsDataList = (props: ContentModelsDataListProps) => {
                             </ListItemText>
                             <ListItemMeta className={rightAlign}>
                                 <ListActions>
-                                    <EditIcon onClick={editRecord(contentModel)} />
+                                    <Tooltip content={t`View content`} placement={"top"}>
+                                        <IconButton
+                                            icon={<ViewListIcon className={viewEntriesIconStyles}/>}
+                                            label={"View entries"}
```
doitadrian, line 150: Translator missing here.

PR #989, opened 2020-06-11
Fix: Page Builder - updating domain in PB settings doesn't get immediately reflected in the rest of the app

### LINKED ISSUE #962: Page Builder - updating domain in PB settings doesn't get immediately reflected in the rest of the app
Basically, once you update the domain in PB settings, the change should be immediately reflected in the rest of the Page Builder.

Currently, if you update the settings, and open the Page Builder editor, and click on the Preview button, the old domain would still be used. It would only work if the user refreshed the whole page, which is, of course, not a solution.

Make sure everything works immediately on the domain update.

Hint: it's possible that, when fetching settings via GQL, that the `id` field is not included in the response, which prevents Apollo from proper caching / updating the cache.

Ashu96:
## Related Issue
Closes #962 

## Your solution
update `pageBuilder` settings in the cache after `updateSettings` mutation in file
`packages/app-page-builder/src/admin/plugins/settings/components/generalSettings/GeneralSettings.tsx`

## How Has This Been Tested?
Manually, via the Admin app UI

## Screenshots:
https://www.loom.com/share/81a083db4f404e3cb725e4d0090c1016

### DIFF
--- a/packages/app-page-builder/src/admin/plugins/settings/components/generalSettings/GeneralSettings.tsx
+++ b/packages/app-page-builder/src/admin/plugins/settings/components/generalSettings/GeneralSettings.tsx
@@ -9,7 +9,7 @@ import { Query, Mutation } from "react-apollo";
 import { useSnackbar } from "@webiny/app-admin/hooks/useSnackbar";
 import graphql from "./graphql";
 import { CircularProgress } from "@webiny/ui/Progress";
-import { get } from "lodash";
+import { get, set } from "lodash";
 import { validation } from "@webiny/validation";
 
 import {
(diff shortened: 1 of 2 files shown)


packages/app-page-builder/src/admin/plugins/settings/components/generalSettings/GeneralSettings.tsx
```diff
@@ -18,6 +18,7 @@ import {
     SimpleFormContent,
     SimpleFormHeader
 } from "@webiny/app-admin/components/SimpleForm";
+import {DOMAIN_QUERY} from "@webiny/app-page-builder/admin/hooks/usePageBuilderSettings/usePageBuilderSettings";
```
doitadrian, line 21: Prettier, sorrrry :)

PR #1010, opened 2020-06-15
fix(snackbar): fix close button in snackbar at account page

### LINKED ISSUE #766: The Error snackbar from Account settings is not closeable
## This is:
- Bug

## Specifications
  - OS: Windows 10
  - Browser: Chrome

## Expected Behavior
When clicking the `Close` button, the `Snackbar` element should vanish. Also, the `Close` message is currently black on a dark gray background (see the screenshot below), which makes it nearly invisible and is thus very difficult to use: its text color should be more user friendly. 🙂

## Actual Behavior
Clicking on the `Close` button does nothing.

1. Run the `admin` app (this may be the `examples/apps/admin` app) and navigate to the Account's settings which you may also find at URL route `/admin/account`.
2. Edit `packages/app-security/src/admin/views/Account.tsx`'s `onSubmit` handler by making an error appear. Your code should transform from:
`const { error } = response.security.updateCurrentUser;`
into, for example:
`const { error } = { message: "This is an arbitrary error" }`
3. Click on `Update Account`. You should see the snackbar show up exactly like in the screenshot below.

## Additional info
You may see a list of `Tokens` in my screenshot. That's a new feature which I am currently working on - don't worry if you don't see it in the app yet. 🍻

image

afzalsayed96:
Sets the message to null on clicking close thus closing snackbar

## Related Issue

Closes #766 

## Your solution

It seems that the functionality for close button wasn't implemented. The solution was to import `SnackbarAction` component and attach `onClick` handler to it.

## How Has This Been Tested?

Tested locally and verified that snackbar closes on clicking `close`

## Screenshots (if relevant):
As per the current implementation of Snackbar component - `open` prop is handled with `message` itself.

```JSX
const SnackbarMain = () => {
    const ui = useUi();
    const message = get(ui, "snackbar.message");
    const options = get(ui, "snackbar.options", {});

    const hideSnackbar = useCallback(() => {
        ui.setState(ui => ({ ...ui, snackbar: null }));
    }, [ui]);

    return <Snackbar open={!!message} onClose={hideSnackbar} message={message} {...options} />;
};
```

In which case to close the snackbar, we must set message to `null` (or falsy). This causes the message to disappear while snackbar is still in closing transition. It is better visible when the message is very large. See attached GIF below (slowed down to 20% speed for emphasis).

ezgif-6-d5a5dacbca2a

Most users wouldn't notice on normal speed and it could be taken care of in the future.

### DIFF
--- a/packages/app-security/src/admin/views/Account.tsx
+++ b/packages/app-security/src/admin/views/Account.tsx
@@ -15,6 +15,7 @@ import AvatarImage from "./Components/AvatarImage";
 import { validation } from "@webiny/validation";
 import { useSecurity } from "@webiny/app-security/hooks/useSecurity";
 import AccuntTokens from "./AccountTokens";
+import { SnackbarAction } from "@webiny/ui/Snackbar";
 
 import {
     SimpleForm,
@@ -102,7 +103,7 @@ const UserAccountForm = () => {
         setState({ loading: false });
         if (error) {
             return showSnackbar(error.message, {
-                action: "Close"
+                action: <SnackbarAction label="Close" onClick={() => showSnackbar(null)} />
             });
         }
 
--- a/packages/ui/src/Snackbar/index.ts
+++ b/packages/ui/src/Snackbar/index.ts
@@ -1,1 +1,1 @@
-export { Snackbar } from "./Snackbar";
+export { Snackbar, SnackbarAction } from "./Snackbar";


Pavel910:
@afzalsayed96 hey could you please create a new PR against the `master` branch? We'd really appreciate it. We messed something up with the repo and that would be the easiest solution. Thanks!

Also, our main branch for development is now `master`, so please update your fork for further PRs. Cheers!

Issue #829, opened 2020-05-08
Headless CMS - security scopes


doitadrian:
When creating new content models, users should be able to define which user in the system can edit which content models / entries.

We should probably generate scopes dynamically based on `modelId`, and wrap all resolvers with  `hasScope` function accordingly.

Just a high-level idea, didn't go into much details here though.

doitadrian:
Note that we should also have the ability to disable JWT/scopes auth for certain resolvers.

doitadrian:
Please reopen this issue once you add more information and updates here. If this is not the case and you need some help, feel free to seek help from our [Slack Community](https://www.webiny.com/slack) or ping one of the reviewers. Thank you for your contributions!

PR #1018, opened 2020-06-17
Supporting jwt auth in /read and /preview apis + [New] Headless CMS Security scopes

### LINKED ISSUE #994: Headless CMS - Supporting authorization using JWT/PAT and public content
We got several inquiries about building an ecommerce app using Headless CMS. Some of the requested features are posing a blocker at this point in time:
1) having some/all of the content models publicly accessible 
2) being able to access the API using user's JWT, and not the access token

What are the things we need to consider to allow this type of use-case? @webiny/developers

### LINKED ISSUE #1034: Headless CMS Security Scopes
*Addressing the Security Scopes mentioned in #829 and #1018*

Here's how I envision headless security scopes:

##### Toggling Model security scopes
- we add a new section under **Security > Headless CMS** called **Models** *(see ss #1)*
- in this section, you select an existing environment *(see ss #2)*
- then, you select **read** / **preview** / **manage api** (imagine 3 different tabs) and you see a list with all resolvers associated to - that env & api type
- all resolvers above have 1 "Authorization Required" checkbox
- (checkboxes can be disabled and set to true/false for values we don't want to modify, ie: allow full AT access to read/preview api)

##### New user permissions
- we add a new accordion section in /users admin app route where we can set which headless cms resolvers the selected user can use (see ss #3)
- just as above, we will see a list with all resolvers
- the **environment** is selectable via a `<select>` input or sth similar

### Back End
- valid JWTs & PATs are automatically resolved into `context.user.id`, `context.user.access.scopes`, `context.user.access.fullAccess`... In the same fashion we can also resolve a new `context.user.access.headlessCmsScopes` field which contains our aforementioned cms permissions set via checkboxes
```js
eg:

context.user.access.headlessCmsScopes = {
    production: [
        "createBlogPost",
        "getBlogPost",
         ...
    ], 
    bassCannon: [
        "listBlogPosts",
         ...
    ],
    ...
}
```

### User stories solved *([see #1018](https://github.com/webiny/webiny-js/pull/1018#issuecomment-645175156))*
~~**us [1]** - already done~~
**us [1.1]** - we can add a disable / delete button in **/settings/cms/accessTokens?id=...**
**us [1.2]** - we can add a button that disables all **Security > Headless CMS > Models > /read API** checkboxes and sets them to false by default
~~**us [2]** - fixed in my PR #1018~~ (but i could rename the plugin from "context-cms-validate-access-token" to "context-cms-authorize" as it'll be more clear)
**us [3]** - whatever.
- we could give add an option under **Settings > Headless CMS** that allows users to set if the APIs are public by default, thus setting the earlier checkbox's default value (mentioned above at [1.2]). 
- we could also give them a note stating `Bro, your API is public by default. make sure you check your security scopes or you're gonna get your corporation's data crawled by russian hackers and exposed on liveleak.com when they create a new model`

**us [4]** - (actual **TODO**) will solve as explained above if we find it a good approach
**us [5]** - (actual **TODO**) will solve as explained above if we find it a good approach

### Questions / Remarks
**ps / question #1**: "The MANAGE API only needs JWT auth." actually, PATs should (and do currently) authenticate the user as well as the JWT does
**ps #2**: i didn't use hasScope (mentioned in #829) because i didn't need to

image
image
image

Fsalker:
## Related Issue
resolves #1034
addresses part of #994

## Your solution
I looked into `context.user` which is populated automatically when a valid JWT is passed in `context.event.headers.[Aa]uthorization`. Admins will also have `access.fullAccess` field `true` - they will be allowed to access the API.

I used this information to authorize users using JWTs on all 3 APIs (manage, read, preview).

## How Has This Been Tested?
Locally with my admin account's JWT and another created account's (non-admin) JWT

## Screenshots:
A user couldn't be fetched in `context.user` from the authorization header:
image

A user was filled in `context.user` along with its `id` and `access` fields:
image

### DIFF
--- a/packages/api-headless-cms/src/content/plugins/index.ts
+++ b/packages/api-headless-cms/src/content/plugins/index.ts
@@ -55,6 +55,17 @@ export default (
                 return;
             }
 
+            if (
+                context.user &&
+                context.user.id &&
+                context.user.access &&
+                context.user.access.fullAccess
+            ) {
+                // A valid JWT was provided and context.user was populated with our user's data, including their user ID.
+                // If they have fullAccess, then they're admins which are allowed to access the API...
+                return;
+            }
+
             if (context.cms.READ || context.cms.PREVIEW) {
                 const accessToken =
                     context.event.headers["authorization"] ||
@@ -76,10 +87,10 @@ export default (
                         `You are not authorized to access "${currentEnvironment.name}" environment!`
                     );
                 }
-            }
-
-            if (context.cms.MANAGE && !context.user) {
-                throw new Error("Not authorized!");
+            } else if (context.cms.MANAGE) {
+                throw new Error(
+                    "Not authorized! Make sure you provided a valid JWT in the 'authorization' header."
+                );
             }
         }
     },
--- a/packages/app-headless-cms/src/admin/plugins/index.ts
+++ b/packages/app-headless-cms/src/admin/plugins/index.ts
@@ -12,6 +12,7 @@ import contentRevisions from "./contentDetails/contentRevisions";
 import contentModelEditorPlugins from "./../editor/plugins";
 import appTemplateRenderer from "./appTemplatePlugins";
 import welcomeScreenWidget from "./welcomeScreenWidget";
+import scopesList from "./scopesList";
 
 export default () => [
     install,
@@ -31,5 +32,6 @@ export default () => [
     contentModelEditorPlugins,
     appTemplateRenderer,
 
-    welcomeScreenWidget
+    welcomeScreenWidget,
+    scopesList
 ];
--- a/packages/app-headless-cms/src/admin/plugins/scopesList.ts
+++ b/packages/app-headless-cms/src/admin/plugins/scopesList.ts
@@ -1,0 +1,30 @@
+import { i18n } from "@webiny/app/i18n";
+import { SecurityScopesListPlugin } from "@webiny/app-security/types";
+
+const t = i18n.ns("app-security/admin/scopesList");
+
+const generateCmsScopes = async () => {
+    try {
+        // TODO: add actual scopes here. There are two options:
+        //  -query the db, get <environments>, generate the scopes
+        //  -receive <scopes> as a parameter
+        return [
+            {
+                scope: "cms:read:production:potatoes",
+                title: t`Cms "potatoes" /read api permissions`,
+                description: t`Allows you to use the /read api against model "potatoes"`
+            }
+        ];
+    } catch (e) {
+        console.log("Failed to fetch cms scopes:");
+        console.log(e);
+    }
+};
+
+export default [
+    {
+        name: "security-scopes-list-access-tokens",
+        type: "security-scopes-list",
+        scopes: generateCmsScopes
+    } as SecurityScopesListPlugin
+];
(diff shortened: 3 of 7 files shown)


packages/api-headless-cms/src/content/plugins/index.ts
```diff
@@ -55,6 +55,17 @@ export default (
                 return;
             }
 
+            if (
```
doitadrian, line 58: What happens when a user doesn't have a full-access but does have the required scopes?

doitadrian:
I believe we've come at a point in which we are adding logic as-needed, which doesn't necessarily always work and sometimes leads to hacks / hard-coding. In other words, I think we're going to need to restructure this auth section a bit.

There are a couple of things user's have asked:

1. Users want to use access tokens (READ/MANAGE APIs)
2. Users want to use JWTs
3. Users want to have public access on some models, or the whole READ API
4. Users want to have some admins just to be able to edit the content, and for others just to read it.

### Comments
1. For now, we can leave ATs as-is, in a sense that once somebody has AT, he can access everything on the READ / PREVIEW API. But ideally, I believe we should also enable users to attach scopes to it.

1.1 EXTRA FEAT - We should have the option to disable AT, is this possible maybe already? Or you just delete it?

1.2 EXTRA FEAT - We should maybe have the option to remove AT auth completely.

2. Sometimes, there are users that want to access the site that's backed  up by our CMS, and since every user might have different permissions in terms of accessing the content, we need to support JWT auth as well. So, if a valid JWT is received, the GQL resolvers should only be triggered if it contains a valid security scope. [This](https://github.com/webiny/webiny-js/issues/829) issue needs to be resolved for this.

3. Since the users want to have public access on some models, we should add a new setting in the content model editor. I guess we'll need to add per-API-type/per-resolver auth switch. 

Users would be able to, for example, select READ API, and see the generated resolvers, and uncheck the auth where needed. Something like nested checkboxes UI would work here OK I believe.

I'd only do this for the READ API maybe (PREVIEW follows selected settings automatically).

This way, the user can enable public list/read only for models that they need, and the rest might still only be accessable with a valid AT/JWT.

4. This issue again - https://github.com/webiny/webiny-js/issues/829

Pavel910:
@Fsalker I agree with @doitadrian. Maybe let's get this scopes thing implemented #829 because this won't be the last time we get that request from users.

Do a little research to see what's necessary to get this done, write it down in an issue, we'll review it and implement if it isn't a month worth of work.

doitadrian:
BTW one question that's present in this issue - are scopes per environment?

So for example, one user might have `authors:read` on environment A, but not on environment B. So, should scopes be generated like `env-name:authors:read` or maybe even `env-alias-name:read` ?

doitadrian:
Try doing some research @Fsalker, and collect all possible questions you might have, then we can jump on a call together with @Pavel910 and define everything.

packages/api-headless-cms/src/content/plugins/index.ts
Pavel910: @Fsalker I don't really get it; why do you even need to check the `id` and `access` fields? The idea is that if `context.user !== null`, then the user provided a JWT or a PAT and is authenticated. Which at this point is all we need. That should work for both `read` and `manage`.  

`context.user` is only set when: there is a valid JWT/PAT. That's it. 

At this point we're only checking if a user is `authenticated`. We don't care about authorization until we solve those other issues we talked about.

PR #1024, opened 2020-06-17
Headless CMS - Environment aliases are now automatically transformed to lowercase

### LINKED ISSUE #1015: Headless CMS - make sure to lowercase environment slugs
Mange API URLs like `/cms/read/Development` should not happen.

Make sure the environment slugs, used in URLs, are always lower-cased, so the above URL should actually be `/cms/read/development`.

EmilK15:
## Related Issue
https://github.com/webiny/webiny-js/issues/1015
Closes #1015 

## Your solution

In Progress, will update once complete.


packages/api-headless-cms/src/plugins/models/environmentAlias.model.ts
```diff
@@ -8,16 +8,28 @@ import {
     ref,
     withName,
     withHooks,
-    withProps
+    withProps,
+    onSet
 } from "@webiny/commodo";
+import slugify from "slugify";
+
+const toSlug = text =>
```
doitadrian, line 16: Just figured we are using the same thing in `packages/api-headless-cms/src/content/plugins/models/contentModelGroup.model.ts:15`. Please extract this function into a separate file, so we can reuse the same thing in both cases.

You can put it here:
image

packages/api-headless-cms/__tests__/environmentAliases.test.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+`;
+
+describe("Environment Alias test", () => {
+    const database = new Database();
+    const { invoke } = useGqlHandler({ database });
+    const initialEnvironment = { id: mdbid() };
+
+    beforeAll(async () => {
+        await database.collection("CmsEnvironment").insert({
```
doitadrian, line 58: Please check how we've created the test environment in this file (and use the same approach):
`packages/api-headless-cms/__tests__/unlockingFields.test.js:13`

EmilK15:
Fixed the corresponding comments.

doitadrian:
Hey @EmilK15, looking at the changed files - all is looking the same.

Can you check if the code was actually pushed?

Thanks!

PR #1037, opened 2020-06-19
Admin - when trying to visit an unexisting route, users will be redirected to the default route

### LINKED ISSUE #768: Admin - handle non-existing routes
## This is:

- Bug

## Detailed Description
If a user opens a non-existing route in the Admin app, a blank page is shown.

image

## Possible solution
Let's create a simple "Not found" page.

An example that shows how a default route is registered can be found [here](https://github.com/webiny/webiny-js/blob/master/packages/app-page-builder/src/site/plugins/index.tsx#L14). So basically, it's a route that doesn't have the `path` prop defined.

At this moment, I believe this route could be part of the `@webiny/app-admin` plugins. If so, it should be added here `packages/app-admin/src/plugins/index.ts`.

Anshuman71:
## Related Issue
Closes #768 

## Your solution
Redirect unknown routes to the default route.

## How Has This Been Tested?
Manually

## Screenshots (if relevant):
[Gif](https://www.loom.com/share/6b7bbcd382d9479c8b2a78dca3613259)

### DIFF
--- a/packages/app-template-admin/src/index.tsx
+++ b/packages/app-template-admin/src/index.tsx
@@ -62,6 +62,7 @@ export default createTemplate<AdminAppOptions>(opts => {
                             path="/"
                             render={() => <Redirect to={opts.defaultRoute || "/"} />}
                         />
+                        <Route component={() => <Redirect to={opts.defaultRoute || "/"} /> } />
                     </BrowserRouter>
                 );
             }


doitadrian:
The PR looks good to me.

I'd merge this @Pavel910.

Pavel910:
@doitadrian yep, please do!

@Anshuman71 thanks for the PR and the video! 🚀 🍻

PR #1039, opened 2020-06-19
Feat: Page Builder - simplify domains management

### LINKED ISSUE #961: Page Builder - simplify domains management
We've noticed users are struggling a bit with this concept of apps running on different localhost domains. 

The first instance where they're actually asked to do something with the domain is in the Page Builder installation process, where they need to type in the domain on which their site is running.

This is really something we want to hide from them, as it's not easy to grasp. So, in the page builder installation process, let's hide the `domain` input, and behind the scenes, just prefill it with hardcoded `http://localhost:3000` (only if localhost, otherwise use standard CDN URL).

Sure, this might not work for all users (because their site might be running on a different port), that's why we are also introducing the following mechanism.

In the page builder app, users can preview their pages, by clicking on the Preview button:

1) In the page builder editor
image

2) In the page view:
image

When clicking on the button, IF in localhost, let's issue a behind-the-scenes request that will actually ping the saved domain (it's saved in PB settings) and if there is no valid HTTP response, show a warning dialog to the user, saying that he should either start the app, or change the domain in page builder.

Of course, let's make this as easy and understandable for the user. We could also link an article if needed.

Ashu96:
## Related Issue
Closes #961 

## Your solution

1. Remove `domain` input in the PB installation process and by default set it to `http://localhost:3000`if app is running locally.
2. Show confirmation dialog on page preview, if the site is not running

## How Has This Been Tested?
Manually, using the Admin app

## Screenshots
pb-installtion-no-domain-input
---
image
---
https://www.loom.com/share/839212c825514c888ebadceeb67031c5


packages/app-page-builder/src/admin/plugins/pageDetails/header/pageOptionsMenu/PageOptionsMenu.tsx
```diff
@@ -15,6 +16,7 @@ import { setHomePage } from "./graphql";
 import { useConfirmationDialog } from "@webiny/app-admin/hooks/useConfirmationDialog";
 import { getPlugins } from "@webiny/plugins";
 import { PbPageDetailsHeaderRightOptionsMenuItemPlugin } from "@webiny/app-page-builder/types";
+import { useConfigureDomainDialog } from "@webiny/app-page-builder/utils/useConfigureDomain";
```
doitadrian, line 19: Instead of `/app-page-builder/utils/`, put it into the `admin/hooks` folder, where all other hooks are located as well.

packages/app-page-builder/src/admin/plugins/pageDetails/header/pageOptionsMenu/PageOptionsMenu.tsx
```diff
@@ -44,9 +48,21 @@ const PageOptionsMenu = props => {
         )
     });
 
+    const { showConfigureDomainDialog } = useConfigureDomainDialog(getDomain(), refreshSiteStatus);
+
     // We must prevent opening in new tab - Cypress doesn't work with new tabs.
     const target = window.Cypress ? "_self" : "_blank";
 
+    const handlePreviewClick = () => {
```
doitadrian, line 56: Let's wrap this with the `useCallback`. You can use the `[page.id]` for the second arg.

packages/app-page-builder/src/admin/plugins/pageDetails/pageRevisions/Revision.tsx
```diff
@@ -20,12 +20,14 @@ import { ReactComponent as BeenHereIcon } from "@webiny/app-page-builder/admin/a
 import { ReactComponent as GestureIcon } from "@webiny/app-page-builder/admin/assets/gesture.svg";
 import { useRevisionHandlers } from "./useRevisionHandlers";
 import { usePageBuilderSettings } from "@webiny/app-page-builder/admin/hooks/usePageBuilderSettings";
+import {useSiteStatus} from "@webiny/app-page-builder/admin/hooks/useSiteStatus";
```
doitadrian, line 23: BTW check formatting here.

packages/app-page-builder/src/admin/plugins/pageDetails/header/pageOptionsMenu/PageOptionsMenu.tsx
```diff
@@ -44,6 +48,11 @@ const PageOptionsMenu = props => {
         )
     });
 
+    const { showConfirmation: showSettingsConfirmation } = useConfirmationDialog({
```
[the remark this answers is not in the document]
Ashu96: Sorry! This one is outdated. Github glitch maybe :sweat_smile:

PR #1053, opened 2020-06-22
Feat/environment view

### LINKED ISSUE #863: Headless CMS - Add a view with all API URLs
## This is:
- Feature request

## Description
We need to have a place where users can find a list of all URLs to all APIs and environments. Or at least have it per-environment.

When I was using Contentful I was going nuts trying to construct the API URL. I want it to be accessible within 2-3 clicks at most. It can be a separate view, a Dialog, whatever. We just need to make it very easily accessible.

For example:
```
Environment: Production
Content Delivery API:  CDN_URL/cms/read/production
Content Preview API: CDN_URL/cms/preview/production
Content Management API: CDN_URL/cms/manage/production
```

EmilK15:
## Related Issue
Closes #863 

## Your solution
Added information icon next to the name of the environment.
Added EnvironmentInfoDialog component which will display and allow users to copy the URL of their api endpoints.

## How Has This Been Tested?

Made sure that if you switch to different environments that the dialog also gets updated
## Screenshots (if relevant):
image

image


doitadrian:
I'd go with a green icon instead of a black one on the menu. Can't really see it. Not sure how the other icons are being shown in dark mode. 🤔

BTW, not sure if you've seen this comment: https://github.com/webiny/webiny-js/issues/863#issuecomment-640585346

Let's check those, and the remaining stuff after we do the current release. I like the general idea in the dialog, but I fell some additional styling might be added (for example, different fonts for the subtitle and the actual link).

EmilK15:
Okay yea I can add it to the places we mentioned before: 
i.e Add CMS links in environment aliases views.
Also add it in the Access Tokens form.

doitadrian:
BTW, the [dialog](https://user-images.githubusercontent.com/9532882/85323884-de8b8100-b496-11ea-8a9d-07c271302a9e.png) is wrong here, because it's not showing all of the possible env aliases that the current env might have.

Check [this](https://github.com/webiny/webiny-js/issues/863#issuecomment-632765313) comment.

EmilK15:
Updated the the EnvironmentInfoDialog so that it now loads all relevant aliases to the environment you clicked.
Styled the text so it looks much nicer and different.
image

EmilK15:
See the below menu with the added icon.
Adrian and I were talking about what the layout should be in a way that doesn't intrude on the environment name being cut off:
image
One way is we can remove the left Icon and reduce the space specifically for the information icon and go from there.
We could also move the information icon elsewhere so that part of the menu isn't too crowded.
Some other opinions might be helpful to come to a good solution.

EmilK15:
After talking with the team, I'll be changing the icon to be added to the `FooterMenu`.

EmilK15:
Added the API Information to the FooterMenu and can see the updated menu here:
image

EmilK15:
Fixed API Information so it looks like this now:
image
Now EnvironmentInfoDialog looks like the following:
image

EmilK15:
Also added information icon to Environment Aliases:

EmilK15:
Below is a gif of the menus that were added/updated
environmentInfoDialog

doitadrian:
Sorry, but is it possible to provide larger GIFs? :)

EmilK15:
While we find a better GIF editor, I'll attach all relevant screenshots:
image
image
image
image
image
image
image

doitadrian:
What do you think @webiny/developers ?

What I would change...

In the environment aliases / URLs dialog:
1. I'd remove the underline on the type-of-API label (e.g. Content Delivery API). We didn't use this style anywhere in the system, and not only that, underlining might lead the user to think it's a link. So I'd remove the underline, and maybe try to play with the `Typography` component a little, see if something looks nice.

2. Would put a link to the actual links, so that user can immediately click on it, on which the API is opened - in a new tab.

3. Would add a bit of spacing between each section (API alias). I think a single `</br>` would do the trick (or margins via CSS, even better).

The rest looks fine I'd say.

4. The only thing that works a bit weird to me is the (i) icon in the Env Aliases list. Not sure if it's only because of Windows.

Is this an icon from the https://material.io/resources/icons/?style=baseline set ?

EmilK15:
Alright, made the following changes:
I tried the icon set you mentioned and it only looked nice on the footer menu, the icon within the aliases looked off. And I think having consistency for those icons would be great so I stuck with the previous icons I used:
Updated the info dialog like this:
image
I kept the link color orange so users can clearly know what to click on to open the page.
I also kept the Copy Button on the off chance it wasn't clear the urls were links.
Also found a small bug on clicking different aliases within the same environment and fixed that :).

doitadrian:
It looks good visually if you ask me.

Although one thing that surprised me a bit was the fact that we have the `GraphQL API` in it. Because users might get confused about what it actually represents, I'd try to make a distinction between it and the Headless CMS URLs, which are also per-env, which makes it even more complicated.

I'd make a couple of ammends:
1. Make **GraphQL API** a bit larger (the same as **Alias: dev2**).
1.1. I'd also add a tooltip near the label, so that hovering over it, shows an explanation of what the **GraphQL API** actually represents.
2. Below it, I'd add a label **Headless CMS - {ENV NAME}** (maybe {ENV NAME} can be shown in a tooltip as well)
3. After that, I'd have what you already have, although the **Alias: dev2** would take over the size of the **Content Deliver API** label. And the actual font sizes for labels like **Content Deliver API** would get a bit smaller.

That's just my opinion. Sorry for taking so long to merge this, but I'm just not convinced by the solution that we currently have.

Maybe that's just me, so, of course @webiny/developers, if you feel otherwise and that we can just merge this, we can do that as well.

Pavel910:
@doitadrian the idea was to have one single place with all API URLs. So users don't have to use terminal to find the main GraphQL API URL. I agree that we need to iterate on this visually, just to make real distinction between the 2 sections, but I would definitely keep this information all in one place.

doitadrian:
Agreed, I'm also for keeping åll of the info in one place. 👍 👍

EmilK15:
Made the following changes as:
Environment Aliases
image
image
When clicking API information:
image

packages/app-admin/src/components/views/EnvironmentAliases/graphql.ts
```diff
@@ -0,0 +1,136 @@
+import gql from "graphql-tag";
+
+const fields = `
```
doitadrian, line 3: These queries are specific to `app-headless-cms`, and are made in the `app-admin` package.

This should not be here.

I suggest you leave this dialog here, but make it also pluggable, so that the additional Headless CMS URLs section can be added from that app, via a plugin.

I see you have the `EnvironmentInfoDialog` component there. I would rename it to `APIInformationDialog`, and add  the Headless CMS section via a new `admin-api-information-dialog` plugin.

Pavel910: @EmilK15 this is the same thing you did for the `WelcomeScreen`. Create a skeleton which uses plugins, but the plugins themselves must come from other apps. App dependencies must not be hardcoded inside the packages, meaning, `app-admin` must NOT depend on any other app, because it's a skeleton, not an actual implementation. That's the whole point of plugins.

doitadrian: Thanks @Pavel910 for additional explanation.

EmilK15: Great points,
So to recap the plan is for that the component within `app-admin` will render a plugin that will be defined within `app-headless-cms` which will contain the environment information content.
In doing this we won't need to make a query that was specific to `app-headless-cms` from `app-admin`.
Makes sense to me

doitadrian: Cool, let us know if you'll need any help with this.

EmilK15:
Moved method into plugin: Added different views depending on what you click on:
Api Documentation: (now shows GraphQL API, no other view does this)
image
Environment Aliases view now only shows the links for the selected Alias: (no graphql api)
image
Environments now shows links for every alias specific to that environment (no graphql api)
image

PR #1055, opened 2020-06-23
feat: i18n for the mailchimp app

### LINKED ISSUE #618: I18N for the Mailchimp app
Since we want to support localizing base Webiny apps in the future, we need to make sure the labels are wrapped using the `i18n` library (example below).

## This is:
- Feature request

## Detailed Description
The Mailchimp app is located in the following folder: `packages/app-mailchimp/src`.

In there you will find `admin` and `render` folders, which contain a set of plugins for the Page Builder in Admin, and rendering on a real site (respectively).

Both folders contain some strings that should be wrapped.

In `render` folder: 
- `packages/app-mailchimp/src/render/components/MailchimpDefaultForm.js`

In `admin` folder:
- all of the `Mailchimp*` components in the `packages/app-mailchimp/src/admin/components` folder

Make sure to import the `i18n` lib on top of your files, and create an i18n namespace like so (feel free to use other if you feel like it):

```js
import { i18n } from "@webiny/app/i18n";

// For admin you could use the following namespace:
const t = i18n.ns("app-mailchimp/admin");

// For render you could use the following namespace:
const t = i18n.ns("app-mailchimp");
```

After that, just wrap all of the strings with received `t` function, eg. `t("Save  settings")`.

Anshuman71:
## Related Issue
Closes #618 

## Your solution
```tsx
import { i18n } from "@webiny/app/i18n";
const t = i18n.ns("app-mailchimp/admin");
```
Now wrap strings into `t` template literal

## How Has This Been Tested?
Manually

### DIFF
--- a/packages/app-mailchimp/src/admin/components/MailchimpElement.tsx
+++ b/packages/app-mailchimp/src/admin/components/MailchimpElement.tsx
@@ -4,6 +4,8 @@ import { Form } from "@webiny/form";
 import { get } from "lodash";
 import { getPlugins } from "@webiny/plugins";
 import { PbPageElementMailchimpComponentPlugin } from "../../types";
+import { i18n } from "@webiny/app/i18n";
+const t = i18n.ns("app-mailchimp/admin");
 
 const MailchimpElement = React.memo(({ element }: any) => {
     const selected = get(element, "data.settings.component", get(element, "settings.component"));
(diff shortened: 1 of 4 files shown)


doitadrian:
Great job @Anshuman71, many thanks! 🤗🍻

PR #1057, opened 2020-06-23
feat: i18n support for app-google-tag-manager

### LINKED ISSUE #619: I18N for the Google Tag Manager app
Since we want to support localizing base Webiny apps in the future, we need to make sure the labels are wrapped using the `i18n` library (example below).

## This is:
- Feature request

## Detailed Description
The Google Tag Manager app is located in the following folder: `packages/app-google-tag-manager/src`.

In there you will find `admin` and `render` folders, which contain a set of plugins for the Page Builder in Admin, and rendering on a real site (respectively).

But only `admin` contains plugins that actually need I18N, specifically these files: 
- `packages/app-google-tag-manager/src/admin/index.js`
- `packages/app-google-tag-manager/src/admin/components/GoogleTagManagerSettings.js`.

Make sure to import the `i18n` lib on top of your files, and create an i18n namespace like so:

```js
import { i18n } from "@webiny/app/i18n";
const t = i18n.ns("app-google-tag-manager/admin");
```

After that, just wrap all of the strings with received `t` function, eg. `t("Settings updated successfully.")`.

Anshuman71:
## Related Issue
Closes #619 

## Your solution
```tsx
import { i18n } from "@webiny/app/i18n";
const t = i18n.ns("app-google-tag-manager/admin");
```
Now wrap strings into `t` template literal

## How Has This Been Tested?
Manually

### DIFF
--- a/packages/app-google-tag-manager/src/admin/components/GoogleTagManagerSettings.tsx
+++ b/packages/app-google-tag-manager/src/admin/components/GoogleTagManagerSettings.tsx
@@ -9,6 +9,7 @@ import { Input } from "@webiny/ui/Input";
 import graphql from "./graphql";
 import { CircularProgress } from "@webiny/ui/Progress";
 import get from "lodash.get";
+import { i18n } from "@webiny/app/i18n";
 
 import {
     SimpleForm,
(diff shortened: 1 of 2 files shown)


doitadrian:
We are in a middle of a release. Will check it a bit later. 

Thanks!

Anshuman71:
@doitadrian sure. I have the PR in place for similar issues. Good work you all team @webiny

PR #1062, opened 2020-06-23
Fix/cwp error handling

### LINKED ISSUE #1049: npx create-webiny-project new-project 安装错误
## This is:

- Bug
- Feature request
- Question
- Idea

## Specifications

  - version:
  - OS:
  - Browser:

1.
image

2.
3.

## Possible Solution

image``

EmilK15:
## Related Issue
Closes #1049

## Your solution
Added a message that pops up once the project setup fails to clean the project.
I changed where the logs are piped out, so that it is not within the cleaned folder.
Added small tweak for us to support specified tags for create-webiny-project.
Added post-install running git add .gitignore and git commit -m "committed .gitignore".

## How Has This Been Tested?
Ran through create-webiny-project setup.


packages/create-webiny-project/init.js
```diff
@@ -213,10 +213,9 @@ module.exports = async function({ root, appName, templateName, tag, log }) {
                         cwd: root,
                         maxBuffer: "500_000_000"
                     };
-
                     let logStream;
                     if (log) {
-                        logStream = fs.createWriteStream(path.join(root, "logs.txt"), {
+                        logStream = fs.createWriteStream(path.join("./", "cwp-logs.txt"), {
```
Pavel910, line 218: @EmilK15 what do you think about using `--log` in a way that we need to specify a path to log file? For example:
`--log=./new-project.log` This way the user controls the location of the file.

packages/cwp-template-full/index.js
```diff
@@ -22,6 +23,10 @@ module.exports = async ({ appName, root }) => {
         );
     }
 
+    //Commit .gitignore
+    execa.sync("git", ["add", ".gitignore"], { cwd: root });
+    execa.sync("git", ["commit", "-m", `"Commited .gitignore"`], { cwd: root });
```
Pavel910, line 28: Please update the commit message to follow conventional commit standard. `chore: initialize git`.
`Commited` is a past tense, it's not the standard.

packages/create-webiny-project/init.js
```diff
@@ -238,19 +237,19 @@ module.exports = async function({ root, appName, templateName, tag, log }) {
...
-    try {
-        await tasks.run();
-    } catch (err) {
-        console.error("Unable to complete project initialization", err);
-        return;
-    }
+    await tasks.run().catch(async err => {
+        console.error(err);
+        console.log("\nCleaning up project...");
+        await execa("rm", ["-r", root]);
```
Pavel910, line 249: Will `rm` work  in Git Bash on Windows, without WSL?
I think we should use a tool like [rimraf](https://www.npmjs.com/package/rimraf) here, which is cross-platform.

Pavel910 (changes requested):

packages/create-webiny-project/init.js
EmilK15: maybe, We can definitely have that functionality present.
Add a bit of description text within the --help function; but if they don't specify a path we could default to the current directory.

Pavel910: Yeah, defaults are fine as well.

packages/create-webiny-project/init.js
EmilK15: I wanted to confirm this so I tested it out on different shells:
rm works on my powershell, and out of the box with git bash (You can configure when you install Git Bash so that you do get this feature for powershell I believe)
I also found this which points out some flaws in rimraf:
http://www.ashleysheridan.co.uk/blog/Why+You+Should+Not+Use+Rimraf
Though the fact its a bit slow isn't a big deal for the small size of the project.
I also added the bubbling up of errors so the output log has proper outputs.
I added some help text for the log flag, swapped it to use strings instead and accepts path or file names.

Pavel910: Great! 👍

packages/create-webiny-project/index.js
```diff
@@ -15,15 +15,17 @@ console.log = str => originalLog(indent(str, 2));
...
 yargs
     .usage("Usage: $0 <project-name> [options]")
     .version(packageJson.version)
     .demandCommand(1)
-    .example("$0 my-project --template=full")
+    .example("$0 my-project")
     .example(
         "$0 create-webiny-project --template=../path/to/template --tag=../path/to/webiny/files"
     )
+    .example("$0 create-webiny-project --log=./my-logs.txt")
```
Pavel910, line 28: @EmilK15 just caught my eye that these command examples look weird. Looks like they're missing the project name parameter. If we consider `$0` to be `create-webiny-project`, then having a project name `create-webiny-project` looks weird, and in the examples above the `create-webiny-project` string is not even mentioned, so it refers to `$0`.

packages/create-webiny-project/init.js
```diff
@@ -227,30 +231,36 @@ module.exports = async function({ root, appName, templateName, tag, log }) {
...
-        return;
-    }
+    await tasks.run().catch(async err => {
+        if (log.startsWith(".") || log.startsWitch("file:")) {
+            basePath = log;
+        }
+        fs.appendFileSync(path.join(basePath), JSON.stringify(err, null, 2) + os.EOL);
+        console.error(err);
+        console.log("\nCleaning up project...");
+        await execa("rm", ["-r", root]);
```
Pavel910, line 260: Shouldn't this use `rimraf` as well ?

Pavel910 (changes requested):

EmilK15:
Edited the yargs to actually show the examples when you run the --help flag:
image

Pavel910:
@EmilK15 2 minor things:

1. A word gone missing in this sentence :D
image

2. Before the example if `--template=../path...` please add another one which shows `--template=cms`

And that's about it.

EmilK15:
image

Pavel910:
@EmilK15 looks great to me! Good work!

PR #1066, opened 2020-06-23
fix: hiding side nav links from unwarranted users using secure view

### LINKED ISSUE #957: Settings menu items visible to unauthorized user
## This is:
- Bug

## Expected Behavior
When creating a new user without any access permissions (no roles assigned), the user should not see menu items under Settings menu.

## Actual Behavior
The user is able to see the whole Settings menu, but when clicking on a menu item it renders a `Not Authorized` view.

image

1. Create a new user via Security -> Users module, don't assign any roles to the new user.
2. Log in with the new user credentials
3. Open the main navigation

## Possible Solution
I 'm guessing that `scopes` are not applied to the menu items, that's why they're visible.

TommyJackson85:
## Related Issue

Closes #957

## Your solution

Added ```SecureView``` jsx tags from the app security components library, as parent tags to each section of the Side Nav bar where the links need warranted access.

## How Has This Been Tested?
Tested manually:
From the unwarranted user's account and the initial users account (which added the unwarranted user).
I checked both to see if there was any differences.
I ran ```watch``` on packages: app-headless-cms, app-page-builder, app-form-builder
While I edited each package.
From the unwarranted user's account, all sections are gone.

Things left to do:
A message needs to appear when the dropping down settings.
More testing on the user account, maybe add warranted access to see if links reappear afterwards.

## Screenshots (if relevant):
Unwarranted access doesnt display links.
Screenshot 2020-06-24 at 00 23 25

Screenshot 2020-06-24 at 00 23 15


TommyJackson85:
Does the Headless CMS have a required settings type for secure view's scopes?

I set it to ```[forms:settings]``` though ```const ROLE_H_CMS_SETTINGS```. Maybe this const name needs to be changed too. Any ideas?

Pavel910:
@TommyJackson85 great job! As for CMS settings, set it to `const ROLE_CMS_SETTINGS = ["cms:settings"];`

Note that I renamed `ROLE_H_CMS_SETTINGS` to `ROLE_CMS_SETTINGS`, the `H` is not necessary as there will never be another CMS.

Thanks! Looking forward to merging this 🚀

TommyJackson85:
@Pavel910
What about a dropdown message from the settings dropdown when a user has no permissions?
like...
"Your account has no permissions yet."  Or what ever you would suggest.

Pavel910:
@TommyJackson85 it's difficult to know if Settings menu has any items visible, or is completely empty. Each menu plugin is standalone and handles only itself. We'll have to think about how to approach this.

This is where the main `Settings` menu is being rendered: `packages/app-admin/src/plugins/init.tsx:24`. As you can see, it just loops over the `settingsPlugins` and renders those plugins, which may or may not render any DOM depending on the scope.

TommyJackson85:
> @TommyJackson85 it's difficult to know if Settings menu has any items visible, or is completely empty. Each menu plugin is standalone and handles only itself. We'll have to think about how to approach this.
> 
> This is where the main `Settings` menu is being rendered: `packages/app-admin/src/plugins/init.tsx:24`. As you can see, it just loops over the `settingsPlugins` and renders those plugins, which may or may not render any DOM depending on the scope.

Would it be better to create a separate issue for this, once this one is merged?

Pavel910:
@TommyJackson85 yeah totally. Feel free to open a separate issue and we can see how to approach the problem 👍

Pavel910:
Merged! @TommyJackson85 thank you for your time and contribution!! 🍻

TommyJackson85:
No problem, thank you for the opportunity. :)
Will create a new issue now for the message now.

Issue #931, opened 2020-06-04
Headless CMS - ref field - do not allow model change


doitadrian:
When editing ref field, prevent changes on the linked model type. If data exists, changing this would break the API.

On API, throw an error if user tries to change this.
On UI, disable the select input.

Issue #938, opened 2020-06-04
Headless CMS - prevent deletion of models that are linked


doitadrian:
Imagine we have two CMs - `A` and `B`.

`B` has a ref field, that links modes of type `A`. That means that in the GraphQL schema, it will use the `B` type.

Now, if we delete `B` type, the GraphQL schema will immediately become invalid, simply because there is no `B` type.

We must prevent deletions in this case.

PR #1072, opened 2020-06-26
Fix: Headless CMS- update locked field

### LINKED ISSUE #930: Headless CMS - date/time field - do not allow change of date/time type
If data exists, we cannot change the type of data stored.

Throw an error if the user tries to change it on the API side. On the UI, disable the select input.

Ashu96:
## Related Issue
Closes #930 #931 

## Your solution
Use `locked fields` to check for invariant and validate `content model` update

## How Has This Been Tested?
Manually, using the Admin app

## Screenshots:
https://www.loom.com/share/1ee7b4b20a8f4c77b0e749fa7dd3ad75


doitadrian:
We should add a couple of Jest tests for this.

Try testing the ref model type and date/time format changing. Ensure that a proper error message is thrown from the API.

I created a similar test (`packages/api-headless-cms/__tests__/refFieldReferencedIn.test.js:25`) when I was doing https://github.com/webiny/webiny-js/issues/938.

PR #1075, opened 2020-06-26
fix: Page Builder - preventing category deletion with linked pages.

### LINKED ISSUE #1054: Page Builder - prevent deletion of a category
Prevent deletion of a category if there are pages linked with it.

The check should be done in this file:
```packages/api-page-builder/src/plugins/models/pbCategory.model.ts:9```

### PR DESCRIPTION
**Pull request by TommyJackson85**
## Related Issue

Closes #1054

## Your solution
Catch page Categories with existing linked pages, so when attempting to delete a category with pages it will throw an error.

## How Has This Been Tested?
Not tested yet. Just a base for the code based on an [existing example](packages/api-headless-cms/src/content/plugins/models/contentModel.model.ts:81) has been added.
Much of its content doesn't apply to what is needed to be done.

## Screenshots (if relevant):
Screen shot of existing example for reference:
Screenshot 2020-06-27 at 00 06 23

### DIFF
--- a/packages/api-page-builder/src/plugins/models/pbCategory.model.ts
+++ b/packages/api-page-builder/src/plugins/models/pbCategory.model.ts
@@ -3,9 +3,6 @@ import { validation } from "@webiny/validation";
 
 import { withFields, string, withHooks, withName } from "@webiny/commodo";
 
-/*import { PbContext } from "@webiny/api-page-builder/types";
-
-export default ({ createBase, context } : { createBase: Function, context: PbContext }) =>*/
 export default ({ createBase, context }) =>
     flow(
         withName("PbCategory"),
--- a/packages/api-page-builder/src/plugins/models/pbCategory.model.ts
+++ b/packages/api-page-builder/src/plugins/models/pbCategory.model.ts
@@ -1,15 +1,15 @@
 import { flow } from "lodash";
 import { validation } from "@webiny/validation";
 
-import { withFields, string, withName } from "@webiny/commodo";
+import { withFields, string, withHooks, withName } from "@webiny/commodo";
 
 export default ({ createBase }) =>
     flow(
         withName("PbCategory"),
         withHooks({
             async beforeDelete() {
-                const Model = context.models[this.modelId];
-                if (await Model.findOne()) {
+                const { PbPage } = context.models[this.modelId];
+                if (await PbPage.findOne()) {
                     throw new Error(
                         "Cannot delete category because some pages are linked to it."
                     );
--- a/packages/api-page-builder/src/plugins/models/pbCategory.model.ts
+++ b/packages/api-page-builder/src/plugins/models/pbCategory.model.ts
@@ -3,9 +3,10 @@ import { validation } from "@webiny/validation";
 
 import { withFields, string, withHooks, withName } from "@webiny/commodo";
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/api-page-builder/src/plugins/models/pbCategory.model.ts [hunk 1]
```diff
@@ -6,6 +6,16 @@ import { withFields, string, withName } from "@webiny/commodo";
 export default ({ createBase }) =>
     flow(
         withName("PbCategory"),
+        withHooks({
+            async beforeDelete() {
+                const Model = context.models[this.modelId];
```
doitadrian, line 11: You can see all of the available models here `packages/api-page-builder/src/plugins/models.ts`

So, change `const Model = context.models[this.modelId];` to `const { PbPage } = context.models;` and do a `findOne` with the `PbPage` model below.

doitadrian: Btw, `withHooks` needs to be imported as well.

TommyJackson85:
> Please check my comment. Thanks!

Yeah it was just a rough commit, just to make sure I am going about it the right way and because I had to finish up last night. Thanks for the help! Commited your suggested pages.

Just for further understanding as I am new to this repository, when creating a page does it store the category into the database? I assume categories have their own models which updates a list of pages its linked with when saving pages yeah?

I will test the application out now.

TommyJackson85:
So the Build and Test is failing because Context cant be found.
I'm attempting to import the context but is there a context type created for this situation?

My rough attempt at importing a context:
Screenshot 2020-06-27 at 11 58 14

From the types file, would this be suitable as a context?:
Screenshot 2020-06-27 at 11 58 41

doitadrian:
Will take a look at this today.

doitadrian:
@TommyJackson85 check this file `packages/api-page-builder/src/plugins/models.ts:35`.

You can see there the `context` is not passed, so just add it and everything should work.

doitadrian:
Let me know if you need some help @TommyJackson85. :)

TommyJackson85:
> Let me know if you need some help @TommyJackson85. :)

Thanks :)  I will try and work on it now.

TommyJackson85:
@doitadrian What do you think now? I still have to delete a comment out bit of code.

On a side not, I can't manually test this yet as I have this error when navigating to the page builder:

Screenshot 2020-07-02 at 22 51 51

Could this potentially be a database error? Do you know a way I can completely reset my database from the terminal?

Ashu96:
Hey @TommyJackson85 
Looks like the model code is still wrong.

Use `const { PbPage } = context.models;` instead of `const { PbPage } = context.models[this.modelId];`

the same thing that @doitadrian mentioned in the previous comment.

TommyJackson85:
Sorry, I forgot I wrote that actually. Let my try.

TommyJackson85:
Ok I’ll get around to it later. :)

TommyJackson85:
@doitadrian I removed the comment. Merge the changes if you want but I am still getting a TypeError: Cannot read property 'name' of null  from reading```page.category.name```, when loading page-builder/pages.
So I wasn't able to manually test the application, maybe you could test it for me. It might be just a database issue with Mongodb. I might need to delete some of the objects.

TommyJackson85:
@doitadrian @Ashu96  thanks for helping me through it!

PR #1084, opened 2020-06-30
fix(app-template): revert change to getting route plugins

EmilK15:
## Related Issue
Fixes the bug introduced by: https://github.com/webiny/webiny-js/commit/9b5841b15c2b3103d13dd80ed09446a9a047039b

## Your solution
Reverts the method of which to grab the route plugins to the previous method

## How Has This Been Tested?
Manually:

## Screenshots (if relevant):
image

### DIFF
--- a/packages/app-template/src/Routes.tsx
+++ b/packages/app-template/src/Routes.tsx
@@ -5,14 +5,11 @@ import { RoutePlugin } from "@webiny/app/types";
 
 export const Routes = () => {
     const plugins = getPlugins<RoutePlugin>("route").sort((a, b) => {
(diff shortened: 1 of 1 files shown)


Pavel910:
Not merging this one. A proper fix is pushed into `master`. Sorting works correctly, the problem was caused by `app-plugin-admin-welcome-screen` which was registering `route-welcome` using `/` path. So in the end we had 2 routes with `/` path - which of course is not possible.

So the proper way of rendering something on `/` is to register a plugin with `name=route-root` which will override the default `/` content.

PR #1078, opened 2020-06-28
fix: Hiding widgets / links from unauthorized users on welcome screen.

TommyJackson85:
## Related Issue

Closes #1074

## Your solution

Attempting to hide ```welcomeScreenWidget```s from unauthorised users, as they should not have access to if they're not authorised to use them.\
Using ```app-form-builder``` as a first attempt.

My initial attempt was to do it from each welcomeScreenWidget but I only could manage to hide the link button. Here is what it looked like:

Screenshot 2020-06-28 at 13 19 07

So I then applied SecureView to each mapped out Widget plugin from ```app-plugin-admin-welcome-screen/src/components/Welcome.tsx``` file, while passing in a scopes key from the app-form-builder welcomeScreenWidget plugin.
From both unauthorized and authorized accounts, here are the results as it appears to work:

Screenshot 2020-06-28 at 13 49 00

Screenshot 2020-06-28 at 13 48 00

@EmilK15 what do you think? Should I keep the Secure View from the Welcome.tsx file instead and just pass in Secure View scopes from welcomeScreenWidget? I am still yet to do this to other widgets. Just doing one as a first attempt.

## How Has This Been Tested?
I am testing visually and manually through both authorized and unauthorized accounts

PR #1088, opened 2020-06-30
🙀 Simplify Security Plugins

### LINKED ISSUE #1034: Headless CMS Security Scopes

Fsalker:
## TODO
- [ ] update docs to reflect latest plugins changes
- [ ] polish Headless CMS Access Tokens form

## About
This PR introduces a couple of significant changes to the whole framework.

### 1. Changes in the `api-security` plugins
We simplified the security plugins, where the developer now only needs to register a single `type: security` plugin, which handles both authentication and authorization processes, via the `SecurityIdentity` class. This will certainly allow easier integration with other auth providers, like Auth0, Okta, you name it.

### 2. Headless CMS - Access Tokens use security scopes
Previously, if a user wanted to use read/preview API, they would need to create an access token, which would immediately grant full access. Now users can control what can be accessed via the access token, by assigning it needed security scopes.

### 3. Security scopes more flexible
Users can now use `*` when creating scopes. For example, `security:*` would give full permissions to `api-security` API. This also means just assigning the `*` scope gives the full-access to the API.

### 4. Simplified `context` plugins
To alter the `context` object, developers used the `type: context` plugins, which offered `preApply`, `apply`, and `postApply` methods. With this PR, we decided to reduce everything to a single `apply` method, which makes things clearer and also helps to prevent circular dependency situations.

### 5. A single context object
Until this PR, the `context` object sent to the Apollo GraphQL Server would be recreated, which may lead to confusion / inconsistencies. That's why the object created in the handler will now just be passed to the Apollo GraphQL Server, or in other words, a single `context` object will always be present.

### 6. Optimized PAT validation
In order to authenticate via PAT, we have to invoke the `validatePat` function, even when it was happening on the `api-security` handler itself, which can be done via a direct database call too. If we can avoid a function call, we should certainly do it, so we made the necessary changes.

### 7. Refactored tests
We had a couple of different variations on how we approached writing tests in a couple of `api-*` packages. We refactored the old variations and made them work with the latest approach we decided to go with (creating assertions against the complete handler, not just against the schema). The `api-headless-cms` package still contains some tests written in this "old" way, but at this point, it's OK, it's not critical. 

Another cool improvement in this department is that the auth is now working as in a real-world now. We no longer follow the approach `if (ENV === test) then ignore security`. You will need to pass JWTs in order to access and test protected resolvers.


Fsalker:
**Main features so far**
-refactored authentication & authorization: now we have `type: authorization` and `type: authentication` plugins in order to support JWT, PAT (personal access token), AT auth (in the future, we will easily add Okta etc)
-added `slug` field to `CmsEnvironment` model
-updated `CmsEnvironment` test to include slugs
-ran prettier

**[WIP]**
-adding `listContentModels` field to `CmsEnvironment`
-updating `CmsEnvironment` test to include `listContentModels`

Fsalker:
**New commits' summary**
-added `scopes` to `CmsAccessToken` (works well so far) + tests for these `scopes`
-[WIP] attempted to fix `listContentModels`. will get back to it later

Fsalker:
**Latest commits**
-refactored `scopes` in `access tokens`' commodo/gql so that we can set scopes (as originally and logically intended...) 
-updated AT & env tests to include
-added `scopes` to front end
image
image
-added `slug` to cms envs in the FE. now they're fully functional
image
image

**Remaining WIP**
-fixing `context.models.CmsEnvironment` in `packages\api-headless-cms\src\content\plugins\authentication.ts` & sorting out `preApply` / `apply` plugins. we need it in order to generate cms scopes
-fixing `CmsContentModel` in `packages\api-headless-cms\src\plugins\models\environment.model.ts`. we need it to get `modelId` from each model belonging to an env (ps: oddly, CmsContentModel entries don't have an `environment` field, but the entries in mongodb do; i'll figure this out some other time)

TommyJackson85:
Not sure if it's been done, but this file needs the correct CRUD scope passed into the scope array:
```packages/app-headless-cms/src/admin/plugins/welcomeScreenWidget.tsx```
Check ```line 22``` to find the array. Here it is:
```const plugin: AdminWelcomeScreenWidgetPlugin = {
    type: "admin-welcome-screen-widget",
    name: "admin-welcome-screen-widget-headless-cms",
    scopes: [],
    widget: {
```
It is to prevent  welcome widget from being viewed by unauthorised viewers.

For comparison purposes, see these similar examples:
Screenshot 2020-07-13 at 16 12 22

Screenshot 2020-07-13 at 16 12 36

Not sure if this PR is the right place for it but it did previously prevent me from finishing [Pull request 1078](https://github.com/webiny/webiny-js/pull/1078/files). Thanks anyway :)

Pavel910:
@TommyJackson85 There are no scopes currently so an empty array is fine. We'll sort it out in the future, there's a lot going on with security at the moment.

TommyJackson85:
> @TommyJackson85 There are no scopes currently so an empty array is fine. We'll sort it out in the future, there's a lot going on with security at the moment.

Ok, sure at a later stage so. If you want ill delete my suggestion and this comment. I don't want to complicate things.

Pavel910:
@TommyJackson85 No need to delete anything :) Even better, you could open a separate issue for that particular problem so we can fix it when CMS scopes are ready 🚀

Pavel910:
Closing this PR, it has morphed into something a lot bigger, coming in v5.

PR #1095, opened 2020-07-02
Fix/environment incorrect url

### LINKED ISSUE #936: Headless CMS - handle incorrect API URLs
Accessing an invalid URL, like the following...
```https://dugilt40hnfag.cloudfront.net/cms/reXYZad/5ecbedb4e78b2b00082b8b63```

...causes the following error to be thrown:
image

### PR DESCRIPTION
**Pull request by EmilK15**
## Related Issue

Closes #936 

## Your solution
Added try/catch block around the logic that would try to grab environments based on the alias name provided.
Wrote test case to make sure this behavior is enforced and checked for.

## How Has This Been Tested?
Looked at urls with incorrect aliases to see if error message was updated

## Screenshots (if relevant):
image
image

### DIFF
--- a/packages/api-headless-cms/src/content/plugins/models.ts
+++ b/packages/api-headless-cms/src/content/plugins/models.ts
@@ -50,18 +50,19 @@ export default () => {
         // Before continuing with the rest of the models, we must load the environment and assign it to the context.
         let environment = null;
         let environmentAlias = null;
-        if (context.cms.environment && typeof context.cms.environment === "string") {
-            if (context.models.CmsEnvironment.isId(context.cms.environment)) {
-                environment = await context.models.CmsEnvironment.findById(context.cms.environment);
-            } else {
-                environmentAlias = await context.models.CmsEnvironmentAlias.findOne({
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/api-headless-cms/src/content/plugins/models.ts [hunk 1]
```diff
@@ -50,18 +50,19 @@ export default () => {
...
-            } else {
-                environmentAlias = await context.models.CmsEnvironmentAlias.findOne({
-                    query: { slug: context.cms.environment }
-                });
-                environment = await environmentAlias.environment;
-            }
-        }
 
-        if (!environment) {
+        try {
```
doitadrian, line 54: This won't work, because if the user passed an ID of unexisting env, line `57` would get executed, meaning `environment` would be `null` (`findById` will return `null`). After this, the next line we have is the `  context.cms.environment = environment.slug;`, which would again throw a mystery error to the user. :)

I'd stick with the proposition we discussed over the call:
image

doitadrian:
Hoping we can merge this for the upcoming release. Let me know tomorrow if you'll have any questions.

Thanks @EmilK15

PR #1130, opened 2020-07-15
wip(headless): add test case to include setting multiple values as index

### LINKED ISSUE #910: Headless CMS - remove GQL filtering for `multipleValues` fields
Fields with `multipleValues` enabled cannot be searched at the moment. That's why we don't need filters like these:

image

Check all fields, and remove dynamic creation of any filters if the field has `multipleValues` enabled.

P.S. We're gonna need to revisit this topic in the future.

EmilK15:
## Related Issue
Closes #910

## Your solution
Essentially added a test case to test for when a field within Content Models has `multipleValues` attribute to true, it should NOT be able to be set as an index.

## How Has This Been Tested?
Writing a test case for the following scenario above and fix the endpoint so that it won't allow it.

### DIFF
--- a/packages/api-headless-cms/__tests__/multipleValues.test.js
+++ b/packages/api-headless-cms/__tests__/multipleValues.test.js
@@ -121,31 +121,4 @@ describe("Multiple Values Test", () => {
             `Cannot change "multipleValues" for the "tags" field because it's already in use in created content.`
         );
     });
-
-    it(`should not be able to set "multipleValues" as index`, async () => {
-        const { createContentModel, updateContentModel, content } = environment(ids.environment);
-
-        let error;
-        const productContentModel = await createContentModel(
-            mocks.blogWithTagsSetToMultipleValue({ contentModelGroupId: ids.contentModelGroup })
-        );
-
-        let newSome;
-        try {
-            newSome = await updateContentModel({
-                id: productContentModel.id,
-                data: {
-                    indexes: {
-                        fields: ["tags"]
-                    }
-                }
-            });
-        } catch (e) {
-            error = e;
-        }
-
-        expect(error.message).toBe(
-            `Cannot create an index with a field with "multipleValues" set to true`
-        );
-    });
 });
--- a/packages/api-headless-cms/__tests__/multipleValues.test.js
+++ b/packages/api-headless-cms/__tests__/multipleValues.test.js
@@ -121,4 +121,27 @@ describe("Multiple Values Test", () => {
             `Cannot change "multipleValues" for the "tags" field because it's already in use in created content.`
         );
     });
+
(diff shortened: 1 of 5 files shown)


packages/api-headless-cms/src/content/plugins/models/contentModel.model.ts
```diff
@@ -212,10 +212,20 @@ export default ({ createBase, context }: { createBase: Function; context: CmsCon
                             if (field === "id") {
                                 continue;
                             }
-
                             if (!this.fields.find(item => item.fieldId === field)) {
                                 break indexesFor;
                             }
+
```
doitadrian, line 218: 1. Please put a space between `//` and `if`.
2. Make the word `if` => `If`.
3. Error message - change it to `Cannot create a new index with the "${item.fieldId}" field included because the field accepts multiple values.`

packages/api-headless-cms/__tests__/multipleValues.test.js
```diff
@@ -121,4 +121,31 @@ describe("Multiple Values Test", () => {
             `Cannot change "multipleValues" for the "tags" field because it's already in use in created content.`
         );
     });
+
+    it(`should not be able to set "multipleValues" as index`, async () => {
```
doitadrian, line 125: `should not be able to set "multipleValues" as index` => `should prevent the creation of an index that includes a multiple values field in the fields list`

packages/api-headless-cms/__tests__/multipleValues.test.js
```diff
@@ -121,4 +121,31 @@ describe("Multiple Values Test", () => {
...
+        const { createContentModel, updateContentModel, content } = environment(ids.environment);
+
+        let error;
+        const productContentModel = await createContentModel(
+            mocks.blogWithTagsSetToMultipleValue({ contentModelGroupId: ids.contentModelGroup })
```
doitadrian, line 130: While the mock you've chosen here works, let's copy it into a separate file. Notice that basically each test file has a corresponding mock file in the `mocks` folder. Let's keep it that way.

This helps us to avoid breaking tests if adjustments on the mocks are needed (e.g. me breaking your test because I had to change something on it, and vice versa).

doitadrian: Other than that, the test looks good. 👍

adrians5j (changes requested):
A couple of minor change requests, nothing serious.

Please take a look when you get a chance.

packages/api-headless-cms/__tests__/multipleValues.test.js
EmilK15: See new changes to PR

adrians5j (approved):
Nice! 🎉

PR #1132, opened 2020-07-16
Fix: field labels not shown for current locale

### LINKED ISSUE #1096: Headless CMS - field labels not shown for current locale
Changing the locale doesn't change the actual labels (we did set them in the field dialog).

image

Ashu96:
## Related Issue

Closes #1096 

## Your solution
Use `getValue` function from `useI18N` hook to get values like `label and description` for current locale.

## How Has This Been Tested?
Manually using the Admin app.

## Screenshots (if relevant):
https://www.loom.com/share/6144458813ec4d4dbb38693b464942de

### DIFF
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/Time.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/Time.tsx
@@ -3,18 +3,7 @@ import Input from "./Input";
 
 const Time = props => {
     // "09:00:00"
-    return (
-        <Input
-            {...props}
-            field={{
-                ...props.field,
-                label: props.label || props.field.label
-            }}
-            type={"time"}
-            step={5}
-            trailingIcon={props.trailingIcon}
-        />
-    );
+    return <Input {...props} type={"time"} step={5} trailingIcon={props.trailingIcon} />;
 };
 
 export default Time;
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/Time.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/Time.tsx
@@ -8,7 +8,7 @@ const Time = props => {
             {...props}
             field={{
                 ...props.field,
-                label: props.label
+                label: props.label || props.field.label
             }}
             type={"time"}
             step={5}
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/dateTimeFields.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/dateTimeFields.tsx
@@ -1,9 +1,9 @@
 import React from "react";
 import { CmsEditorFieldRendererPlugin } from "@webiny/app-headless-cms/types";
-import { Input } from "@webiny/ui/Input";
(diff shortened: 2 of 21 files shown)


packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/Time.tsx
```diff
@@ -8,7 +8,7 @@ const Time = props => {
             {...props}
             field={{
                 ...props.field,
-                label: props.label
+                label: props.label || props.field.label
```
doitadrian, line 11: Why do we need both props.label && props.field.label? 🤔

doitadrian, line 11: Why do we need `props.label || props.field.label` ? 🤔

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/dateTimeFields.tsx
```diff
@@ -60,16 +62,21 @@ const plugin: CmsEditorFieldRendererPlugin = {
...
                                 />
                             );
                         }
 
                         return (
                             <Input
-                                {...bind.index}
+                                bind={{
+                                    ...bind.index,
+                                    label: t`Value {number}`({ number: index + 1 })
```
doitadrian, line 74: `label` prop can't be passed to `bind`, am I missing something here?

packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileField.tsx
```diff
@@ -16,16 +16,25 @@ const plugin: CmsEditorFieldRendererPlugin = {
...
-                        <Label>
-                            <I18NValue value={field.label} />
-                        </Label>
-                        <Bind>{bind => <SingleFile field={field} bind={bind} />}</Bind>
+                        <Label>{label}</Label>
+                        <Bind>
+                            {bind => (
+                                <SingleFile
+                                    field={field}
+                                    bind={{ ...bind, description: helpText }}
```
doitadrian, line 34: `description` prop can't be passed to `bind`, am I missing something here?

adrians5j (changes requested):
Please check my comments, maybe I'm missing something. 🤔

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/Time.tsx
Ashu96: `Time` component is used in two different places.

- As standalone `Time` field in case of `DateTime` format is set to `Time only`. The label is what we specify in `field settings`. 
- But, when `Time` is used to compose `DateTimeWithoutTimezone` and `DateTimeWithTimezone` it has label specified by that component. For example, if the field label is `sale ends` in the case of `DateTimeWithTimezone`, the label for `Time` field will be `sale ends time`.

Ashu96: I was doing something wrong. It doesn't have to be this way.
Thanks for the feedback @doitadrian :)

adrians5j (approved):

PR #1136, opened 2020-07-16
Fix: Hiding settings tab if no permissions / user roles set

### LINKED ISSUE #1073: fix: Hide Settings tab within the Side nav for user accounts with no permissions.
## This is:

- Bug

## Specifications

  - version:
  - Browser:

## Expected Behavior
When creating/adding a new user (from an administrative user account) without giving new user any access permissions (no roles assigned), that new user (when logged in) should not be able to see any Settings tab / dropdown.

## Actual Behavior
After logging in (from a user account with no permissions), the settings dropdown bar can be seen with no links to click when dropped down. Furthermore there is a big empty space below it wether closed or opened.
I can't even tell when it is open or closed from these two images:

Screenshot 2020-06-26 at 12 39 17
Screenshot 2020-06-26 at 12 39 06

1. Create a new user via Security -> Users module, don't assign any roles to the new user.
2 . Log in with the new user credentials
3. Open the main navigation
4. Take note of the side and the settings dropdown within it.

## Detailed Description

The settings tab and drop down should be removed, and MAYBE replaced with a guide, or other use that is engaging to the new user, but for now we should focus on removing the Settings tab.

## Possible Solution

Me and @Pavel910 were discussing this in [Pull Request: 1066](https://github.com/webiny/webiny-js/pull/1066).

@Pavel910 noted one of the challenges:
> @TommyJackson85 it's difficult to know if Settings menu has any items visible, or is completely empty. Each menu plugin is standalone and handles only itself. We'll have to think about how to approach this.
> 
> This is where the main `Settings` menu is being rendered: `packages/app-admin/src/plugins/init.tsx:24`. As you can see, it just loops over the `settingsPlugins` and renders those plugins, which may or may not render any DOM depending on the scope.

From this initial post, we could discuss further possible solutions.

TommyJackson85:
## Related Issue

Closes #1073 

## Your solution

Use the ```hasScopes``` function from each plugin's ```render```, to check each permission. Set a ```permitted``` key to true if permitted.
Determine if there is at least 1  permission scope set from the app-admin init file. Do this by using the ```some``` function on all permitted keys to find atleast one true value.

This is an updated method in comparison to the previous attempt of using hasScopes from ```app-admin``` which was causing security related errors. The previous attempt was done successfully in [PR 1078](https://github.com/webiny/webiny-js/pull/1078) to display a separate message (instead of welcome widgets) for user accounts with no permissions.

## How Has This Been Tested?

Checking both administrative accounts and users with no permissions:

- Settings tab is removed for none authorised accounts successfully.
- Settings tab remains in place for administrators accounts and when toggled a menu appears.

- I may need to test accounts with only 1 and/or 2 ```settings``` permissions.

## Screenshots (if relevant):
For none authorized users, Settings tab removed:
Screenshot 2020-07-18 at 19 32 31

### DIFF
--- a/packages/app-admin/src/plugins/init.tsx
+++ b/packages/app-admin/src/plugins/init.tsx
@@ -4,7 +4,8 @@ import { registerPlugins, getPlugins } from "@webiny/plugins";
 import { ReactComponent as SettingsIcon } from "@webiny/app-admin/assets/icons/round-settings-24px.svg";
 import { WebinyInitPlugin } from "@webiny/app/types";
 import { AdminMenuSettingsPlugin } from "@webiny/app-admin/types";
-import { hasScopes } from "@webiny/app-security"; 
+
+import { hasScopes } from "@webiny/app-security"; // remove later
 
 const t = i18n.namespace("app-admin/menus");
 
@@ -14,14 +15,19 @@ const plugin: WebinyInitPlugin = {
     init() {
         // Settings
         // Apps / integrations can register settings plugins and add menu items like the following.
-        const settingsPlugins = getPlugins<AdminMenuSettingsPlugin>("admin-menu-settings");
 
         registerPlugins({
             name: "menu-settings",
             type: "admin-menu",
             order: 100,
             render({ Menu, Section, Item }) {    
+                const settingsPlugins = getPlugins<AdminMenuSettingsPlugin>("admin-menu-settings");
+                
+                //replace pl.scopes with pl.permitted to check a boolean value instead
                 const canSeeAnySettings = settingsPlugins.some(pl => hasScopes(pl.scopes, { forceBoolean: true }));
+                
+                console.log("checking PB Settings permission from init file:::::::")
+                console.log(hasScopes(["pb:settings"], { forceBoolean: true }));
                 console.log("canSeeAnyScope");
                 console.log(canSeeAnySettings);
                 console.log(settingsPlugins);
--- a/packages/app-admin/src/plugins/init.tsx
+++ b/packages/app-admin/src/plugins/init.tsx
@@ -4,7 +4,6 @@ import { registerPlugins, getPlugins } from "@webiny/plugins";
 import { ReactComponent as SettingsIcon } from "@webiny/app-admin/assets/icons/round-settings-24px.svg";
 import { WebinyInitPlugin } from "@webiny/app/types";
 import { AdminMenuSettingsPlugin } from "@webiny/app-admin/types";
-import { hasScopes } from "@webiny/app-security"; // remove later
 
 const t = i18n.namespace("app-admin/menus");
 
--- a/packages/app-admin/src/plugins/init.tsx
+++ b/packages/app-admin/src/plugins/init.tsx
@@ -4,7 +4,6 @@ import { registerPlugins, getPlugins } from "@webiny/plugins";
 import { ReactComponent as SettingsIcon } from "@webiny/app-admin/assets/icons/round-settings-24px.svg";
 import { WebinyInitPlugin } from "@webiny/app/types";
 import { AdminMenuSettingsPlugin } from "@webiny/app-admin/types";
-
 import { hasScopes } from "@webiny/app-security"; // remove later
 
 const t = i18n.namespace("app-admin/menus");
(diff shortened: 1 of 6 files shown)


doitadrian:
I converted it back to draft @TommyJackson85 .

packages/app-admin/src/plugins/init.tsx
```diff
@@ -4,6 +4,7 @@ import { registerPlugins, getPlugins } from "@webiny/plugins";
 import { ReactComponent as SettingsIcon } from "@webiny/app-admin/assets/icons/round-settings-24px.svg";
 import { WebinyInitPlugin } from "@webiny/app/types";
 import { AdminMenuSettingsPlugin } from "@webiny/app-admin/types";
+import { hasScopes } from "@webiny/app-security";
```
doitadrian, line 7: The Admin app should not depend on the `app-security` app.

doitadrian: Instead of relying on the `hasScopes` function and the `scopes` property you've added here (not a bad idea), maybe we could just see if any of the registered plugins have returned something.

In other words, can you check... is it possible to determine the `canSeeAnySettings` by examining the output of all registered plugins? 

If all are `null`, maybe we can consider that as an indicator that we don't have to render the Settings item at all.

What do you think @TommyJackson85 ?

adrians5j (changes requested):
Overall I like the way of thinking here, but I think we should try to do it in a bit simpler way.

Please look at my comment, and let me know your thoughts when you get a chance.

Thank you!

P.S. The errors below happened because of the newly added `app-security` to `app-admin`. I believe all errors should go away once you remove it.

packages/app-admin/src/plugins/init.tsx
TommyJackson85: For sure I can look into it.
Just for further context, what would be the purpose of doing this? Is it for better performance and speed? Just asking for learning purposes.
I did use hasScopes for for all welcome widgets in ```app-plugin-admin-welcome-screen``` to see if a separte message needs to be displayed. Do you think we would do the same to the welcome widgets? Maybe create a new issue for it?

doitadrian: First, I wanted to try to avoid adding additional properties to the plugin, and keep it as simple as possible. The other, more important thing, is that we don't want to have security-app functionality in the admin-app. The admin-app should be as standalone as possible, and introducing a security-app related prop here would break that. Ideally, if you wanted to just use the admin-app for creating a simple admin interface, we should be able to do it.

Can you give a reference or two regarding the `app-plugin-admin-welcome-screen` ?

TommyJackson85: @doitadrian 
On an unauthorised account, and from running this is in the console..
```const settingsPlugins = getPlugins<AdminMenuSettingsPlugin>("admin-menu-settings");```
I get:
Screenshot 2020-07-16 at 20 46 14

Doesn't seem to be anything displaying. What about doing an individual ```hasScopes``` security check for each one from each respective file there I initially passed the scope then block the plugin from being passed through?
OR maybe from the init function we could filter them out based on boolean values (i.e. a potential key passed called permitted to use.)

TommyJackson85:
@doitadrian I would'nt want to add too many keys, if I could pass a different value type I would replace the ```scopes``` key with it.

I ran hasScopes on```"pb:settings"``` from both render functions of ```app-admin``` .. ```init```  and ```app-page-builder``` settings/index to get some comparison results, using an Admin account.

Both are returning true. I set a new key called ```permitted``` equaling the hasScopes returned value.

Screenshot 2020-07-17 at 18 37 32
Screenshot 2020-07-17 at 18 37 47
Screenshot 2020-07-17 at 18 38 03

From what I understand, hasScopes has to be ran from render to check users scopes.

So now I could apply, this new permitted key to the remaining plugins.
Then from the init file I can check for the ```permitted``` keys instead. If all are false then the settings tab is gone.

TommyJackson85:
@doitadrian I followed up with my previous suggested changes. It appears to be working well now.

Users with no permissions set now have no settings tab displayed. There is no message for now, incase a user account could have no Settings permissions but might have permissions for other tabs. Atleast now user wont be toggling an empty Settings tab.
Users with atleast one Settings type permission can still see the Settings tab which when dropped down shows all the sections they have permission to use.

As you see from the code changes:
The setting of the ```permitted``` key to it's correct data, has to be done from ```render``` within each Plugin in order to receive the the correct data from ```hasScopes```. It is set to a  boolean value.

From ```app-admin```..```init.tsx``` file, the ```canSeeAnySettings``` variable doesn't need to use useScopes anymore, it can rely on the ```permitted``` key instead to check for atleast one ```true``` value from all plugins. However, as it needs to call render on on each plugin to receive the permitted key, I had to call render on each before being passed through the init.tsx render function. 

In doing this I created a new mapped out list of object data through a new constant called ```renderedSettingsPlugins``` which itself is then mapped out as JSX instead of the initial ```settingsPlugins```. This is just so we arent unecessarily calling render again on each plugin.

As all checks are passing, I will submit this for review. :)

doitadrian:
Thanks @TommyJackson85, I checked the code, and I think we're gonna have to come up with a bit more "non-hackish" solution. Setting the `this.permitted` on every render looks weird if you ask me :)

I'll take a look at this today, and let you know about the alternative approach.

TommyJackson85:
> Thanks @TommyJackson85, I checked the code, and although I see what you wanted to achieve there, I still think we're gonna have to come up with a bit more "non-hackish" solution. Setting the `this.permitted` on every render definitely looks a bit weird if you ask me 🙃
> 
> I'll take a look at this today, and let you know about the alternative approach.

For sure. But we are still only calling render once from the ```init``` file. Only difference is we are mapping the rendered data as new object data BEFORE passing it through jsx. It's why I also include the```name:``` key and variables in the object data so we didnt have to call render again. 
Originally render was called within JSX.
Just highlighting this if you're worried I was calling render again to just to get the ```permitted``` key and data. :)

One thing we could do is include the ```permitted``` key as part of the returned Rendered objects (render() calls back object data), As apposed to adding it to the initial object where render is called. 
I think this would be more secure, plus I know using ```this``` is often confusing haha! I aknowledge this is unclean and confusing::
 ``` this.permitted = hasScopes(ROLE_FORMS_SETTINGS, { forceBoolean: true }); ```

If you don't mind, don't close this PR. I would still like to solve this and will be a good learning experience for me, haha! Just give me some suggestions. Maybe if I find out where the returned object data / model (from render()) comes from, we could pass ```permitted``` value there instead.

TommyJackson85:
@doitadrian permitted key is now found in the rendered object data, i.e. ```this.permitted``` is removed and instead I am passing a new key prop ```permitted``` on SecureView.
The plugins are rendered once only from ```app-admin```, in a ```data``` object which is used for display as well as to check permissions.

PR #1137, opened 2020-07-16
Fix/wrong queries

### LINKED ISSUE #1090: Headless CMS - wrong queries generated
If a model was named e.g. B, then these queries are wrong:
image

Basically, the queries are incorrectly generated, it should be `listAs`, not `listAS`.

I checked the `pluralize` function, it's simply how it works by default. If all letters are uppercase, then the plural suffix letter will also be uppercase.

All in all, didn't have the time to check what is actually causing this error.

EmilK15:
## Related Issue
Closes #1090

## Your solution
Added a parameter to change the name and model if the length of the name is 1 to it won't automatically capitalize as is the default behavior with `pluralize`.

## How Has This Been Tested?
Followed the steps in the issue itself and no error occured

## Screenshots (if relevant):
image


packages/api-headless-cms/src/content/plugins/schema/createManageResolvers.ts
```diff
@@ -27,11 +27,15 @@ export const createManageResolvers: CreateManageResolvers = ({
 }) => {
     const typeName = createTypeName(model.modelId);
     const mTypeName = createManageTypeName(typeName);
+    let pluralizedTypeName = pluralize(typeName);
```
doitadrian, line 30: 1. We could write this like so and make it a bit more efficient:

```ts
let pluralizedTypeName;
if (typeName.length === 1) {
  pluralizedTypeName = `${typeName}s`;
} else {
  pluralizedTypeName = pluralize(typeName);
}
```

Bottom line is, we can call the pluralize function only if actually needed.

2. Since we are c/p-ing this code into several places, let's extract it into a separate helper function.

You can put it into the `utils` folder (check imports at the top of the file, e.g. `packages/api-headless-cms/src/content/plugins/schema/createManageResolvers.ts:7`).

3. Add a short comment in it, why we are doing this.

4. Write a short test please.

adrians5j (changes requested):
Please check my comment. Just a small polish and a test, and we should be good to go.

packages/api-headless-cms/src/content/plugins/utils/pluralizedTypeName.ts
```diff
@@ -0,0 +1,12 @@
+import pluralize from "pluralize";
+
+// This will make it so if the content model name is a single capitalized letter
+// Ex. A, pluralize will not capitalize the name to AS, and instead make it As
+
+export const pluralizedTypeName = (typeName: string) => {
+    let pluralizedTypeName = pluralize(typeName);
```
doitadrian, line 7: We mentioned we want this pluralize call written differently. Is there a specific reason why we left it like this? Am I missing something? :)

packages/api-headless-cms/__tests__/pluralizeContentModel.test.js
```diff
@@ -0,0 +1,29 @@
...
+    it("should be able to create a content model with a single capitalized letter and not have the pluralized version be capitalized", async () => {
+        const { content, createContentModel } = environment(initial.environment.id);
+
+        const contentModel = await createContentModel(
+            mocks.contentModel({ contentModelGroupId: initial.contentModelGroup.id })
+        );
+
+        const ps = await content("p");
+
+        const createdP = await ps.create(mocks.createP);
```
doitadrian, line 25: Let's try this, instead of just calling `ps.list()`, write the query yourself. So `listPs { ... }`.

For example, the following tests are defining GQL queries inline, and executing them using the `invoke` function:
packages/api-headless-cms/__tests__/accessTokens.test.js:51

adrians5j (changes requested):
Please check a couple of comments.

doitadrian:
Any updates on this? @EmilK15

EmilK15:
> Any updates on this? Need help? @EmilK15

I'll spend another hour or two, if I can't figure it out I'll send you a note.

packages/api-headless-cms/src/content/plugins/utils/pluralizedTypeName.ts
```diff
@@ -0,0 +1,14 @@
+import pluralize from "pluralize";
+
+// This will make it so if the content model name is a single capitalized letter
```
doitadrian, line 3: I don't understand this comment.

> // Ex. A, pluralize will not capitalize the name to AS, and instead make it As

I mean, pluralize will turn `A` to `AS`, not `As`.

Did I just read it wrong?

packages/api-headless-cms/__tests__/pluralizeContentModel.test.js
```diff
@@ -0,0 +1,44 @@
...
+
+        let [{ data }] = await invoke({
+            body: {
+                query: LIST_P
+            }
+        });
+
+        expect(data.listPs.data.id).toEqual(mocks.createP.id);
```
doitadrian, line 42: I don't understand how this works, when `mocks.createP.id` doesn't exist (I'm talking about `.id` portion)? 

Wasn't this supposed to be `mocks.createdP.id` ?

packages/api-headless-cms/__tests__/mocks/pluralizeContentModel.js
```diff
@@ -0,0 +1,104 @@
...
+                    },
+                    {
+                        locale: locales.de.id,
+                        value: [false, false, false, true, true]
+                    }
+                ]
+            }
+        }
+    },
+    createdPList: productId => [
```
doitadrian, line 75: Is this used anywhere or is this left here accidentally?

adrians5j (changes requested):
Please check some of my comments.

packages/api-headless-cms/src/content/plugins/utils/pluralizedTypeName.ts
EmilK15: You are right here, I swapped the two by accident. Good catch!

adrians5j (approved):

PR #1148, opened 2020-07-19
Fix/api info improvements

### LINKED ISSUE #1128: Headless CMS: API information improvements
Add improvements to the `API information` section in the admin app.

## This is:
- Bug

## Specifications

  - version: 4.5.0
  - OS: Ubuntu
  - Browser: Chrome 80

## Expected Behavior

1. API information should resolve immediately after creating an alias.
2. The placement of tooltip in the API information dialog should be near to the title, as shown in the screenshot.

## Actual Behavior
1. API information is resolving after page reload.
image

2. The placement of tooltip in the API information dialog is wrong.
image (4)
image (3)

For sync bug:
1. Login into the Admin app and navigate to the environment aliases section.
2. Create a new alias and then click on the info icon, the dialog doesn't have the information.
For visual bug:
1. Login into the Admin app and navigate to the API information section.
2. Check the position of Tooltip in the API information dialog.

EmilK15:
## Related Issue
Closes #1128

## Your solution
Added information icon so that the tooltip now matches that instead of floating text
Added refetch for environment alias so when you create an alias the links render appropriately/

## How Has This Been Tested?
see below:
2nd photo that was after I created a new environment alias and then clicked on the information icon and saw the url's render accordingly.
## Screenshots (if relevant):
image
image


packages/app-headless-cms/src/admin/views/EnvironmentAliases/EnvironmentAliases.tsx
```diff
@@ -21,15 +21,18 @@ function EnvironmentAliases() {
...
                     ]
                 }
             }}
             read={READ_ENVIRONMENT_ALIAS}
             create={{
                 mutation: CREATE_ENVIRONMENT_ALIAS,
                 options: {
-                    update: updateCacheAfterCreate
+                    update: updateCacheAfterCreate,
+                    refetchQueries: [
```
doitadrian, line 33: Since we're already doing it for updating environments, let's also update the cache with necessary changes in the `updateCacheAfterCreate`, instead of making a separate call to API.

EmilK15:
Changing the icon shows the following when using the material icon svg:
image
image
image

packages/app-headless-cms/src/admin/views/EnvironmentAliases/environmentAliasesUtils.ts
```diff
@@ -15,13 +16,46 @@ export const updateCacheAfterCreate = (cache, updated) => {
...
+                }
+            }
+        ];
+
+        cache.writeQuery({
+            ...environmentListAliasParams,
+            data: aliases
+        });
+    } catch (err) {
+        console.log(err);
```
doitadrian, line 49: Just a short question, can you explain this `try` / `catch`? The `writeQuery` / `readQuery` functions throw errors in specific cases?

packages/app-admin/src/plugins/Menu/gqlApiInformation.tsx
```diff
@@ -26,17 +34,19 @@ const plugin: ApiInformationDialogPlugin = {
...
-                            <span>
-                                This link allows you to access content created by different
-                                application across Webiny like Page Builder or Form Builder.
-                            </span>
-                        }
-                    >
-                        <Typography use={"headline6"}>GraphQL API:</Typography>
-                    </Tooltip>
+                    <div className={style.api}>
+                        <Typography use={"headline6"} >GraphQL API:</Typography>
```
doitadrian, line 38: Please remove `:`, so it's just:

> GraphQL API ℹ️

Looks a bit weird if we put the ℹ️  before `:`.

packages/app-headless-cms/src/admin/views/EnvironmentAliases/environmentAliasesUtils.ts
EmilK15: I saw the `try/catch` for cloning the parameters of the `gqlParams`, so I figured it might be a safe bet to wrap the code in a similar `try/catch` in case the `writeQuery` format no longer matches the right format for `LIST_ENVIRONMENT_ALIASES`. Though we probably don't need console.log there for sure

doitadrian:
Looks good.

BTW, I just noticed this space here:
image

EmilK15:
> Looks good.
> 
> BTW, I just noticed this space here:
> image
> 
> Is it possible to remove it somehow?

Will take a look and let you know

packages/app-headless-cms/src/admin/views/EnvironmentAliases/environmentAliasesUtils.ts
doitadrian: I think we can remove `try` / `catch`. Not sure why we'd want to swallow the error if it happens.

Ofc, correct me if I'm missing something.

doitadrian:
Thank you @EmilK15!

EmilK15:
> > Looks good.
> > BTW, I just noticed this space here:
> > image
> > Is it possible to remove it somehow?
> 
> Will take a look and let you know

yea took a look, not sure why that is. CSS is consistent for all of those parts, like no left padding/margins or anything. It might just be the letters but they have the same font. Honestly this is a very very very small issue, I wouldn't worry over this aspect.

doitadrian:
Thought it was an easy fix, but yeah, if not, then let's ignore it.

PR #1171, opened 2020-07-30
Feat/1143 cli webiny info

### LINKED ISSUE #1143: Add a CLI command to view project information
## This is:
- Feature request

## Description
Webiny CLI needs to contain a command to list all relevant URLs for all environments and stacks as it is very difficult to find this information when you need it. Ideally, it would be `yarn webiny info`, and it would list all the data for each stack and every environment that is deployed for each stack.

## Possible Solution
Since every project can be different, and bootstrapped using different project templates, we can't create a generic command for this, but we can create a placeholder command, which will in turn trigger hooks registered by project template.

We already have project template hooks, in this form:
```js
// webiny.root.js
require("@webiny/cwp-template-full/hooks/api")(),
require("@webiny/cwp-template-full/hooks/apps")(),
```

These can internally be upgraded to add a hook plugin for the `info` command.
The `info` command can be added directly to the `cli` itself, also as a plugin, so developers can easily override it if they have a custom way of printing project information. This command would load the hooks of type `hook-project-info` which would individually print project information how they see fit.

Fsalker:
## Related Issue
closes #1143 

## Demo / Screenshot

image

### DIFF
--- a/packages/cli/commands/index.js
+++ b/packages/cli/commands/index.js
@@ -8,10 +8,12 @@ if (!process.env.DEBUG.includes(WEBINY)) {
     process.env.DEBUG += `,${WEBINY}`;
 }
 
+const info = require("./info");
 const run = require("./run");
 
 module.exports.createCommands = (yargs, context) => {
     context.plugins.register(run);
+    context.plugins.register(info);
 
     context.loadUserPlugins();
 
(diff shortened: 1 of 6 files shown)


doitadrian:
Updated output a bit:

**CMS template**
image

**Full template**
image

Issue #1167, opened 2020-07-28
Create a project template for a blank admin app with API


Pavel910:
## This is:
- Feature request

## Detailed Description
Webiny currently offers 2 project templates to use with `create-webiny-project`: [cwp-template-full](https://github.com/webiny/webiny-js/tree/master/packages/cwp-template-full) and [cwp-template-cms](https://github.com/webiny/webiny-js/tree/master/packages/cwp-template-cms).

We need to create a new project template, `cwp-template-basic`, which will create a solid starting point for any developer, without any of our current apps. It will only include an `api` setup with resources necessary to run the `admin` app, and the `admin` app itself, which will use [app-template-admin](https://github.com/webiny/webiny-js/tree/master/packages/app-template-admin).

To learn about React app templates, watch this video: https://www.youtube.com/watch?v=EQxNsDSdpsQ&t=1s
(don't confuse `project template` with `app template`, those are 2 very different things).

## Expected outcome
When users run `npx create-webiny-project my-project --template=basic`, they need to get the described project structure ready for deployment and development.

## Bonus
Tweak CLI hooks to output relevant data and remove unnecessary spaces (make it look decent):

![image](https://user-images.githubusercontent.com/3920893/87779976-08358f00-c82e-11ea-988a-6e7b5b5446b5.png)

## Notes
- you don't need to invent anything for this task
- everything you need is already present in our repository
- don't hesitate to ask for references and instructions
- spend some time getting familiar with existing packages, what they do, how they work
- go through the docs and [Youtube videos](https://www.youtube.com/c/Webiny/videos) to even better understand Webiny

Issue #1166, opened 2020-07-28
Use camelCased version of the appName where needed


doitadrian:
The other day, I ran `webiny create-webiny-project wby-4.5.1`, which caused an issue when trying to create S3 buckets and a new database in runtime.

### How to resolve?
Upon assigning the `MONGODB_NAME` value in the `env.json` files, we should at least `camelCase` it, so we don't end up with an invalid database name. We should do the same for the S3 bucket name.

### Implementation

When creating a new Webiny project, users can choose between `cms` and `full` templates. So, the fix should be implemented in the following files:
`packages/cwp-template-cms/index.js`
`packages/cwp-template-full/index.js`

doitadrian:
Please reopen this issue once you add more information and updates here. If this is not the case and you need some help, feel free to seek help from our [Slack Community](https://www.webiny.com/slack) or ping one of the reviewers. Thank you for your contributions!

ng29:
@doitadrian 
Can i work on this one ?

doitadrian:
No need, because in v5 this is irrelevant. 

But you can maybe try to solve this one -  https://github.com/webiny/webiny-js/issues/1294. What do you say?

ng29:
Just checked that issue @doitadrian can give it a shot, but to be honest, I have started aws within a month ago so might not be win win for me.

But yeah will give it a shot for sure :)

PR #1172, opened 2020-07-31
feat: added cwp-template-basic

Astrix3:
Issue: https://github.com/webiny/webiny-js/issues/1167

1. Created a webiny template **cwp-template-basic**.
2. It includes an api setup with resources needed to run the admin app.
3. After this is published to npm, use this `npx create-webiny-project my-project --template=basic`  command to create a admin app.

- You can also clone/fork the repo.
- Run **yarn** in the folder containing the cloned code, this will install all the dependencies and perform tests.
- Run **yarn setup-repo** . It will setup the environment and build packages.
- In the directory contain you repo run `node ./<folder containing cloned code>/packages/create-webiny-project <your project name> --template=./<folder containing cloned code>/packages/cwp-template-basic`. This will create a project which you can use.
eg: `node ./webiny/packages/create-webiny-project my-project --template=./webiny/packages/cwp-template-basic`

Tested manually:
1. Create project using the above command.
create_project_using_template_basic
2. Deploy project
webiny_template_deploy
3. Run the project
webiny_template_start_local
4. Project 
image
image


packages/cwp-template-basic/README.md
```diff
@@ -0,0 +1,7 @@
+# cwp-template
```
doitadrian, line 1: I'm aware this was just a c/p, but while we're at it, let's update this README with some meaningful info. You can do that for the cms/full templates as well.

doitadrian:
At this point, I'm not 100% on the name `cwp-template-basic`, is the `basic` word we want to have here? 🤔

Astrix3:
cwp-template-basic, name was suggested by pavel, so i kept as it is.

doitadrian:
OK @Astrix3

packages/cwp-template-basic/README.md
Astrix3: should i add, the service/apis it have and how to make use of it, like similar to the things mentioned in PR

doitadrian: Yeah why not.

Explain what this template consists of and why it exists - for the users that don't need any of the official Webiny apps (PB,FB,CMS...), and just want to start building their own custom app.

doitadrian:
BTW add the package you've created to the list here:
`project-root/tsconfig.build.json`

Pavel910:
@doitadrian @Astrix3 we could call it `cwp-template-starter`, as in `starter kit`, many projects use that word. What do you think?

doitadrian:
`cwp-template-starter` sounds good, especially if other projects are using it as well.

Although I must admin, I like `cwp-template-admin` as well ☺️

Pavel910:
@Astrix3 let's just review the package name for a second, to decide on the package name.

This is what the command will look like:
```js
// cwp-template-starter
npx create-webiny-project my-project --template=starter

// cwp-template-admin
npx create-webiny-project my-project --template=admin
```

I'm thinking that in the future, we may even have a template which only contains an API, without the actual admin app, so `starter` is a little bit vague.

Let's go with `cwp-template-admin`, as it precisely describes what is being created: an `admin` app (with API).

Please rename what's necessary and we can then merge this PR. Thanks.

Astrix3:
Sure, i will update it to admin

Astrix3:
The pull request consist of #1166, #1172 and #1167 .
The latest commit also include the readme docs for  all there templates that are admin, cms, full.
As per the request the template is update to admin from basic.
similarly the installation will also change from basic to admin. i.e
`npx create-webiny-project my-project --template=admin` will be used to create a admin app.

packages/cwp-template-admin/index.js
```diff
@@ -0,0 +1,71 @@
...
+    return this.replace(/^([A-Z])|[\s-_](\w)/g, function(match, p1, p2) {
+        if (p2) {
+            return p2.toUpperCase();
+        }
+        return p1.toLowerCase();
+    });
+};
+
+module.exports = async ({ appName, root }) => {
+    const appNam = appName.toCamelCase();
```
Pavel910, line 24: Let's solve this in a better way. Let's move the handling of this to `create-webiny-project`, so we don't duplicate this logic in every template. Then, let's not do camelcase, but instead let's not disallow special characters and "." in the project name. Only allow `[a-zA-Z0-9-_]`. 

Handling it in one place will help us to modify this later if necessary, without updating the code in 3+ packages. Then you can safely remove this `toCamelCase` entirely from these template packages, because `appName` will already be properly formatted when it enters this function.

Astrix3: problem with [a-zA-Z0-9-_] would be, s3 does not accept _ as a character in name. So we will have [a-zA-Z0-9] as name rest all is removed? or should i throw error not a valid name?

Pavel910: We handle that separately so S3 buckets are safe:

```
const s3BucketName = (projectId, appName, env) => {
    return `${projectId}-${appName.toLowerCase().replace(/_/g, "-")}-${env}`;
};
```

TommyJackson85:
> The pull request consist of #1166, #1172 and #1167 .
 @Astrix3 do you mean this PR will close them? I will work on issue #1166 if this doesn't solve it.

doitadrian:
Yes, the fix should be included in this fix.

I'm preparing the this week's release, maaaaaybe I'll have some time to check this one, we'll see 🤞

TommyJackson85:
> Yes, the fix should be included in this PR.
> 
> I'm preparing this week's release, maaaaaybe I'll have some time to check this one, we'll see 🤞

Ok, good to know. I won't work on it so.

doitadrian:
Closing this one, we'll revisit it in v5 branch.

PR #1197, opened 2020-08-21
fix: email addresses with hyphens shouldn't wrap in the UserMenu

### LINKED ISSUE #1196: UserMenu: email addresses with hyphens wrap and break the layout
The UserInfo component in the UserMenu prevents the user's name from wrapping, but not the email. Of course email addresses don't usually contain whitespace, but a hyphen is fairly common and counts as a whitespace character for the purposes of text wrapping.

## This is:
- Bug

## Specifications

  - version: v4.9.0
  - OS: Any
  - Browser: Any (but tested in Chrome and Safari)

## Expected Behavior
An email address containing a hyphen should not wrap.

## Actual Behavior
The email address does wrap, then by taking up 2 lines squeezes the content to the edges of the fixed height header. If the half-address parts and name are short enough it also causes the menu items below to wrap too - e.g. Dark mode, Send feedback and even Sign out can end up on 2 lines each. It might be worth separately giving the whole menu a minimum width.

1. Open CMS (or other app), signed in with a user that has a hyphen in their email address.
2. Click on the user initials (top right)
3. See the email address has wrapped to 2 lines!

## Possible Solution
We can solve this by adding the same CSS the user name already has to the email address to prevent wrapping.

markwilcox:
## Related Issue
Closes #1196

## Your solution
Add the same CSS to the email address span that the user name header above already has.

## How Has This Been Tested?
Tested in Chrome and Safari only (are there any other relevant rendering engines now?). Just a quick visual check.

## Screenshots (if relevant):
Screenshot 2020-08-21 at 15 28 24

### DIFF
--- a/packages/app-security/src/admin/plugins/UserMenu/UserInfo.tsx
+++ b/packages/app-security/src/admin/plugins/UserMenu/UserInfo.tsx
@@ -52,6 +52,12 @@ const linkStyles = css({
             textOverflow: "ellipsis",
             maxWidth: 180,
             display: "block"
+        },
+        ".mdc-typography--subtitle2": {
+            overflow: "hidden",
+            whiteSpace: "nowrap",
+            textOverflow: "ellipsis",
+            maxWidth: 180
         }
     }
 });


Pavel910:
@markwilcox Thanks for the PR! Sorry we missed it before :( Merging it right now for the release this week :)

PR #1224, opened 2020-09-10
Feat: App Admin: hide settings group if empty

### LINKED ISSUE #1223: App Admin: Hide settings section if it is empty
## This is:

- Bug
- Feature request

## Specifications

  - version: 4.11.0

## Expected Behavior

If there is no `admin-menu-settings` plugins, it shouldn't be rendered.
Also if all the settings items rendering `null` (because of permission) it shouldn't render the settings.

## Actual Behavior

It is rendered empty.

Remove all the `admin-menu-settings` and check the menu.

## Detailed Description

I use the app-admin as a dashboard for my client. And all my settings has permission guard, and it looks odd, when a user with no permission for settings see an empty settings section.

I could bypass it by overwriting the `webiny-init-settings` plugin, but it would be great if it would be automatic.

## Possible Solution

https://github.com/webiny/webiny-js/blob/ee1dadbcece0929a4fe259f9051554249012c043/packages/app-admin/src/plugins/init.tsx#L22-L32
```typescript
            render({ Menu, Section, Item }) {
                const items = settingsPlugins
                    .map(plugin => ({ name: plugin.name, element: plugin.render({ Section, Item }) }))
                    .filter(({ element }) => !!element);

                return (
                    <Menu name="settings" label={t`Settings`} icon={<SettingsIcon />}>
                        {items.map(({ name, element }) => (
                            <React.Fragment key={name}>
                                {element}
                            </React.Fragment>
                        ))}
                    </Menu>
                );
            }
```

latotty:
## Related Issue

Closes #1223

## Your solution

Checking the amount of actually rendered settings plugins. And if the list is empty, we return null.

## How Has This Been Tested?

Locally agains my codebase. Tested with no settings plugin, and with settings plugin which rendered null.

### DIFF
--- a/packages/app-admin/src/plugins/init.tsx
+++ b/packages/app-admin/src/plugins/init.tsx
@@ -20,12 +20,21 @@ const plugin: WebinyInitPlugin = {
             type: "admin-menu",
             order: 100,
             render({ Menu, Section, Item }) {
+                const items = settingsPlugins
+                    .map(plugin => ({
+                        name: plugin.name,
+                        element: plugin.render({ Section, Item })
+                    }))
+                    .filter(({ element }) => !!element);
+
+                if (!items.length) {
+                    return null;
+                }
+
                 return (
                     <Menu name="settings" label={t`Settings`} icon={<SettingsIcon />}>
-                        {settingsPlugins.map(plugin => (
-                            <React.Fragment key={plugin.name}>
-                                {plugin.render({ Section, Item })}
-                            </React.Fragment>
+                        {items.map(({ name, element }) => (
+                            <React.Fragment key={name}>{element}</React.Fragment>
                         ))}
                     </Menu>
                 );


Pavel910:
@latotty does this handle the lack of permissions, or just absence of settings plugins?

latotty:
> @latotty does this handle the lack of permissions, or just absence of settings plugins?

Both, I've tested it

Given that lack of permission means the entry is rendering null, I've checked some part of the code and it was the case.

Pavel910:
@latotty Perfect! Let's merge this one for the next release 👍

TommyJackson85:
@latotty Interesting. In the [pull request 1136](https://github.com/webiny/webiny-js/pull/1136/files) which I was working on, I don't think I was getting any ```null``` returns from each when calling render on plugins a user didn't have permission to. Anyway, yours is much cleaner than mine so well done! 👍 

@Pavel910 some checks are failing here, is it still ok to merge?

doitadrian:
This file needs to be checked: `packages/app-admin/src/plugins/init.tsx`. Try formatting it with prettier.

The other error is very weird, it looks like a configuration error, not an actual error in the test.

latotty:
> This file needs to be checked: `packages/app-admin/src/plugins/init.tsx`. Try formatting it with prettier.
> 
> The other error is very weird, it looks like a configuration error, not an actual error in the test.
> 
> Ignore it for now, we'll take a look at that.

Ohh sorry, fixed 😄 got used to auto format with eslint 😄

PR #1237, opened 2020-09-15
fix: prettier and eslint run for v5 branch

brunozoric:
## Related Issue
No issue

## Your solution
Ran the prettier and eslint to fix all style related stuff. Also removed eslint rules for webiny/i18n namespaces.

## How Has This Been Tested?
Ran `yarn build`, `yarn webiny deploy api --env=local` and `yarn start`.


packages/cli-plugin-scaffold-node-package/template/src/index.ts
```diff
@@ -3,4 +3,5 @@
  * Reference it from other packages via the following import statement:
  * import { PACKAGE_NAME_CAMEL_CASED } from "PACKAGE_NAME";
  */
+/* eslint-disable @typescript-eslint/no-empty-function */
```
doitadrian, line 6: This file is a part of the Node Package scaffold, and with the `webiny scaffold` tool, it is copied to the user's project 

The thing is, there is no value in having the `eslint-disable` above the function, because users will start writing code in it (or they will remove it and write a class / plugin / whatever), and the `eslint-disable` immediately becomes redundant.

If you want to avoid `eslint` errors here, either ignore the file or maybe even better, just return some value from the function, e.g. `return Math.random();` or something cool.

packages/cli-plugin-scaffold-react-package/template/src/index.tsx
```diff
@@ -3,4 +3,5 @@
  * Reference it from other packages via the following import statement:
  * import { NewComponent } from "PACKAGE_NAME";
  */
+/* eslint-disable @typescript-eslint/no-empty-function */
```
doitadrian, line 6: See `packages/cli-plugin-scaffold-node-package/template/src/index.ts` comment.

doitadrian (changes requested):
Looks good, just one minor thing, please check the comments.

packages/cli-plugin-scaffold-node-package/template/src/index.ts
brunozoric: returned Math.random() as suggested.

packages/cli-plugin-scaffold-react-package/template/src/index.tsx
brunozoric: returned Math.random() as suggested.

packages/cli-plugin-scaffold-react-package/template/src/index.tsx
```diff
@@ -3,4 +3,6 @@
  * Reference it from other packages via the following import statement:
  * import { NewComponent } from "PACKAGE_NAME";
  */
-export const NewComponent = function() {};
+export const NewComponent = function() {
+    return Math.random();
```
doitadrian, line 7: Let's return some kind of DOM here, as the function represents a React component. E.g. 

```js
return <div>This is my new component.</div>;
```

doitadrian (changes requested):
One minor detail and I'd say we're good to go.

package.json
```diff
@@ -137,8 +137,8 @@
     "lint-staged": "lint-staged",
     "postinstall": "node ./scripts/linkPackages.js",
     "prepublishOnly": "node scripts/prepublishOnly.js",
-    "prettier": "prettier --config .prettierrc.js --write",
-    "prettier:check": "prettier \"packages/**/src/**/*.{js,ts,tsx}\" --check --config .prettierrc.js",
+    "prettier": "prettier \"*.{js,jsx,ts,tsx}\" \"cypress/**/*.{js,jsx,ts,tsx}\" \"packages/**/src/**/*.{js,jsx,ts,tsx}\" --config .prettierrc.js --write",
+    "prettier:check": "prettier \"./*.{js,jsx,ts,tsx}\" \"cypress/**/*.{js,jsx,ts,tsx}\" \"packages/**/src/**/*.{js,jsx,ts,tsx}\" --check --config .prettierrc.js",
```
doitadrian, line 141: This looked like it could be simplified, so I tried the following:

```
"prettier:check": "prettier **/*.{js,ts,tsx} --check --config .prettierrc.js",
```

The result is:
image

This is **master** branch BTW, so results may be different in v5 branch.

Anyhow, can you try the updated `prettier:check` command maybe and see what results you get?

packages/cli-plugin-scaffold-react-package/template/tsconfig.json
```diff
@@ -1,3 +1,6 @@
 {
-  "extends": "../../tsconfig.json"
+  "extends": "../../tsconfig.json",
+	"compilerOptions": {
```
doitadrian, line 3: Something looks off here (prettier?).

packages/project-utils/bundling/app/config/webpack.config.js
```diff
@@ -234,13 +234,16 @@ module.exports = function(webpackEnv, { paths, babelCustomizer }) {
                             safari10: true
                         },
                         // Added for profiling in devtools
+                        /* eslint-disable-next-line @typescript-eslint/camelcase */
```
doitadrian, line 237: Is it possible to apply the `/* eslint-disable-next-line @typescript-eslint/camelcase */` rule just once, at the top of the file? Just curious.

doitadrian (changes requested):
A couple of minor comments, let me know if you have any questions, thanks! 🍻

Issue #784, opened 2020-04-06
Support for DynamoDB - data restructuring - Page Builder - querying images


doitadrian:
Continuing from #781, this issue gives more details about solving the "Page Builder - querying latest pages" issue.

## From the introduction issue
Both pages and page elements (`PbPage` and `PbPageElement` collections in the database) contain images and currently, they are stored as simple IDs in the database. That means every time a page is being loaded, the database needs to be queried / the files need to be loaded. 

So if we had 10 files placed in a page, that's basically 10 database calls. And what's even worse, is that the Page Builder is directly querying the `Files` collection, owned by the File Manager service. In theory, the database that the File Manager is relying on might be different than the one Page Builder is using, so this definitely shouldn't be allowed.

The following images show some example data:

![image](https://user-images.githubusercontent.com/5121148/78454754-4e5d4d00-769a-11ea-9c9c-61eb5475bfaf.png)
![image](https://user-images.githubusercontent.com/5121148/78454769-692fc180-769a-11ea-9bc6-546d3572777c.png)

## Possible Solution
The idea is to replace the shown IDs with file "keys". What is a file "key" ?

It's basically a file name, but it is unique and immutable once created. The file key is created upon file upload. Here we can see some examples:

![image](https://user-images.githubusercontent.com/5121148/78522805-db092780-77ce-11ea-968a-a305663d7c77.png)

By replacing the IDs with the keys, we won't be needing to query the database to get it the key that's behind the ID. This key basically all that is needed in order for us to construct the full URL. The only other thing we need is the `srcPrefix`, which is stored in File Manager's settings:

![image](https://user-images.githubusercontent.com/5121148/78523104-fa548480-77cf-11ea-8c14-1b40e022c2f7.png)

How to get that? Well, we don't want to get it by doing a database query on the `Settings` collection and make a similar mistake we previously did (querying the `Files` collection directly from the Page Builder service).

We should fetch it via the File Manager's GraphQL endpoint with a simple GraphQL query. Note that [`srcPrefix`](https://github.com/webiny/webiny-js/blob/master/packages/api-files/src/plugins/models/filesSettings.model.ts) cannot be extracted via GraphQL and that you'll need to expose it first.

Once you have that, in the Page Builder, you'll be able to do something like [this](https://github.com/webiny/webiny-js/blob/master/packages/api-i18n/src/plugins/service.ts). Here we can see that we are using a GraphQL client to query the I18N service for needed data. The same logic should happen in the page builder.

[...]

PR #1238, opened 2020-09-15
feat: file is saved as a object containing more data instead of just id

### LINKED ISSUE #614: Handling of files in the `content` field of `PbPage` model
## This is:
- Idea

## Actual Behavior

Before we start off, let's give some background.

## Background
In PageBuilder, we have `PpPage` model and in it, the `content` field.

The `content` field may store images. The following is an excerpt in
which we can see how a file is saved in the database.

```
(...)
    {
      "id": "Of2JQrobK",
      "data": {
        "image": {
          "file": "5db3410646e6da8d61cc1393"
        }
      },
      "elements": [],
      "path": "0.0.0.0.2",
      "type": "image"
    }
(...)
```

This value of `file` key here is a result of several steps:
1. In the Page Builder, user picks an image from File Manager. Note that the File Manager doesn't return the file's ID, but a full data object: `{ src, size, id, type, key }`. So once the client sends the revision update mutation, this whole object is sent.

2. On the backend, we have a set of plugins that completely traverse received `content`, and do the replacing upon saving to the database. So instead of the full received object, only `id` is saved, like in the first example shown above.

3. Again, when fetching data from the database, we have the opposite process, in which a set of plugins convert these IDs into `{ src, id }` objects, so that the frontend again has the initial structure od data. Note that only `src` and `id` values are returned, no need really for other keys. But it can be added if needed.

## The problem
The problem is currently happening in step 3. Currently, when we want to replace the ID with the object, we need to load the file, and we do that by directly accessing the database connection, and doing a query on `File` collection, which is not an ideal approach.

The second approach that was tried here was doing a GQL query on `api-files` service, which is much better in terms of microservices approach. But this time, the problem is that getting/listing pages becomes slow since each page has to do an HTTP request (1 page = 1 HTTP request that fetched multiple files). We have an even worse case when you do a `listPages`, in which you might query for 10 or more pages. A lot of HTTP requests and a lot of latency is being added.

So is there a third way that we could solve this?

## Possible Solution
To solve the above-explained problem, we could save files a bit differently.

Instead of saving the file `id` to the database, we would save file `key`. This is good because field `key` never changes and since we have it, we can easily construct the full URL to the file upon receiving the page data from the database.

We would only need to fetch the prefix for all files, which is stored in `api-files` service's settings, and which can be fetched via GQL. This would need to be fetched only once so it's not super intensive on the backend.

So to summarize, this would free us from making GQL requests (fetching all files by their IDs), since we would be able to construct full paths directly in code.

### Problem - File Manager File-Replacement Feature 
Question: do we want a file-replacement feature in the File Manager? In other words, can a user have a single image selected in multiple pages, and then go to the File Manager and choose that same image and replace it with another one? So basically `id` of the file stays the same, and the `key` to which it points changes.

If so, the above solution won't work, since it proposes saving file `key` instead of `id`. And changing the key of a single file via File Manager would not apply changes to the `content` field in pages.

brunozoric:
## Related Issue
Closes #614 #784 

## Status
- [x] correct storing/fetching the data in the existing modifiers and use `SettingsManager` to get the `srcPrefix`, always once
- [ ] adapt installation process to the newest changes
- [x] **[moved to #1250]** host installation files on a public S3  bucket

## Your solution
We are now saving file object that contains __type, id, name, size, type and meta object (height, width, aspectRatio, private).
When fetching, there is only one extra call to get srcPrefix so we can create src of the image.

## How Has This Been Tested?
Project was built from the start - yarn install -> setup-repo -> deploy -> save page, save settings general tab, save settings social tab


packages/api-page-builder/src/plugins/pageContentModifiers.ts
```diff
@@ -14,41 +36,31 @@ export default [
         type: "pb-page-element-modifier",
         elementType: "*",
         getStorageValue({ element }) {
-            const file = get(element, "data.settings.background.image.file.id");
-            if (file) {
-                set(element, "data.settings.background.image.file", file);
-            }
+            const file = element?.data?.settings?.background?.image?.file;
+            element.data.settings.background.image.file = createFileForDatabase(file);
```
Pavel910, line 40: @brunozoric line 39 uses optional chaining, but line 40 assigns directly to the deeply nested key. Maybe we should use `set` to assign file value?

doitadrian:
P.S. Don't forget to test the PB installation. 🙂  @brunozoric

doitadrian:
BTW I see the installation isn't mentioned here, but I was sure I did mention it.

So I went searching, and this is basically a duplicate issue, but this one mentions the installation 🙂 

https://github.com/webiny/webiny-js/issues/784

doitadrian:
@brunozoric @Pavel910 

Once we're done with the basic setup, we should also finally move the installation files into our public S3 bucket and have our installation process pull files from it. We don't need to bother users with uploading the ZIP in every new project.

I think we can just create a simple S3 bucket in our main AWS account, and host the files there. I remember @Pavel910 mentioning having a new folder for every version of Webiny, but at this point, since these files have NEVER changed, I think we can just drag and drop the ZIP, and get the URL.

Pavel910:
@doitadrian we'll do that as the very last step before v5 release. We also need to migrate to new Slate, which will change the structure of data, and we'll need to update installation files.

doitadrian:
Excellent @Pavel910, I'll write this as a new issue then.

PR #1245, opened 2020-09-16
fix: eslint warnings for v5

brunozoric:
## Related Issue
eslint warnings in v5- no defined issue

## Your solution
Unused variables were removed, imports were renamed and removed and some lines put under eslint disable (where really necessary).
For some function arguments a type was created with what it contains since it is not used but we want to know what is inside it.
There is some object destructuring when some property does not need (or must not) be used later on when passing that object and eslint is giving error about var not being used - here I set eslint-disable for that line. Maybe lodash omit can be used?

## How Has This Been Tested?
`yarn` -> `yarn build` -> `yarn webiny deploy api --env=local` -> `yarn start`


Ashu96 (commented):
Looks good to me :+1: 
Nicely done @brunozoric  :zap: 

I really liked the simplified `scripts` inside `package.json`. For example,
`"eslint:fix": "yarn eslint --fix",`

package.json
```diff
@@ -138,9 +138,9 @@
     "lint-staged": "lint-staged",
     "postinstall": "node ./scripts/linkPackages.js",
     "prepublishOnly": "node scripts/prepublishOnly.js",
-    "prettier": "prettier --config .prettierrc.js --write",
-    "prettier:check": "prettier \"**/**/*.{js,jsx,ts,tsx}\" --check --config .prettierrc.js",
-    "prettier:fix": "prettier \"**/**/*.{js,jsx,ts,tsx}\" --config .prettierrc.js --write",
+    "prettier": "prettier \"**/**/*.{js,jsx,ts,tsx}\" --config .prettierrc.js",
+    "prettier:check": "yarn prettier --check",
+    "prettier:fix": "yarn prettier --write",
```
doitadrian, line 143: Looks good, please just use the command here as well:
image

packages/project-utils/bundling/ssr/buildAppSSR.js
```diff
@@ -3,6 +3,7 @@ const fs = require("fs-extra");
 
 module.exports.buildAppSSR = async (
     { app = null, indexHtmlPath = null, createAppPath = null, output, ...options },
+    // eslint-disable-next-line
```
doitadrian, line 6: I think we can remove this line / the unused `context` below. I don't see it being used anywhere.

packages/project-utils/bundling/ssr/buildAppSSRFromSource.js
```diff
@@ -5,6 +5,7 @@ const { setupOutput } = require("./utils");
 
 module.exports.buildAppSSRFromSource = async (
     { handler = null, renderer = null, output = null, ...options },
+    // eslint-disable-next-line
```
doitadrian, line 8: Samo as `packages/project-utils/bundling/ssr/buildAppSSR.js`.

doitadrian (changes requested):
Very nice cleanup 👏 

Just check the three super-tiny comments, and we merge this.

doitadrian (approved):

doitadrian:
Awesome @brunozoric 👏

PR #1207, opened 2020-09-01
Feat: Menu element plugin package for page builder

TommyJackson85:
## Related Issue

Closes #1176

## Your solution

Creating a menu plugin package used for the app page builder package.

Creating element plugin within the app page builder.

## How Has This Been Tested?
In progess.

## Screenshots (if relevant):
None yet.

PR #1246, opened 2020-09-16
FEAT: Menu element plugin for page builder [SECOND PR]

### LINKED ISSUE #1176: Menu element plugin for Page Builder
## This is:
- Feature request

## Detailed Description
We want to create a new element plugin for Page Builder, which will allow us to use Menus, defined in the Menus module, as part of the page content. 

image

Once the user creates a new menu, he should be able to go to the page editor, and add a menu element, select which menu he wants to render, and select a component to render the menu.

To implement this, create a new package in the repository called `app-plugin-page-builder-menu-element`. Test your package by adding the plugin to the `admin` app. Also, add the render plugins to the `site` app and test rendering on the site.

## References
The following references contain everything you need to implement this feature. To try out similar plugins, try adding a Form or a Pages List element to your page.

*MUST READ*: A tutorial on creating Page Builder element plugins:
https://docs.webiny.com/docs/guides/creating-iframe-element-plugin

A very similar plugin for Mailchimp: 
https://github.com/webiny/webiny-js/tree/master/packages/app-mailchimp

An existing implementation of a very similar plugin to embed forms into the page editor:
https://github.com/webiny/webiny-js/tree/master/packages/app-form-builder/src/page-builder

An example of `PagesList` element and how it loads different renderers via plugins:
https://github.com/webiny/webiny-js/blob/master/packages/app-page-builder/src/render/plugins/elements/pagesList/index.tsx

A reference to `PagesList` element plugin for the editor:
https://github.com/webiny/webiny-js/tree/master/packages/app-page-builder/src/editor/plugins/elements/pagesList

TommyJackson85:
## Related Issue

Closes #1176

## Your solution

THIS IS A NEW PR IN REPLACEMENT TO [PREVIOUS PR #1207](https://github.com/webiny/webiny-js/pull/1207) DUE TO A SEVERE BUG.

Creating a menu plugin package used for the app page builder package.

Creating element plugin within the app page builder.

## How Has This Been Tested?
In progess.

## Screenshots (if relevant):
None yet.


TommyJackson85:
Previous PR had a bug which set me back on fixing this issue. This is a fresh PR with the most updated files of Webiny plus the relevant changes for a Menu element from the [previous PR's](https://github.com/webiny/webiny-js/pull/1207) commit e1e3776 (not all changes).
Previous PR is to be used for reference only.

TommyJackson85:
@Pavel910 What do you think of my work so far?
tutorial-1

Take a look at the these files too. These are the files that I have been working on. Let me know if anything I wrote is a little confusing.

***Editor files:***
packages/app-page-builder/src/editor/plugins/elements/menu/index.tsx
packages/app-page-builder/src/editor/plugins/elements/menu/Menu.tsx
packages/app-page-builder/src/editor/plugins/elements/menu/MenuDesignSettings.tsx
packages/app-page-builder/src/editor/plugins/elements/menu/graphql.ts

***Render files:***
packages/app-page-builder/src/render/plugins/elements/menu/index.tsx
packages/app-page-builder/src/render/plugins/elements/menu/Menu.tsx
packages/app-page-builder/src/render/plugins/elements/menu/components/MenuTemplate.tsx
packages/app-page-builder/src/render/plugins/elements/menu/graphql.ts

***types:***
packages/app-page-builder/src/types.ts

packages/app-page-builder/src/editor/plugins/elements/menu/Menu.tsx
```diff
@@ -0,0 +1,84 @@
...
+});
+
+export type MenuElementProps = {
+    isActive: boolean;
+    element: PbElement;
+};
+
+const Menu = (props: MenuElementProps) => {
+    //menu data passed through here
+    const { element, isActive } = props;
```
brunozoric, line 28: You can use {element, isActive} directly as function argument

packages/app-page-builder/src/editor/plugins/elements/menu/Menu.tsx
```diff
@@ -0,0 +1,84 @@
...
+    console.log("MENU PROPS ::::;;;;");
+    console.log(props);
+ 
+    const data = get(element, "data") || {};
```
brunozoric, line 32: Use optional instead of get if you are unsure that element actually is a object. So it looks like this: `const data = element?.data || {}`;
From MenuElementProps it is notable that element must be a PbElement and it has data property. So, basically, it would be enough to write `const data = element.data;`.

packages/app-page-builder/src/editor/plugins/elements/menu/Menu.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+    let menuId;
+   
+    if ('settings' in data) {
+        //SIMULAR TO:: const menu = get(element, "data.settings.menu") || {};
+        menuId = data.settings.menu.element;
+    } else {
+        return <div>Selected menu component not found!</div>;
+    }
+    console.log(`menu ID: ${menuId}`);
```
brunozoric, line 48: For this part, i would write it totally different since you do not care about menuId yet, eg:
```
if (!data?.settings) {
  return <div>Selected menu component not found!</div>
}
```

packages/app-page-builder/src/editor/plugins/elements/menu/Menu.tsx
```diff
@@ -0,0 +1,84 @@
...
+    let render = <span>Menu not selected.</span>;
+
+    if (menuId){
+        const props = {
+            preview: true,
+            menu: menuId,
+        };
+        render = <MenuComponent {...props} />;
+    }
+
```
brunozoric, line 65: Here, a render var would actually be a function, eg:
```
const render = (menuId?: string) => {
  if (!menuId) {
    return <span>Menu not selected.</span>;
  }
  return <MenuComponent menu={menuId} preview={true} />;
};
```
And then just pass the value that should be menuId:
```
render(data?.settings?.menu?.element);
```

packages/app-page-builder/src/render/plugins/elements/menu/components/MenuItems/findObject.ts
```diff
@@ -0,0 +1,26 @@
+import { findIndex, each } from "lodash";
+
+/**
+ * Recursively search for an object with given ID in the given source array.
+ */
+const findObject = (source: Array<any>, id: string): any => {
```
brunozoric, line 6: This is ok, but let's try to avoid lodash when possible :)
Here is how it would look with raw JS:
```
type FindObjectResult<T> = {
  source: T[],
  index: number,
  item: T,
};
const findObject = <T extends any>(source: T[], id: string): FindObjectResult<any> | null => {
  if (source.length === 0) {
    return null;
  }
  const index = source.findIndex(item => item.id === id);
  if (index >= 0) {
    return {source, index, item: source[index]};
  }
  return source.reduce((acc: FindObjectResult<T> | null, item: T) => {
    if (!item.children) {
      return acc;
    }
    return findObject(item.children, id);
  }, null);
};
```

Didn't run the fn but it should work :)

brunozoric (commented):
Looks good overall (didn't run it) and I left you some comments on what to improve and refactor.

packages/app-page-builder/src/editor/plugins/elements/menu/Menu.tsx
TommyJackson85: @brunozoric Im sure its an object, so I am just going to call get and remove ```|| {} ;```

TommyJackson85:
@brunozoric I basically copied the menu form from from the menu builder page on the webiny project (http://localhost:3001/page-builder/menus) but I am not sure if this is what we want to have displayed. 
Is the menu meant to be used for nav bars and side navs with it's menu items? Is there other uses for them?
This is whats displayed when rendering the page.

Screenshot 2020-10-11 at 21 47 38
Screenshot 2020-10-11 at 21 47 57

TommyJackson85:
@brunozoric I added a new option input for the user to choose a component to render the menu data. I also added a new render plugin which will be used for a Menu Navigation component with dropdown functions but its still being worked on. I intend to remove the Menu Form component option later as its not relevent. I wanted to show a gif display of how it the form currently looks but theres a separate bug from the master branch causing it to break, in which just posted about on the slack channel.

TommyJackson85:
@brunozoric Form now has a component input field to choose a component to render menu with. Still working on Menu Navigation which currently only displays a string as you can see below. I intend to remove Menu Form as I feel it is not relevant for this task but I am keeping it for testing for the time being.
tutorial-1

Pavel910:
v5 changed the whole editor API, so I'm closing this and we'll come back to this topic later.

PR #1286, opened 2020-10-05
fix: remove duplicate entries from gitignore

jbampton:


Pavel910:
Closing this as we no longer have those packages.

PR #1287, opened 2020-10-05
fix: typos

jbampton:


Pavel910:
These packages were refactored in v5, so this is no longer applicable. Closing.

PR #1316, opened 2020-10-20
fix: correct behavior of plugins retrieval functions

### LINKED ISSUE #1308: Fix PluginsContainer.byType method arg and return
## This is:
- Bug

## Specifications

  - version: all
  - OS: all
  - Browser: all

## Expected Behavior
`byType` method returns all plugins if type is undefined

## Actual Behavior
Method returns all plugins if no type is sent which is not a good thing to do.
It is called `byType` so returning all plugins when type is undefined is wrong.
Check in: https://github.com/webiny/webiny-js/blob/master/packages/plugins/src/PluginsContainer.ts#L48

## Possible Solution
`byType()` method should accept type argument that must be a string, cannot be undefined.
Introduce `all()` method that will return all plugins - return type should be `Plugin[]`.
While already working on this:
* add access modifiers to methods and to plugins attribute
* switch plugins property to Map (of type `Map<string, Plugin>`) and refactor PluginsContainer and assign function accordingly

## Note
Beware that changing byType behaviour there will be issues when someone wanted to get all the plugins with that method so check everything

jmakhack:
## Related Issue

Closes #1308 

## Your solution

Changed the `byType` method to no longer accept `undefined` as a valid type. Added the `all` method to `PluginsContainer` as the correct way to retrieve all plugins. 

## How Has This Been Tested?

I ran my changes against the existing tests, making sure that `plugins.test.js`, `modelFields.test.js`, `validators.test.js`, and `contentModelToSDL.test.js` passed in particular.

I also checked all uses of the `byType` method within the codebase to ensure that it was never being passed a value of `undefined` as a means for retrieving all plugins. As a result of this check, I updated the `getPlugins` function in `index.ts` to return all plugins when passed a value of `undefined`. This is expected behavior based on the tests found in `plugins.test.js`.

## Screenshots (if relevant):
N/A

### DIFF
--- a/packages/plugins/src/PluginsContainer.ts
+++ b/packages/plugins/src/PluginsContainer.ts
@@ -45,12 +45,12 @@ export class PluginsContainer {
         return this.plugins[name] as T;
     }
 
-    byType<T extends Plugin>(type?: string): T[] {
-        if (!type) {
-            return Object.values(this.plugins) as T[];
-        }
-
+    byType<T extends Plugin>(type: string): T[] {
         return Object.values(this.plugins).filter((pl: Plugin) => pl.type === type) as T[];
+    }
+
+    all<T extends Plugin>(): T[] {
+        return Object.values(this.plugins) as T[];
     }
 
     register(...args: any): void {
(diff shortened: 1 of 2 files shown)


brunozoric (approved):
Everything looks good :)

PR #1317, opened 2020-10-21
Wait for environment to load before mounting content form

Pavel910:
## Related Issue
When you open a content form to create a new entry and reload the browser, API calls are made to the wrong API endpoint because CMS environments are not yet loaded and the appropriate Apollo client instance is not yet set.

## Your solution
Wait for CMS Apollo client instance to be available before mounting the content form.

## How Has This Been Tested?
Manually, by reloading the browser and observing the network requests.


Pavel910:
This PR is no longer relevant as we no longer have environments.

PR #1330, opened 2020-11-08
fix(api-security): handle authentication errors

### LINKED ISSUE #1300: Invalid Authorization header causing errors
## This is:

- Bug

## Expected Behavior
I should just get auth error when sending this:
image

## Actual Behavior
```2020-10-10T06:49:57.856Z	b455b200-9a98-43ac-ab8a-1560d6758683	INFO	dobeo res {
  statusCode: 500,
  body: `{"error":{"name":"TypeError","message":"Cannot destructure property 'header' of '_jsonwebtoken.default.decode(...)' as it is null.","stack":"TypeError: Cannot destructure property 'header' of '_jsonwebtoken.default.decode(...)' as it is null.\\n    at Object.authenticate (/var/task/handler.js:59191:11)\\n    at async Object.apply (/var/task/handler.js:59065:24)\\n    at async Runtime.handler (/var/task/handler.js:59647:9)"}}`,
  headers: { 'Cache-Control': 'no-store', 'Content-Type': 'application/json' }
}
```

daniel-shuy:
## Related Issue

Closes #1300
See #1331 for the same fix for `v4`

## Your solution

Currently, uncaught errors in `SecurityAuthenticationPlugin`s will cause APIs to return HTTP `500`, and expose the stack trace to the client.
This PR adds a try-catch block to ignore errors in `SecurityAuthenticationPlugin`s, essentially treating them as a failed authentication.

### DIFF
--- a/packages/api-security/src/authenticator/index.ts
+++ b/packages/api-security/src/authenticator/index.ts
@@ -53,10 +53,16 @@ export default () => [
             );
 
             for (let i = 0; i < authenticationPlugins.length; i++) {
-                const identity = await authenticationPlugins[i].authenticate(context);
-                if (identity instanceof SecurityIdentity) {
-                    context.security.identity = identity;
-                    return;
+                try {
+                    const identity = await authenticationPlugins[i].authenticate(context);
+                    if (identity instanceof SecurityIdentity) {
+                        context.security.identity = identity;
+                        return;
+                    }
+                } catch (e) {
+                    // Authentication errors should not exposed to the client,
+                    // and should be treated as an authentication failure
+                    continue;
                 }
             }
         }


packages/api-security/src/authenticator/index.ts
```diff
@@ -53,10 +53,16 @@ export default () => [
...
+                try {
+                    const identity = await authenticationPlugins[i].authenticate(context);
+                    if (identity instanceof SecurityIdentity) {
+                        context.security.identity = identity;
+                        return;
+                    }
+                } catch (e) {
+                    // Authentication errors should not exposed to the client,
+                    // and should be treated as an authentication failure
+                    continue;
```
daniel-shuy, line 65: Question to maintainers: Should the authentication errors be logged?

daniel-shuy:
@Pavel910 could I at least get an explanation on why the PR is closed?

Pavel910:
@daniel-shuy sure; the response from GQL no longer returns 500 - ever; it's always 200, so your error will get propagated to the browser. You'll get a proper error in the `errors` array, as per GraphQL specs.

daniel-shuy:
@Pavel910 great! Thanks

Pavel910:
@daniel-shuy yeah, this isn't something that should be solved on security level, because any plugin can throw an error, not just authentication. Returning `500` will break response from Lambda, it simply won't be returned through API Gateway. So we must always return 200 and add errors to `errors` array, to be handled on the client.

Sorry for not commenting before, was cleaning many PRs and it seemed as an old one, wasn't even sure you're still following Webiny since we didn't merge it already.

Cheers 🍻

daniel-shuy:
@Pavel910 yup that's a much cleaner solution, I understand, no worries, cheers!

PR #1331, opened 2020-11-08
fix(api-security): handle authentication errors

### LINKED ISSUE #1300: Invalid Authorization header causing errors

daniel-shuy:
## Related Issue

Closes #1300
See #1330 for the same fix for `v5`

## Your solution

Currently, uncaught errors in `SecurityPlugin`s will cause APIs to return HTTP `500`, and expose the stack trace to the client.
This PR adds a try-catch block to ignore errors in `SecurityPlugin`s, essentially treating them as a failed authentication.

### DIFF
--- a/packages/api-security/src/plugins/security.ts
+++ b/packages/api-security/src/plugins/security.ts
@@ -19,7 +19,13 @@ export default (options: SecurityOptions) => [
 
             const securityPlugins = context.plugins.byType<SecurityPlugin>("graphql-security");
             for (let i = 0; i < securityPlugins.length; i++) {
-                await securityPlugins[i].authenticate(context);
+                try {
+                    await securityPlugins[i].authenticate(context);
+                } catch (e) {
+                    // Authentication errors should not exposed to the client,
+                    // and should be treated as an authentication failure
+                    continue;
+                }
             }
 
             if (options.public === false && !context.user) {


packages/api-security/src/plugins/security.ts
```diff
@@ -19,7 +19,13 @@ export default (options: SecurityOptions) => [
 
             const securityPlugins = context.plugins.byType<SecurityPlugin>("graphql-security");
             for (let i = 0; i < securityPlugins.length; i++) {
-                await securityPlugins[i].authenticate(context);
+                try {
+                    await securityPlugins[i].authenticate(context);
+                } catch (e) {
+                    // Authentication errors should not exposed to the client,
+                    // and should be treated as an authentication failure
+                    continue;
```
daniel-shuy, line 27: Question to maintainers: Should the authentication errors be logged?

daniel-shuy:
@Pavel910 could I at least get an explanation on why the PR is closed?

PR #1513, opened 2021-03-16
fix: store and reuse settings to/from a per tenant cache

Ashu96:
This PR fixes the redundant DB call on every file query.

## Your solution
Introduce a per tenant cache to store the settings.

## How Has This Been Tested?
Manually using logs.

## Screenshots (if relevant):
N/A


adrians5j (commented):
The question I have here...

Why is the `getSettings` function called on every request, and not called only when actually needed? Correct me if I'm wrong, but I think that was the initial issue we were trying to resolve here.

Ashu96:
> The question I have here...
> 
> Why is the `getSettings` function called on every request, and not called only when actually needed? Correct me if I'm wrong, but I think that was the initial issue we were trying to resolve here.

Thanks for reviewing the PR @doitadrian 

I've addressed the initial issue by chaning `getSettings` to load on-demand in FileStorage.

packages/api-file-manager/src/plugins/crud/settings.crud.ts
```diff
@@ -36,13 +36,22 @@ export default (context: FileManagerContext): SettingsCRUD => {
     const PK_SETTINGS = () => `T#${security.getTenant().id}#FM#SETTINGS`;
     const SK_SETTINGS = () => `default`;
 
+    const cache = {};
```
adrians5j: Let's remove this caching. I think it's really not that needed, and who knows if down the road could cause some issues / inconsistencies for the user (we don't have cache invalidation logic here).

packages/api-file-manager/src/plugins/context.ts
```diff
@@ -16,16 +16,13 @@ export default {
...
         const fileStoragePlugin = context.plugins.byName("api-file-manager-storage");
 
-        const settings = settingsCRUD(context);
-
         context.fileManager = {
             ...context.fileManager,
             files: filesCRUD(context),
             system: systemCRUD(context),
-            settings,
+            settings: settingsCRUD(context),
```
adrians5j, line 20: We are sending `context`, and we also use the `fileStorageUpload` constant to pass the plugin to the `FileStorage` class. 

If that's the case, let's just get the plugin within the `FileStorage` class.

adrians5j (changes requested):
Minor stuff, please take a look when you get a chance.

🍻

packages/api-file-manager/src/plugins/crud/settings.crud.ts
Ashu96: Sure. I'm always up for less code :)

adrians5j (approved):
Looks good!

PR #1533, opened 2021-03-25
fix(cli-plugin-scaffold-graphql-service): graphql types naming

brunozoric:
## Related Issue
The problem when using graphql scaffold multiple times is that CreatedByResponse, InstallResponse and UninstallResponse are not prefixed with the created model name (Target) so graphql breaks due to multiple definitions.

## Your solution
Prefix the responses with Target (model name).

## How Has This Been Tested?
Created Toy, Car and Beer APIs and deployed them. Ran the InstallMutation and UninstallMutation.

### DIFF
--- a/packages/cli-plugin-scaffold-graphql-service/template/src/graphql.ts
+++ b/packages/cli-plugin-scaffold-graphql-service/template/src/graphql.ts
@@ -119,11 +119,11 @@ const graphql = /* GraphQL */ `
 
         uninstall: TargetUninstallResponse!
     }
-    # we need to extend the original Query with targets so we can write a query like { targets: { listTargets {...} } }
+    # We need to extend the original Query with targets so we can write a query like { targets: { listTargets {...} } }
     extend type Query {
         targets: TargetQuery
     }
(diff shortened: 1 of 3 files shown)


packages/cli-plugin-scaffold-graphql-service/template/src/graphql.ts
```diff
@@ -115,9 +115,9 @@ const graphql = /* GraphQL */ `
 
         deleteTarget(id: ID!): TargetDeleteResponse!
 
-        install: InstallResponse!
+        install: TargetInstallResponse!
 
-        uninstall: UninstallResponse!
+        uninstall: TargetUninstallResponse!
     }
     # we need to extend the original Query with targets so we can write a query like { targets: { listTargets {...} } }
```
adrians5j: All comments start with upper case, except this one. Let's just have it more "unified", and start with "We" instead of "we".

adrians5j (approved):
Just check one super-minor comment, and feel free to merge this @brunozoric .

PR #1542, opened 2021-03-30
Fix: Rich text editor visual bug

Ashu96:
## Changes

This PR includes a couple of small visual bug fixes:

1. The context menu of the `rich text` editor is getting obscured if directly below is another `rich text` editor.
2. The `unpublish` revision message in CMS now has the missing revision version.
3. The File field renderer was missing some bottom spacing when there is another field below it. This PR contains a fix for that.

File field issue:
image

## How Has This Been Tested?

Manually using the admin app:

1. Rich text editor
image

2. File field renderer
image


packages/ui/src/ImageUpload/Image.tsx
```diff
@@ -25,11 +25,13 @@ type Props = {
     style?: { [key: string]: any };
     renderImagePreview?: (props: any) => React.ReactElement<any>;
     round?: boolean;
+    imageContainerStyle?: { [key: string]: any };
```
adrians5j: Let's use `React.CSSProperties` instead of `{ [key: string]: any };`.

Example:
```
const style: React.CSSProperties = {
    position: "relative",
    width: 200,
    margin: "0 auto"
};
```

packages/ui/src/ImageUpload/Image.tsx
```diff
@@ -109,9 +111,9 @@ class Image extends React.Component<Props> {
     }
 
     render() {
-        const { value, disabled } = this.props;
+        const { value, disabled, imageContainerStyle } = this.props;
```
adrians5j: Let's rename `imageContainerStyle` to just `containerStyle`.

adrians5j (changes requested):
Looking good pretty much. Just two minor details.

packages/ui/src/ImageUpload/Image.tsx
Ashu96: Nice!
Should we also change the `style` props as well to follow the same type instead of `{ [key: string]: any }`?

adrians5j: Let's do it.

PR #1543, opened 2021-03-30
Fix: UI improvements in new account form

Ashu96:
## Changes
This PR introduces the following two UI improvements in the new user account form:
- make the `group` accordion-item open by default so that the validation message is more accessible to user
- add `minLength:8` validator to password field

## How Has This Been Tested?

Manually using the admin app:

image

### DIFF
--- a/packages/app-plugin-security-cognito/src/identityProvider/cognitoPasswordValidator.ts
+++ b/packages/app-plugin-security-cognito/src/identityProvider/cognitoPasswordValidator.ts
@@ -25,7 +25,7 @@ export const createCognitoPasswordValidator = (policy: PasswordPolicy) => (value
     }
 
     if (policy.requireUppercase && !requireUppercase.test(value)) {
-        throw new ValidationError("Value must contain a uppercase character.");
+        throw new ValidationError("Value must contain an uppercase character.");
     }
 
     if (policy.requireSymbols && !requireSymbols.test(value)) {
--- a/packages/app-plugin-security-cognito/src/identityProvider/cognitoPasswordValidator.ts
+++ b/packages/app-plugin-security-cognito/src/identityProvider/cognitoPasswordValidator.ts
@@ -1,24 +1,7 @@
-import { validation, ValidationError } from "@webiny/validation";
+import { ValidationError } from "@webiny/validation";
 import { PasswordPolicy } from "../types";
 
-const COGNITO_PASSWORD_POLICY = "cognitoPasswordPolicy";
-
-export const cognitoPasswordValidator = (policy: PasswordPolicy): string => {
-    // Create custom validator
-    const cognitoPasswordValidator = createCognitoPasswordValidator(policy);
-
-    validation.setValidator(COGNITO_PASSWORD_POLICY, cognitoPasswordValidator);
-
-    let validator = COGNITO_PASSWORD_POLICY;
-
-    if (policy.minimumLength !== undefined) {
-        validator += `,minLength:${policy.minimumLength}`;
-    }
-
-    return validator;
-};
-
-const createCognitoPasswordValidator = (policy: PasswordPolicy) => (value: any): void => {
+export const createCognitoPasswordValidator = (policy: PasswordPolicy) => (value: any): void => {
     if (!value) {
         return;
     }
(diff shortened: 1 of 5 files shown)


packages/app-plugin-security-cognito/src/identityProvider/index.tsx
```diff
@@ -51,7 +51,11 @@ export default (): PluginCollection => [
                 <Cell span={12}>
                     <Bind
                         name="password"
-                        validators={data.createdOn ? undefined : validation.create("required")}
+                        validators={
+                            data.createdOn
+                                ? validation.create("minLength:8")
+                                : validation.create("required,minLength:8")
```
adrians5j: For this one, I'd like @Pavel910 to chime in.

I believe it's not smart to hardcode this here because this rule can change in the Cognito settings. Am I right?

Maybe we could just add a parameter here (`apps/admin/code/src/plugins/security.ts:27`):

image

Of course, if not specified, then use the `minLength:8` or whatever default Cognito setting is.

adrians5j (changes requested):
We need to discuss one detail here. Hoping @Pavel910 can provide more info here.

packages/app-plugin-security-cognito/src/identityProvider/index.tsx
Pavel910: This is what password policy can look like:
image

So we should support:
```
{
                minimumLength: 8,
                requireLowercase: false,
                requireNumbers: false,
                requireSymbols: false,
                requireUppercase: false,
 }
```
and generate validators based on that.

Ashu96: @Pavel910 I've updated the PR with the suggested upgrade. Please feel free to review.
Thanks :+1:

packages/app-plugin-security-cognito/src/identityProvider/cognitoPasswordValidator.ts
```diff
@@ -0,0 +1,49 @@
+import { validation, ValidationError } from "@webiny/validation";
+import { PasswordPolicy } from "../types";
+
+const COGNITO_PASSWORD_POLICY = "cognitoPasswordPolicy";
+
+export const cognitoPasswordValidator = (policy: PasswordPolicy): string => {
+    // Create custom validator
+    const cognitoPasswordValidator = createCognitoPasswordValidator(policy);
+
+    validation.setValidator(COGNITO_PASSWORD_POLICY, cognitoPasswordValidator);
```
Pavel910: I wouldn't do this at all. Nobody outside of this component will ever use this validator, so no need to register it.

packages/app-plugin-security-cognito/src/identityProvider/cognitoPasswordValidator.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    let validator = COGNITO_PASSWORD_POLICY;
+
+    if (policy.minimumLength !== undefined) {
+        validator += `,minLength:${policy.minimumLength}`;
+    }
+
```
Pavel910: The minLength can be included directly in the cognito assword validation function.

Pavel910 (changes requested):

packages/app-plugin-security-cognito/src/identityProvider/cognitoPasswordValidator.ts
```diff
@@ -0,0 +1,36 @@
...
+    if (policy.requireLowercase && !requireLowercase.test(value)) {
+        throw new ValidationError("Value must contain a lowercase character.");
+    }
+
+    if (policy.requireNumbers && !requireNumber.test(value)) {
+        throw new ValidationError("Value must contain a number.");
+    }
+
+    if (policy.requireUppercase && !requireUppercase.test(value)) {
+        throw new ValidationError("Value must contain a uppercase character.");
```
Pavel910: `an uppercase character`

packages/app-plugin-security-cognito/src/identityProvider/index.tsx
```diff
@@ -47,12 +59,19 @@ export default (): PluginCollection => [
     {
         type: "security-user-form",
         render({ Bind, data }) {
+            const passwordValidators = [];
+
+            if (options && options.passwordPolicy) {
```
Pavel910: What if I send a `{}` as `options` ? `defaultArgs` will not be applied and passwordPolicy will not be set.
Make sure you assign `passwordPolicy` into `options` if it's not set.

Pavel910 (changes requested):

Pavel910 (approved):

PR #1624, opened 2021-05-17
Pass the --debug argument to Pulumi CLI (webiny deploy and webiny destroy commands)

### LINKED ISSUE #1619: Pass the `--debug` argument to Pulumi CLI (`webiny deploy` and `webiny destroy` commands)
## This is:

- Bug

## The Issue
When users run the `yarn webiny deploy {folder} --env {env} --debug` or `yarn webiny destroy {folder} --env {env} --debug` commands, the specified `--debug` flag isn't actually passed to the underlying Pulumi CLI command that we run. At the moment, this flag is only passed to the [build-packages](https://github.com/webiny/webiny-js/blob/next/packages/cli-plugin-deploy-pulumi/commands/deploy.js#L49) command.

As an escape hatch, if users wanted to get the Pulumi debug logs, they would have to run the `webiny pulumi` command, for example:

```bash
yarn webiny pulumi api --env dev -- up --yes --debug
```

Definitely not ideal.

#### Why do we care about this?
Often, passing this flag to the Pulumi CLI can reveal useful error-related information to the user in his terminal.

## Possible Solution
Let's make sure this flag is passed to the Pulumi CLI as well.

There are a couple of places in which we must make amends.

#### Deployment previews
Found [here](https://github.com/webiny/webiny-js/blob/next/packages/cli-plugin-deploy-pulumi/commands/deploy.js#L110).

We must add the `args` property like so:

```ts
args: {
  debug: inputs.debug
}
```

#### Deployments
Found [here](https://github.com/webiny/webiny-js/blob/next/packages/cli-plugin-deploy-pulumi/commands/deploy.js#L123). We just need to add the missing `debug: inputs.debug` into the `args` property.

#### Destroys
Found [here](https://github.com/webiny/webiny-js/blob/next/packages/cli-plugin-deploy-pulumi/commands/destroy.js#L50).

Like in the first **Deployment previews** place, we must add the `args` property like so:

```ts
args: {
  debug: inputs.debug
}
```

ankurvr:
## Changes

Closes #1619 

## How Has This Been Tested?
I fixed the issue as described in the issue.

### DIFF
--- a/packages/app-security-tenancy/src/plugins/installation.tsx
+++ b/packages/app-security-tenancy/src/plugins/installation.tsx
@@ -102,12 +102,18 @@ const Install = ({ onInstalled }) => {
                             )}
                             <Cell span={12}>Let&apos;s create your admin user:</Cell>
                             <Cell span={12}>
-                                <Bind name="firstName" validators={validation.create("required,minLength:2")}>
+                                <Bind
+                                    name="firstName"
+                                    validators={validation.create("required,minLength:2")}
+                                >
                                     <Input label={t`First Name`} />
                                 </Bind>
                             </Cell>
                             <Cell span={12}>
-                                <Bind name="lastName" validators={validation.create("required,minLength:2")}>
+                                <Bind
+                                    name="lastName"
+                                    validators={validation.create("required,minLength:2")}
+                                >
                                     <Input label={t`Last Name`} />
                                 </Bind>
                             </Cell>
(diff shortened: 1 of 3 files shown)


adrians5j:
Looks good, although there are some [prettier issues](https://github.com/webiny/webiny-js/pull/1624/checks?check_run_id=2599386712), let's check those before merging.

ankurvr:
Good to go for this.

adrians5j:
Awesome @ankurvr !

Let's merge this into the remaining PR, and we're good to go!

PR #1634, opened 2021-05-21
fix: tooltip hover type of file

### LINKED ISSUE #1604: File Manager - Incorrect tooltip in file details sidebar
## This is:
- Bug

## Expected Behavior
In the case of a file, the tooltip should show `Delete File`:

File manager: delete file action

## Actual Behavior
In the case of a file, the tooltip is showing `Delete Image`:

File manager: delete image action

1. Visit the admin app and navigate to the `File Manager` view by clicking the link in the drawer:
Screenshot 2021-05-10 at 5 57 52 PM
 
2. In the file manager view, upload at least one image and a non-image file if there isn't already.
3. Click on the gear/clog icon to open up the file details sidebar and hover on the delete icon to activate the tooltip.

## Detailed Description
We need to make the delete action tooltip's content dynamic with respect to the file type. If the file is an image (based on type/extension) use `Delete Image` otherwise use `Delete File`.

The code responsible for rendering delete action in file detail sidebar is located at: https://github.com/webiny/webiny-js/blob/next/packages/app-admin/src/components/FileManager/FileDetails.tsx#L267

## Action Items:
- [ ] Update tooltip's content based on file type (image or non-image) for delete action in file details sidebar

imrishabh18:
## Changes

Closes #1604 

## How Has This Been Tested?

Hovered over different files and tested.

## Documentation

This checks if the file is of any of the type jpeg, png, gif or jpg then the hover shows `Delete image` otherwise `Delete file`

### DIFF
--- a/packages/app-admin/src/components/FileManager/FileDetails.tsx
+++ b/packages/app-admin/src/components/FileManager/FileDetails.tsx
@@ -148,18 +148,19 @@ type FileDetailsProps = {
     };
     [key: string]: any;
 };
+
+const isImage = file => {
+    const fileType = mime.getType(file && file.name);
+
+    if (fileType && typeof fileType === "string") {
+        return fileType.includes("image");
+    }
+
+    return false;
+};
+
 export default function FileDetails(props: FileDetailsProps) {
     const { file, uploadFile, validateFiles } = props;
-
-    const isImage = file => {
-        const fileType = mime.getType(file && file.name);
-
-        if (fileType && typeof fileType === "string") {
-            return fileType.includes("image");
-        }
-
-        return false;
-    };
 
     const filePlugin = getFileTypePlugin(file);
     const actions = get(filePlugin, "fileDetails.actions") || [];
--- a/packages/app-admin/src/components/FileManager/FileDetails.tsx
+++ b/packages/app-admin/src/components/FileManager/FileDetails.tsx
@@ -28,6 +28,7 @@ import { useSecurity } from "@webiny/app-security";
 import { ConfirmationDialog } from "@webiny/ui/ConfirmationDialog";
 import { DELETE_FILE, LIST_FILES, LIST_TAGS } from "./graphql";
 import { i18n } from "@webiny/app/i18n";
+import mime from "mime";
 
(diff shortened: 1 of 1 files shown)


Ashu96:
Thanks for the PR @imrishabh18 

Looks like this solution doesn't check for all image types. For example, `SVG` image won't  pass the `isImage` test.
Maybe we can use the `mime` npm package here to do more robust checking, something like this:
```js
const isImage = file => {
    const fileType = mime.getType(file && file.name);

    if (fileType && typeof fileType === "string") {
        return fileType.includes("image");
    }

    return false;
};
```

Also, please check the prettier issue :)
Thanks 👍

imrishabh18:
Is this good? @Ashu96 @Pavel910

packages/app-admin/src/components/FileManager/FileDetails.tsx
```diff
@@ -149,7 +150,17 @@ type FileDetailsProps = {
 };
 export default function FileDetails(props: FileDetailsProps) {
     const { file, uploadFile, validateFiles } = props;
-    const isImage = file.name.match(/.(jpg|jpeg|png|gif)$/i)
+
+    const isImage = file => {
```
Ashu96: Good work @imrishabh18 👍 

I would just move this function outside of React component because it doesn't depend upon it.
Rest looks good.

Ashu96 (changes requested):
Looks good, just need a minor change.

packages/app-admin/src/components/FileManager/FileDetails.tsx
imrishabh18: Okay, Can you please tell how to figure out which functions to put outside of the functional component and when to define it inside the component? It's confusing to me. As I define all my functions inside the component 😅

Pavel910: If a function can be standalone, meaning - it doesn't need anything but its parameters to work, you can move it outside of a component. That way, React component will be always calling the same function by reference and won't be re-creating the function on each rerender.

Ashu96 (approved):
Nice work @imrishabh18 👍 
Will merge it soon.

PR #1712, opened 2021-06-11
Fix: I18N update locale

### LINKED ISSUE #1694: I18N: Unable to update an existing locale
## This is:
- Bug

## Detailed Description
When I try setting an existing locale as default, I received the following error:

Screenshot_2021-05-27_at_11 29 18_AM

Looking at the network request, it seems `create` request is triggered instead of `update`:

```js
[{
    "operationName": "createI18NLocale",
    "variables": {
        "data": {
            "code": "en-US",
            "default": true,
            "createdOn": null
        }
    },
    "query": "mutation createI18NLocale($data: I18NLocaleInput!) {\n  i18n {\n    locale: createI18NLocale(data: $data) {\n      data {\n        code\n        default\n        createdOn\n        __typename\n      }\n      error {\n        code\n        message\n        data\n        __typename\n      }\n      __typename\n    }\n    __typename\n  }\n}\n"
}]
```

We use `createdOn` to differentiate between the `create` and `update` operation: `packages/app-i18n/src/admin/views/I18NLocales/I18NLocalesForm.tsx:64`

But, `createdOn` is not being saved currently:

Screenshot_2021-05-27_at_11 25 08_AM

## Action Items:

- [ ] add `createdOn` and  `createdBy` fields on the locale record
- [ ] write integration test for the same
- [ ] write cypress test for the same

### LINKED ISSUE #1711: I18N: Deleting selected content locale breaks the admin app
## This is:
- Bug

## Specifications
  - version: 5.8.0

## Expected Behavior
Should delete locale and select the `"default"` locale in locale selector.

## Actual Behavior
 Deleting selected content locale breaks the admin app:

https://user-images.githubusercontent.com/13612227/121664426-17b46b00-cac5-11eb-9b58-d79a6d9d0b0a.mov

Ashu96:
## Changes
Closes #1694
Closes #1711 

## How Has This Been Tested?
Manually and jest test

## Screenshots

https://user-images.githubusercontent.com/13612227/121665128-d5d7f480-cac5-11eb-9296-9e2719a0e044.mov


packages/app-i18n/src/admin/views/locales/hooks/useLocalesList.ts
```diff
@@ -104,6 +104,12 @@ export const useLocalesList: UseLocalesListHook = (config: Config) => {
 
                 showSnackbar(t`Locale "{code}" deleted.`({ code: item.code }));
 
+                if (getCurrentLocale("content") === item.code) {
```
adrians5j, line 107: Nice.

packages/api-i18n/src/graphql/crud/locales.crud.ts
```diff
@@ -40,13 +40,21 @@ export default (context: Context<DbContext, TenancyContext, I18NContext>) => {
...
                 data: {
                     PK: PK_LOCALE(),
                     SK: data.code,
                     code: data.code,
-                    default: data.default
+                    default: data.default,
+                    createdOn: new Date().toISOString(),
+                    createdBy: {
+                        id: identity.id,
+                        displayName: identity.displayName,
```
adrians5j, line 55: Nice.

adrians5j: BTW where is the code that sets the previous default locale as a non-default one? Or maybe, where is the code that sets this one as the default one? 

🤔

adrians5j: I just saw it's in the GQL resolver function. If you'll have the time, maybe we could migrate those updateDefault calls there into this file (like all other apps have it).

packages/api-i18n/src/graphql/graphql/locales.ts
```diff
@@ -26,7 +27,10 @@ const plugin: GraphQLSchemaPlugin<I18NContext & SecurityContext> = {
             input I18NLocaleInput {
                 code: String
                 default: Boolean
-                createdOn: DateTime
```
adrians5j, line 29: Nice.

packages/api-i18n/src/graphql/graphql/locales.ts
```diff
@@ -124,12 +128,16 @@ const plugin: GraphQLSchemaPlugin<I18NContext & SecurityContext> = {
                     if (data.default) {
                         await i18n.locales.updateDefault(data.code);
                     }
-
-                    return new Response(data);
+                    const locale = await i18n.locales.getByCode(data.code);
```
adrians5j, line 131: If we'd migrate the logic into crud file, we'd probably be able to avoid this extra DB call.

packages/api-i18n/src/graphql/graphql/locales.ts
```diff
@@ -142,15 +150,20 @@ const plugin: GraphQLSchemaPlugin<I18NContext & SecurityContext> = {
...
 
-                    if (locale.default) {
+                    await i18n.locales.update(code, data);
+
+                    if (data.default) {
                         await i18n.locales.updateDefault(code);
                     }
-
-                    return new Response(locale);
+                    const updatedLocale = await i18n.locales.getByCode(code);
```
adrians5j, line 165: If we'd migrate the logic into crud file, we'd probably be able to avoid this extra DB call.

adrians5j (approved):
Looks good, just a couple of comments. But it's pretty much good to go.

What do you think about mentioned code migrations/refactor? You have an extra hour-two to maybe try get this in order?

Ashu96:
Thanks for the feedback @doitadrian  👍 

> What do you think about mentioned code migrations/refactor? You have an extra hour-two to maybe try get this in order?

I would love to do the suggested changes. But, I'm not sure if we can make it in this release (due to bandwidth). Will try 💯

PR #1735, opened 2021-06-24
Fix: file manager API delete file

Ashu96:
Now deleting a file via File Manager will also delete it from storage i.e S3

Changes:
- Update tests to use `until` instead of `sleep`


Pavel910:
@brunozoric since you're also working on File Manager, please take a look at this PR, and transfer what's necessary into your work.

Pavel910 (approved):

PR #1738, opened 2021-06-26
Update Lambda runtime and sharp layer

### LINKED ISSUE #1684: Deprecated AWS Lambda runtime node.js 10
## This is:

- Idea

AWS lambda will stop supporting runtime Node.js 10.x at the end of July 2021. Currently webiny creates lambda function fm-image-transformer-xxxxx that uses this version.

https://docs.aws.amazon.com/lambda/latest/dg/runtime-support-policy.html

Pavel910:
## Changes
Closes #1684 

Also added an upgrade script to check all Lambda functions in Pulumi configuration and update them to use `nodejs12.x` runtime and the new `sharp` layer.

## How Has This Been Tested?
Deployed a project using the new setup and used FileManager to upload and resize several images. Everything seems to be working as it should.


scripts/layers/layers.json
```diff
@@ -16,5 +16,41 @@
...
+    "eu-west-2": "arn:aws:lambda:eu-west-2:764866452798:layer:chrome-aws-lambda:20",
+    "eu-west-3": "arn:aws:lambda:eu-west-3:764866452798:layer:chrome-aws-lambda:20",
+    "sa-east-1": "arn:aws:lambda:sa-east-1:764866452798:layer:chrome-aws-lambda:20",
+    "us-east-1": "arn:aws:lambda:us-east-1:764866452798:layer:chrome-aws-lambda:22",
+    "us-east-2": "arn:aws:lambda:us-east-2:764866452798:layer:chrome-aws-lambda:20",
+    "us-west-1": "arn:aws:lambda:us-west-1:764866452798:layer:chrome-aws-lambda:20",
+    "us-west-2": "arn:aws:lambda:us-west-2:764866452798:layer:chrome-aws-lambda:20"
+  },
+  "sharp": {
+    "us-east-1": "arn:aws:lambda:us-east-1:632417926021:layer:sharp:10",
```
ianvonholt, line 39: Should this be layer `10` or `1`?

Pavel910: `10` is correct. The thing with layers is that they preserve the version number for a specific layer name.
So if we used `sharp` in the past, in that region, and we published 9 versions, then deleted the layer, and then created a new layer with the same name, it will be version 10, not version 1. Because layers are identified by name, not some random ID.

For Webiny, consider layer version an internal thing, and just use that `getLayerArn` utility to get the latest layer ARN at deploy time. Those versions are published as an npm package and you'll always be using the latest tested layer.

PR #1744, opened 2021-06-30
feat(api-headless-cms): add compression to long-text cms field

Ashu96:
This PR introduces the compression feature for `"long-text"` cms field.

Note: This feature will compression all `"long-text"` fields i.e users will not have to selectively choose to compression some fields while leaving others as it is.


brunozoric (approved):
I tested the compression/decompression of the text on node versions 12, 14 and 15. There are no differences. I think that error which was happening was due to compressed string was created on Mac and CI is running on linux. Hence the differences between compressed strings. Everything seems fine to me here.

brunozoric:
One more thing that should be tested is deployment of this plugin to some existing system.

PR #1805, opened 2021-07-19
feat: add AWS region + project name in webiny info command output

### LINKED ISSUE #1649: Update Webiny info command
## This is:
- Improvement

## Detailed Description:
At the moment when we run the Webiny info command using:
```
yarn webiny info --env=<ENV>
```
we only display the URLs and the environment name.
With this task, we want to expand that and add `AWS region` and `project name` to the list.

## Possible Solution:

We need to make the necessary changes in the [packages/cwp-template-aws/cli/info/index.js](https://github.com/webiny/webiny-js/blob/next/packages/cwp-template-aws/cli/info/index.js#L6) file.

Inside `printEnvOutput` function, we can do the following:
- get the AWS region from the output of `await getStackOutput("api", env)` call, for example, `output.region`
- get the project name from the `context` argument, for example, `context.project.name`

After that, we can simply log them after `Environment`.

--- 
**Here is how it should look:**

Screenshot 2021-05-25 at 10 34 45 AM

NomanGul:
## Changes
Closes #1649 

when not deployed
image

when deployed
image

### DIFF
--- a/packages/cwp-template-aws/cli/info/index.js
+++ b/packages/cwp-template-aws/cli/info/index.js
@@ -4,12 +4,19 @@ const { green } = require("chalk");
 const line = `-------------------------`;
 
 const printEnvOutput = async (env, context) => {
+    let output = await getStackOutput("api", env);
+
     console.log(line);
-    console.log(`Environment: ${green(env)}`);
+    console.log(
+        [
+            `Environment: ${green(env)}`,
+            `AWS region: ${green(output.region ?? "not deployed yet")}`,
+            `Project name: ${green(context.project.name)}`
+        ].join("\n")
+    );
     console.log(line);
 
     let stacksDeployedCount = 0;
-    let output = await getStackOutput("api", env);
     if (output) {
         stacksDeployedCount++;
         console.log(


adrians5j:
Looks good.. just thinking about the green color for "not deployed yet".

Not sure if gray or just a regular color would be better here? Any thoughts maybe?

Also, the `undefined` in Headless CMS URLs is also weird. Just noticed this.

Maybe we can tackle this in a separate PR? 🤔

packages/cwp-template-aws/cli/info/index.js
```diff
@@ -4,12 +4,19 @@ const { green } = require("chalk");
...
+            `Environment: ${green(env)}`,
+            `AWS region: ${green(output.region ?? "not deployed yet")}`,
+            `Project name: ${green(context.project.name)}`
+        ].join("\n")
+    );
     console.log(line);
 
     let stacksDeployedCount = 0;
-    let output = await getStackOutput("api", env);
     if (output) {
```
NomanGul, line 20: @doitadrian Headless CMS URLs are showing undefined because we are checking for output that is an empty object and then logging URLs. and an empty object evaluates to true in JS if statement.

we can use lodash isEmpty method to fix it.

NomanGul:
@doitadrian gray color contrast is not clearly visible and has low accessibility.

image

The regular color is looking fine.

image

What do you think?

packages/cwp-template-aws/cli/info/index.js
adrians5j: I'd just go with a label like "Not Available" when the CDN URL is not defined. I think that would make sense. We can try and see how it looks maybe.

Sorry for delay BTW! 🙁

SyedUmerHasan (approved):

adrians5j:
Closing this due to inactivity.

PR #1877, opened 2021-08-24
fix(cli-plugin-scaffold-ci): repositories fetched limit

### LINKED ISSUE #1785: CI Setup only gives you access to your first page of repositories
In the CI/CD scaffolding script where if the github account selected has enough repositories where it starts paginating them, only the first page of repositories, sorted alphabetically, are available as options to the CI/CD script.  We got past this by changing the repo to start with an “a”, but this seems unintended.  The issue is in the results - what you are calling looks to be paginated, but it looks like it's not handling the results in that way.

## This is:
- Bug
OR 
- Feature request

## Specifications

  - version: v5.10.0
  - OS: Mac OS 11.4
  - Browser: Terminal

## Expected Behavior

I should see all of my repositories that are visible, not just the first page (30 or so)

## Actual Behavior

I only get access to using my first page of repos, listed alphabetically

1. Go load yourself up with more than 30 active repos
2. Then when you go to setup the CI process and it needs to attach to a repo, it's not there.

## Detailed Description

There's a place where the page size needs to be set to something a bit more reasonable - but really it should give you access to ALL of them.

## Possible Solution

https://github.com/webiny/webiny-js/blob/next/packages/cli-plugin-scaffold-ci/src/githubActions/index.ts#L143

Change this to iterate over ALL repos to build the list - instead of any finite number.  Pointer from @Pavel910:

https://octokit.github.io/rest.js/v18#repos-list-for-authenticated-user

IzioDev:
## Changes

This change the behavior of repositories fetching, it does not limit to a certain number anymore but instead, browse all repositories the user has.

Closes #1785

## How Has This Been Tested?

Yes, manually, I can now access all my repository in the ci scaffold

### DIFF
--- a/packages/cli-plugin-scaffold-ci/src/githubActions/fetchAllRepositories.ts
+++ b/packages/cli-plugin-scaffold-ci/src/githubActions/fetchAllRepositories.ts
@@ -1,0 +1,35 @@
+import { Octokit } from "octokit";
+
+// TODO: try to add octokit types properly based on
+// https://www.npmjs.com/package/@octokit/types
+type GithubRepository = {
+    full_name: string;
+    name: string;
+    owner: { login: string };
+};
+
+export default async (args: { octokit: Octokit }): Promise<GithubRepository[]> => {
+    const octokit = args.octokit;
+
+    let page = 0;
+    const results: GithubRepository[] = [];
+
+    while (true) {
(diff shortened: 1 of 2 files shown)


adrians5j (approved):
Looks good @IzioDev !

PR #1889, opened 2021-08-30
feat(app-page-builder): support multiple pd-plugin-loader extension

etamity:
## Changes

Currently, app-page-builder only supports a single `pd-plugins-loader`, this pr to makes it support multiple `pd-plugins-loader`, which mean, you can extend app-page-builder without modifying app-page-builder source code.
## How Has This Been Tested?
Manually test.

create new `.ts` file and copy paste code below:
```ts
import { plugins } from "@webiny/plugins";
const extenstionPlugin =  {
        type: "pb-plugins-loader",
        async loadEditorPlugins() {
            console.log("loadEditorPlugins tekclaw");
            // return (await import("./plugins/editorPlugins")).default;
        },
        async loadRenderPlugins() {
            console.log("loadRenderPlugins tekclaw");
           // return (await import("./plugins/renderPlugins")).default;
        }
    };
plugins.register(extenstionPlugin);
``` 
When you reload page-builder, you should be able to see the logs.

### DIFF
--- a/packages/app-page-builder/src/admin/components/EditorPluginsLoader.tsx
+++ b/packages/app-page-builder/src/admin/components/EditorPluginsLoader.tsx
@@ -14,7 +14,19 @@ export function EditorPluginsLoader({ children, location }) {
     );
 
     async function loadPlugins() {
-        const [{ loadEditorPlugins, loadRenderPlugins }] = plugins.byType("pb-plugins-loader");
+        const pbPlugins = plugins.byType("pb-plugins-loader");
+        const loadEditorPlugins = async () =>
+            await Promise.all(
+                pbPlugins
+                    .map(plugin => plugin.loadEditorPlugins && plugin.loadEditorPlugins())
+                    .filter(Boolean)
+            );
+        const loadRenderPlugins = async () =>
+            await Promise.all(
+                pbPlugins
+                    .map(plugin => plugin.loadEditorPlugins && plugin.loadRenderPlugins())
+                    .filter(Boolean)
+            );
 
         // If we are on pages list route, import plugins required to render the page content.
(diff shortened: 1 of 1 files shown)


Pavel910:
@etamity the reason we were only using 1 plugin is because that plugin is defined within user's project, not within app-page-builder. [This location ](https://github.com/webiny/webiny-js/blob/next/apps/admin/code/src/plugins/pageBuilder.ts#L22-L31) can be freely modified by developers. That part of code exists to enable code-splitting of big chunks of plugins. If you just want to add more plugins to page builder, you would add them into  `./plugins/editorPlugins` and `./plugins/renderPlugins`.

Is there a specific use case you're unable to achieve with the existing implementation and a single plugin?

etamity:
@Pavel910 I am creating a new npm package that contains a bundle of plugins, e.g, I created bundle plugins that support bootstrap and custom HTML code (in feature will support even more feature plugins). so other developers can just install my plugin, and simple add:
```ts
import { plugins } from "@webiny/plugins";
import extensionPlugins from "@extension/elements";
plugins.register(extensionPlugins);
```

Here is the demo:
https://www.youtube.com/watch?v=1gDsgeGxjD0

Pavel910:
Brilliant! Great work 🚀 

I see, you don't want to bother importing individual plugins into existing files. That makes sense. Thanks for clarifying 👍

etamity:
@Pavel910 It seems failed at the `Publish to Verdaccio` stage, do I need to push the branch in this repo instead of a fork?

Pavel910:
@etamity we'll check this out and get back to you. Forks should work fine, it's something in our CI.

Pavel910:
@etamity with everything going on, I forgot about this PR. You should've poked me about it earlier :(
It's merged now! Thanks for the contribution!

PR #1902, opened 2021-09-08
fix: problem with richtexteditor RTL support

### LINKED ISSUE #1892: Rich Text elements that can have their alignment set do not render correctly using Rich Text Renderer when the direction property is set to `rtl` on the HTML tag.
### Version

5.14

### Operating System

_No response_

### Browser

_No response_

1. create an article in an RTL language
2. pass the article content to the Rich Text Renderer component.
3. add the dir="rtl" property to the HTML tag 
4. see the content remain stuck with ltr alignment.

Text alignment should 'just work' when using the dir tag.

### What do you see instead?

_No response_

### Additional information

_No response_

### Possible solution

If the rich text rendered uses "start" and "end" instead of "left" and "right" for the textAlign property, it works as expected, but this isn't the right place for the update - it should be made in the editor, I think. There may be consequences to that I am unaware of however!

jimjoes:
## Changes
RTE now stores 'start' and 'end' instead of 'left' and 'right' in the textAlign property for paragraphs.

As this property is used directly by the richtext renderer package (which applies it inline in the JSX as a textAlign property) the browser is now able to correctly align paragraphs for RTL languages because it interprets start/end according to the html tag's property direction (which is ltr by default).

This PR also includes accordingly updated CSS for the editor.

Fixes #1892

## How Has This Been Tested?
Checked that the textAlign property of paragraphs is now start/end. Also checked (using the webiny-richtext-renderer package) that existing paragraphs with left/right specified still render correctly and that changing the HTML direction property makes new paragraphs (i.e. ones using start/end) correctly changes the alignment of paragraphs.

### DIFF
--- a/packages/app-admin/src/components/RichTextEditor/tools/header/index.ts
+++ b/packages/app-admin/src/components/RichTextEditor/tools/header/index.ts
@@ -117,7 +117,7 @@ class Header {
 
         newData.text = data.text || "";
         newData.level = parseInt(data.level) || this.defaultLevel.number;
-        newData.textAlign = data.textAlign || TextAlign.LEFT;
+        newData.textAlign = data.textAlign || TextAlign.START;
 
         return newData;
     }
@@ -329,7 +329,7 @@ class Header {
      * @returns {TextAlign} textAlign
      */
     getTextAlign(className) {
-        let textAlign = TextAlign.LEFT;
+        let textAlign = TextAlign.START;
         // Match className with alignment
         this.alignments.forEach(alignment => {
             if (className.includes(`ce-header-text--${alignment.name}`)) {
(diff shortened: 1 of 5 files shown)


Pavel910:
@econtentmaps Thanks for the PR!

What about users who already have  content with left/right classes? This should be backwards compatible. We need to at least keep the old CSS classes.

Also, check the CI log, you have some TS errors there.

jimjoes:
No content-related classes have been changed. Existing content is still aligned as it was, because the rich text renderer simply applies the string it finds in the paragraph content block to an inline style in the JSX directly (so 'left' or 'start'). The style will still be applied correctly whether it's left/right or start/end. TS errors fixed - missed a couple of references to 'left' in the header.ts file, which are now corrected. Some CSS in the editor was updated, but that doesn't impact content, unless I've missed something?

PR #1926, opened 2021-09-23
fix: disable help processing until after plugins are imported

### LINKED ISSUE #1925: --help does not work
### Version

5.14.0

### Operating System

Mac

### Browser

Firefox

yarn webiny --help

The full help output

### What do you see instead?

The default help output with no subcommands

### Possible solution

The problem is the yarn.argv.env at the start of the cli script is called before any plugins are loaded, which causes --help processing.  The fix is to call yarn.help(false) before any calls to yarn.argv prior to loading plugins and then to call yarn.help() afterwards to re-enable help processing.

snstanton:
## Changes

Closes #1925

## How Has This Been Tested?

I have manually tested this.

## Documentation

No documentation changes are needed.

### DIFF
--- a/packages/cli/cli.js
+++ b/packages/cli/cli.js
@@ -8,6 +8,10 @@ const { boolean } = require("boolean");
 // This way we ensure all of the environment variables are not loaded too late.
 const project = getProject();
 let paths = [path.join(project.root, ".env")];
+
+// Disable help processing until after plugins are imported
+yargs.help(false);
+
 if (yargs.argv.env) {
     paths.push(path.join(project.root, `.env.${yargs.argv.env}`));
 }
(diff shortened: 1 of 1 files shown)


adrians5j:
Checking this.

adrians5j:
Checked this, and it really is because of this line:
https://github.com/webiny/webiny-js/blob/next/packages/cli/cli.js#L11-L13

The moment I removed all `yargs` calls at the top of the file, help started working fine.

Once merged, I will just try moving the "help disabling" line a couple of lines up, so it's not in the middle of the initial env vars loading / setup. But all in all, this is good. 👍🏻 

Hope it's OK I add you @snstanton to our list of contributors (located in our main README.md).

Thanks a bunch! 🍻 🍻

PR #1991, opened 2021-10-25
feat(app-page-builder): extend button element to support custom click-handlers

jimjoes:
Extends button element to allow 'click-handler' plugins to be set as the onClick handler.

## Changes
Added a new click-handler type for the onClick handler plugins. This specifies the button's handler function. plus any parameters the user can set that will be passed to the function.
Created a 'null' clickHandler. This serves two purposes - acts as a simple example and the 'no action' default. Not sure where to put this.
Renamed the 'link' button settings plugin to 'action'
Used Link as the basis for the new settings plugin.
Added a boolean switch to allow the user to select whether the button's action should be link or a clickHandler.
Added new 'clickHandler' options:  1) a drop down that allows the user to select one from a list of all registered 'click-handler' plugins and 2) parameters fields as specified in the plugin's 'parameters' array.
Updated button renderer to parse settings and render the button with either a clickHandler attached to the parent div or the link component.

Both settings and renderer components should correctly deal with buttons with the old 'link' settings.  

## How Has This Been Tested?
Added buttons and created a three custom handlers (both simple console logs and a more involved stripe integration).

## Documentation
Once the PR is accepted I'll be happy to write some documentation for it.


packages/app-page-builder/src/editor/plugins/elementSettings/index.ts
```diff
@@ -14,20 +15,24 @@ import save from "./save";
...
-    clone,
-    deleteElement,
-    width,
-    height,
-    save,
-    link,
-    grid
-];
+export default () => {
+    console.log("is this where it's being registered?");
```
adrians5j, line 19: Redundant clog. Please remove.

packages/app-page-builder/src/editor/plugins/elements/button/index.tsx
```diff
@@ -19,13 +19,15 @@ const buttonWrapper = css({
...
         "pb-editor-page-element-style-settings-button",
-        "pb-editor-page-element-style-settings-link",
+        "pb-editor-page-element-style-settings-action",
         "pb-editor-page-element-style-settings-horizontal-align-flex",
         "pb-editor-page-element-style-settings-margin",
         "pb-editor-page-element-settings-clone",
         "pb-editor-page-element-settings-delete"
     ];
 
+    // "pb-editor-page-element-style-settings-link",
```
adrians5j, line 29: Redundant clog. Please remove.

packages/app-page-builder/src/types.ts
```diff
@@ -251,6 +251,15 @@ export type PbRenderElementAttributesPlugin = Plugin & {
     }) => { [key: string]: string };
 };
 
+type ActionParameter = { label: string; defaultValue: any };
+
+export type PbButtonElementClickHandlerPlugin = Plugin & {
+    type: "pb-button-element-click-handler";
```
adrians5j, line 257: Let's make the `type` more in sync with the rest, like `pb-page-element-images-list-component`. So in our case, let's go with `pb-page-element-button-click-handler`.

packages/app-page-builder/src/types.ts
```diff
@@ -251,6 +251,15 @@ export type PbRenderElementAttributesPlugin = Plugin & {
...
+    title: string;
+    parameters: ActionParameter[];
+    handler: Function;
```
adrians5j, line 260: Can we provide a better type here?

packages/app-page-builder/src/render/plugins/elements/button/Button.tsx
```diff
@@ -9,10 +10,22 @@ const Button = ({ element }: { element: PbElement }) => {
     const {
         responsiveDisplayMode: { displayMode }
     } = React.useContext<PageBuilderContextValue>(PageBuilderContext);
-    const { type = "default", icon = {}, link = {} } = element.data || {};
+    const { type = "default", icon = {}, action = {}, link = {} } = element.data || {};
     const { svg = null } = icon;
     const { position = "left" } = icon;
 
+    const plugin = plugins.byName(action.clickHandler);
```
adrians5j, line 17: Wrap in `useMemo` please.

packages/app-page-builder/src/render/plugins/elements/button/Button.tsx
```diff
@@ -9,10 +10,22 @@ const Button = ({ element }: { element: PbElement }) => {
...
+
+    let href: string, newTab: boolean;
+
```
adrians5j, line 20: Let's just add a quick comment, saying we're preserving backwards compatibility.

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
```diff
@@ -0,0 +1,221 @@
...
+
+    return (
+        <Accordion title={"Action"} defaultValue={defaultAccordionValue}>
+            <Form
+                data={{ href, newTab, clickHandler, actionType, parameters }}
+                onChange={updateSettings}
+            >
+                {({ Bind }) => {
+                    const actionTypeOptions = [
+                        { id: "link", name: "Href Link" },
```
adrians5j, line 80: Let's go with just "Link"

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
```diff
@@ -0,0 +1,221 @@
...
+    element: PbEditorElement;
+};
+const ActionSettingsComponent: React.FunctionComponent<
+    ActionSettingsPropsType & PbEditorPageElementSettingsRenderComponentProps
+> = ({ element, defaultAccordionValue }) => {
+    const handler = useEventActionHandler();
+
+    let href: string, newTab: boolean;
+
+    if (element.data?.link && !element.data?.action) {
```
adrians5j, line 47: Let's just add a quick comment, saying we're preserving backwards compatibility

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
```diff
@@ -0,0 +1,221 @@
...
+            <Form
+                data={{ href, newTab, clickHandler, actionType, parameters }}
+                onChange={updateSettings}
+            >
+                {({ Bind }) => {
+                    const actionTypeOptions = [
+                        { id: "link", name: "Href Link" },
+                        { id: "onClickHandler", name: "Click Handler" }
+                    ];
+                    const clickHandlers = plugins.byType<PbButtonElementClickHandlerPlugin>(
```
adrians5j, line 83: Let's wrap in `useMemo`. You can move this up above the `return` line (72).

apps/admin/code/src/plugins/pageBuilder/editorPlugins.ts
```diff
@@ -72,7 +73,11 @@ import { gridPresets } from "@webiny/app-page-builder/editor/plugins/gridPresets
 // event actions
 import actionPlugins from "@webiny/app-page-builder/editor/recoil/actions/plugins";
 
+// The default null clickHandler for buttons - you can add your own custom plugins here to extend the functionality of the button element
+import nullHandler from "./nullHandler";
+
 export default [
+    nullHandler,
```
adrians5j, line 80: I saw there's this `nullHandler` in a couple of places. I'm only going to leave a comment about it here.

To make it a bit simpler, let's remove it, and in the settings component, let's maybe just insert a "None" option manually. 

P.S. I'm aware that the `Select` component is a bit tricky, so maybe you can also try replacing it with Autocomplete component. More than once, I've seen this component working better than Select.

adrians5j (changes requested):
Here's the initial review @econtentmaps. All in all, looks good, and I think it's gonna be a very nice addition.

BTW could we also see this in action somewhere? Or maybe even better, could you attach a short video of the new feature? 

Thank you again for your time and patience!

packages/app-page-builder/src/types.ts
jimjoes: Appreciate some guidance on this one. A handler function will be somewhat arbitrary, is there a better way to type it? Is that the problem you're referring to?

adrians5j:
Closing this - not relevant anymore.

PR #1997, opened 2021-10-27
Broken sort settings field for List of Pages

### LINKED ISSUE #1994: Broken sort settings field for List of Pages
### Version

5.16.0

### Operating System

Windows 10

### Browser

Chrome

Simply drag and drop a `List of Pages` page builder element into the editor.

pagelistissue1

I expect to see a Grid of pages.

### What do you see instead?

It always renders as `No pages match this criteria`.

pagelistissue2

### Additional information

`List of Pages` page builder element does not work since the `sort` field of the GraphQL query has changed (somewhere after `v5.12.0`).

### Possible solution

The current GraphQL query might look like this is:
```graphql
{
  pageBuilder {
    listPublishedPages(sort: { publishedOn : "desc" }) {
      data {
        title
      }
    }
  }
}
```

It appears that the correct GraphQL query for this would be:
```graphql
{
  pageBuilder {
    listPublishedPages(sort: publishedOn_DESC) {
      data {
        title
      }
    }
  }
}
```

Notice how the sort field is now an enum of type `PbListPagesSort`:
```graphql
enum PbListPagesSort {
  id_ASC
  id_DESC
  savedOn_ASC
  savedOn_DESC
  createdOn_ASC
  createdOn_DESC
  publishedOn_ASC
  publishedOn_DESC
  title_ASC
  title_DESC
}
```

dibenso:
## Changes
change `sort` argument type to `[PbListPagesSort!]` and remove `page` arguments from `LIST_PUBLISHED_PAGES` query used in `pagesLists` for render and editor plugins. Also remove no longer existing fields from `meta` in `LIST_PUBLISHED_PAGES`:            

```typescript
export const LIST_PUBLISHED_PAGES = gql`
    query ListPublishedPages(
        $where: PbListPublishedPagesWhereInput
        $limit: Int
        $sort: [PbListPagesSort!]
        $exclude: [String]
    ) {
        pageBuilder {
            listPublishedPages(where: $where, sort: $sort, limit: $limit, exclude: $exclude) {
                data {
                    id
                    title
                    path
                    url
                    snippet
                    tags
                    images {
                        general {
                            src
                        }
                    }
                    publishedOn
                    createdBy {
                        id
                        displayName
                    }
                    category {
                        slug
                        name
                    }
                }
                meta {
                    totalCount
                }
            }
        }
    }
`;
```

## How Has This Been Tested?
Manually. By Adding the `list of pages` page builder element to page and validating that it works as expected (before breaking). Would be happy to add some test if anyone can point me in the right direction 😄

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elements/pagesList/PagesList.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elements/pagesList/PagesList.tsx
@@ -15,9 +15,8 @@ const PagesList = props => {
     );
     const pageList = components.find(cmp => cmp.componentName === component);
     const { theme } = usePageBuilder();
-
-    // only assign to `setPage` since `page` is no longer a valid variable in `LIST_PUBLISHED_PAGES`.
-    const [, setPage] = useState(1);
+    const [cursors, setCursors] = useState([null]);
+    const [page, setPage] = useState(0);
     const pageAtomValue = useRecoilValue(pageAtom);
 
     if (!pageList) {
@@ -41,6 +40,7 @@ const PagesList = props => {
             }
         },
         limit: parseInt(vars.resultsPerPage),
+        after: cursors[page],
         exclude: [pageAtomValue.path]
     };
 
@@ -65,17 +65,19 @@ const PagesList = props => {
 
     const listPublishedPages = get(data, "pageBuilder.listPublishedPages");
 
-    /**
-     * How to handle these two checks if `meta` field no longer has `previousPage` and `nextPage` fields?
-     */
     let prevPage = null;
-    if (listPublishedPages.meta.previousPage) {
-        prevPage = () => setPage(listPublishedPages.meta.previousPage);
+    if (page >= 1) {
+        prevPage = () => {
+            setPage(page => page - 1);
+        };
     }
 
     let nextPage = null;
-    if (listPublishedPages.meta.nextPage) {
-        nextPage = () => setPage(listPublishedPages.meta.nextPage);
+    if (listPublishedPages.meta.cursor) {
+        nextPage = () => {
+            setCursors(cursors => [...cursors, listPublishedPages.meta.cursor]);
+            setPage(page => page + 1);
+        };
     }
 
     return (
--- a/packages/app-page-builder/src/editor/plugins/elements/pagesList/PagesList.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elements/pagesList/PagesList.tsx
@@ -15,7 +15,9 @@ const PagesList = props => {
     );
     const pageList = components.find(cmp => cmp.componentName === component);
     const { theme } = usePageBuilder();
-    const [page, setPage] = useState(1);
+
+    // only assign to `setPage` since `page` is no longer a valid variable in `LIST_PUBLISHED_PAGES`.
+    const [, setPage] = useState(1);
     const pageAtomValue = useRecoilValue(pageAtom);
 
     if (!pageList) {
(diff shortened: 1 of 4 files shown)


packages/app-page-builder/src/editor/plugins/elements/pagesList/graphql.ts
```diff
@@ -3,19 +3,12 @@ import gql from "graphql-tag";
 export const LIST_PUBLISHED_PAGES = gql`
     query ListPublishedPages(
         $where: PbListPublishedPagesWhereInput
-        $sort: PbListPagesSortInput
         $limit: Int
-        $page: Int
+        $sort: [PbListPagesSort!]
```
Ashu96, line 7: Good work on this @dibenso  👍 

Please also include the `after` argument because it is essential to implement pagination:

image

packages/app-page-builder/src/editor/plugins/elements/pagesList/graphql.ts
```diff
@@ -39,13 +32,7 @@ export const LIST_PUBLISHED_PAGES = gql`
                     }
                 }
                 meta {
-                    from
-                    to
-                    page
-                    nextPage
-                    previousPage
                     totalCount
```
Ashu96, line 42: Also add `cursor` and `hasMoreItems` properties in `meta`. They will be useful for implementing pagination :)

packages/app-page-builder/src/editor/plugins/elements/pagesList/PagesList.tsx
```diff
@@ -64,6 +65,9 @@ const PagesList = props => {
 
     const listPublishedPages = get(data, "pageBuilder.listPublishedPages");
 
+    /**
+     * How to handle these two checks if `meta` field no longer has `previousPage` and `nextPage` fields?
```
Ashu96: Good question @dibenso 👍 

In the new schema, we'll use `meta.hasMoreItems` to check whether there are more items
and `meta.cursor` to fetch more items by passing it as `after` variable in the GQL query.

One way to implement pagination is to keep a list of such cursors and then maintain an index to paginate through pages i.e. add the new cursor to the list and increment the index or page for `nextPage` and decrement the index for `prevPage`. For example:

```js
 const [cursors, setCursors] = useState([null]);
 const [page, setPage] = useState(0);

// Pass cursor as after argument
 const variables = {
        sort,
        where: {
            category: vars.category,
            tags: {
                query: vars.tags,
                rule: vars.tagsRule
            }
        },
        limit: parseInt(vars.resultsPerPage),
        after: cursors[page],
        exclude: [pageAtomValue.path]
    };

// And then, paginate using the cursor and pseudo page

let prevPage = null;
    if (page >= 1) {
        prevPage = () => {
            setPage(page => page - 1);
        };
    }

    let nextPage = null;
    if (listPublishedPages.meta.cursor) {
        nextPage = () => {
            setCursors(cursors => [...cursors, listPublishedPages.meta.cursor]);
            setPage(page => page + 1);
        };
    }

```

packages/app-page-builder/src/render/plugins/elements/pagesList/PagesList.tsx
```diff
@@ -91,6 +90,9 @@ const PagesListRender = props => {
 
     const listPublishedPages = get(data, "pageBuilder.listPublishedPages");
 
+    /**
```
Ashu96: We can use the same approach here as mentioned above.

Ashu96 (changes requested):
The PR looks good! Thanks for contributing @dibenso 💯 

I've left a few comments to answer your questions. I hope it helps.
If you need any help regarding this, please feel free to ask :)

dibenso:
Thanks @Ashu96 very helpful review! I've included what you've described to implement pagination.

Ashu96 (approved):
All looks good @dibenso 👍 

Let's wait for the CI checks to pass.

PR #2022, opened 2021-11-12
feat(app-page-builder): extend button element to support custom click-handlers (supersedes #1991)

jimjoes:
## Changes
Added a new click-handler type for the onClick handler plugins. This specifies the button's handler function. plus any parameters the user can set that will be passed to the function.
Created a 'null' clickHandler. This serves two purposes - acts as a simple example and the 'no action' default. Not sure where to put this.
Renamed the 'link' button settings plugin to 'action'
Used Link as the basis for the new settings plugin.
Added a boolean switch to allow the user to select whether the button's action should be link or a clickHandler.
Added new 'clickHandler' options: 1) a drop down that allows the user to select one from a list of all registered 'click-handler' plugins and 2) parameters fields as specified in the plugin's 'parameters' array.
Updated button renderer to parse settings and render the button with either a clickHandler attached to the parent div or the link component.

Both settings and renderer components should correctly deal with buttons with the old 'link' settings.

## How Has This Been Tested?
Added buttons and created three custom handlers (both simple console logs and a more involved stripe integration).

## Documentation
Once the PR is accepted I'll be happy to write some documentation for it.


packages/app-page-builder/src/render/plugins/elements/button/Button.tsx
```diff
@@ -9,10 +10,26 @@ const Button = ({ element }: { element: PbElement }) => {
     const {
         responsiveDisplayMode: { displayMode }
     } = React.useContext<PageBuilderContextValue>(PageBuilderContext);
-    const { type = "default", icon = {}, link = {} } = element.data || {};
+    const { type = "default", icon = {}, action = {}, link = {} } = element.data || {};
     const { svg = null } = icon;
     const { position = "left" } = icon;
 
+    console.log("action: ", action.clickHandler);
```
adrians5j: Please remove these logs here.

packages/app-page-builder/src/render/plugins/elements/button/Button.tsx
```diff
@@ -32,23 +49,41 @@ const Button = ({ element }: { element: PbElement }) => {
...
-                            className={getAllClasses(...classes)}
-                        >
-                            {content}
-                        </Link>
-                    </div>
+                    <>
+                        {action.actionType === "onClickHandler" ? (
+                            <div
+                                style={{ display: "flex", justifyContent }}
+                                onClick={
```
adrians5j: Instead of creating the function on each render here, let's move this above, and create the callback via `useCallback` hook. We could do it below the `if (link && !action) { ... } else { ...}` block, maybe line 33.

packages/app-page-builder/src/render/plugins/elements/button/Button.tsx
```diff
@@ -9,10 +10,26 @@ const Button = ({ element }: { element: PbElement }) => {
...
         responsiveDisplayMode: { displayMode }
     } = React.useContext<PageBuilderContextValue>(PageBuilderContext);
-    const { type = "default", icon = {}, link = {} } = element.data || {};
+    const { type = "default", icon = {}, action = {}, link = {} } = element.data || {};
     const { svg = null } = icon;
     const { position = "left" } = icon;
 
+    console.log("action: ", action.clickHandler);
+    console.log("plugin: ", plugins.byName(action.clickHandler));
+    const plugin = useMemo(() => plugins.byName(action.clickHandler), []);
```
adrians5j: Hm... it seems this will only memoize the plugin that was initial returned from the `plugins.byName` call. What happens if user changes the handler? 

In other words, let's ensure the plugin is recalculated every time the name changes. So, instead of passing `[]` as the 2nd arg to`useMemo`, let's pass `[action.clickHandler]`.

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
```diff
@@ -0,0 +1,228 @@
...
+                                                    };
+                                                });
+                                                return (
+                                                    <AutoComplete
+                                                        value={selectOptions.find(
+                                                            option => option.id === value
+                                                        )}
+                                                        placeholder="Select a handler"
+                                                        options={selectOptions}
+                                                        onChange={option => {
```
adrians5j: Can't we just pass `onChange` here, instead of `option => { ... }`?

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
```diff
@@ -0,0 +1,228 @@
...
+                                            {({ value, onChange }) => {
+                                                const selectOptions = clickHandlers.map(plugin => {
+                                                    return {
+                                                        id: plugin.name,
+                                                        name: plugin.title
+                                                    };
+                                                });
+                                                return (
+                                                    <AutoComplete
+                                                        value={selectOptions.find(
```
adrians5j: Let's move the value calculation above the `return` statement.

adrians5j (changes requested):
Almost there @econtentmaps! Please take a look at a couple minor details here.

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
jimjoes: We can. I'm more consistently debugging directly in chrome instead of peppering the code with console.logs now (which is why this was a function - I wanted to see what was passed to the onChange function so added a console.log here). #everydayisaschoolday :)

adrians5j:
@econtentmaps This looks good!

Thanks for the contribution! :) 🍻 

P.S. I just noticed we don't have you listed in the contributors list in our README.md, so I'll do that right now!

PR #2066, opened 2021-11-29
New `acl` and `cacheControl` params on the `uploadFolderToS3` Utility Function

### LINKED ISSUE #2065: uploadFolderToS3 script conflicts with private buckets
### Version

5.18.0

### Operating System

MacOS

### Browser

Firefox

The new uploadFolderToS3 mechanism is a nice speed up.  Unfortunately it's also making it harder to make the s3 buckets private.  Our IT insists on using private acls for buckets served by a cdn since there's no need to have public-read if it's not being directly accessed.  We have a bucket public access block that prevents public acls on objects.  This conflicts with the acl on the putObject call in the upload script.  Previously I could patch this because the code was inline, but now that it's been factored out, I will need to patch the cli-plugin-deploy-pulumi package.

It should be possible to mark the apps/admin bucket as private since it's only served by cloudfront.  If the App were passed an aws.cloudfront.OriginAccessIdentity parameter, then it could make the bucket private, set the BucketPublicAccessBlock to deny public acls, and add a policy that grants access to the CDN OAI.  This would also require the uploadFolderToS3 function to use a private acl to match.

### What do you see instead?

After marking the bucket private, the upload function fails because it attempts to set a forbidden public-read acl.

### Additional information

_No response_

### Possible solution

_No response_

snstanton:
…derToS3

Allow for specifying the acl to use when uploading files to s3.

## Changes

Closes #2065

## How Has This Been Tested?

I have installed into our environment using this patch.

### DIFF
--- a/packages/cli-plugin-deploy-pulumi/utils/aws/uploadFolderToS3.js
+++ b/packages/cli-plugin-deploy-pulumi/utils/aws/uploadFolderToS3.js
@@ -26,7 +26,9 @@ module.exports = async ({
     bucket,
     onFileUploadSuccess,
     onFileUploadError,
-    onFileUploadSkip
+    onFileUploadSkip,
+    acl = "public-read",
+    cacheControl = "max-age=31536000"
 }) => {
     const s3 = new S3Client({ region: process.env.AWS_REGION });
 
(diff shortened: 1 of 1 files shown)


adrians5j (changes requested):
Looks good, although, while we're at it, let's also expose `cacheControl` (as the default value, let's use the existing `max-age=31536000`).

Once this is merged, I'll add the TS types for this so it's easier to use.

adrians5j (approved):

PR #2076, opened 2021-12-02
fix(cwp-template-aws): logging incorrect profile during deploy

kedrzu:
## Changes
When running webiny deploy command, if AWS credentials are passed via environment variables (AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY), then the profile doesn't get loaded and we get this minor error in the logs.

Closes https://webiny.atlassian.net/browse/WEB-393

## How Has This Been Tested?
Set AWS credentials via `.env` file and checked the console output on deploy.

## Documentation
Not needed

### DIFF
--- a/packages/cwp-template-aws/cli/aws/checkCredentials.js
+++ b/packages/cwp-template-aws/cli/aws/checkCredentials.js
@@ -23,8 +23,6 @@ module.exports = {
             process.exit(1);
         }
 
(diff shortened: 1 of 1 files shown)


kedrzu:
Some PR check is failing because it looks for a branch in `webiny-js` repo I guess, and this is from fork.

PR #2079, opened 2021-12-02
Added hooks for before & after build

kedrzu:
## Changes
Added hooks `hook-before-build` and `hook-after-build`, that allow running a custom code during builds.
It's an extension of already existing `hook-before-deploy` and `hook-after-deploy`.

## How Has This Been Tested?
Defined a custom hook plugin to check if every hook is running in proper moment.

## Documentation
I have another task to add documentation for deploy & build hooks.


adrians5j (approved):

adrians5j:
Looks good, we'll merge this for 5.20.0, because we need the docs.

PR #2087, opened 2021-12-03
fix(@webiny/app-page-builder & apps/admin): resolve remaining issues …

### LINKED ISSUE #2052: Finish new Page Builder Button / Action Settings
> This issue is a continuation of https://github.com/webiny/webiny-js/pull/2022, which wasn't released with `5.18.0` due to a couple of polishes that still need to be done.
 
Hey @econtentmaps ,

here are a couple of things we'll just need to take a look at before making the new feature official.

1. The most important issue is the fact that on new projects, in the button settings, we cannot see both previous Link nor the new Action settings. This is related to the fact that the new action plugins is not registered, I believe [here](https://github.com/webiny/webiny-js/blob/next/packages/app-page-builder/src/editor/plugins/elementSettings/index.ts#L31).

image

BTW, it's interesting though that we've added a change [here](https://github.com/webiny/webiny-js/pull/2022/files#diff-bc7fbf9c46443f4ced176c031a3318e81da39ebb042ab4337dc8bfd55276dd6a), but yeah, that plugin doesn't really exist because it was not included in the first file.

2. The 2nd, a bit trickier, problem is:  what happens to projects that used the Link settings for button?

We cannot just remove the `pb-editor-page-element-style-settings-link` plugin, and introduce `pb-editor-page-element-style-settings-action`. Because in the database, if the `pb-editor-page-element-style-settings-link` was used, settings would no longer be correctly shown in the editor.

Ultimately, I think that, for backwards compatibility purposes, we'll need to stick with the `pb-editor-page-element-style-settings-link` as the name of the settings.

3. When an action doesn't require any params, the label "None Required" is not in style with the rest of the UI. Let's see if we can use the `Typography` component so that's looking a bit nicer.

4. The input that's used for selecting the action / callback is larger then others. Let's try and make it smaller, e.g.:
image

jimjoes:
…with button custom handlers

Resolves: #2052

See #2052 comments for explanations of changes and testing

### DIFF
--- a/apps/admin/code/src/plugins/pageBuilder/editorPlugins.ts
+++ b/apps/admin/code/src/plugins/pageBuilder/editorPlugins.ts
@@ -56,6 +56,7 @@ import height from "@webiny/app-page-builder/editor/plugins/elementSettings/heig
 import align from "@webiny/app-page-builder/editor/plugins/elementSettings/align";
 import save from "@webiny/app-page-builder/editor/plugins/elementSettings/save";
 import link from "@webiny/app-page-builder/editor/plugins/elementSettings/link";
+import action from "@webiny/app-page-builder/editor/plugins/elementSettings/action";
 import gridSettings from "@webiny/app-page-builder/editor/plugins/elementSettings/grid";
 import textSettings from "@webiny/app-page-builder/editor/plugins/elementSettings/text";
 import visibility from "@webiny/app-page-builder/editor/plugins/elementSettings/visibility";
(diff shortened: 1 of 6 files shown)


adrians5j:
Please just resolve the `yarn.lock` conflict on your end so we can merge this. 

I'll do one additional round of testing before we release 5.19.0.

@econtentmaps

adrians5j:
Not relevant anymore. Closing.

PR #2091, opened 2021-12-03
HTTP 500 error is thrown when JWT token is invalid or expired

kedrzu:
## Changes
When you pass an auth token that is invalid or expired, you get 500 HTTP response, instead of basic authorization error.
Probably the best way to handle this would be to fallback gracefully, when invalid token is passed. This way for endpoints that require authorization, user would get basic auth error, and for endpoints that accept both logged and not logged users, you would fall back to `not-logged` instead of pushing back with some kind of error.

## How Has This Been Tested?
Requested GQL resources with invalid & expired tokens.


adrians5j:
Closing for now and creating a separate internal issue.

PR #2092, opened 2021-12-03
Rerender page from the page builder

kedrzu:
## Changes
Right now, the only way for users to re-render a page is to create a new revision of it and publish it again, which is pretty annoying. This adds a feature of re-rendering the page from the page preview:
image

Also there is a new dedicated mutation `pageBuilder.rerenderPage` that you can use in your own scripts.

## How Has This Been Tested?
Manual tests of the new `rerenderPage` mutation and admin app.


kedrzu:
We decided to remove the button in the UI and leave only GQL mutation. This way we basically restrict it from non-technical users, that will not be aware what is the purpose of the button.
For technical users & developers - you will still be able to rerender the page with a simple gql query, for example using our embedded GQL playground.

adrians5j (approved):
Looks good.

packages/api-page-builder/src/graphql/graphql/pages.gql.ts
```diff
@@ -1,11 +1,18 @@
...
+} from "@webiny/handler-graphql/responses";
 import { GraphQLSchemaPlugin } from "@webiny/handler-graphql/types";
 import { Page, PbContext } from "~/types";
 import Error from "@webiny/error";
 import resolve from "./utils/resolve";
 import pageSettings from "./pages/pageSettings";
 import { fetchEmbed, findProvider } from "./pages/oEmbed";
 import lodashGet from "lodash/get";
+import { PageSecurityPermission } from "@webiny/api-page-builder/types";
+import checkBasePermissions from "@webiny/api-page-builder/graphql/crud/utils/checkBasePermissions";
```
adrians5j: BTW @kedrzu, check these two imports. Use "~" when importing from package root.

Wouldn't mention this, but CI/CD threw an error related to these two imports. Using "~" will fix it.

adrians5j (changes requested):

packages/api-page-builder/src/graphql/graphql/pages.gql.ts
adrians5j: Also, `PageSecurityPermission` can be added to the import statement in line 8.

PR #2097, opened 2021-12-06
Refactor plugins into class based

kedrzu:
## Changes
Currently, plugins are written as plain objects with the required `type` property, that serves as a discriminator.
This has several drawbacks:
- typings are hard, we lose the whole power of Typescript
- there is no intellisense, no way to tell which properties are required for the plugin config
- documentation for plugins is difficult to create, no tool will auto-generate a reference docs for that

We want to migrate to more class-based plugins, so every plugin instance will be created with a constructor `new MyPlugin({ ... })`. This way we can ensure good TypeScript types, and inline docs for each plugin parameter. It will be also easier to navigate between available plugins in the reference documentation (that is about to be created).

Basically, this task is a big epic, that will require a lot of work, to be finished, so probably we won't do it in one shot.
This mechanism will be backward compatible, so it should not break your code.

## How Has This Been Tested?
❌Not tested yet, draft only

## Documentation
⌛Will have to update current plugin docs after making the changes.

### DIFF
--- a/packages/app-page-builder/src/admin/utils/createElementPlugin.tsx
+++ b/packages/app-page-builder/src/admin/utils/createElementPlugin.tsx
@@ -5,7 +5,7 @@ import { PbEditorElement, PbEditorPageElementPlugin } from "../../types";
 import Title from "./components/Title";
 
 export default (el: PbEditorElement) => {
-    const elementPlugins = plugins.byType<PbEditorPageElementPlugin>("pb-editor-page-element");
+    const elementPlugins = plugins.byType(PbEditorPageElementPlugin);
     const rootPlugin = elementPlugins.find(pl => pl.elementType === el.content.type);
 
     if (!rootPlugin) {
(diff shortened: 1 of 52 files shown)


packages/app-page-builder/src/admin/utils/createElementPlugin.tsx
```diff
@@ -14,31 +14,32 @@ export default (el: PbEditorElement) => {
...
-        type: "pb-editor-page-element",
-        elementType: name,
-        target: rootPlugin.target,
-        toolbar: {
-            title({ refresh }) {
-                return <Title plugin={name} title={el.name} id={el.id} refresh={refresh} />;
+    plugins.register(
+        new PbEditorPageElementPlugin({
+            name,
+            // title: el.name,
```
kedrzu, line 20: There is no such property in typings, nor it's filled in any other place.

packages/app-page-builder/src/types.ts
```diff
@@ -357,7 +368,11 @@ export type PbEditorPageElementPlugin = Plugin & {
         parent?: PbEditorElement
     ) => Omit<PbEditorElement, "id">;
     // A function to render an element in the editor.
-    render: (params: { theme?: PbTheme; element: PbEditorElement; isActive: boolean }) => ReactNode;
+    render?: (params: {
```
kedrzu, line 371: There is one place in which `render` function is not provided. Can't tell if it's intended or not. Check `packages/app-page-builder/src/admin/utils/createElementPlugin.tsx` file - that's the one.

adrians5j:
Closing this for now, not a priority.

PR #2101, opened 2021-12-08
Upgraded webpack to version 5

kedrzu:
## Changes
Upgrading to webpack 5 gives faster builds & smaller bundles.
Upgrade process is rather complicated, because of many inter-dependencies.

I had to update several other packages.

## How Has This Been Tested?
Built & run admin application in standard and watch mode.


Pavel910:
Tried this branch, and here are some findings:

1) Error in the console, and some weird `svg` error in the browser when you visit the PB Editor:

image

2) Can we make the bundle output as compact as possible? Similar to how we have it with webpack-4 (see the screenshot):

image

With this new branch, there's tons of noise in the output, which is not useful. 
We should hide that by default for development (watch) mode, and of course, print it as we currently do for an actual production build.

3) There's also this warning:
```
<w> [webpack-dev-server] "hot: true" automatically applies HMR plugin, you don't have to add it manually to your webpack configuration.
```

4) If possible, let's try getting React Fast Refresh to work, it will be really useful, especially with the new admin framework I'm working on, where we'll no longer have non-React plugins.

5) Curious why `WebpackBar` is not showing, it's a nice addition to see that something is actually happening instead of staring at static terminal window. [EDIT]: maybe @doitadrian disabled it because of how `watch` command works with those panels? 

6) What's this TODO about?
```
TODO: replace with eslint-webpack-plugin
```
I see you already enabled that plugin. Does this require some testing, or?

7) Let's output `svg` files into a `static/media` folder, as it does with the WP4 setup, so it doesn't make a mess in the build folder:
image

adrians5j:
Some context for point 5 of @Pavel910's review.

There is definitely some custom logging behavior applied when watching packages via the `webiny watch` command. This is basically because when running the command, receiving logs from different Webpack / bundling watch sessions at the same time can be overwhelming and hard to read.

So, the general idea was - when doing a watch of a single package - leave all of the logs. Even `WebpackBar` can be shown in that case. On the other hand, if 2 or more packages are used, we reduce the information that's being sent from the initialized watch processes, and we simply output essential watch-related info. That way we make the logs a bit clearer for the user.

That being said, while looking through the code, I noticed a bug here:
https://github.com/webiny/webiny-js/blob/next/packages/cli-plugin-deploy-pulumi/commands/watch/watchPackages.js#L99

Please change this to:

```ts
worker.postMessage(
                    JSON.stringify({
                        options: { env, debug, logs: !multipleWatches },
                        package: current
                    })
                );
```

So, the `logs` property needs to be `true` only when not doing a watch over multiple packages. Note that this `logs` value is ultimately sent to `watch` commands located in `webiny.config.ts` files (hence [this](https://github.com/webiny/webiny-js/blob/next/packages/project-utils/bundling/function/createBuildConfig.js#L49)).

The watch/deploy commands have been heavily revamped in the current Webiny 5.18.0 release, so it's definitely possible we'll encounter a couple of these rough edges that need to be corrected.

In any case, I know I might've been a bit brief when explaining things above, so if you'll need any extra info around this, let me know.

.yarnrc.yml
```diff
@@ -19,4 +19,8 @@ packageExtensions:
     peerDependencies:
       react-dom: ^16.14.0
 
+plugins:
```
adrians5j, line 22: Can you clarify this change please?

api/code/fileManager/download/tsconfig.json
```diff
@@ -1,9 +1,9 @@
 {
-  "extends": "../../tsconfig.json",
+  "extends": "../../../tsconfig.json",
```
adrians5j, line 2: I see in a log of file we're doing the same change.

So, extending the `"../../../tsconfig.json"`, and in `references`, using `tsconfig.build.json`. Can you please explain why we had to do these changes? Just so we're all on the same page.

api/code/graphql/src/security.ts
```diff
@@ -13,7 +13,7 @@ import createAdminUsersApp from "@webiny/api-admin-users-cognito";
 import { syncWithCognito } from "@webiny/api-admin-users-cognito/syncWithCognito";
 import { createStorageOperations as createAdminUsersStorageOperations } from "@webiny/api-admin-users-cognito-so-ddb";
 
-export default ({ documentClient }) => [
+export default ({ documentClient }: { documentClient: any }) => [
```
adrians5j: Can we assign the `DocumentClient` class (imported from AWS SDK) here instead of `any`?

api/code/headlessCMS/src/security.ts
```diff
@@ -10,7 +10,7 @@ import parentTenantGroupAuthorization from "@webiny/api-security/plugins/parentT
 import cognitoAuthentication from "@webiny/api-security-cognito";
 import anonymousAuthorization from "@webiny/api-security/plugins/anonymousAuthorization";
 
-export default ({ documentClient }) => [
+export default ({ documentClient }: { documentClient: any }) => [
```
adrians5j: The same goes here. Can we use DocumentClient class? Just so I don't repeat myself (and annoy you 😉 ), I won't mention the same thing in the following files.

api/code/prerenderingService/flush/src/index.ts
```diff
@@ -12,7 +12,7 @@ const documentClient = new DocumentClient({
 export const handler = createHandler(
     flushPlugins({
         storageOperations: createPrerenderingServiceStorageOperations({
-            table: table => {
+            table: (table: any) => {
```
adrians5j: It's actually `TableModifier`, not `any`. But on the other hand, not sure why this is needed in the first place because I can see the types defined in `packages/api-prerendering-service-so-ddb/src/types.ts`.

adrians5j: The same question goes for the same issue in other files.

api/tsconfig.json
```diff
@@ -10,7 +10,8 @@
     "pretty": true,
     "noFallthroughCasesInSwitch": true,
     "noImplicitReturns": true,
-    "forceConsistentCasingInFileNames": true
+    "forceConsistentCasingInFileNames": true,
```
adrians5j, line 13: I'll have to read a bit what these rules do.

All in all, just a note that almost all changes to configs that we do on the `webiny-js` repo level, should also be propagated to user's project (`cwp-template-aws` package). Not 100% of the time, but mostly that will be the case.

Maybe you already did it but I just didn't get to that part of the PR. Still, I decided to mention it so I don't forget.

apps/admin/code/package.json
```diff
@@ -52,7 +52,9 @@
   "scripts": {
     "analyze": "source-map-explorer build/static/js/main.*",
     "watch": "cross-env PORT=3001 yarn webiny run watch",
-    "build": "yarn webiny run build"
+    "build": "yarn webiny run build",
+    "build:dev": "yarn webiny run build --env dev",
```
adrians5j: Let's remove these custom scripts. No need to add additional stuff we need to explain. If users will need these commands, they can add them themselves.

apps/admin/code/package.json
```diff
@@ -33,16 +33,16 @@
...
     "apollo-link-batch-http": "^1.2.14",
     "apollo-utilities": "^1.3.4",
     "core-js": "^3.0.1",
     "cross-fetch": "^3.0.4",
     "invariant": "^2.2.4",
     "prop-types": "^15.7.2",
-    "react": "^16.14.0",
-    "react-dom": "^16.14.0",
+    "react": "^17.0.2",
+    "react-dom": "^17.0.2",
```
adrians5j, line 26: Nice.

package.json
```diff
@@ -65,14 +65,14 @@
     "del": "^6.0.0",
     "empty": "^0.10.1",
     "env-ci": "2.0.1",
-    "eslint": "^7.18.0",
-    "eslint-config-standard": "^16.0.2",
-    "eslint-import-resolver-babel-module": "^5.2.0",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-jest": "^23.1.1",
+    "eslint": "^8.4.1",
```
adrians5j, line 60: Good stuff, although I must say it might've been more smart to do these project-level updates within a separate dedicated PR.

packages/api-prerendering-service-so-ddb/tsconfig.json
```diff
@@ -1,35 +1,29 @@
 {
-  "extends": "../../tsconfig",
+  "extends": "../../tsconfig.json",
```
adrians5j, line 2: BTW what did cause the items to be alphabetically sorted? Some tool or you did this manually?

packages/app-form-builder/src/admin/views/Editor.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import HTML5Backend from "react-dnd-html5-backend";
+import { HTML5Backend } from "react-dnd-html5-backend";
```
adrians5j, line 2: I guess this is the proper way of doing it? Just asking in case this was done by accident, sorry.

packages/app-i18n-content/package.json
```diff
@@ -42,11 +41,6 @@
     "build": "yarn webiny run build",
     "watch": "yarn webiny run watch"
   },
-  "svgo": {
```
adrians5j, line 46: BTW just a confirmation, the `svgo` is no longer needed right?

packages/cwp-template-aws/template/common/api/code/prerenderingService/flush/src/index.ts
```diff
@@ -12,7 +12,7 @@ const documentClient = new DocumentClient({
 export const handler = createHandler(
     flushPlugins({
         storageOperations: createPrerenderingServiceStorageOperations({
-            table: table => {
+            table: (table: any) => {
```
adrians5j: Just a reminder - check `any` type here.

packages/cwp-template-aws/template/common/example.babel.react.js
```diff
@@ -43,16 +43,6 @@ module.exports = ({ path }) => ({
                 useBuiltIns: true
             }
         ],
-        [
```
adrians5j, line 46: I see you've propagated the changes to CWP template as well. Cool 👍🏻

packages/cwp-template-aws/template/ddb-es/api/code/graphql/src/security.ts
```diff
@@ -13,7 +13,7 @@ import createAdminUsersApp from "@webiny/api-admin-users-cognito";
 import { syncWithCognito } from "@webiny/api-admin-users-cognito/syncWithCognito";
 import { createStorageOperations as createAdminUsersStorageOperations } from "@webiny/api-admin-users-cognito-so-ddb";
 
-export default ({ documentClient }) => [
+export default ({ documentClient }: { documentClient: any }) => [
```
adrians5j: Just a reminder, please check `any` type here.

tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "./tsconfig.build.json",
   "compilerOptions": {
-    "noEmit": true,
+    "noEmit": false,
```
adrians5j: Would like to learn why this is `false` now. Can you clarify please?

scripts/generateTsConfigsInApi.js
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
-const { getPackages, getPackage } = require("./utils/getPackages");
-const { relative } = require("path");
+const { getPackages, getPackage, PROJECT_ROOT } = require("./utils/getPackages");
```
adrians5j, line 2: What is this file? Why do we need this? Was this a temporary script or it's something we might need in the future? 🤔

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -47,7 +50,9 @@ module.exports = options => {
                 if (messages.errors.length > 1) {
                     messages.errors.length = 1;
                 }
-                return reject(new Error(messages.errors.join("\n\n")));
+
+                console.error(messages.errors.join("\n\n"));
```
adrians5j, line 55: Is this console log necessary here?

packages/cli-plugin-deploy-pulumi/commands/watch/watchPackages.js
```diff
@@ -96,7 +98,7 @@ module.exports = async ({ inputs, output, context }) => {
 
                 worker.postMessage(
                     JSON.stringify({
-                        options: { env, debug, logs },
+                        options: { env, debug, logs: !multipleWatches },
```
adrians5j: Let's make this `logs: !multipleWatches || logs`.

In other words, show full compilation logs if we're only watching a single package (`!multipleWatches`) or if specifically instructed by the user (`|| logs`).

.babel.react.js
```diff
@@ -43,16 +43,6 @@ module.exports = ({ path }) => ({
                 useBuiltIns: true
             }
         ],
-        [
-            "babel-plugin-named-asset-import",
```
adrians5j, line 47: I think I already asked about this, but just in case, these are no longer needed?

adrians5j (changes requested):
First iteration done.

.yarnrc.yml
kedrzu: I was using `yarn upgrade-interactive` as a simple tool to upgrade some packages for all of our projects. With the new version of `yarn` it's a separate tool. I left that, because I thought that we may make a use of it.

api/code/fileManager/download/tsconfig.json
kedrzu: Some of the paths were simply wrong and they pointed to non-existing files. That's because the script for generating tsconfigs didn't take into account the depth of the project directory, and for api functions this depth varies. I fixed the script, so it works well.

Referencing directly `tsconfig.build.json` is a very important one, because otherwise, we have all the packages referenced by `compilerOptions.paths` configuration. This doesn't work, because it treats referenced projects' files as they belong to the current project, so for example `~/` in referenced project files is calculated relative to the current project, not the referenced one.

Basically, we had a lot of the wrong configurations and my IDE (vscode) was all red on some of `tsconfigs`. Babel seems to be very liberal in terms of how it treats this configuration, TS is not.

api/code/graphql/src/security.ts
kedrzu: Sure, will do that. I wasn't sure what should be here and TS didn't allow for implicit `any` types.

api/code/prerenderingService/flush/src/index.ts
kedrzu: I will check that, maybe `createPrerenderingServiceStorageOperations` has not that well-written declaration and type inferring fails.

packages/api-prerendering-service-so-ddb/tsconfig.json
kedrzu: We have a script that generates all tsconfigs. I simply run it.

packages/app-form-builder/src/admin/views/Editor.tsx
kedrzu: Yes, API of the library changed. I made a slight upgrade of the library, because it wasn't going well with other libs. It's not the latest one though, because in latest there are some breaking changes to the API.

packages/app-i18n-content/package.json
kedrzu: It is, but I moved `svgo` configuration from separate `package.json` files to `svgr` loader config in webpack to be more consistent.

tsconfig.json
kedrzu: If you use typescript project references, you can't set `noEmit`. TS simply raises an error.

scripts/generateTsConfigsInApi.js
kedrzu: @Pavel910 told me we have `tsconfigs` generated and I found, that we do it in this file. It adds typescript project references and `compilerOptions.paths` basing on `package.json` references.

kedrzu: As I mentioned before, I had to fix few things here, because extended config paths were wrong sometimes.

packages/project-utils/bundling/function/buildFunction.js
kedrzu: I replaced `reject` with error with a simple console log and `reject` with no param, because it gives cleaner output. Otherwise in case of error you would see a stack trace leading you to this exact line, which is not very useful for anything.

But yes, you are right, I can do a reject with string just the same way as in code above.

.babel.react.js
kedrzu: These vere used to preserve loader directives inside imports, like `@svgr/webpack!./local-file.svg`.
There were only few usages of this, and I changed them to be more consistent with everything else.

Because we have `svgr` set on webpack level, we don't need to explicitely define this loader. This makes the configuration a little bit simpler and consistent. Also it allows you to remove all the reduntant `svgo` configuration from all of the projects.

api/code/prerenderingService/flush/src/index.ts
kedrzu: Problem was in fact with env variable which may possibly be undefined, and TS couldn't properly inference type.
image

tsconfig.json
Pavel910: @kedrzu `tsconfig.json` is used by the IDE (Webstorm, for the majority of the team) and this will output files into `src` folder.
Build scripts are configured to work with `tsconfig.build.json`. This needs to be double-checked, because this small change can easily mess up the whole repo setup.

That's basically the only reason we have 2 tsconfig files. One is for the IDE, and one is for actual build process.

scripts/generateTsConfigsInApi.js
Pavel910: @doitadrian it's similar to `generageTsConfigsInPackages` :) I was sick of `api` folder pointing to declaration files when I Cmd+Click the import path.

.babel.react.js
adrians5j: Nice.

tsconfig.json
adrians5j: Please double-check this @kedrzu, as @Pavel910 mentioned, the `tsconfig.json` is used for IDEs, not within actual build processes.

packages/app-form-builder/src/admin/views/Editor.tsx
adrians5j: Cool.

packages/api-prerendering-service-so-ddb/tsconfig.json
adrians5j: Cool.

.yarnrc.yml
adrians5j: OK, I guess it doesn't hurt.

adrians5j (dismissed):
Once we resolve the question related to `tsconfig.json`, I think we can start thinking about integrating this.

Note that a lot of files were adjusted and package version updated here, so I might've missed something. Plus the adjustments on the Webpack-related code are something you'd really need to dig deep to fully understand, so I didn't go too deep there.

But that's why we'll give it a good amount of testing first.

tsconfig.json
kedrzu: In which IDE there is automatic build with TSC? I personally use vscode and it doesn't work this way.

Pavel910: @kedrzu please read my message again :)

Pavel910: If this flag is `on`, it will make a mess of your project, if your tsconfig is configured to emit declarations:
image

kedrzu: Sorry, I must have missed it :)
In the latest change, I enabled `noEmit: true`, but disabled project references to rely purely on `paths`. 
It works fine for vscode and as far as I understand, @Pavel910 confirmed, that works good for webstorm as well.

Pavel910:
Closing this due to unexplainable build issue in GH Actions. Want to try a fresh PR.

PR #2102, opened 2021-12-08
Package compilation made faster

kedrzu:
## Changes
I managed to get all the packages to be compiled by TSC in one go, so it's much faster and allows watching all the packages,
which should make development a lot easier.

I had quite a lot of TS errors and I needed to fix them.

Note, that babel is not running here, only TSC, so we need to verify, if packages are processed by babel later during webpack build.

This change was made on a top of upgrading webpack, so I would mitigate any conflicts. I will rebase it later.
If you want to read the code and see changes, check the last commit.

## How Has This Been Tested?
run `yarn tsc --build` or `yarn watch` - everything will get compiled. Then you can run for example watch for admin app.


adrians5j:
Closing for now. At the moment, not a priority.

PR #2105, opened 2021-12-09
Disable public access to application S3 buckets

kedrzu:
## Changes
There was this issue raised by the community:
image

Instead of restricting HTTP access to S3 bucket it's possible to completely disable public access to S3 bucket and only allow cloudfront distribution to access it via proper origin identity.

Currently I've made this change only for `apps/admin` and `apps/website` projects, and didn't touch template projects.
We should think about proper way of migration. Putting S3 bucket into pulumi component probably causes creating a new bucket, so this may be a pain during migration (or maybe not).

## How Has This Been Tested?
Deployed and checked all apps.
I may have to make further tests to this S3 bucket create-from-scratch issue.


apps/website/cli/deploy.ts
```diff
@@ -39,6 +39,7 @@ export default {
         await uploadFolderToS3({
             path: buildFolderPath,
             bucket: websiteOutput.appStorage,
+            acl: "private",
```
kedrzu, line 42: Files can no longer be uploaded with `public` access, or we get 403 error during upload.

apps/website/cli/renderWebsite.ts
```diff
@@ -20,6 +20,10 @@ export default {
 
         // 2. Get exports from `site` stack, for `args.env` environment.
         const apiOutput = getStackOutput({ folder: "api-ddb", env: args.env });
+        if (!apiOutput) {
```
kedrzu: This little thing blows up with errors during every deploy. @Pavel910 says it's some old stuff and should be removed entirely. I leave it for now, because I don't know the project that well to clean it up.

apps/website/pulumi/delivery.ts
```diff
@@ -76,6 +51,13 @@ class Delivery {
                 defaultTtl: 30,
                 maxTtl: 30
             },
+            customErrorResponses: [
+                {
+                    errorCode: 404,
+                    responseCode: 404,
+                    responsePagePath: "/_NOT_FOUND_PAGE_/index.html"
```
kedrzu, line 58: This one does not fully work for my setup, but maybe I'm missing something:
image

.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
+{
```
kedrzu, line 1: Added some local configuration for anyone using VS code (like me) 😉

.vscode/settings.json
```diff
@@ -0,0 +1,14 @@
...
+    ".webiny": true
+  },
+  "search.exclude": {
+    "**/dist": true,
+    "**/node_modules": true,
+    "yarn.lock": true,
+    ".pulumi": true
+  },
+  "files.eol": "\n"
+}
```
kedrzu, line 14: Added some local IDE configuration for anyone using VS code (like me) 😉

apps/website/cli/renderWebsite.ts
Pavel910: Folder name `api-ddb` should be just `api`, and that's it.

apps/admin/cli/deploy.ts
```diff
@@ -39,6 +39,7 @@ export default {
         await uploadFolderToS3({
             path: buildFolderPath,
             bucket: adminOutput.appStorage,
+            acl: "private",
```
adrians5j, line 42: Let's make this setting the default one so that the user doesn't have to pass it manually. Ultimately, it's probably better to have a safe default value instead of more convenient one.

adrians5j: So, internally, make the default `acl` value to be `private`, and then scan existing `deploy.ts` files and simply remove the `acl` property.

Note that you'll need to do this 4 times in total - 2x in our webiny-js repo, a 2x in `cwp-template-aws` package, which is what gets executed when users run `npx create-webiny-project ...`.

apps/website/cli/deploy.ts
adrians5j: Cool.

apps/website/cli/renderWebsite.ts
adrians5j: This should be `api` as @Pavel910 suggested.

apps/website/cli/updatePbSettings.ts
```diff
@@ -18,6 +18,10 @@ export default {
 
         // 2. Get exports from `api` stack, again, for `args.env` environment.
         const apiOutput = getStackOutput({ folder: "api-ddb", env: args.env });
+        if (!apiOutput) {
+            // website was not yet deployed
```
adrians5j: Let's put something more meaningful here, like maybe:

```ts
context.warning(`Cannot update Page Builder application's settings - ${context.warning.hl("API")} project application not deployed.`);
```

apps/admin/pulumi/index.ts
```diff
@@ -10,11 +9,44 @@ export = async () => {
         WbyEnvironment: process.env.WEBINY_ENV as string
     });
 
-    const app = new App();
-    const cloudfront = new Cloudfront({ appS3Bucket: app.bucket });
+    const bucket = new CloudFrontBucket("admin-app");
```
adrians5j, line 12: I think this makes sense. But you know what, the other day @Pavel910 and I discussed the possible end goal with all of this Pulumi code, and the conclusion was: let's have a couple of these standard components that we use in Webiny projects. This way users are less mentally overloaded by the whole setup.

For example, for Admin Area, let's have the `SPA` component. This one we'd use with the Admin Area project app.

For the Website app, we wouldn't be able to use `SPA` because it has a bit different architecture. So, we could maybe create the `PageBuilderWebsite` component, that would internally utilize the `SPA` one and possibly some other components.

All in all, I think it might be better for us to immediately utilize this approach instead of doing another project upgrade down the road. Two birds with one stone.

adrians5j (changes requested):
First iteration done.

apps/admin/cli/deploy.ts
kedrzu: I'm not sure we can make it private already, because users need to change their own pulumi code to make private S3 bucket access available. Without proper code migration their apps will stop working. That is why I started with modifying templates only.
When we migrate to simpler pulumi configuration with ready-made components, then I think we can make this change safely.

adrians5j: Makes sense. Good catch. 👍🏻

apps/website/cli/updatePbSettings.ts
```diff
@@ -17,7 +17,7 @@ export default {
         const websiteOutput = getStackOutput({ folder: "apps/website", env: args.env });
 
         // 2. Get exports from `api` stack, again, for `args.env` environment.
-        const apiOutput = getStackOutput({ folder: "api-ddb", env: args.env });
+        const apiOutput = getStackOutput({ folder: "api", env: args.env });
```
adrians5j, line 20: There was an `if` here in my last review. Let's bring it back as the warning for the user will be informative.

adrians5j (changes requested):
Besides one small thingie I noted, let's revisit https://github.com/webiny/webiny-js/pull/2105#discussion_r774541425 again.

It would make sense to implement this immediately - less upgrades / fuss for our users down the road. Let me know if you need anything additional here (call, separate Jira task with more specs, ...).

PR #2109, opened 2021-12-10
feat(api-headless-cms): take locale from cms context

### LINKED ISSUE #2107: i18n not working in headless cms for dynamodb only project
### Version

5.18.1

### Operating System

macOs Monterey

### Browser

Firefox

1. Updating a field in one locale updates it for all locales
- Set up a new Webiny project using the scaffolder. 
- Choose DynamoDB only project.
- Deploy the webiny admin app and the webiny api.
- Create one or more cms models (either through a plugin or through the admin application), this can be a model with one text field.
- Add an entry for that model and fill in the field.
- Add a new locale.
- Go to the model and switch locales.
- Update the entry's text field for the newly created locale.

2. Creating an entry in a new locale corrupts the entry (and consequentially the model)
- Set up a new Webiny project using the scaffolder.
- Choose DynamoDB only project.
- Deploy the webiny admin app and the webiny api.
- Create a new model using the admin app.
- Add a new locale
- Go to the model and switch to the newly created locale
- Create a new entry

1. Updating a field in one locale updates it for all locales
- I can update the entry fields for each locale separately.

2. Creating an entry in a new locale corrupts the entry (and consequentially the model)
- I can create an entry in a locale and edit it's fields for every locale.

### What do you see instead?

1. Updating a field in one locale updates it for all locales
- The field of the entry is updated for every locale

2. Creating an entry in a new locale corrupts the entry (and consequentially the model)
- I get a message that the entry cannot be found, but it stays visible in the left column (see screenshot). I cannot remove the entry either. If I try to remove the model, I a the message that the model still has entries. 

image

I encountered both situation 1. and 2. in the same project.

### Additional information

I was able to reproduce this issue myself in a fresh project without our custom plugin models.

### Possible solution

_No response_

brunozoric:
## Changes
Closes #2107

## How Has This Been Tested?
Jest and manually

### DIFF
--- a/packages/api-headless-cms-ddb-es/src/operations/model/index.ts
+++ b/packages/api-headless-cms-ddb-es/src/operations/model/index.ts
@@ -145,7 +145,6 @@ export const createModelsStorageOperations = (params: Params): CmsModelStorageOp
 
         try {
             await entity.delete(keys);
-            return model;
         } catch (ex) {
             throw new WebinyError(
                 ex.messatge || "Could not delete model.",
@@ -157,6 +156,26 @@ export const createModelsStorageOperations = (params: Params): CmsModelStorageOp
                 }
             );
         }
+        const { index } = configurations.es({
+            model
+        });
+        /**
+         * In case of DynamoDB error we need to remove the index we created.
+         */
+        try {
+            await elasticsearch.indices.delete({
+                index
+            });
+        } catch (ex) {
+            throw new WebinyError(
+                `Could not delete elasticsearch index "${index}" after model record was deleted.`,
+                "DELETE_MODEL_INDICE_ERROR",
+                {
+                    error: ex
+                }
+            );
+        }
+        return model;
     };
 
     const get = async (params: CmsModelStorageOperationsGetParams) => {
--- a/packages/api-headless-cms-ddb-es/src/operations/model/index.ts
+++ b/packages/api-headless-cms-ddb-es/src/operations/model/index.ts
@@ -127,7 +127,7 @@ export const createModelsStorageOperations = (params: Params): CmsModelStorageOp
             return model;
         } catch (ex) {
             throw new WebinyError(
-                ex.messatge || "Could not update model.",
+                ex.message || "Could not update model.",
                 ex.code || "MODEL_UPDATE_ERROR",
                 {
                     error: ex,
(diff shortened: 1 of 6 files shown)


packages/api-headless-cms-ddb/src/operations/model/index.ts
```diff
@@ -19,6 +19,11 @@ interface PartitionKeysParams {
 }
 const createPartitionKey = (params: PartitionKeysParams): string => {
     const { tenant, locale } = params;
+    if (!tenant) {
+        throw new WebinyError(`Missing tenant variable when creating model partitionKey`);
+    } else if (!locale) {
+        throw new WebinyError(`Missing tenant variable when creating model partitionKey`);
```
Pavel910: `Missing tenant...`, shouldn't this be `Missing locale variable..` ?

packages/api-headless-cms/src/plugins/context.ts
```diff
@@ -3,7 +3,7 @@ import { CmsContext } from "~/types";
 
 export default () => {
     return new ContextPlugin<CmsContext>(context => {
-        const locale = context.i18n.getCurrentLocale();
+        const locale = context.i18nContent.getLocale();
```
Pavel910, line 6: What is the difference between these 2 lines of code?

brunozoric: `i18n.getCurrentLocale()` does not get the selected locale in the UI. it always returns en-us

Pavel910: We should open this as a separate issue and ask @doitadrian for input. This behavior of `i18n.getCurrentLocale()` sounds odd/broken.

adrians5j: Discussed this over a call. The change here is correct.

packages/api-headless-cms-ddb-es/src/operations/model/index.ts
```diff
@@ -145,7 +145,6 @@ export const createModelsStorageOperations = (params: Params): CmsModelStorageOp
 
         try {
             await entity.delete(keys);
-            return model;
         } catch (ex) {
             throw new WebinyError(
                 ex.messatge || "Could not delete model.",
```
adrians5j: Check `messatge`

packages/api-headless-cms-ddb/src/operations/model/index.ts
```diff
@@ -19,6 +19,11 @@ interface PartitionKeysParams {
 }
 const createPartitionKey = (params: PartitionKeysParams): string => {
     const { tenant, locale } = params;
+    if (!tenant) {
```
adrians5j, line 22: Saw that you ended your error messages with a "." above (`packages/api-headless-cms-ddb-es/src/operations/model/index.ts`), so let's do the same here. :)

adrians5j (changes requested):
Two minor things to look at, feel free to merge once resolved.

adrians5j (approved):

PR #2117, opened 2021-12-15
refactor(api-headless-cms): make listEntries internal use method

brunozoric:
## Changes
The `context.cms.listEntries()` should be used only for our internal coding. Added comments about that.
Added check if published / latest were sent into the `listEntries` method - cannot search if trying to search both, or none.

## How Has This Been Tested?
Jest and manually.


Ashu96 (approved):
LGTM 👍

PR #2128, opened 2021-12-16
feat: telemetry push

Ewan-Roberts:
## Changes

- Added ability to send telemetry data to WCP endpoint
- Added telemetry code to wrap clients lambda functions
- Added telemetry code pulling for getting latest telemetry code from remote bucket

AWAITING:
- Clients WCP_API_KEY is to be set

## How Has This Been Tested?

- Jest

## Documentation

TBC this should possibly be added to legal documentation or security pages


packages/project-utils/package.json
```diff
@@ -41,6 +42,7 @@
     "get-yarn-workspaces": "1.0.2",
     "glob": "^7.1.2",
     "html-webpack-plugin": "4.0.0-beta.5",
+    "https": "^1.0.0",
```
adrians5j: Hm... why `https` here?

packages/project-utils/package.json
```diff
@@ -1,5 +1,6 @@
 {
   "name": "@webiny/project-utils",
+  "telemetryVersion": "1.0.1",
```
adrians5j: Can you describe this a bit to me?

package.json
```diff
@@ -228,5 +228,9 @@
     "@pulumi/pulumi": "3.19.0",
     "@pulumi/aws": "4.28.0"
   },
-  "packageManager": "yarn@3.1.0"
+  "packageManager": "yarn@3.1.0",
```
adrians5j: 1. These new dependencies should not be defined here, but in the actual package / `package.json` that is using them. 2. `https` is a native Node module, doesn't need to be included in list of dependencies.

packages/project-utils/package.json
adrians5j: BTW as mentioned in other comment, native Node modules do not need to be declared as dependencies. Feel free to remove this.

api/code/graphql/package.json
```diff
@@ -47,5 +47,8 @@
         "@webiny/api-tenant-manager"
       ]
     }
+  },
+  "devDependencies": {
+    "graphql-request": "^3.4.0"
```
adrians5j: Hmm.. why is this here?

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -54,4 +73,21 @@ module.exports = options => {
             resolve();
         });
     });
+
+    const handlerFile = fs.readFileSync(options.cwd + "/build/handler.js", {
```
adrians5j: Use `path.join` to join paths. This is mainly for compatibility with multiple OSes.

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -54,4 +73,21 @@ module.exports = options => {
...
+        encoding: "utf8",
+        flag: "r"
+    });
+    const includesGraphQl = handlerFile.includes("handler-graphql");
```
adrians5j: How does this perform BTW? Is it an intensive operation / an operation that takes a lot of time to execute? Just wondering...

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -54,4 +73,21 @@ module.exports = options => {
...
+
+    if (includesGraphQl) {
+        try {
+            await getLatestTelemetryCode;
```
adrians5j: You mean `await getLatestTelemetryCode();` ?

adrians5j: Let's rename this function. `get` prefix in function name usually indicates that the function is returning a value, but in this case, the function is really just adapting the manipulating built code.

Let's call it for example `injectHandlerTelemetry`.

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -54,4 +73,21 @@ module.exports = options => {
...
+        } catch (err) {
+            console.log("Error loading telemtry code");
```
adrians5j: These should throw an error, and not do `console.log`.

adrians5j (changes requested):
First iteration.

Let's check these couple of quick fixes first. BTW then `package-lock.json` should not be in the repo. We use `yarn`, which relies on `yarn.lock`.

packages/project-utils/package.json
Ewan-Roberts: So I wanted a place to store the local telemetry version number so I can compare it to the remote version so thought it best to store it in the `package.json`. Does that make sense or were you thinking of storing it somewhere else?

packages/project-utils/bundling/function/buildFunction.js
Ewan-Roberts: this bundle was on the smaller size but its pretty quick at 10ms on average

package.json
Ewan-Roberts: so I added this due to an error thrown by adio: 
```
✖ "adio" found some errors. Please fix them and try committing again.
1. @webiny/project-utils (/Users/ewan/Development/webiny/webiny-js/packages/project-utils)
Packages used in source code, but not listed in package.json:
1. https
```
I wonder why adio is picking this up as a required library 🤔 
I have kept it in so i can commit and push for the time being

packages/project-utils/package.json
Ewan-Roberts: same as https://github.com/webiny/webiny-js/pull/2128#discussion_r776326749

packages/project-utils/bundling/function/telemetry.js
```diff
@@ -0,0 +1,43 @@
...
+    const latestTelemetryVersion = telemetryVersions.pop();
+
+    if (packageData.telemetryVersion === latestTelemetryVersion) {
+        return;
+    }
+
+    const latestCodeData = await s3
+        .getObject({
+            Bucket: "telemetry-code",
+            Key: "telemetry-versions/" + latestVersion
```
Ewan-Roberts: hmmmm maybe a dumb qustion but will this work on another cilents instance? Its a public bucket, but bucket names are not unique, so how does aws know where to get it from?

packages/project-utils/package.json
adrians5j: Ah...

Yeah, just add it here `.adiorc.js:14`. That should then let you commit without setting `https` in deps.

packages/project-utils/package.json
adrians5j: Hm.. maybe I'm missing something.. would it work if we just relied on `@webiny/project-utils`'s version? So on the backend, we just take that version into account and apply the needed logic accordingly?

BTW on backend, there is the `process.env.WEBINY_VERSION` env var, or even better: `context.WEBINY_VERSION` (`packages/handler/src/types.ts`).

Maybe we can discuss this further.

But in any case, let's write a short document in Notion / WCP category on this topic. 

Something like - what are the problems we're trying to achieve this. What happens in the client is not up-to-date. And so on.

packages/project-utils/bundling/function/buildFunction.js
adrians5j: Nice :)

I just wanted to confirm it's not like 30secs or something like that :)

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -54,4 +73,20 @@ module.exports = options => {
             resolve();
         });
     });
+
+    const handlerFile = fs.readFileSync(path.join(options.cwd, "/build/handler.js"), {
+        encoding: "utf8",
+        flag: "r"
+    });
+    const includesGraphQl = handlerFile.includes("handler-graphql");
```
adrians5j: We'll need to include `handler-graphql-content-model` (`packages/api-headless-cms/src/content/graphQLHandlerFactory.ts:127`) too, because this is the plugin name of the dynamic Headless CMS GraphQL API (so, the `handler-graphql` package is not being used for that).

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -54,4 +73,20 @@ module.exports = options => {
...
+
+    if (includesGraphQl) {
+        try {
+            await injectHandlerTelemetry();
+        } catch (err) {
+            throw new Error(err);
```
adrians5j: I guess we can remove the `catch (err)` since the only thing we're doing within the catch block is again throwing the same `err`. :)

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -54,4 +73,20 @@ module.exports = options => {
             resolve();
         });
     });
+
+    const handlerFile = fs.readFileSync(path.join(options.cwd, "/build/handler.js"), {
```
adrians5j: I just noticed we'll need to add a check here, before performing this handler code-related stuff.

Basically, if the `id` of the project is not present in root `webiny.project.ts` file, the whole dev/deploy experience should work as if the user is not using WCP at all.

This is how things work in my yet-to-be-merged branch.

So, ultimately, what we need to do here is do a check - if `!project.config.id`, then exit.

The `project` is something you can get by calling `getProject`, for example: `packages/cli-plugin-scaffold-react-component/src/index.ts:117`. Note that you'll need to pass the `cwd` from `options` object to it.

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -1,8 +1,27 @@
...
+const fs = require("fs");
+const telemetry = require("./telemetry");
 
-module.exports = options => {
+async function injectHandlerTelemetry() {
+    await telemetry.getLatestTelemetryFunction();
+
+    fs.copyFileSync(
+        path.join(cwd, "build", "handler.js"),
+        path.join(cwd, "build", "handler.original.js")
```
adrians5j: Let's rename `handler.original.js` to `_handler.js`. It's nothing super important, but my thinking is maybe the less the user knows here, the better.

adrians5j (changes requested):
I left a couple of comments. But you know what.

I think for now I'm gonna pause the review, so that we can, as mentioned in one of the comments below, first create a short document that describes how telemetry client will be retrieved on deployments, and how it works.

Let's do that first, and then we'll see if some more important changes will be needed, and if this code (mainly `packages/project-utils/bundling/function/telemetry.js`) will need to change.

packages/project-utils/bundling/function/buildFunction.js
Ewan-Roberts: Maybe I miss understand, this will capture anything with the string "handler-graphql" so this should be included no?

packages/app-page-builder/src/editor/plugins/elementSettings/align/HorizontalAlignFlexSettings.tsx
```diff
@@ -67,78 +67,77 @@ enum AlignmentsTypeEnum {
...
-        const element = useRecoilValue(elementWithChildrenByIdSelector(activeElementId));
-        const fallbackValue = useMemo(
-            () =>
-                applyFallbackDisplayMode(displayMode, mode =>
-                    get(element, `${DATA_NAMESPACE}.${mode}`)
-                ),
-            [displayMode]
-        );
-        const align = get(element, propName, fallbackValue || AlignmentsTypeEnum.CENTER);
+const HorizontalAlignFlexSettings: React.FunctionComponent<
```
Ewan-Roberts: Anything under app-page-builder can be ignored, they are prettier fixes

packages/project-utils/bundling/function/telemetry.js
```diff
@@ -0,0 +1,16 @@
+const fs = require("fs");
+const fetch = require("node-fetch");
+
+const TELEMETRY_BUCKET_URL = "https://wcp-telemetry-function-7cbe312.s3.amazonaws.com/latest";
```
adrians5j: Hm.. let's do this in the following way...

In the WCP repo where you're doing the hosting of the client code... let's rather have a CF distribution that is configured to serve files from S3 when `/clients/*` is requested. The files are ofc pulled from S3.

So, this URL will always be something like `xyz.cloudfront.net/clients/latest`. Something along those lines.

Also, please introduce an option to override this URL via an env variable, so that we can do some testing (automated or manual) in the future here.

I know for example I added the following two so that I can perform some testing with Webiny CLI stuff:

```bash
WCP_API_URL=https://d2d4evj0p7twjc.cloudfront.net/graphql
WCP_APP_URL=http://localhost:3000
```

So.. maybe we could also have like `WCP_API_CLIENTS_URL` or something like that.

packages/project-utils/bundling/function/telemetryFunction.js
```diff
@@ -0,0 +1,116 @@
+const userFunction = require("./_handler.js");
+const https = require("https");
+const packageData = require("../../package.json");
+const AWS = require("aws-sdk");
+AWS.config.update({ region: "us-east-1" });
```
adrians5j: Ughh we definitely don't want to globally override AWS SDK's region. What if a Lambda function was deployed into `eu-central-1`?  This could definitely affect the user if he's working with services within a different region.

Can we have an isolated instance of the AWS SDK, just so we don't affect the user?

packages/project-utils/bundling/function/buildFunction.js
adrians5j: Nope, that's the thing (and a good question you've asked).

Interestingly enough, the Headless CMS Lambda function doesn't include the `handler-graphql` string. We'll see how to maybe better resolve this, by potentially not even relying on `handler-graphql`, but simply putting our own magic string into the code, which will be explicitly be used for these purposes. I'll write this into jira.

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -1,8 +1,25 @@
...
 const chalk = require("chalk");
+const fs = require("fs");
+const telemetry = require("./telemetry");
+import { getProject } from "@webiny/cli/utils";
 
-module.exports = options => {
+async function injectHandlerTelemetry() {
+    await telemetry.updateTelemetryFunction();
+
+    fs.copyFileSync(path.join(cwd, "build", "handler.js"), path.join(cwd, "build", "_handler.js"));
```
adrians5j: Wait, where is this `cwd` coming from? 🤔

adrians5j (dismissed):
Let's continue by resolving some of the things reported in this iteration.

Thank you!🍻

packages/project-utils/bundling/function/telemetryFunction.js
Ewan-Roberts: whoops this is an older version than the one in s3

packages/project-utils/bundling/function/telemetry.js
Ewan-Roberts: The CF distribution was very time consuming and had no progress even with help from the team, i have put it into a ticket so i can address it on your return: https://webiny.atlassian.net/browse/WEB-707

packages/project-utils/bundling/function/buildFunction.js
Ewan-Roberts: For whoever reviews this, this is an outstanding question. I was thinking of adding a comment with `// add-function-to-telemetry-tracking` to the functions we want tracked and search for that string in the code

Pavel910: For the `alpha`, you can hardcode just the `graphql` and `headlessCMS` functions. As we're working on the new project setup, you will be able to hook into any project function via a plugin, so no string hardcoding will be necessary.

It's something new we're developing, so we'll have way more control over the generated code for the users (basically, there will be no code in the project at all, and we'll be generating stuff on fly). This means that you'll be able to intercept lambda templates, modify stuff with ASTs, or jsut add WCP plugins on the fly.

Pavel910: Resolved this over Slack and a call: to avoid problems with comments, which can be stripped away by webpack, let's register a dummy plugin, and use its `type` as your identifier which will tell you that you need to wrap the function with a tracker. Something like:

```js
{ type: "wcp-telemetry-tracker" }
```

My suggestions is to stick it in here: `packages/api-tenancy/src/index.ts:41`
This package is used by GraphQL, Headless CMS, and PB import/export functions. This will cover the WCP Alpha needs, and in the future, the new project setup will allow us to make the implementation a lot smarter than that.

packages/project-utils/bundling/function/telemetryFunction.js
```diff
@@ -0,0 +1,115 @@
...
+        return;
+    }
+
+    timerRunning = true;
+
+    setInterval(async () => {
+        const timeInFiveMinutes = Date.now() + MAXIMUM_MINUTES_UNTIL_REQUEST_FIRED * 60000;
+        if (timeInFiveMinutes > initialTime) {
+            if (localData.logs.length > 0) {
+                await postTelemetryData(localData);
```
Pavel910, line 65: if this `postTelemetryData` throws an error, will this break the original Lambda execution? 🤔

packages/project-utils/bundling/function/buildFunction.js
```diff
@@ -60,4 +77,22 @@ module.exports = options => {
...
         });
     });
+
+    const project = getProject({ cwd });
+
+    if (!project.config.id) {
+        return result;
+    }
+
+    const handlerFile = fs.readFileSync(path.join(options.cwd, "/build/handler.js"), {
```
Pavel910: Bundles can be several MBs large, around 3-4MB. So this should not be synchronous. Use an asynchronous read, and feel free to use the `fs-extra` package for all FS related stuff, we already have it in the `project-utils` package, it exposes proper promisified functions.

The reason is, we're building ~14 functions, some of which are large, so reading 14 large files in a synchronous manner can make the terminal freeze for a few seconds.

Pavel910 (commented):
Left some questions and comments @Ewan-Roberts.

packages/project-utils/bundling/function/telemetryFunction.js
Ewan-Roberts: Good point, I wrapped this in a try catch

PR #2140, opened 2021-12-21
Enable sourcemaps for lambda functions

kedrzu:
## Changes

Lambda functions are hard to debug.  Even if we get stacktraces, they tell us nothing about the root of the problem, because of TS transpilation and bundling in the process of building the lambda function.

Added support for source maps within lambda functions and additional error logging for graphql, that includes stack traces in the error response.

Everything you need to do is to add a `DEBUG=true` environmental variable (for example inside your `.env` file).

## How Has This Been Tested?
Throw an error in GQL resolver and check results. Should include nice stacktrace:
image

## Documentation
We should maybe add some entry in docs, @doitadrian what do you think?


kedrzu:
I fixed generating source maps for packages and added support for reading these sourcemaps with `source-map-loader` but it doesn't seem to fully work, errors are shown to occur in `pages.gql.js` not `pages.gql.ts`. Maybe after upgrading webpack to 5 I will make it a little better.

Also, I enabled source maps by default on prod - didn't find any performance overhead here, maybe during cold-start, but it's hard to measure. It can be disabled with env variable `SOURCE_MAPS=false`.

packages/project-utils/packages/buildPackage.js
```diff
@@ -84,12 +84,16 @@ const babelCompile = async ({ cwd }) => {
...
     for (let i = 0; i < compilations.length; i++) {
-        const [file, { code, map }] = await compilations[i];
+        const [file, result] = await compilations[i];
+        const { code, map } = result;
 
         const paths = getDistFilePaths({ file, cwd });
         fs.mkdirSync(dirname(paths.code), { recursive: true });
         writes.push(fs.promises.writeFile(paths.code, code, "utf8"));
-        writes.push(paths.map, map, "utf8");
+        if (map) {
```
adrians5j: Can you clarify why this `if` is needed here? 

Can we put a quick comment here?

package.json
```diff
@@ -107,7 +107,6 @@
     "rimraf": "^3.0.2",
     "rxjs": "^6.5.5",
     "semver": "^6.3.0",
-    "source-map-support": "^0.5.0",
```
adrians5j, line 101: Nice cleanup 👍🏻

packages/handler-graphql/src/responses.ts
```diff
@@ -17,13 +17,26 @@ export class ErrorResponse {
...
-    constructor(params: ErrorResponseParams) {
+    constructor(params: ErrorResponseParams | (Error & Partial<ErrorResponseParams>)) {
         this.data = null;
-        this.error = {
-            code: params.code || defaultParams.code,
-            message: params.message || defaultParams.message,
-            data: params.data || defaultParams.data
-        };
+
+        if (params instanceof Error) {
```
adrians5j: Let's replace this with something like the following:


```ts
this.data = null;

let data = null;
if (params.data || params.stack) {
  data = {};
  if (params.data) {
    data = { ...params.data };
  }

  const debug = process.env.DEBUG === "true";
  if (params.stack && debug) {
    data = { ...data, stack: params.stack };
  }
}

this.error = {
  code: params.code || defaultParams.code,
  message: params.message || defaultParams.message,
  data
};
```

Note that within the backend / GraphQL code, we are almost always not throwing an instance of the native `Error` class, but of [`WebinyError`](https://github.com/webiny/webiny-js/blob/next/packages/error/src/Error.ts#L7), which can contain all three - `message`, `code`, and `data`.

packages/project-utils/bundling/function/createWatchConfig.js
```diff
@@ -8,25 +8,29 @@ module.exports = options => {
...
 
-    const { overrides, debug, cwd } = options;
+    const { overrides, cwd } = options;
 
     // Customize Babel options.
     if (typeof overrides.babel === "function") {
         babelOptions = overrides.babel(babelOptions);
     }
 
+    const sourceMaps = process.env.SOURCE_MAPS !== "false";
```
adrians5j: You know what... when I think about it, let's not even check an environment variable, and have this hardcoded to `true`.

The reason being.... users can always adjust the Webpack configs on their own (via their `webiny.config.ts` files). They have the ability to turn off this feature there.

Ultimately, it's probably better to leverage existing features we already have in place. This way we also don't have to write any documentation that cover this magic env variable.

adrians5j (changes requested):
In general, looks good. But let's revisit a couple of things before we merge this.

packages/project-utils/bundling/function/createWatchConfig.js
kedrzu: It's used in several places, so user would have to override all of them to effectively turn it off. I'll better move it to `options` object.

packages/handler-graphql/src/responses.ts
kedrzu: Current code would support that well, becasue `WebinyError` extends native `Error` - it even has to, otherwise stack trace would not be preserved. But I'll make the code a little more bit explicit about that.

packages/project-utils/packages/buildPackage.js
kedrzu: Sure. Previously there was a bug that prevented from generating source maps at all, because `writes` array was supposed to be filled with promises, and instead parameters for `writeFile` went there.
In case where maps are not enabled, `map` value is null, so we should not write it into file. In our setup it's always enabled, but user may want to use our `buildPackage` function with his own babel setup and disabled sourcemaps.

Maybe we should force source map generation at the level of generating packages, not `.babelrc` file? This way user will not be able to override that uninentionally. They won't miss anything on always generating source maps.

adrians5j: Yeah I think that makes sense. Let's always have them, and not let the user worry about it. I guess then we could undo the change we did in `babelrc` file?

packages/handler-graphql/src/responses.ts
adrians5j: Let's see what you'll end up with and then we'll cover it again if needed. 

P.S. Have in mind that the `data` property of `this.error` should not return either stack trace or data that has been provided by the thrower. It could contain both - some custom data assigned by the thrower and `stack` if the received object contains it and `DEBUG===true`.

packages/project-utils/bundling/function/createWatchConfig.js
adrians5j: Yes, but one thing to note here - it's possible that the number of `webiny.config.ts` and workspaces will reduce in the future (been experimenting with this in our WCP repo), so ultimately, users would only need to do a single adjustment, not 10+ which is now the case.

Just something to have in mind.

packages/error/src/Error.ts
```diff
@@ -5,18 +5,24 @@ interface ErrorOptions<TData> {
...
 export default class WError<TData = any> extends Error {
-    message: string;
     code?: string;
     data?: TData;
-    constructor(message: string | ErrorOptions<TData>, code: string = null, data: TData = null) {
-        super();
 
-        if (typeof message === "string") {
-            this.message = message;
+    constructor(message: string, code?: string, data?: TData);
```
adrians5j, line 11: Cool.

packages/project-utils/packages/buildPackage.js
```diff
@@ -89,7 +94,12 @@ const babelCompile = async ({ cwd }) => {
 
         const paths = getDistFilePaths({ file, cwd });
         fs.mkdirSync(dirname(paths.code), { recursive: true });
+
+        // save the compiled JS file
         writes.push(fs.promises.writeFile(paths.code, code, "utf8"));
+
+        // save source maps file
+        // source map generation may be disabled, in this case we don't want to generate the file
```
adrians5j: All looks good although this comment is now confusing. Says `source map generation may be disabled,`, but that's not true, isn't it? It's always `true`. 

Base on the `true` that we're now passing for source maps, does the `if` statement still make sense?

BTW please just use upper case for the first letter, when starting a sentence in comments. I also like to end the comment with `.`, but that's less important. Without it it's fine too (I've seen we have both styles in the codebase - both with the ending `.` and without it).

packages/project-utils/bundling/function/createWatchConfig.js
```diff
@@ -15,7 +15,7 @@ module.exports = options => {
         babelOptions = overrides.babel(babelOptions);
     }
 
-    const sourceMaps = process.env.SOURCE_MAPS !== "false";
+    const sourceMaps = options.sourceMaps !== false;
```
adrians5j: Looks good, please just update `BuildFunctionConfig` in `packages/project-utils/index.d.ts`.

adrians5j (changes requested):
A couple of smaller things.

packages/project-utils/packages/buildPackage.js
kedrzu: That's true, this `if` is not relevant since I force-enabled source maps.

adrians5j (approved):

adrians5j:
@kedrzu Looks good, please just resolve the conflicts when you get a chance, and we'll merge this!

packages/handler-graphql/src/responses.ts
```diff
@@ -17,13 +17,39 @@ export class ErrorResponse {
...
+            // usually we don't use env variables directly inside code
+            // but otherwise it would make it very tedious to pass this through multiple layers of code
+            const debug = process.env.DEBUG === "true";
+
+            this.error = {
+                code: defaultParams.code,
```
adrians5j: Let's do `params.code || defaultParams.code` here, the same way we did it with the `message` prop. Then we can remove the below `if ("code" in ...` check.

packages/handler-graphql/src/responses.ts
```diff
@@ -17,13 +17,39 @@ export class ErrorResponse {
...
+                message: params.message || defaultParams.message,
+                data: debug ? { stacktrace: params.stack } : defaultParams.data
+            };
+
+            if ("code" in params && params.code) {
+                this.error.code = params.code;
+            }
+
+            if ("data" in params && params.data) {
```
adrians5j: We need to revisit this part.

If the `data` property contained the `stacktrace` property (assigned on line 31), the following line would overwrite it (line 41):
```ts
this.error.data = params.data;
```

Because of this and the fact that `data` can sometimes be an array, I think it'd be best if we didn't push the stack trace into the `data` property. Instead, we can just introduce a separate `stack` property and that's it. 

So, if debugging is enabled, it would get populated. Otherwise it'd be `null`.

What do you think?

adrians5j (changes requested):
Left two comments, hope I didn't miss anything.

Thanks!

packages/project-utils/bundling/function/webpack.config.js
```diff
@@ -10,27 +10,31 @@ module.exports = options => {
...
     let babelOptions = require("./babelrc");
     // Customize Babel options.
     if (typeof overrides.babel === "function") {
         babelOptions = overrides.babel(babelOptions);
     }
 
+    const sourceMaps = options.sourceMaps !== false;
+
     const definitions = overrides.define ? JSON.parse(overrides.define) : {};
     const tsCheksEnabled = process.env.WEBINY_ENABLE_TS_CHECKS === "true";
```
adrians5j: BTW there's a typo here: `tsCheksEnabled`. 

Just wondering, does your IDE highlight grammar errors? If not, it would definitely be smart to enable grammar checking. Webstorm does it out of the box, not sure about other IDEs.

adrians5j (changes requested):
Added one minor comment regarding a grammar error.

packages/handler-graphql/src/responses.ts
kedrzu: I simplified it, but couldn't remove `"code" in params` check, because of TS typings. Error can be a simple `Error` instance that does not have `code` or `data` props. I could make type casting to `any` but it's not pretty.

packages/handler-graphql/src/responses.ts
```diff
@@ -16,14 +16,29 @@ export class ErrorResponse {
...
-    constructor(params: ErrorResponseParams) {
+    constructor(params: ErrorResponseParams | Error | (Error & ErrorResponseParams)) {
         this.data = null;
-        this.error = {
-            code: params.code || defaultParams.code,
-            message: params.message || defaultParams.message,
-            data: params.data || defaultParams.data
-        };
+
+        if (params instanceof Error) {
```
adrians5j: I decided to dive a bit deeper into this code block, as it seemed a bit more complex than it should be.

Ended up with the following solution:

```ts
interface ErrorResponseParams {
  message: string;
  code?: string;
  data?: any;
  stack?: string;
}

const defaultParams: ErrorResponseParams = {
  code: "",
  message: "",
  data: null,
  stack: null,
};

export class ErrorResponse {
  data: null;
  error: {
    code: string;
    message: string;
    data: any;
    stack: string;
  };
  constructor(params: ErrorResponseParams) {
    this.data = null;

    const debug = process.env.DEBUG === "true";

    let stack = defaultParams.stack;
    if (debug && params.stack) {
      stack = params.stack;
    }

    this.error = {
      code: params.code || defaultParams.code,
      message: params.message || defaultParams.message,
      data: params.data || defaultParams.data,
      stack
    };
  }
}
```

The most important change I guess would be the `constructor(params: ErrorResponseParams) {` simplification. Ultimately, it doesn't matter if the constructor accepts `Error` or not. It's important that `Error` is just compatible with the `ErrorResponseParams` interface, which is true. Once that's done, the below code gets simplified significantly.


I'm hoping I didn't miss something here, let me know if I did!

packages/handler-graphql/src/responses.ts
```diff
@@ -16,14 +16,29 @@ export class ErrorResponse {
...
+                data: ("data" in params && params.data) || defaultParams.data,
+                stack: debug ? params.stack : undefined
+            };
+        } else {
+            this.error = {
+                code: params.code || defaultParams.code,
+                message: params.message || defaultParams.message,
+                data: params.data || defaultParams.data
+            };
+        }
```
adrians5j, line 33: BTW the same addition should happen in the `ListErrorResponse` class.

packages/handler-graphql/src/responses.ts
```diff
@@ -16,14 +16,29 @@ export class ErrorResponse {
...
     }
```
adrians5j, line 41: Another thing - we've added the `stack` here in these classes, but not on the GraphQL layer. We want to be able to retrieve this newly added `stack` property via a new `stack` GraphQL field.

Unfortunately we'll need to add this field one by one on all defined `*Error` GraphQL types. You can find these with the following regex search:

image


BTW maybe we could finally introduce a generic `Error` type? Something to think about and potentially tackle in a separate PR in the near future. For now even a new issue would be OK. All in all, not a high priority.

adrians5j (changes requested):
Please check my latest review. Let me know if I've overlooked something.

Thank you! 🍻

packages/handler-graphql/src/responses.ts
```diff
@@ -37,37 +44,45 @@ export class NotFoundResponse extends ErrorResponse {
...
 export class ListErrorResponse {
-    data: null;
-    meta: null;
-    error: {
+    public readonly data: null;
+    public readonly meta: null;
+    public readonly error: {
         code: string;
         message: string;
         data?: any;
```
adrians5j: Let's remove `?` from `data?` (just to be more unified with the above ErrorResponse).

packages/handler-graphql/src/responses.ts
```diff
@@ -11,18 +12,24 @@ const defaultParams: ErrorResponseParams = {
 };
 
 export class ErrorResponse {
-    data: any;
-    error: {
+    public readonly data: null;
```
adrians5j, line 16: Just wanted to report this but you've already fixed this. Nice!

packages/handler-graphql/src/responses.ts
```diff
@@ -37,37 +44,45 @@ export class NotFoundResponse extends ErrorResponse {
 }
 
 export class ListErrorResponse {
-    data: null;
-    meta: null;
-    error: {
+    public readonly data: null;
+    public readonly meta: null;
```
adrians5j, line 55: `readonly` - nice!

packages/handler-graphql/src/responses.ts
```diff
@@ -11,18 +12,24 @@ const defaultParams: ErrorResponseParams = {
...
         this.data = null;
+
+        const debug = process.env.DEBUG === "true";
+
         this.error = {
             code: params.code || defaultParams.code,
             message: params.message || defaultParams.message,
-            data: params.data || defaultParams.data
+            data: params.data || defaultParams.data,
+            stack: debug ? params.stack : defaultParams.stack
```
adrians5j: Not super problematic/important, but still, maybe we can be explicit here and assign `null` when `stack` property doesn't exist.

This is the reason I went with a separate `if` statement here in my previous comment:

```ts
// This part here ensures `stack` is either `string` or `null`.
let stack = defaultParams.stack;
if (debug && params.stack) {
  stack = params.stack;
}
```

In your case, we'd have `undefined` in the `this.error.stack` property if the `debug` flag is true and `params.stack` property doesn't exist.

Initially I went also with ternary operator ...

```ts
stack: debug ? params.stack || defaultParams.stack : defaultParams.stack
```

... but since it was a bit hard to read, I decided to put it into a separate `if` above the `this.error` assignment.

Does this make sense?

adrians5j (changes requested):
Good job, almost there! 🍻

adrians5j:
@brunozoric Sorry to pull you in.

I need to ask a favor - can you please double check last 4 commits (all on Mar 4, 2022), just so all is in order with the latest TS work you've been doing? From my side, I was just about to merge this PR, but now there are these new post-merge-with-next-branch changes, which I think it'd be best you double-check.

If all is good from your side, it's good to merge from my side.

Thanks :)

brunozoric:
@doitadrian ts is building, no new any types added, so its ok. went through comits and everything seems fine

brunozoric (approved):

PR #2141, opened 2021-12-21
fix(api-page-builder-import-export): pb refactoring

brunozoric:
## Changes
Page Builder import/export was failing due to changed locations of methods.

## How Has This Been Tested?
Jest and manual testing.


Ashu96 (approved):
LGTM 👍

PR #2144, opened 2021-12-23
feat(api-headless-cms): make graphql fields required if field has required validator

### LINKED ISSUE #2142: Make fields that are required non-nullable in GraphQL

We use TypeScript code generators for our GraphQL API that is backed by Webiny's, but the schema for fields are never marked as non-nullable (marked with `!`). This makes all the typings optional, even though the fields are required in the UI.

We are noticing this for both UI-defined models, as for models defined with the Content Model plugin.

### Describe the solution you'd like.

Ideally, all fields marked as required (with a required validation in the Content Model plugin), should have their typings set to non-nullable according to GraphQL specs.

Wrapping the Webiny schema and manually rewriting types to be non-nullable. This requires your own server that wraps Webiny and causes a hard-coupling between both.

brunozoric:
## Changes
Closes #2142

## How Has This Been Tested?
Jest and manually.

## Documentation
Add a note in the changelog.


StampixSMO:
@brunozoric We have updated our Webiny to 5.21.0 and we do not seem our existing fields that have a required validation in them, actually become required. Any idea?

Thanks!

brunozoric:
hi @StampixSMO 
I just created a testing model, which has title field as required, and here is the input in the graphql to create the entry:
```
input TestingInput {
  title: String!
  longText: String
  richText: JSON
}
```

StampixSMO:
@brunozoric Could it be that we have to re-create our models for this? We are updating this in our deployed environments, so perhaps the schemas are fixed at this point and the `!` is not coming through?

brunozoric:
@StampixSMO everything is generated on-the-fly, so you should not be updating your models.
Only thing you could possibly do is to just go into some model and save it again. Schema is cached, so let's see if that could be the issue. If problem persiss, write into our Community Slack #hepl channel.

PR #2177, opened 2022-01-06
feat(handler): wait for context property

brunozoric:
## Changes
A possibility to wait for the context property to be defined and then trigger a callback.

## How Has This Been Tested?
Jest


Pavel910 (approved):
This look ok, and solves a burning problem. I'll merge this into v5, and test with my current problem immediately.

PR #2193, opened 2022-01-13
fix: fixed an issue with admin app after webpack was upgraded

kedrzu:
## Changes
After upgrading webpack some views in admin app began to fail.
Turned out, there was an unnecessary alias for `react-dom` inside webpack config, that was adding hot reloading ability before `react-refresh`.

Also added support for sourcemaps, because some babel setting was missing.

## How Has This Been Tested?
Run the admin app, check content models and form builder views.


Pavel910 (approved):
Looks good, I think this is now ready for another cypress run.

PR #2204, opened 2022-01-19
fix(api-prerendering-service): injectApolloState escapes "<"

taifok:
## Changes
When there is any value in Apollo State is `</script>`, the original `injectApolloState` will inject "</script>" and break the page.

I just change it based on https://github.com/apollographql/apollo-client/issues/5567


Pavel910:
@taifok thank you! This will be included in the upcoming 5.23.! 👍

PR #2205, opened 2022-01-19
Simplify pulumi code

kedrzu:
## Changes
Goal is to make the pulumi code in users' projects smaller:
- remove all the boilerplate code, that is same for 95% of users
- preserve ability to configure the service down to every single AWS resource, if user wants to have more control

Idea is to pull all of pulumi code into separate components, that have good configurability and even possibility to override configs for every pulumi resource with a simple function.

Note, that this PR is made on a top of another one https://github.com/webiny/webiny-js/pull/2105, so not every change regards this particular issue. Please read comments to code - they will guide you through what's important.

## How Has This Been Tested?
Build & deploy admin app.


apps/admin/pulumi/index.ts
```diff
@@ -10,11 +9,50 @@ export = async () => {
...
     });
 
-    const app = new App();
-    const cloudfront = new Cloudfront({ appS3Bucket: app.bucket });
+    const bucket = new CloudFrontBucket("admin-app", {
+        bucketConfig(config) {
+            config.versioning = {
+                enabled: true
+            };
+        }
```
kedrzu, line 17: This is how you can override any resource configuration. You have some smart defaults, but you can change anything.

packages/pulumi-aws/src/components/CloudFrontBucket.ts
```diff
@@ -0,0 +1,70 @@
...
+import * as aws from "@pulumi/aws";
+import { PulumiInputValue } from "../types";
+import { createResource, ResourceOverride } from "./createResource";
+
+type OriginConfig = PulumiInputValue<
+    PulumiInputValue<aws.cloudfront.DistributionArgs["origins"]>[number]
+>;
+
+export interface CloudFrontBucketConfig {
+    bucketConfig: ResourceOverride<typeof aws.s3.Bucket>;
```
kedrzu: In resource config you can define functions to override any settings.

packages/pulumi-aws/src/components/CloudFrontBucket.ts
```diff
@@ -0,0 +1,70 @@
...
+        super("webiny:aws:CloudFrontBucket", name);
+
+        // Origin Identity is a kind of AWS user that represents Cloudfront distribution
+        // We can add IAM policies to it later, to allow accessing private S3 bucket
+        this.originIdentity = new aws.cloudfront.OriginAccessIdentity(`${name}-origin-identity`);
+
+        this.s3Bucket = createResource(aws.s3.Bucket, name, config?.bucketConfig, {
+            acl: aws.s3.CannedAcl.Private,
+            forceDestroy: true
+        });
```
kedrzu: This function you use instead of calling `new aws.s3.Bucket`. This is a syntactic sugar really. It will run config override logic and call constructor inside. Provides some nice typing.

apps/admin/pulumi/index.ts
```diff
@@ -10,11 +9,50 @@ export = async () => {
...
+    });
```
kedrzu: This is how you can override any resource configuration. You have some smart defaults, but you can change anything.

packages/pulumi-aws/src/apps/customDomain.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export function applyCustomDomain(
+    cloudfront: PulumiAppResource<typeof aws.cloudfront.Distribution>,
+    params: CustomDomainParams
+) {
+    cloudfront.config.aliases([params.domain]);
+    cloudfront.config.viewerCertificate({
+        acmCertificateArn: params.acmCertificateArn,
+        sslSupportMethod: params.acmCertificateArn
```
adrians5j: Can we double check this? Should this line really be `sslSupportMethod: params.acmCertificateArn`?

adrians5j: P.S. Pavel just sent me an example usage:

```
sslSupportMethod: "sni-only"
```

packages/pulumi-sdk/src/ApplicationBuilderGeneric.ts
```diff
@@ -0,0 +1,87 @@
...
+                }
+            },
+            {
+                workDir: pulumiWorkDir,
+                projectSettings: {
+                    name: this.config.name,
+                    runtime: "nodejs",
+                    description: this.config.description
+                },
+                secretsProvider: PULUMI_SECRETS_PROVIDER,
```
adrians5j, line 54: The `PULUMI_SECRETS_PROVIDER` expects `PULUMI_CONFIG_PASSPHRASE` to be also present. Should we send it via the `envVars` property?

tsconfig.build.json
```diff
@@ -6,7 +6,7 @@
     "allowSyntheticDefaultImports": true,
     "moduleResolution": "node",
     "module": "esnext",
-    "lib": ["esnext"],
+    "lib": ["esnext", "dom"],
```
adrians5j: With [this PR](https://github.com/webiny/webiny-js/pull/2317), we've added `@types/web`, so the `dom` addition should not be neccessary anymore.

More specific ref: https://github.com/webiny/webiny-js/blob/next/packages/app/package.json#L21

adrians5j (changes requested):
First iteration done. 

Please check a couple of comments @Pavel910 and I left.

packages/pulumi-sdk/src/ApplicationBuilderGeneric.ts
kedrzu: Currently `PULUMI_CONFIG_PASSPHRASE` is taken out of `process.env` and passed to execa `env` options:
image
Because `pulumi` automatically passes all `env` variables down the line, it's unnecessary to do it explicitly, however, it may be nice for the sake of readability.

tsconfig.build.json
kedrzu: Thanks, works now. I added this because I had compilation errors, but after merging `next` it went to normal.

packages/pulumi-aws/src/apps/customDomain.ts
kedrzu: You're right. Basically, you have access to `cloudfront` config during app configuration, so I added this just for the sake of simplicity for the user.

adrians5j: Cool 👍🏻

packages/pulumi-sdk/src/ApplicationBuilderGeneric.ts
adrians5j: OK 👍🏻

adrians5j (commented):
@Pavel910 and I gave this another look. Looks mergable.

But before doing that, will we tackle the separate ddb+es within this PR or? How will we approach it?

Thanks!

kedrzu:
> @Pavel910 and I gave this another look. Looks mergable.
> 
> But before doing that, will we tackle the separate ddb+es within this PR or? How will we approach it?
> 
> Thanks!

I created a PR regarding ES: https://github.com/webiny/webiny-js/pull/2340
Now it covers only infrastructure, because code part is entirely in the user's project, which makes it difficult to make generic.

PR #2215, opened 2022-01-27
Feat: CMS Create Ref Field Entry

Ashu96:
This PR introduces a new feature in `app-headless-cms` whereby users can create a new reference field entry(if not found) right from a content entry form without having to go back to the reference model entry form to create one. In other words,
users now have the ability to generate reference field entries on the fly.

## Changes
- updated both `refField` and `refFields` components to support the new feature
- updated `AutoComplete` and `MultiAutoComplete` components prop
- remove duplicate `helpText` from `fileField` component

## How Has This Been Tested?
Manually 

---

https://user-images.githubusercontent.com/13612227/149706459-06591e24-a040-4c84-be69-05464d42b968.mov

### DIFF
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/ref/components/MissingEntryHelpText.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/ref/components/MissingEntryHelpText.tsx
@@ -10,12 +10,20 @@ const t = i18n.ns("app-headless-cms/admin/fields/ref");
 const missingEntryLabel = t`If you can't find the intended reference value in the target model,
          please close this dialog and populate the {newEntryLink} in the target model first.`;
 
+const referenceMultipleModelsLabel = t`The creation of reference values from within this view is only supported
+ when a single reference model is selected. To reference values from multiple models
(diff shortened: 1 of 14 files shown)


packages/app-headless-cms/src/admin/plugins/fieldRenderers/ref/components/MissingEntryHelpText.tsx
```diff
@@ -10,13 +10,21 @@ const t = i18n.ns("app-headless-cms/admin/fields/ref");
 const missingEntryLabel = t`If you can't find the intended reference value in the target model,
          please close this dialog and populate the {newEntryLink} in the target model first.`;
 
+const referenceMultipleModelsLabel = t`The creation of reference values from within this view is only supported
+ when a single reference model is selected. To reference values from multiple models
+ please make sure the referenced values exist before setting the reference`;
```
Pavel910: Add a `,` before please, and a `.` at the end of the sentence.
```
To reference values from multiple models, please make sure the referenced values exist before setting the reference.
```

Pavel910 (changes requested):

Pavel910 (approved):

PR #2237, opened 2022-02-15
`@webiny/api-authentication-cognito` - `undefined undefined` Present In the `displayName` Property

adrians5j:
## Changes
Prior to this PR, upon retrieving identity information within `packages/api-authentication-cognito/src/index.ts:46`, if `given_name` or `family_name` is missing, the `undefined` string ends up in the `displayName` property. For example:

```json
{
  "id": "xyz",
  "email": "abc@webiny.com",
  "displayName": "undefined undefined"
}
```

From now on, if both `given_name` and `family_name` is missing, the `displayName` will be `null`. 

Also, if only one of the two is present, only one will be used as `displayName`. For example, if only `given_name` is present, `displayName` will be equal to it.

## How Has This Been Tested?
Jest.

## Documentation
Changelog.

### DIFF
--- a/packages/api-authentication-cognito/__tests__/getIdentity.test.ts
+++ b/packages/api-authentication-cognito/__tests__/getIdentity.test.ts
@@ -1,0 +1,80 @@
+import { getIdentity } from "@webiny/api-authentication-cognito";
+
+describe("getIdentity Test", () => {
+    test("full identity information should be returned correctly", async () => {
+        expect(
+            getIdentity({
+                token: {
+                    id: "xyz",
+                    family_name: "Family",
+                    given_name: "Given",
+                    email: "family@given.com"
+                },
+                identityType: "user"
+            })
+        ).toEqual({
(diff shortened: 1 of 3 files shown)


Pavel910 (approved):
Makes perfect sense. I'm approving the PR, but could you comment on where/how you ran into the problem with this? 🤔 Probably custom security?

adrians5j:
Yeah, sry forgot to do that.

In our app, users enter their name after they've completed the Cognito signup. 

And even though, immediately after the successful Cognito signup, users are presented with the personal info screen where they can enter their name, it's still possible to have a user who did not complete this step. The moment this happens, you start seeing these 'undefined undefined' in the UI, when displaying authors of actions, for example the creator of an environment.

P.S. We still have one issue to fix on our side and, actually, when that's resolved, this PR would most probably not matter that much to us. But ultimately, I still decided to tackle this particular edge  case.

PR #2246, opened 2022-02-22
Preparing for blue/green deployments

kedrzu:
## Changes
This is a change that prepares us to introduce blue/green deployment functionality.
It mainly consists of moving API to new pulumi architecture and extracting persistent resources into a separate app called `storage`.

## How Has This Been Tested?
Not tested yet, work in progress.

## Documentation
Will need one.


packages/pulumi-aws/src/apps/api/ApiPageBuilder.ts
```diff
@@ -0,0 +1,462 @@
...
+        createInstallationZip(pbInstallationZipPath);
+
+        new aws.s3.BucketObject("./pbInstallation.zip", {
+            key: "pbInstallation.zip",
+            acl: "public-read",
+            bucket: params.fileManagerBucketId,
+            contentType: "application/octet-stream",
+            source: new pulumi.asset.FileAsset(pbInstallationZipPath)
+        });
+    });
```
Pavel910, line 34: @kedrzu does this mean that it will be creating and uploading this zip on every deploy?

packages/pulumi-aws/src/apps/storage/StorageApp.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export interface StorageAppConfig extends Partial<ApplicationHooks> {
+    config?(app: StorageApp, ctx: ApplicationContext): void;
+    protect?(ctx: ApplicationContext): boolean;
+}
+
+export const StorageApp = defineApp({
+    name: "Storage",
+    config(app, config: StorageAppConfig) {
+        const protect = config.protect?.(app.ctx) ?? app.ctx.env === "dev";
```
Pavel910: `dev` should be `prod` here I think?

kedrzu: Thanks for finding it :) 
It should of course be `!==`. 

That is the question, do we assume, that by default there is one `prod` environment that is protected and all other are considered non-production, or we assume that there is `dev` environment, and all other are considered production?

kedrzu: The first seems to be more common, but the latter is safer.

packages/pulumi-aws/src/apps/api/ApiPageBuilder.ts
kedrzu: Probably - that's basically a rip off from here: https://github.com/webiny/webiny-js/blob/next/api/pulumi/dev/pageBuilder.ts#L35

packages/pulumi-aws/src/apps/storage/StorageApp.ts
Pavel910: As it is today, we assume all envs to be `dev` unless it's a `prod` env, but the developers can control that in their current project setups. With the new setup, I was looking at resolving it with that map I showed you, remember this?
```
devEnvironments: ["dev", "test"],
prodEnvironment: ["prod", "staging", "client"]
```

I'm glad we landed on this topic, and we can now discuss how do we want to approach this. Feel free to lay out your ideas on this topic.

packages/pulumi-aws/src/apps/api/ApiPageBuilder.ts
Pavel910: Ah got it, so it's the same as it currently is. That's cool. We'll revisit it later. Thanks.

PR #2254, opened 2022-03-01
Validate Headless CMS model SDL and model names

Pavel910:
## Changes
This PR introduces better Headless CMS content model schema validation, which validates schema integrity before saving content model to the database. On the UI side, we now have better validation of content model names, by utilizing GQL introspection and verifying the uniqueness of a GraphQL type.

I also added the `srcPrefix` to File Manager settings, as several users needed to update it, and it was previously not exposed neither via API nor UI.

image
image
image

## How Has This Been Tested?
Jest tests.

## Documentation
Not necessary.


packages/api-headless-cms/__tests__/contentAPI/contentModelToSDL.read.test.ts
```diff
@@ -5,9 +5,12 @@ import contentModels from "./mocks/contentModels";
 import categorySDL from "./snapshots/category.read";
 import productSDL from "./snapshots/product.read";
 import reviewSDL from "./snapshots/review.read";
+import { CmsModelFieldToGraphQLPlugin } from "~/types";
 
 describe("READ - ContentModel to SDL", () => {
-    const fieldTypePlugins = graphqlFieldPlugins().reduce((acc, pl) => {
+    const fieldTypePlugins = graphqlFieldPlugins().reduce<
+        Record<string, CmsModelFieldToGraphQLPlugin>
+    >((acc, pl) => {
```
endymion1818, line 13: is `pl` a standard definition we use elsewhere? Otherwise would it make it more self documenting to use the full name even though its more verbose?

packages/api-headless-cms/package.json
```diff
@@ -39,9 +39,11 @@
     "@webiny/utils": "^5.23.1",
     "@webiny/validation": "^5.23.1",
     "boolean": "^3.0.2",
+    "code-frame": "^5.0.0",
```
endymion1818: Is it worth using @babel/code-frame as there are type definitions for it and it has had more recent updates?

packages/app-headless-cms/src/admin/components/FieldEditor/EditFieldDialog.tsx
```diff
@@ -1,7 +1,7 @@
 import React, { useState, useEffect, useCallback } from "react";
 import { cloneDeep } from "lodash";
 import { css } from "emotion";
-import { Dialog, DialogContent, DialogTitle, DialogActions, DialogButton } from "@webiny/ui/Dialog";
+import { Dialog, DialogContent, DialogTitle, DialogActions } from "@webiny/ui/Dialog";
```
doitadrian, line 4: `DialogButton` sux :|

packages/api-headless-cms/package.json
Pavel910: @endymion1818 Bundle size difference is significant for something that will rarely kick in: 14Kb vs 700B. Honestly I'm not even sure we need this at all, but let's have that in the error response for now. TS types are irrelevant here, because it's a single function call. The function is so small that we could even have our own utility for it.

https://bundlephobia.com/package/@babel/code-frame@7.16.7
vs
https://bundlephobia.com/package/code-frame@5.0.0

packages/api-headless-cms/__tests__/contentAPI/contentModelToSDL.read.test.ts
Pavel910: We don't have a standard for variable names at all 😅

endymion1818: No worries maybe something we could look at at some point, but I do think it's worth being more verbose for the sake of clarity.

packages/api-headless-cms/package.json
endymion1818: Yes I saw that. OK no worries see your point 👍

packages/api-headless-cms/src/content/plugins/graphqlFields/object.ts
```diff
@@ -29,11 +29,8 @@ const typeFromField = (params: TypeFromFieldParams): TypeFromFieldResponse | nul
     const mTypeName = createManageTypeName(typeName);
 
     // `field` is an "object" field
-    const fields = field.settings.fields as CmsModelField[];
+    const fields: CmsModelField[] = field.settings.fields || [];
 
-    if (!fields || fields.length === 0) {
```
doitadrian, line 34: Just to confirm - we want to remove this check?

Asking because of the code below:

image

Looks like now there's a possibility that a type can be included in the schema, without any fields. Previously that wasn't possible because of the check that would return `null` if no fields are defined.

I might be missing something here.

adrians5j (changes requested):
Checked it. Liking the GraphQL schema construction on content model update. 👍🏻

adrians5j (approved):
Accidentally hit Request changes. Meant to approve. :)

packages/api-headless-cms/__tests__/contentAPI/contentModelToSDL.read.test.ts
Pavel910: It's a good point though, I'll keep that in mind for all the future code I'm writing. I think it has more to do with the laziness than any particular style :D

brunozoric (approved):

packages/api-headless-cms/package.json
Pavel910: @endymion1818 I reviewed this once more, and completely removed `code-frame`. Instead, I'm returning the full SDL, without any formatting. That way if something goes wrong, the developer can share the full SDL with us.

PR #2255, opened 2022-03-02
Feat: Add support for scoped files in File Manager using tags

Ashu96:
## Changes
This PR introduces the support for scoped files in File Manager using tags.
Now, we use `scope:` as the naming convention for tagging a file in a particular scope.
For example, a file with the tag `scope:apw` means it is scoped within the Advanced Publishing Workflow application.

And we'll use the new `tag_startsWith` and `tag_not_startsWith` properties in the `listFiles` and `listTags` GQL query to filter them.

## How Has This Been Tested?

Jest tests


brunozoric:
I'm still not 100% positive that this is the correct way to go. We will now have filtering on each list files or tags request, in both cases, either we request APW files or regular ones. And not sure what to do about it.
Otherwise this looks good.

brunozoric (approved):

PR #2262, opened 2022-03-08
fix(react-rich-text-renderer): Compile to commonJS

### LINKED ISSUE #2261: Unable to use @webiny/rich-text-renderer with NextJS applications
### Version

5.24.0

### Operating System

MacOS 12.2.1

### Browser

All

using react-rich-text-renderer with a NextJS application (using the blog starter: https://github.com/vercel/next.js/tree/canary/examples/blog-starter) results in the following error:

```
> Build error occurred
/Users/benjaminread/Developer/nextjs-starter-webiny/node_modules/@webiny/react-rich-text-renderer/index.js:1
import React from "react";
^^^^^^

SyntaxError: Cannot use import statement outside a module
```
This is happening because the module is not compiled to CommonJS

I'm using it like this in line with the documentation:
```
import { RichTextRenderer } from "@webiny/react-rich-text-renderer";

export default function PostBody({ content }) {
  return (
    <div className="max-w-2xl mx-auto">
      <div className={markdownStyles['markdown']}>
        <RichTextRenderer data={content} />
      </div>
    </div>
  )
}
```

I am able to use the package with applications such as NextJS

### What do you see instead?

App build step fails.

### Additional information

_No response_

### Possible solution

I'll submit a PR in a few minutes.

endymion1818:
## Benefits:
We can use this package outside the Webiny ecosystem (eg. for using the HeadlessCMS to build a
NextJS application)

## Changes
Closes #2261

## How Has This Been Tested?
I tested the build with a NextJS application and it works.

## Documentation
I'll add this to the Changelog

### DIFF
--- a/packages/react-rich-text-renderer/.babelrc.js
+++ b/packages/react-rich-text-renderer/.babelrc.js
@@ -1,1 +1,8 @@
-module.exports = require("../../.babel.react")({ path: __dirname });
+const defaults = require("../../.babel.react")({
+    path: __dirname
+});
+
+module.exports = {
+    ...defaults,
+    plugins: [...defaults.plugins, "@babel/plugin-transform-modules-commonjs"]
+};
--- a/packages/react-rich-text-renderer/package.json
+++ b/packages/react-rich-text-renderer/package.json
@@ -9,6 +9,7 @@
   "author": "Webiny Ltd",
   "license": "MIT",
   "dependencies": {
+    "@babel/plugin-transform-modules-commonjs": "^7.16.8",
     "@babel/runtime": "^7.16.3",
     "@editorjs/editorjs": "^2.20.1",
     "@types/react": "^16.14.0",
--- a/packages/react-rich-text-renderer/package.json
+++ b/packages/react-rich-text-renderer/package.json
@@ -9,7 +9,6 @@
   "author": "Webiny Ltd",
   "license": "MIT",
   "dependencies": {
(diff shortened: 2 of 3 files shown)


packages/react-rich-text-renderer/.babelrc.js
```diff
@@ -1 +1,8 @@
-module.exports = require("../../.babel.react")({ path: __dirname });
+const defaults = require("../../.babel.react")({
+    path: __dirname
+});
+
+module.exports = {
+    ...defaults,
+    plugins: [...defaults.plugins, "@babel/plugin-transform-modules-commonjs"]
```
Pavel910, line 7: @endymion1818 This babel plugin dependency needs to be added to the `package.json` for that respective package.

packages/react-rich-text-renderer/package.json
```diff
@@ -9,6 +9,7 @@
   "author": "Webiny Ltd",
   "license": "MIT",
   "dependencies": {
+    "@babel/plugin-transform-modules-commonjs": "^7.16.8",
```
Pavel910: @endymion1818 this should be in `devDependencies` because it's only relevant for us, at publish time. Once the package is published, users don't need this in their projects. It's not a dependency for the core package logic.

Pavel910 (changes requested):

Pavel910 (approved):

PR #2266, opened 2022-03-10
feat(api-headless-cms): get published or latest revision via manage get

brunozoric:
## Changes
User can now get either published or latest CMS entry via Manage API get query.
PR includes CMS Entry reference field entry status note.

Examples for article:
```
// exact revision
query GetArticle {
    getArticle(revision: "abc#0001") {
        ...
    }
}
// latest revision
query GetArticle {
    getArticle(entryId: "abc", status: latest) {
        ...
    }
}
// latest revision
query GetArticle {
    getArticle(entryId: "abc") {
        ...
    }
}
// published revision
query GetArticle {
    getArticle(entryId: "abc", status: published) {
        ...
    }
}
```

## How Has This Been Tested?
Jest and manually.


SvenAlHamad:
Functionally this meets the required criteria. My only worry is from the DX side that the param name `revision` is a bit misleading as the user is technically providing `entryId`.

packages/api-headless-cms/src/content/plugins/schema/resolvers/manage/resolveGet.ts
```diff
@@ -23,7 +23,7 @@ interface ArgsValues {
 const possibleTypes = ["published", "latest"];
 
 const getValuesFromArgs = (args?: ValuesFromArgsParams): ArgsValues => {
-    const { type, revision, entryId } = args || {};
+    const { status, revision, entryId } = args || {};
     if (!revision && !entryId) {
         throw new WebinyError(
             "Missing both of GraphQL query arguments: revision and entryId. Must have one.",
```
Pavel910:
```
You must pass a 'revision' or an 'entryId' argument.
```

packages/api-headless-cms/src/content/plugins/schema/resolvers/manage/resolveGet.ts
```diff
@@ -41,7 +41,7 @@ const getValuesFromArgs = (args?: ValuesFromArgsParams): ArgsValues => {
             }
         );
     }
-    if (type && possibleTypes.includes(type) === false) {
+    if (status && possibleTypes.includes(status) === false) {
         throw new WebinyError(
             `Status cannot be anything other than ${possibleTypes.join(", ")}.`,
```
Pavel910:
```
Status can only be one of the following values:  ${possibleTypes.join(" | ")}.
```


Btw. can we define `status` as an `enum`  so GraphQL takes care of the validation for us?

Pavel910 (changes requested):

packages/api-headless-cms/src/content/plugins/schema/resolvers/manage/resolveGet.ts
brunozoric: It is an enum (check [here](https://github.com/webiny/webiny-js/blob/b934fd6af046cb74af7d080e92bb8e6e89136e30/packages/api-headless-cms/src/content/plugins/schema/baseSchema.ts#L75)). Check in the code is just to be sure string passed is ok.

PR #2289, opened 2022-03-25
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
## Changes

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

## How Has This Been Tested?
I tested the return of each statement to make sure it's still the same as before


brunozoric:
Hi @CommanderRoot 
While slice works, correct solution would be `String.substring()`.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring

CommanderRoot:
While you can use `substr()`, `substring()` and `slice()` nearly the same way they all behave a bit differently.
`slice()` is [imo stricter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring#differences_between_substring_and_slice) than `substring()` as it doesn't move input parameters around. It's also generally a bit faster.
Both have their usecase, especially if the second parameter is normally positive but can be negative, but generally `slice()` is the better option.

PR #2299, opened 2022-03-30
Disable app scaffolds

Pavel910:
## Changes
This PR disables the 3 scaffolds  for new v5 projects: full-stack apps, graphql api, and react app.
These 3 will not be supported in v6, so by disabling these, we're reducing the blast radius

## Documentation
By disabling these scaffolds, we're also retiring [this tutorial](https://www.webiny.com/docs/custom-app-tutorial/overview/introduction).


adrians5j (approved):
My only question is - do we want to do the same in our repo, in our `webiny.project.ts` ?

Pavel910:
> My only question is - do we want to do the same in our repo, in our `webiny.project.ts` ?

Right, yes. Will do that.

PR #2300, opened 2022-03-30
Fix composition API types

Pavel910:
## Changes
This PR fixes some TS issues with the `Compose` component and `makeComposable` utility.

## How Has This Been Tested?
By using the `<Compose/>` component in the Admin  app.


brunozoric (approved):
Everything looks ok.

PR #2301, opened 2022-03-30
Minor fixes of watch logs, CMS context, and PB lifecycle events.

### LINKED ISSUE #2296: Admin app - update referral links
### Version

5.24.0

### Operating System

MacOS 12.2.1

### Browser

Chrome

In Admin App main view or menu:

- Click on the DOCUMENTATION button
- Click on the TWITTER button 

Link addresses need to be updated based on the recent changes to Webiny assets: 

Documentation button should link to: [https://www.webiny.com/docs/](url)
Twitter button should link to: [https://twitter.com/webinycms](url)

image (4)

image (5)

### What do you see instead?

The links are currently pointing to old URLs that are no longer active:

- docs.webiny.com (old subdomain where docs used to be)
- twitter.com/WebinyPlatform (old Twitter handle, no longer used)

### Additional information

_No response_

### Possible solution

_No response_

Pavel910:
## Changes
Closes #2296

### Logging
This PR fixes console.log aggregation and forwarding to the browser/terminal.
> NOTE: `localtunnel` is glitchy and it can occasionally fail to send the message to your local listener. There's nothing we can do here; I think it's best to ditch it and only have the browser-based logs (in the Dev Tools).

### Other
- fix execution of the `onAfterPageUpdate` lifecycle event in the Page Builder. 
- prevent Headless CMS context construction on OPTIONS request (@brunozoric not sure why this is not handled by your recent addition of request abortion on OPTIONS requests; please check it out during PR review)

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/api-file-manager-s3/tsconfig.build.json
+++ b/packages/api-file-manager-s3/tsconfig.build.json
@@ -3,9 +3,9 @@
   "include": ["src"],
   "references": [
     { "path": "../api-file-manager/tsconfig.build.json" },
+    { "path": "../error/tsconfig.build.json" },
     { "path": "../handler-graphql/tsconfig.build.json" },
-    { "path": "../validation/tsconfig.build.json" },
-    { "path": "../error/tsconfig.build.json" }
+    { "path": "../validation/tsconfig.build.json" }
   ],
   "compilerOptions": {
     "rootDir": "./src",
--- a/packages/api-file-manager-s3/tsconfig.json
+++ b/packages/api-file-manager-s3/tsconfig.json
@@ -3,9 +3,9 @@
   "include": ["src", "__tests__/**/*.ts"],
   "references": [
     { "path": "../api-file-manager" },
+    { "path": "../error" },
     { "path": "../handler-graphql" },
-    { "path": "../validation" },
-    { "path": "../error" }
+    { "path": "../validation" }
   ],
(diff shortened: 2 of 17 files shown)


adrians5j:
Regarding ditching the `-r` functionality... Still not 100% on that one. I was actually using it just now (was debugging a Cognito trigger) and was working OK. Webiny version: 5.22.0. 

Maybe we could add some additional error handling / logging so that the user at least knows what's going on.

Also, who knows, maybe this could potentially be a WCP feature in the future, where we'll be responsible for running the `localtunnel` server.

All in all, still on the fence whether we should ditch this or not.

brunozoric:
@Pavel910
Regarding the OPTIONS request, there is a context plugin which stops the request: https://github.com/webiny/webiny-js/blob/fix/v5/5.25.0-beta.3/packages/api-headless-cms/src/index.ts#L42
There is a test for that: https://github.com/webiny/webiny-js/blob/fix/v5/5.25.0-beta.3/packages/api-headless-cms/__tests__/contentAPI/httpOptions.test.ts

We should have some more universal solution for it. Maybe introduce `before-context` plugin type so we can do things before context plugins start loading.

brunozoric (approved):
We can have checks for the OPTIONS method in the code, no harm to have them, but execution should never reach that code when OPTIONS method.

PR #2319, opened 2022-04-08
feat(page-builder): iframe element

endymion1818:
## Changes
Feature: add new page builder Iframe element

## How Has This Been Tested?
Deployed on AWS

## Documentation
To follow

## Ticket
https://webiny.atlassian.net/browse/WEB-958

## How it Looks
Main settings:
Screen Shot 2022-04-11 at 13 20 27

Plugin settings:
Screen Shot 2022-04-11 at 13 20 40


packages/app-page-builder/src/editor/plugins/elements/media/iframe/IFrame.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+import styled from "@emotion/styled";
+import { ReactComponent as IFrameIcon } from "./iframe-icon.svg";
+import { PbEditorElement } from "~/types";
+import { PageBuilderContext } from "../../../../../contexts/PageBuilder";
```
Pavel910: You can use the `~/` to point to the root of the package. See the `~/types` import above.

packages/app-page-builder/src/render/plugins/elements/iframe/index.tsx
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+import kebabCase from "lodash/kebabCase";
+import IFrame from "./IFrame";
+
+import { PbRenderElementPluginArgs, PbRenderElementPlugin } from "../../../../types";
```
Pavel910: Use the `~/types` import here as well.

Pavel910 (changes requested):
Left a few minor comments. Please attach some screenshots to the PR body, so we can see what the toolbar icons look like.

packages/app-page-builder/src/types.ts
```diff
@@ -119,7 +119,9 @@ export interface PbElementDataSettingsType {
     margin?: PbElementDataSettingsMarginType;
     padding?: PbElementDataSettingsPaddingType;
     height?: {
-        value?: number;
+        [key: string]: {
```
adrians5j: Can you explain this change? Was this a bug? I believe so, but can you remind me?

packages/app-page-builder/src/editor/plugins/elements/media/iframe/IFrameSettings.tsx
```diff
@@ -0,0 +1,87 @@
...
+        handler.trigger(
+            new UpdateElementActionEvent({
+                element,
+                history: true
+            })
+        );
+    };
+
+    const updateSettings = (data: LinkSettingsFormData) => {
+        // Skip update if nothing is change.
```
adrians5j: Nit: small typo (*changed). :)

adrians5j (changes requested):
Code looks good I'd say. Nothing out of the ordinary. Left two minors comments.

I tested it also, found two issues:

1. Visually, on my end at least, the iframe width is reduced to ~300px.
image

2. For some reason, when I add a new iframe, ELEMENTS tabs is automatically focused:

image

On top of the two issues above, we should probably also think about the upgrade script and also updating the upcoming release's change log.

Hope it makes sense. :)

packages/app-page-builder/src/types.ts
endymion1818: Ah good point ... it's not needed anymore, I was previously using 

```
const elementHeight = settings.height[displayMode].value
```

But now with dotProp's `get()` I can revert it.

Pavel910: @endymion1818 I think this is not a bad thing have in types. But, it would be better to use the DisplayMode enum as a key, not `string`. If you just revert this back, you'll have `value?: number`, which is wrong, since you have that DisplayMode key.

endymion1818: After discussion I've added this back as `[key in DisplayMode]` using the `DisplayMode` enum.

Pavel910 (approved):
All good! @endymion1818

packages/app-page-builder/src/editor/plugins/elements/media/iframe/IFrameSettings.tsx
```diff
@@ -0,0 +1,97 @@
...
+        );
+    };
+
+    const updateSettings = (data: LinkSettingsFormData) => {
+        // Skip update if nothing is changed.
+        if (data.url === url) {
+            return;
+        }
+
+        const isValidUrl = data.url && urlRegex.test(data.url);
```
adrians5j: Let's use the `@webiny/validation` package here, which is already imported in this file.

```
const isValidUrl = data.url && validation.validateSync(data.url, "url:allowHref");
```

adrians5j (changes requested):
On the code level, there's one thing we could do better.

On the frontend level, I've detected the following bug:

In the PB editor, I can see the iFrame in full width (correct):
image

But in the preview mode, or when I publish the page, that is not the case, it's fixed width:

image

Published page (https://d23k2vh7fc8l5r.cloudfront.net/untitled-l2r4tgf7/):
image

We should double check that as well before we can merge this.

packages/app-page-builder/src/editor/plugins/elements/media/iframe/IFrameSettings.tsx
endymion1818: Oh nice one thanks

adrians5j (approved):
Looks good!

Thank you @endymion1818 ;)

PR #2327, opened 2022-04-15
Extracting storage app from API

kedrzu:
## Changes
Due to working on staged rollouts, we need to extract an additional application from API. It will contain resources, that are singletons in a multi-stage setup, such as DynamoDB, Cognito User Pool, and File Manager S3 Bucket. 

In the future it will also contain an event bus, so maybe `storage` is not the best name here. Probably `core`?

## How Has This Been Tested?
Manual tests.

## Documentation
TBD


apps/storage/webiny.application.ts
```diff
@@ -0,0 +1,8 @@
+/**
+ * For more information on the Website project application, please see:
```
adrians5j: The comment is not relevant.

packages/cli-plugin-deploy-pulumi/commands/deploy.js
```diff
@@ -82,6 +82,17 @@ module.exports = async (inputs, context) => {
         pulumi: pulumi
     });
 
+    if (stack.app) {
```
adrians5j, line 85: Just a thought - is it possible that this context switching issue is happening because of the fact that we're using both Pulumi Automation API and our old Pulumi SDK at the same time?

packages/pulumi-aws/src/apps/admin/AdminApp.ts
```diff
@@ -66,9 +65,17 @@ export const AdminApp = defineApp({
 
         app.addOutputs({
             appStorage: bucket.bucket.output.id,
+            appDomain: cloudfront.output.domainName,
             appUrl: cloudfront.output.domainName.apply(value => `https://${value}`)
         });
 
+        app.onDeploy(async ({ outputs }) => {
```
adrians5j: Just a question here - why is the `adminUpload` being called via the `onDeploy` and not the `onAfterDeploy` hook?

Maybe it's just a naming inconsistency? 

If so, in order to be more explicit and consistent with the rest of the code, can we rename `onDeploy` to `onAfterDeploy`.

packages/pulumi-aws/src/apps/admin/AdminApp.ts
```diff
@@ -89,14 +96,15 @@ export function createAdminApp(config: AdminAppConfig & ApplicationConfig<AdminA
...
+            const app = new AdminApp(ctx);
+            await app.setup(config);
+            await config.config?.(app, ctx);
             return app;
         },
         beforeBuild: config.beforeBuild,
         afterBuild: config.afterBuild,
         beforeDeploy: config.beforeDeploy,
-        afterDeploy: mergeAppHooks(adminUpload, config.afterDeploy)
+        afterDeploy: config.afterDeploy
```
adrians5j: In general, the naming convention we use when declaring these kind of event handlers is **on{before/after}{ActionName}.** 

In other words, in our case this would be `onBeforeBuild`, `onAfterBuild` ...

Can we rename these in order to be more consistent?

packages/pulumi-aws/src/apps/api/ApiFileManager.ts
```diff
@@ -0,0 +1,174 @@
...
+// @ts-ignore
+import { getLayerArn } from "@webiny/aws-layers";
+import { PulumiApp } from "@webiny/pulumi-sdk";
+
+import { Vpc } from "./ApiVpc";
+import { createLambdaRole } from "./ApiLambdaUtils";
+
+interface FileManagerParams {
+    fileManagerBucketId: pulumi.Input<string>;
+    awsAccountId: pulumi.Input<string>;
```
adrians5j: I see `awsAccountId` is being defined here but not used anywhere. Is this OK? Can we double check?

packages/pulumi-aws/src/apps/api/ApiFileManager.ts
```diff
@@ -0,0 +1,174 @@
...
+            bucket: params.fileManagerBucketId,
+            lambdaFunctions: [
+                {
+                    lambdaFunctionArn: manage.output.arn,
+                    events: ["s3:ObjectRemoved:*"]
+                }
+            ]
+        },
+        opts: {
+            dependsOn: [/* TODO this.bucket, */ manage.output, manageS3LambdaPermission.output]
```
adrians5j: Should we finalize these TODOs here?

packages/pulumi-aws/src/apps/storage/StorageApp.ts
```diff
@@ -0,0 +1,77 @@
...
+        const cognito = createCognitoResources(app, {
+            protect,
+            useEmailAsUsername: legacyConfig.useEmailAsUsername ?? false
+        });
+
+        // Setup file storage bucket
+        const fileManagerBucket = createFileManagerBucket(app, { protect });
+
+        app.addOutputs({
+            fileManagerBucketId: fileManagerBucket.output.id,
```
adrians5j, line 41: I can see we have DynamoDB part covered here. BTW what's going on with the ElasticSearch DB?

Remember - we have two Webiny setups: DynamoDB-only and DynamoDB+ElasticSearch. Do we have both setups covered here or are we missing the latter one?

packages/pulumi-aws/src/apps/storage/StorageApp.ts
```diff
@@ -0,0 +1,77 @@
...
+            primaryDynamodbTableArn: table.output.arn,
+            primaryDynamodbTableName: table.output.name,
+            cognitoUserPoolId: cognito.userPool.output.id,
+            cognitoUserPoolArn: cognito.userPool.output.arn,
+            cognitoUserPoolPasswordPolicy: cognito.userPool.output.passwordPolicy,
+            cognitoAppClientId: cognito.userPoolClient.output.id
+        });
+
+        return {
+            table,
```
adrians5j: Maybe to reduce the amount of changes between legacy and new setups, let's rename the export here to `dynamoDbTable`.

You can check the existing export names in `packages/cwp-template-aws/template/ddb-es/api/pulumi/dev/index.ts:156`. I think being consistent with these should be the best thing to do so that we minimize the possibility of something going wrong upon doing the migration.

packages/pulumi-sdk/src/ApplicationBuilder.ts
```diff
@@ -20,9 +21,10 @@ export interface ApplicationBuilderConfig extends Partial<ApplicationHooks> {
 }
 
 export interface ApplicationStack {
-    refresh(): Promise<void>;
-    preview(): Promise<void>;
-    up(): Promise<void>;
+    app?: PulumiApp;
+    refresh(): Promise<any>;
+    preview(): Promise<any>;
```
adrians5j: Should this be `Promise<void>`?

adrians5j (changes requested):
First iteration done. 

Please check a couple of comments @Pavel910 and I left.

packages/cli-plugin-deploy-pulumi/commands/deploy.js
kedrzu: Nope. Under the hood Automation API uses pulumi CLI as a child process, so basically, that's just a more of a wrapper, than a completely different approach.
Context switching is because we have different pulumi login for every app, so running `getStackOutput` runs `pulumi login` and changes current folder for state files.

packages/pulumi-aws/src/apps/admin/AdminApp.ts
kedrzu: Eventually, I wanted to go for a more functional approach, in which the application is fully set up using this kind of handlers, so any plugin could easily add its own hooks without the need to register the plugin globally and use `if` clauses as is currently done.

Deploy hooks are very CLI-centric - every CLI parameter is passed, if you want to use app output, you need to explicitly retrieve it with `getStackOutput`, it has no typing for params or context. If I wanted to have a consistent way of passing some configurations to app hooks it's much easier to rely on closures we have in the app config function than passing it all the way around, so it would be available to hooks.

kedrzu: I will rename it to `onAfterDeploy`.

packages/pulumi-aws/src/apps/admin/AdminApp.ts
kedrzu: Sure, but I think I need to do it in the upstream branch (https://github.com/webiny/webiny-js/pull/2205).

packages/pulumi-aws/src/apps/api/ApiFileManager.ts
kedrzu: True, thanks for pointing this out.

packages/pulumi-aws/src/apps/api/ApiFileManager.ts
kedrzu: The file manager bucket is no longer inside API, so it cannot be passed as a dependency.
Is it safe to remove the dependency?

packages/pulumi-aws/src/apps/storage/StorageApp.ts
kedrzu: Right, elastic was not covered yet. I think I need to do it in a separate PR, because it's quite a big task.

packages/pulumi-aws/src/apps/storage/StorageApp.ts
kedrzu: Sure, just remember that this is not pulumi output - it's a reference to the application resource to be configured by user or used for something else.

kedrzu: Outputs are added with `app.addOutputs`.

packages/pulumi-sdk/src/ApplicationBuilder.ts
kedrzu: `stack.up()` in pulumi automation returns an object with a lot of data, that may be useful in a deploy script, so I basically passed the result back to the caller. Because Automation API behaves differently than our pulumi wrapper, I couldn't add any meaningful typings right now.
We don't use the result now, but I had some experiments with storage app state files migration and eventually left it like that. We can get rid of it though.

packages/cli-plugin-deploy-pulumi/commands/deploy.js
adrians5j: OK

packages/pulumi-aws/src/apps/api/ApiFileManager.ts
adrians5j: Spoke to @Pavel910 about this. We can remove the TODO.

packages/pulumi-aws/src/apps/storage/StorageApp.ts
adrians5j: What was the approach you decided to take?
Maybe it would be better if we would jump on a call on Monday?

packages/pulumi-sdk/src/ApplicationBuilder.ts
adrians5j: We went through the code a bit.

We think it's better to make it generic / be able to specify the return type upon using the `up` method and probably other methods (`refresh` / `preview`) as well.

So something like:

```ts
await up<{cognitoUserPoolId: string, dynamoDbTableNane: string, ....}>();
```

Does that make sense?

We understand that there are legacy and generic "drivers", but in any case, we both think we should not use `any` nor `void` as a return type here.

adrians5j (commented):
We left some commends. Please take a look a them when you get a chance.

packages/pulumi-sdk/src/ApplicationBuilder.ts
kedrzu: Added types from pulumi automation API. For legacy version it will return undefined.

apps/admin/webiny.application.ts
```diff
@@ -8,7 +8,7 @@ import { createAdminApp } from "@webiny/pulumi-aws";
 export default createAdminApp({
     config(app) {
         app.bucket.config.versioning({
-            enabled: false
+            enabled: true
```
adrians5j: Curious... why is this enabled?

adrians5j (changes requested):
@Pavel910 and gave this another look. All looks good, except one minor thing. Can you double check?

Thanks!

apps/admin/webiny.application.ts
kedrzu: I was fiddling with configurations to force some changes on deployment and committed this accidentally.

The whole configuration in this file is optional, for example, API just runs `createApiApp()` without passing any params. 
I left some admin config params just as an example.

kedrzu: By default versioning is disabled, this is just to show how to change default resource config.

adrians5j (approved):
@Pavel910 and I had another look.

Looks good, thanks!

PR #2328, opened 2022-04-15
Website prerendering using event bus

kedrzu:
## Changes
Working on staged rollouts requires some changes in how we deal with prerendering. Our new approach uses an event bus instead of directly coupling API and website apps.

- API will not know about the website delivery bucket and delivery CloudFront
- Any prerendering requests will be pushed to the event bus
- Prerendering logic is moved from API to the website and connected to the event bus
- We are moving from a custom DB-based queue to one implemented using AWS SQS for superb parallel processing, automatic deduplication, and auto-retry.

## How Has This Been Tested?
Manual tests.

## Documentation
TBD


packages/api-prerendering-service-aws/src/flush/flush.ts
```diff
@@ -0,0 +1,39 @@
...
+import { ArgsContext } from "@webiny/handler-args/types";
+import { Context, HandlerPlugin as DefaultHandlerPlugin } from "@webiny/handler/types";
+
+export type HandlerArgs = EventBridgeEvent<"FlushPages", Args | Args[]>;
+
+export interface HandlerContext extends Context, ArgsContext<HandlerArgs> {
+    //
+}
+
+export type HandlerPlugin = DefaultHandlerPlugin<HandlerContext>;
```
adrians5j, line 14: Can we utilize the `HandlerPlugin` class (from `@webiny/handler` package)?

adrians5j: P.S. The same goes for other handlers.

packages/api-prerendering-service-aws/src/render/subscriber.ts
```diff
@@ -0,0 +1,117 @@
...
+                    if (render.args) {
+                        addRender(render.args);
+                    }
+                }
+            } else {
+                addRender(event);
+            }
+
+            const entries: SendMessageBatchRequestEntry[] = [];
+            // TODO chunk splitting in case of big amount of events
```
adrians5j: Is this TODO still necessary? I can see we have some chunking going on below (L85).

adrians5j (changes requested):
Just some minor nitpicks, but overall the PR looks good.

packages/api-prerendering-service-aws/src/render/subscriber.ts
kedrzu: You're right, I think it went unnoticed on merge

packages/api-prerendering-service-aws/src/flush/flush.ts
kedrzu: Sure, but there is not a single example of how to do that. The only one place that uses `HandlerPlugin` from `@webiny/handler` is `handler.test.ts`.

adrians5j (approved):
@Pavel910 and I gave this another look.

Looks good, we can merge this,

PR #2329, opened 2022-04-15
Staged rollouts infrastructure

kedrzu:
## Changes
Allows to deploy application to multiple stages (variants) and randomly split traffic to variants. Allows better feature rollouts, testing new versions in production, and easy webiny upgrades.

Currently, it's not yet a finished feature. It misses:
- connecting this to WCP
- merging with the new project set up by @Pavel910 

## How Has This Been Tested?
Manual tests.

## Documentation
TBD


apps/website/code/webiny.config.ts
```diff
@@ -31,7 +31,13 @@ export default {
         async build(options) {
             invariant(options.env, NO_ENV_MESSAGE);
 
-            const output = getStackOutput({ folder: "api", env: options.env, map: API_MAP });
+            const output = getStackOutput({
```
adrians5j, line 39: Are we missing variant inclusion within the `watch` command?

gateway.admin.dev.json
```diff
@@ -0,0 +1,10 @@
+{
+    "1": {
+        "domain": "dc6ql0qliia7g.cloudfront.net",
```
adrians5j: These are test values, right? These will go away once we have WCP API integration ready to go?

packages/api-prerendering-service-aws/src/render/subscriber.ts
```diff
@@ -36,6 +36,13 @@ export default (params: HandlerConfig): HandlerPlugin => {
 
             const event = context.invocationArgs.detail;
             const namespace = event.configuration?.db?.namespace ?? "";
+            const variant = event.variant;
+
+            // Check if there is only specific variant rerender is requested.
+            // TODO: pass VARIANT variable in some other way
```
adrians5j: Is this TODO still relevant here?

packages/api-prerendering-service-aws/src/render/subscriber.ts
```diff
@@ -36,6 +36,13 @@ export default (params: HandlerConfig): HandlerPlugin => {
...
+            if (variant && variant !== process.env.VARIANT) {
```
adrians5j: Should we be a bit more explicit when it comes to naming the `VARIANT` env var?

Can we do something like `STAGED_ROLLOUTS_VARIANT` here?

packages/api-prerendering-service/src/render/absoluteAssetUrls.ts
```diff
@@ -0,0 +1,39 @@
+import { NodeAPI, Node } from "posthtml";
+
+export default () => async (tree: NodeAPI) => {
+    // TODO pass this via params or IoC
```
adrians5j: Is this TODO relevant?

packages/api-prerendering-service/src/render/absoluteAssetUrls.ts
```diff
@@ -0,0 +1,39 @@
+import { NodeAPI, Node } from "posthtml";
+
+export default () => async (tree: NodeAPI) => {
```
adrians5j, line 4: Can we do a comment here explaining why we're doing all of this?

packages/aws-helpers/src/cloudfrontFunctions/cookies.ts
```diff
@@ -0,0 +1,31 @@
+import { CloudFrontResponse } from "./types";
+
+export interface CookieParams {
+    name: string;
+    value: string;
+    maxAge?: number;
+}
+
+export function setCookie(response: CloudFrontResponse, cookie: CookieParams) {
```
adrians5j: Let's leave a couple of comments here. For example, can we explain what `multivalue` is?

Also, let's put a brief description about what this function does and how it fits the bigger picture. Feel free to link any of the existing Notion docs.

packages/aws-helpers/src/cloudfrontFunctions/utils.ts
```diff
@@ -0,0 +1,13 @@
+import { CloudFrontRequestHandler, CloudFrontResponseHandler } from "./types";
+
+declare const global: typeof globalThis & {
+    handler: Function;
+};
+
+export function defineCloudfrontFunctionRequestHandler(handler: CloudFrontRequestHandler) {
```
adrians5j, line 15: Can you explain why we're doing this? Why are we setting a handler function into a `global` object?

Looks a bit weird, that's all. :)

packages/aws-helpers/src/stagedRollouts/functions/adminOriginRequest.ts
```diff
@@ -0,0 +1,31 @@
...
+            const variant = result.variant;
+
+            // For file requests we just pass the request to proper origin.
+            if (pointsToFile(request.uri)) {
+                setDomainOrigin(request, variant.domain);
+                return request;
+            }
+
+            // For pages we make a custom HTTP request to the origin and transform page properly.
+            // For example we change asset URLs to be absolute.
```
adrians5j, line 23: Curious about this comment here:
```
// For example we change asset URLs to be absolute.
```

Couldn't we adapt the URLs in the pre-rendering process? I mean, we already do some HTML manipulation there, so my first thought is this should've been done there.

packages/aws-helpers/src/stagedRollouts/functions/viewerRequest.ts
```diff
@@ -0,0 +1,47 @@
...
+
+    // Otherwise we try to retrieve randomized number from user cookie.
+    // This random value will be passed further to the origin request to select based on config.
+    let variantRandom = Number(request.cookies?.[variantRandomKey]?.value);
+    if (isNaN(variantRandom) || variantRandom < 1 || variantRandom > 100) {
+        // If no value is present we simply randomize one.
+        variantRandom = Math.floor(Math.random() * 100 + 1);
+    }
+
+    // TODO take interval from config
```
adrians5j: Is this TODO still relevant?

packages/aws-helpers/src/stagedRollouts/functions/viewerRequest.ts
```diff
@@ -0,0 +1,47 @@
...
+        delete request.headers[variantRandomKey];
+        return request;
+    }
+
+    // Otherwise we try to retrieve randomized number from user cookie.
+    // This random value will be passed further to the origin request to select based on config.
+    let variantRandom = Number(request.cookies?.[variantRandomKey]?.value);
+    if (isNaN(variantRandom) || variantRandom < 1 || variantRandom > 100) {
+        // If no value is present we simply randomize one.
+        variantRandom = Math.floor(Math.random() * 100 + 1);
```
adrians5j, line 38: Can you explain why the +1 here (add comment)? 🤔

packages/aws-helpers/src/stagedRollouts/utils/loadConfig.ts
```diff
@@ -0,0 +1,72 @@
...
+                res.on("end", () => {
+                    resolve(JSON.parse(dataString));
+                });
+            }
+        );
+
+        req.on("error", e => {
+            reject({
+                statusCode: 500,
+                body: e.message.substring(0, 100)
```
adrians5j: Can you explain why `.substring(0, 100)` ?

packages/aws-helpers/src/stagedRollouts/utils/loadConfig.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export interface VariantConfig {
+    domain: string;
+    weight: number;
+}
+
+export type GatewayConfig = Record<string, VariantConfig>;
+
+export async function loadConfig(event: CloudFrontRequestEvent) {
+    const domain = event.Records[0].cf.config.distributionDomainName;
```
adrians5j: Can you explain the `domain` here? What is it?

packages/aws-helpers/src/stagedRollouts/utils/loadConfig.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    return config.config;
+}
+
+function loadConfigCore(domain: string) {
+    return new Promise<GatewayConfig>((resolve, reject) => {
+        let dataString = "";
+
+        const req = get(
```
adrians5j: Ultimately, is this supposed to load the config from WCP API?

packages/aws-helpers/src/stagedRollouts/utils/loadOriginPage.ts
```diff
@@ -0,0 +1,96 @@
...
+import { isHeaderBlacklisted } from "./headerBlacklist";
+import { logDebug } from "./log";
+
+export function loadOriginPage(domain: string, path: string) {
+    return new Promise<CloudFrontResponse>((resolve, reject) => {
+        logDebug(`Pulling page from ${domain}${path}`);
+        let responseBody = "";
+        const req = get(
+            {
+                hostname: domain,
```
adrians5j, line 20: Can we add a comment here explaining the `domain` variable. Is this a domain of a particular variant?

packages/aws-helpers/src/stagedRollouts/utils/loadOriginPage.ts
```diff
@@ -0,0 +1,96 @@
...
+    doc("script").each((_i, el) => {
+        prefixUrl(el, "src", host);
+    });
+
+    doc("body")
+        .find("img")
+        .each((_i, el) => {
+            prefixUrl(el, "src", host);
+
+            // TODO handle srcset
```
adrians5j: Should we resolve this comment now?

packages/aws-helpers/src/stagedRollouts/utils/loadOriginPage.ts
```diff
@@ -0,0 +1,96 @@
...
+                    logDebug(`Page parsed`);
+                    resolve(response);
+                });
+            }
+        );
+
+        req.on("error", e => {
+            reject({
+                statusCode: 500,
+                body: e.message.substring(0, 100)
```
adrians5j: Same question here - `.substring(0, 100)`.

packages/aws-helpers/src/stagedRollouts/utils/loadVariantOrigin.ts
```diff
@@ -0,0 +1,83 @@
...
+export async function loadVariantOrigin(event: CloudFrontRequestEvent) {
+    const cf = event.Records[0].cf;
+    const request = cf.request;
+
+    if (request.uri === configPath) {
+        // requesting the config file, pass it through
+        return request;
+    }
+
+    const config = await loadConfig(event);
```
adrians5j: Looking at the `loadConfig` name, doesn't tell us much. Would it be useful maybe to be a little bit more explicit when naming this function?

packages/aws-helpers/src/stagedRollouts/utils/common.ts
```diff
@@ -0,0 +1,11 @@
+export const variantFixedKey = "webiny-variant-fixed";
+export const variantRandomKey = "webiny-variant-random";
```
adrians5j, line 2: 1st question - these are used for naming cookies right? If so, can we include the word `cookie` in the const name, just so the code is a bit easier to read / understand.

2nd question - do we really need two names here? Would it be possible to simply have `webiny-variant` and that's it?

packages/aws-helpers/src/stagedRollouts/utils/loadConfig.ts
adrians5j: Additionally, we may want to have this function named a bit more explicit (for example, if this will be loading info from WCP, then `Wcp` in the function name might be useful). Alternatively, we can add a short comment describing what the function does.

packages/aws-helpers/src/stagedRollouts/utils/loadVariantOrigin.ts
```diff
@@ -0,0 +1,83 @@
...
+        // If no variant is matching the random value, just return 404.
+        // This should happen only if there is really not a single variant serving traffic.
+        logDebug(`No variant is found`);
+        return notFoundResponse(`No variant is found`);
+    }
+
+    return { variant: variantConfig };
+}
+
+function getRandomVariant(config: GatewayConfig, random: number) {
```
adrians5j, line 49: Just a note (maybe you can provide feedback if needed).

From the WCP side, the idea was to return an object that might look something like this:

```json
[
  { "name": "variant1", "weight": 0.3 },
  { "name": "variant2", "weight": 0.5 },
  { "name": "variant3", "weight": 0.2 }
]
```

So, an object that includes all variants and also weights (configured on the WCP side). 

Will this be OK? Let me know if you need more info.

One thing I'm also curious about is: do you need all of the domains of all deployed apps that the variant encapsulates (one variant has 3 domains - API, admin, website).

packages/aws-helpers/src/stagedRollouts/utils/loadVariantOrigin.ts
```diff
@@ -0,0 +1,83 @@
...
+    let totalWeight = 0;
+
+    const stages = Object.keys(config);
```
adrians5j: What are `stages`? 🙂 

Are these variants?

packages/aws-helpers/src/stagedRollouts/utils/log.ts
```diff
@@ -0,0 +1,7 @@
+export function logDebug(...args: any[]) {
+    // If DEBUG variable is not set during build,
+    // this function should be minified to an empty one and removed completely from the bundle.
+    if (process.env.DEBUG === "true") {
```
adrians5j, line 4: I've noticed we've been using this helper to log both regular logs and errors. 

Can we double check....would it be better if we're logging an error, to use `console.error` maybe?

packages/cli-plugin-deploy-pulumi/commands/index.js
```diff
@@ -27,6 +27,10 @@ module.exports = [
                         describe: `Environment`,
                         type: "string"
                     });
+                    yargs.option("variant", {
```
adrians5j, line 30: I see we have to explicitly set the `variant` into which we're deploying. Just a couple of questions...

1. What happens if staged rollouts were enabled, but `variant` hasn't been passed to the `deploy` command?
2. This may be totally incorrect from my side, but at the moment, WCP can already give you a new variant number. It's basically a GQL mutation, where you say "create variant" and the GQL API responds with a new variant, which has a unique auto-incremented number (wasn't using unique IDs just for sake of simplicity). Maybe a call would be a better idea, so if you're up for it, let me know.

packages/cli-plugin-deploy-pulumi/commands/index.js
```diff
@@ -27,6 +27,10 @@ module.exports = [
...
+                        describe: `Variant (only for staged rollouts)`,
+                        type: "string"
+                    });
                     yargs.option("build", {
                         default: true,
                         describe: `Build packages before deploying`,
```
adrians5j, line 36: Do we need the `variant` param also on the `watch` command?

adrians5j: Adding `variant` here would enable user to target a specific API variant when for example developing frontend apps (admin or website).

packages/cli-plugin-deploy-pulumi/commands/index.js
```diff
@@ -166,6 +170,10 @@ module.exports = [
                         describe: `Environment`,
                         type: "string"
                     });
+                    yargs.option("variant", {
+                        describe: `Variant (staged rollouts only)`,
+                        type: "string"
+                    });
                     yargs.option("json", {
                         describe: `Emit output as JSON`,
                         type: "boolean"
```
adrians5j, line 179: I think we should also have the `variant` param on the `pulumi` command.

packages/project-utils/traverseLoaders.js
```diff
@@ -1,16 +0,0 @@
-/**
- * A utility to recursively traverse loaders and execute the "onLoader" callback.
```
adrians5j: Can you explain why this file was deleted?

packages/pulumi-aws/src/apps/admin/AdminApp.ts
```diff
@@ -76,6 +77,22 @@ export const AdminApp = defineApp({
             });
         });
 
+        // Update variant gateway configuration.
+        const variant = app.ctx.variant;
+        if (variant) {
+            app.onAfterDeploy(async ({ outputs }) => {
+                // After deployment is made we update a static JSON file with a variant configuration.
+                // TODO: We should update WCP config instead of a static file here
+                await updateGatewayConfig({
```
adrians5j, line 86: Sounds good, will provide WCP details once it's ready.

packages/pulumi-aws/src/apps/gateway/GatewayApi.ts
```diff
@@ -0,0 +1,165 @@
...
+            acl: "public-read",
+            forceDestroy: true,
+            website: {
+                indexDocument: "index.html",
+                errorDocument: "index.html"
+            }
+        }
+    });
+
+    // TODO: With WCP we will not need the config file. Config should be loaded from WCP API.
```
adrians5j, line 37: Great, we'll sync on this.

packages/pulumi-aws/src/apps/gateway/GatewayApi.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    const cloudfront = app.addResource(aws.cloudfront.Distribution, {
+        name: `${params.name}-gateway-cdn`,
+        config: {
+            enabled: true,
+            isIpv6Enabled: true,
+            waitForDeployment: false,
+            origins: [
+                {
+                    originId: bucket.output.arn,
```
adrians5j, line 64: This origin will also go away once WCP is implemented?

packages/pulumi-aws/src/apps/gateway/GatewayApi.ts
```diff
@@ -0,0 +1,165 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as aws from "@pulumi/aws";
+
+import { ApplicationContext, loadGatewayConfig, PulumiApp } from "@webiny/pulumi-sdk";
+import { applyCustomDomain, CustomDomainParams } from "../customDomain";
+
+export interface GatewayApiConfig {
```
adrians5j, line 7: BTW was thinking a bit more about this file and the question I now have is - will this file actually be deleted completely once WCP is in place?

packages/pulumi-aws/src/apps/gateway/GatewayApp.ts
```diff
@@ -0,0 +1,77 @@
...
+    admin?: GatewayReactAppConfig;
+    website?: GatewayReactAppConfig;
+}
+
+export const GatewayApp = defineApp({
+    name: "Gateway",
+    config(app, config: GatewayAppConfig) {
+        const lambdas = createLambdas(app);
+
+        const api = createApiAppGateway(app, {
```
adrians5j, line 18: Once question here... 

I see there are three `createXYZGateway` calls, where internally each deploys a separate CloudFront distribution. But in my mind, if I remember correctly, there was only one gateway CDN deployed, that acts as a gateway.

Is the current code just temporary? Is this just because WCP hasn't yet been completed? 

Additional info would be useful.

packages/pulumi-aws/src/apps/gateway/GatewayLambdas.ts
```diff
@@ -0,0 +1,82 @@
...
+                        Action: "sts:AssumeRole",
+                        Principal: aws.iam.Principals.EdgeLambdaPrincipal,
+                        Effect: "Allow"
+                    }
+                ]
+            }
+        }
+    });
+
+    const functions = app.addHandler(() => {
```
adrians5j, line 30: Can you clarify a bit...

When do we use the `addHandler`? And can you describe what it does briefly?

packages/pulumi-aws/src/apps/website/WebsiteApp.ts
```diff
@@ -136,6 +137,7 @@ export const WebsiteApp = defineApp({
             // Among other things, this determines the amount of information we reveal on runtime errors.
             // https://www.webiny.com/docs/how-to-guides/environment-variables/#debug-environment-variable
             DEBUG: String(process.env.DEBUG),
+            VARIANT: app.ctx.variant || "",
```
adrians5j: Should we maybe be a bit more explicit here? Should we maybe use something like "STAGED_ROLLOUTS_VARIANT" or similar?

packages/pulumi-aws/src/apps/website/WebsiteApp.ts
```diff
@@ -163,14 +165,39 @@ export const WebsiteApp = defineApp({
...
+            });
+        });
+
+        // Update variant gateway configuration.
+        const variant = app.ctx.variant;
+        if (variant) {
+            app.onAfterDeploy(async ({ outputs }) => {
+                // After deployment is made we update a static JSON file with a variant configuration.
+                // TODO: We should update WCP config instead of a static file here
+                await updateGatewayConfig({
```
adrians5j, line 192: This is where WCP call will go, great 👍🏻

adrians5j (changes requested):
@Pavel910 and I finally managed to finish the first review. 

For any questions, feel free to ping us via Slack. We can also do a call if needed.

apps/website/code/webiny.config.ts
kedrzu: 👍

gateway.admin.dev.json
kedrzu: yes, this file is created on deploy and uploaded to S3 that serves as gateway app backend.

packages/api-prerendering-service-aws/src/render/subscriber.ts
kedrzu: 👍

packages/api-prerendering-service/src/render/absoluteAssetUrls.ts
kedrzu: 👍

packages/aws-helpers/src/cloudfrontFunctions/cookies.ts
kedrzu: I'll add some explanations to differences between types in Lambda@Edge and Cloudfront Functions. I'll also rename function to `setResponseCookie` - I think it's quite self-explanatory. There is no real logic here, just setting a cookie for HTTP response, but in CloudFront Function, which works a little bit different than Lambda@Edge

packages/aws-helpers/src/cloudfrontFunctions/utils.ts
kedrzu: 👍

packages/aws-helpers/src/stagedRollouts/functions/adminOriginRequest.ts
kedrzu: This is admin app, it's a SPA application with no prerendering. There is really no other place to do that.

kedrzu: We could in principle do this at deployment level, which would be just simplest and most robust, but this way we couldn't achieve these pre-built artifacts.

packages/aws-helpers/src/stagedRollouts/functions/viewerRequest.ts
kedrzu: Yes, but I'll expand it to be little bit more descriptive.
This is not something that is needed, but I left this TODO to find it easier once you decide it's worth pushing on.

kedrzu: I explained it better in notion: https://www.notion.so/webiny/Implementation-Plan-fea7d44bf3554302b0ca3a60eea66965#23a1dde5dd184239bc3a3fa00169de74

packages/aws-helpers/src/stagedRollouts/utils/loadConfig.ts
kedrzu: It doesn't really make sense - I'll remove that.

packages/aws-helpers/src/stagedRollouts/functions/viewerRequest.ts
kedrzu: 👍

packages/aws-helpers/src/stagedRollouts/utils/loadConfig.ts
kedrzu: I explained it a in a comment regarding `configCache` variable, but sure, I will add some additional info on the topic.

packages/aws-helpers/src/stagedRollouts/utils/loadConfig.ts
kedrzu: No, it will not be replaced by WCP endpoint. It's a little bit more complicated - I explained config caching here: https://www.notion.so/webiny/How-traffic-config-is-cached-2c8db57ca2b547a2b2fb1adf378cd191

packages/aws-helpers/src/stagedRollouts/utils/loadOriginPage.ts
kedrzu: Added some JSDOC to the function

packages/aws-helpers/src/stagedRollouts/utils/loadOriginPage.ts
kedrzu: 👍

packages/aws-helpers/src/stagedRollouts/utils/loadOriginPage.ts
kedrzu: I'll remove that

packages/aws-helpers/src/stagedRollouts/utils/loadVariantOrigin.ts
kedrzu: 👍

packages/aws-helpers/src/stagedRollouts/utils/common.ts
kedrzu: They are used for header key as well - you can pass variant either by cookie or header.
Header seems to be better for calling API variants directly from some client.
For better CDN caching under the hood cookie is normalized to header either way. So `viewer-request` reads cookie or header, validates it (or randomizes) and passes further as header.

kedrzu: We need 2 separate ways of telling application what variant we want. First is explicit (fixed variant), so you could set some fixed variant for example for testing purposes.
The other one is used by end clients to be randomly assigned. It's explained here https://www.notion.so/webiny/How-traffic-splitting-works-661f2f321ae74df6896ef700c7559a50

kedrzu: We could in principle use some convention for that, for example having a single header or cookie `webin-variant` that is either a valid variant or some expression like `#random-55#` that is later parsed with regex.

packages/aws-helpers/src/stagedRollouts/utils/loadVariantOrigin.ts
kedrzu: Structure is completely up to you. What I need for a specific application gateway is `weight` and `domain` that will tell me where exactly the variant is hosted.

We can have 3 separate endpoints for each app or a single one that returns all the apps (but in some structure that allows to filter by app), or a single endpoint but with some param `?app=admin`.

packages/aws-helpers/src/stagedRollouts/utils/loadVariantOrigin.ts
kedrzu: Yes, there were some inconsistencies in naming at one point, and I forgot to fix it.

packages/aws-helpers/src/stagedRollouts/utils/log.ts
kedrzu: Yeah, I wasn't sure about that. Logging these kind of errors adds really not much, because it's very difficult to find anything in CloudWatch if you don't know what are you looking for.
For now, because I didn't know what are best practices we use for logging, I added some logs, that helped me debug while developing.

packages/cli-plugin-deploy-pulumi/commands/index.js
kedrzu: I had to pass variant somehow to test the solution, but yeah, we need to figure out how that would fit into the bigger picture with WCP and stuff.
Also, you won't always create a new variant - sometimes you want to update a current one for some reason. 
Proposed workflows I've put here :https://www.notion.so/webiny/User-journeys-41e07c20f47842c5b6f5f872df1f5a84

packages/pulumi-aws/src/apps/gateway/GatewayApi.ts
kedrzu: There will have to be some origin, even if it's not used. Maybe we will have to leave some dummy s3 bucket just for  that case. Or make external origin with non-existing domain.

kedrzu: Or just webiny.com :)

packages/pulumi-aws/src/apps/gateway/GatewayApi.ts
kedrzu: No, why? 
This is a pulumi code for API gateway - to split traffic between different deployed API variants.

packages/pulumi-aws/src/apps/gateway/GatewayApp.ts
kedrzu: No, each application has its own domain and its own cloudfront distro that splits the traffic. It basically mimics current setup. We don't have API/admin/website on a single domain and CloudFront, so traffic splitting gateway works the same.

packages/project-utils/traverseLoaders.js
kedrzu: I haven't found any place in which this is used, but I understand, that this may be some externally-used helper. I'll recover the file.

packages/pulumi-aws/src/apps/website/WebsiteApp.ts
kedrzu: 👍

packages/pulumi-aws/src/apps/gateway/GatewayLambdas.ts
kedrzu: Sure, I'll add more comments.
But it's difficult to understand the whole concept just from comments. maybe I will have to write some notion doc, to describe how new pulumi architecture works.

packages/aws-helpers/src/stagedRollouts/utils/loadVariantOrigin.ts
adrians5j: Cool, thank you @kedrzu.

packages/aws-helpers/src/cloudfrontFunctions/utils.ts
Pavel910: @kedrzu so what's the reason for assigning to `global`? I've never seen this being done (only in Jest tests), would like to learn the reason :)

packages/aws-helpers/src/stagedRollouts/functions/adminOriginRequest.ts
adrians5j: Sounds good. Thanks 👍🏻

packages/cli-plugin-deploy-pulumi/commands/index.js
adrians5j: Can we double check this? We don't see the `variant` being added to the `watch` command.

packages/cli-plugin-deploy-pulumi/commands/index.js
adrians5j: Can we double check this? We don't see the `variant` being added to the `pulumi` command.

packages/project-utils/traverseLoaders.js
adrians5j: It doesn't look like this file was recovered. Can you double check?

BTW this file is not used directly within our `webiny-js` repo, but it is a helper used by Webiny users. In fact, it's mentioned in this doc:  https://www.webiny.com/docs/core-development-concepts/extending-and-customizing/integrate-tailwindcss#header

packages/pulumi-aws/src/apps/gateway/GatewayApp.ts
adrians5j: Awesome, now it's clear. Thanks! 👍🏻

adrians5j (changes requested):
Just a couple of questions in previous comments. Please check those out.

packages/project-utils/traverseLoaders.js
kedrzu: After reverting the change file is no longer in the list of changed files - so it's not deleted anymore.

packages/cli-plugin-deploy-pulumi/commands/index.js
kedrzu: Fixing that 👍

packages/cli-plugin-deploy-pulumi/commands/index.js
kedrzu: There is a major roadblock here - pulumi automation API does not support the `watch` command at all.
So the issue regards not just staged rollouts, but the whole usage of pulumi automation API.
It will be rather difficult to replicate that - we could add a watcher using `chokidar` and redeploy on every change. With some debouncing probably.

Pavel910: Oh, ok, didn't know that. It's ok @kedrzu, I'll take that one over in the upcoming weeks and see what options do we have. Chokidar is ok, yes, but the problem is that Pulumi then has to restart the program, reevaluate the sate and see what needs to be dpeloyed, which takes time. So it won't be instant like it currently is.

But as I said, don't worry about it. We'll come up with something.

PR #2335, opened 2022-04-19
Tool for quick page imports

kedrzu:
## Changes
It's a tool, that I created to test prerendering performance.
It allows to relatively quickly create many pages in webiny for testing purposes. 

Because it uses several third-party libraries, I added it as a package in `/tools` instead of `/scripts` and added `/tools/*` to yarn workspaces config. I made the package private so it won't be published.

## How Has This Been Tested?
Manually. Check `README.md` for details.


adrians5j (approved):
Looks good, will come in handy.

We can merge this.

adrians5j (changes requested):
Hey @kedrzu, please check CI/CD errors. We'll need to resolve those before merging.

PR #2340, opened 2022-04-22
Added elasticsearch config to new pulumi architecture

kedrzu:
## Changes
New pulumi architecture was introduced initially only for DDB-only setup. This PR aims to provide also ElasticSearch configuration. 

Changes are currently only in infrastructure, so no runtime code changes were made. We still need to think how to deal with different webiny setups codewise.

## How Has This Been Tested?
Not yet tested - only infrastructure changed, so could not test impact on GQL endpoint or any other.

## Documentation
TBD


packages/pulumi-aws/src/apps/api/ApiPrerendering.ts
```diff
@@ -16,6 +16,7 @@ interface PreRenderingServiceParams {
     cognitoUserPoolArn: pulumi.Input<string>;
     awsAccountId: pulumi.Input<string>;
     awsRegion: pulumi.Input<string>;
+    elasticsearchDynamodbTableArn: pulumi.Input<string | undefined>;
```
adrians5j: Can we double check... is this really needed with prerendering ?

Because, from what I can remember, prerendering service was only relying on DynamoDb, no matter which setup was used (ddb, or ddb+es).

packages/pulumi-aws/src/apps/storage/StorageElasticSearch.ts
```diff
@@ -0,0 +1,243 @@
...
+                elasticsearchVersion: "7.7",
+                clusterConfig: {
+                    instanceType: "t3.medium.elasticsearch",
+                    instanceCount: 2,
+                    zoneAwarenessEnabled: true,
+                    zoneAwarenessConfig: {
+                        availabilityZoneCount: 2
+                    }
+                },
+                // TODO: vpc is currently only in API
```
adrians5j: Is this comment relevant? I'd say yes 🤔 

ES needs to be deployed into a VPC if `vpc` was set to true.

packages/pulumi-aws/src/apps/storage/StorageElasticSearch.ts
```diff
@@ -0,0 +1,243 @@
...
+                // vpcOptions: {
+                //     subnetIds: [vpc.subnets.private[0].id, vpc.subnets.private[1].id],
+                //     securityGroupIds: [vpc.vpc.defaultSecurityGroupId]
+                // },
+                ebsOptions: {
+                    ebsEnabled: true,
+                    volumeSize: 10,
+                    volumeType: "gp2"
+                },
```
adrians5j, line 43: Another general question - how can we, within a module, access app params? For example, to check if `vpc` is enabled?

packages/pulumi-aws/src/apps/storage/StorageElasticSearch.ts
```diff
@@ -0,0 +1,243 @@
...
+                        ELASTIC_SEARCH_ENDPOINT: domain.output.endpoint
+                    }
+                },
+                description: "Process DynamoDB Stream.",
+                code: new pulumi.asset.AssetArchive({
+                    ".": new pulumi.asset.FileArchive(
+                        path.join(app.ctx.appDir, "code/dynamoToElastic/build")
+                    )
+                })
+                // TODO: vpc is currently only in API
```
adrians5j: Same question here - is this relevant?

adrians5j (changes requested):
@Pavel910 and I did our first review here. Nothing special, just a couple of questions.

Thanks!

packages/pulumi-aws/src/apps/storage/StorageElasticSearch.ts
kedrzu: Yes, VPC is set only inside API and here I found, that with ES setup storage will also require it.
Question is - should we have separate VPC for storage and API? Or is it possible to have a single one in storage and reuse it in other apps? With the first approach, we would have VPC per variant. With the latter, VPC would be shared by all variants.

packages/pulumi-aws/src/apps/storage/StorageElasticSearch.ts
kedrzu: We could return anything from the module or application `config` function. 
So if you need some additional variables, that serve informational purpose, you could return them along with app resources.

packages/pulumi-aws/src/apps/api/ApiPrerendering.ts
kedrzu: Maybe, I haven't verified that. That is simply rewritten from `ddb-es` template: https://github.com/webiny/webiny-js/blob/next/packages/cwp-template-aws/template/ddb-es/api/pulumi/prod/policies.ts#L90

adrians5j:
@kedrzu Can you double check this PR? There are still things that need to be resolved here (merge, PR still in draft, ...).

Thanks!

PR #2344, opened 2022-04-22
[TEST] Cypress tests - login and recover pw tests

razvanvancea:
## Changes

Closes #ISSUE_NUMBER

Added extra cypress tests on the login side of things as it follows:
- recover password E2E for "full-access" user
- authenticated user must logout successfully
- try to access non-existing page as logged user
- try to access admin pages as anonymous user

## How Has This Been Tested?

Tested by running all the tests according to file changes on local machine with passed results.

test_results


package.json
```diff
@@ -226,5 +226,8 @@
     "@types/eslint": "8.2.1",
     "yargs": "^17.3.1"
   },
-  "packageManager": "yarn@3.1.0"
+  "packageManager": "yarn@3.1.0",
+  "dependencies": {
+    "cypress-mailosaur": "^2.7.0"
```
adrians5j: Please move this into `devDependencies`.

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+        cy.visit("/");
+        cy.findByText(/sign in/i).should("exist");
+        cy.findByText(/forgot password?/i).should("exist");
+
+        cy.findByLabelText(/your e-mail/i).type("admin@webiny.com");
+        cy.findByLabelText(/your password/i).type("12345678");
+        cy.findByText(/submit/i).click();
+
+        cy.findByPlaceholderText(/search.../i).should("exist");
+        cy.findByText(/pages/i).should("exist");
```
adrians5j: Can we replace this portion of the code with just the `cy.login` call that we're using in other tests?

adrians5j: The same question goes for `authenticated user must logout successfully` test.

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+
+        cy.findByTestId("logged-in-user-menu-avatar").click();
+        cy.findByText(/Sign out/i).click();
+        cy.findByText(/sign in/i).should("exist");
+        cy.findByText(/forgot password?/i).should("exist");
```
adrians5j: The above code tells us that the user was redirected to the login page, but it doesn't really tell us if the log out worked.

So, what I'm thinking... can we also visit an admin URL and ensure the login is shown instead of the actual page? Basically we can just copy those couple of steps from a previous test, for example:


``` 
cy.visit("/page-builder/pages");
        cy.findByText(/sign in/i).should("exist");
        cy.findByText(/forgot password?/i).should("exist");```

File: cypress/integration/admin/login/login.spec.js [hunk 4]
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+    });
+
+    it('recover password E2E for "full-access" user', () => {
+        const serverId = 'z1fihlo8';
```
adrians5j: Can you explain why this serverId and why it's a hardcoded string?

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+        let passwordResetCode;
+        const newUserAccountEmail = uniqid("cypress_", "@z1fihlo8.mailosaur.net");
+
+        let fullAccessGroupUser;
```
adrians5j: Just a heads up here - I've just approved test execution in our CI/CD, and these couple of lines will be reported by prettier / eslint. Make sure to check the run logs.

https://github.com/webiny/webiny-js/actions/runs/2207349371

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+                    cy.log('email:' + fullAccessGroupUser.email)
+                    cy.log('psw:' + password)
+                    cy.visit("/");
+                    // Verify the access
+                    cy.findByText(`Hi ${user.firstName} ${user.lastName}!`).should("be.visible");
+                    cy.findByText(/To get started - pick one of the actions below:/i).should(
+                        "be.visible"
+                    );
+
+                    // Should have Page Builder card
```
adrians5j: Let's remove these couple of steps - we've already tested that in another test, and we've also done the "To get started..." assertion. I'd sad removing this won't hurt us.

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+                    cy.findByTestId("admin-welcome-screen-widget-page-builder").within(() => {
+                        cy.findByText("Page Builder").should("be.visible");
+                        cy.findByText(/Build a new Page/i).should("be.visible");
+                    });
+
+                    cy.findByTestId("logged-in-user-menu-avatar").click();
+                    cy.findByText(/Sign out/i).click();
+                    cy.findByText(/sign in/i).should("exist");
+
+                    // activate user password
```
adrians5j: One nitpick here - when writing comments, we try to be grammatically correct. So in this case, this would be:

```
// Activate user password.
```

It's not the end of the world of course, but we're trying to pay attention to these small details as much as possible - mainly because Webiny is open source and having nice code can sometimes make a difference.

cypress/integration/admin/login/login.spec.js
adrians5j: Do we want to add this as a env var BTW?

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+                        sentTo: newUserAccountEmail
+                    }).then(email => {
+                        expect(email.subject).to.equal('Your verification code');
+                        passwordResetCode = email.html.body.match(/(\d+)/)[0];
+
+                        cy.log(passwordResetCode)
+
+                        cy.contains('Password Recovery').should('be.visible');
+                        cy.contains('I got the code!').click();
+                        cy.get('div > input[type="text"]').type(passwordResetCode)
```
adrians5j: Can we target by test ID here too? The same way we did in the steps below?

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,146 @@ context("Login Page", () => {
...
+                        cy.get('div[data-testid="new-password-input"] input').type(newPassword);
+                        cy.get('div[data-testid="retype-password-input"] input').type(newPassword);
```
adrians5j: BTW do we really need `div` that is prefixing `[data-testid...` selection?

Instead of `div[data-testid="retype-password-input"]`, can we do just `[data-testid="retype-password-input"]` ? If you ask me, I'd say it's good enough of a selector. If you feel otherwise, feel free to leave it.

Just a nitpick really.

adrians5j (changes requested):
Hey @razvanvancea,

overall, the PR looks good. 🙂 

Please just take a look at a couple of notes I made, nothing major really.

You will also want to check the static code tests errors (prettier, eslint), accessible over this link:
https://github.com/webiny/webiny-js/runs/6164019363?check_suite_focus=true

ℹ️ In order to run prettier locally, you can do `yarn prettier:check` or even `yarn prettier:fix` (the latter will also apply fixes to all issues). You can also do `yarn eslint` to check for ESLint issues.

In any case, good job on this, looking forward to working together on some more tests. 🙂

cypress/integration/admin/login/login.spec.js
razvanvancea: it makes sense, I will keep that in mind. Thank you!

cypress/integration/admin/login/login.spec.js
razvanvancea: I let this assert here on purpose in order to make sure that the test waits until the page is loaded so it can securely click the user avatar in the top-right corner - sometimes cypress is a bit lazy and just try to click elements even though the page is not completely loaded. However I am going to delete it and we can add it later if we see something strange at this point

cypress/integration/admin/login/login.spec.js
razvanvancea: I created this constant as per mailosaur documentation to explicitly point it out ( the harcoded value is an unique value provided for our mailosaur account) - hence it has been used in only 1 place (line no. 132), so per our comment I decided to remove this constant.

2. No, we DO NOT NEED to set serverId it as env var - only the MAILOSAUR_API_KEY as per our chat discussion

cypress/integration/admin/login/login.spec.js
razvanvancea: Indeed it's not mandatory to have the div itself as well, it's just an approach that I've been using because I consider that it makes the selector a bit more readable (under the hood in terms of performance would be the same as long as we avoid chains like div > div > div[data-testid] and so on). I will update the selector as per your suggestion. Cheers!

cypress/integration/admin/login/login.spec.js
razvanvancea: Amazing idea! I'm on it.

cypress/integration/admin/login/login.spec.js
```diff
@@ -17,4 +19,195 @@ context("Login Page", () => {
...
+
+        cy.visit("/cms/content-models");
+        cy.findByText(/sign in/i).should("exist");
+        cy.findByText(/forgot password?/i).should("exist");
+    });
+
+    it("try to access non-existing page as logged user", () => {
+        const newUserAccountEmail = uniqid("cypress_", "@z1fihlo8.mailosaur.net");
+
+        let fullAccessGroupUser;
```
adrians5j: Nitpick: do we really need `fullAccessGroupUser` here? In the L56 we assign the received `user` to it, but we're not utilizing it anywhere except for the following couple of lines.

Unless I'm missing smt here, I'd say we can remove this variable (the same goes for the tests below, where we use the same approach).

adrians5j (changes requested):
Just one minor comment. Other than that - looking good.

cypress/integration/admin/login/login.spec.js
razvanvancea: You are right. I removed fullAccessGroupUser from all the tests. Cheers!

adrians5j:
Just a note on this PR - I'll be merging this once we resolve some failing tests in the `next` branch. Don't want to introduce more complexity before that, by merging more this PR.

Will merge it probably sometime next week, once the issue has been resolved.

PR #2355, opened 2022-04-27
Feat(app-admin): open thankyou page on new install or upgrade

endymion1818:
## Changes
- Add `isFirstInstall` check to `useInstaller()` hook
- Change text in installer completed button
- Add `openWelcomeScreen()` function to the above button
- Only redirect if we are on the root tenant

## How has this been tested?
- Run install script on a new install

### DIFF
--- a/packages/app-admin/src/components/AppInstaller/AppInstaller.tsx
+++ b/packages/app-admin/src/components/AppInstaller/AppInstaller.tsx
@@ -54,6 +54,7 @@ export const AppInstaller: React.FC = ({ children }) => {
         }
         if (!isFirstInstall) {
             window.location.href = `https://site.webiny.com/thank-you/upgrade?version=${wbyVersion}&returnUrl=${window.location.href}`;
+            return;
         }
         window.location.href = `https://site.webiny.com/thank-you/installation?returnUrl=${window.location.href}`;
     };
@@ -178,7 +179,7 @@ export const AppInstaller: React.FC = ({ children }) => {
                             openWelcomeScreen();
                         }}
                     >
-                        Finish install
+                        {isFirstInstall ? "Finish install" : "Finish upgrade"}
                     </ButtonPrimary>
                 </SuccessDialog>
             </Elevation>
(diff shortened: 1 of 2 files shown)


packages/app-admin/src/components/AppInstaller/AppInstaller.tsx
```diff
@@ -40,12 +41,23 @@ export const AppInstaller: React.FC = ({ children }) => {
...
         onUser,
         skippingVersions
     } = useInstaller({ isInstalled: isInstallerCompleted() });
 
+    const openWelcomeScreen = () => {
+        if (!isRootTenant) {
+            return;
+        }
+        if (!isFirstInstall) {
+            window.location.href = `https://site.webiny.com/thank-you/upgrade?version=${wbyVersion}&returnUrl=${window.location.href}`;
```
Pavel910, line 56: There's a big chance that this will not redirect soon enough, and the second `window.location.href` on line `58` will kick in. Just add a `return` after this to stop the execution.

I tried 2 subsequent `window.location.href` in the browser dev tools, and the last one wins.

Second question: is `site.webiny.com` a permanent domain we want to have here?

packages/app-admin/src/components/AppInstaller/AppInstaller.tsx
```diff
@@ -163,9 +175,10 @@ export const AppInstaller: React.FC = ({ children }) => {
                         onClick={() => {
                             markInstallerAsCompleted();
                             setFinished(true);
+                            openWelcomeScreen();
                         }}
                     >
-                        Open Admin Area
+                        Finish install
```
Pavel910: Since we have the `isFirstInstall` flag, maybe we can have this message reflect that, and show `Finish Setup` for first install, and show `Finish Upgrade` for upgrades?

Pavel910 (changes requested):
Left a few comments :)

packages/app-admin/src/components/AppInstaller/AppInstaller.tsx
endymion1818: Ooh nice idea! OK I'll implement that.

packages/app-admin/src/components/AppInstaller/AppInstaller.tsx
endymion1818: Thanks Pavel, good catch, I've updated this.

Yes, site.webiny.com is a domain we own & use for various lead capturing information.

Pavel910 (approved):

PR #2360, opened 2022-04-28
Decomposed storage and API apps into modules

kedrzu:
## Changes
Using app modules is a nice way to decompose an app into set of relatively loosely coupled modules, that can be later used in any plugin.

## How Has This Been Tested?
Built and deployed application


packages/pulumi-aws/src/apps/api/ApiGateway.ts
```diff
@@ -1,50 +1,53 @@
...
-        api,
-        defaultStage,
-        routes: routesResult,
-        addRoute
-    };
+        return {
+            api,
+            stage,
+            routes,
+            addRoute
```
Pavel910, line 42: This is super cool 👍  Exposing customization methods like that... really digging that!

Pavel910 (approved):
This is looking good @kedrzu, some nice organizational ideas for the future in here, will use as a reference 👍

packages/pulumi-aws/src/apps/api/ApiGateway.ts
kedrzu: And I like to define all of that kind of functions at the end of the function call, even after the `return` statement.
It's a cool way of utilizing hoisting to our advantage.

PR #2365, opened 2022-04-29
feat(app-admin): send telemetry to GA on first install

endymion1818:
## Changes
- Added iframe to AppInstaller
- Added `isFirstInstall` boolean to `useInstaller()` hook

## How Has This Been Tested?
- Cypress
- Locally

## Screenshots

1. Update to install screen repositioning ToS & Privacy link: 
Screen Shot 2022-05-05 at 15 14 31
 
2. IFrame on successful install screen
Screen Shot 2022-05-05 at 16 09 00

3. Cookies from www.webiny.com showing `_ga` cookie
Screen Shot 2022-04-29 at 15 15 08

### DIFF
--- a/packages/app-admin-users-cognito/src/plugins/installation.tsx
+++ b/packages/app-admin-users-cognito/src/plugins/installation.tsx
@@ -126,7 +126,7 @@ const Install: React.FC<InstallProps> = ({ onInstalled }) => {
                             <View name={"adminUsers.installation.fields"} props={{ Bind, data }} />
                         </Grid>
 
-                        <Grid style={{ paddingTop: '0px' }}>
+                        <Grid style={{ paddingTop: "0px" }}>
                             <Cell span={12}>
                                 <Bind name="subscribed">
                                     <Checkbox
@@ -144,7 +144,10 @@ const Install: React.FC<InstallProps> = ({ onInstalled }) => {
                     <SimpleFormFooter>
                         <Grid>
                             <Cell span={8}>
-                                <p style={{ textAlign: "left" }}>By submitting the form, you agree to our Terms of Service and acknowledge our {privacyPolicyLink}.</p>
+                                <p style={{ textAlign: "left" }}>
+                                    By submitting the form, you agree to our Terms of Service and
+                                    acknowledge our {privacyPolicyLink}.
+                                </p>
                             </Cell>
                             <Cell span={4}>
                                 <ButtonPrimary
@@ -157,7 +160,6 @@ const Install: React.FC<InstallProps> = ({ onInstalled }) => {
                                 </ButtonPrimary>
                             </Cell>
                         </Grid>
-
                     </SimpleFormFooter>
                 </SimpleForm>
             )}
--- a/packages/app-admin-users-cognito/src/plugins/installation.tsx
+++ b/packages/app-admin-users-cognito/src/plugins/installation.tsx
@@ -149,8 +149,8 @@ const Install: React.FC<InstallProps> = ({ onInstalled }) => {
                             <Cell span={4}>
                                 <ButtonPrimary
                                     data-testid="install-security-button"
-                                    onClick={ev => {
-                                        submit(ev);
+                                    onClick={() => {
+                                        submit();
                                     }}
                                 >
                                     Create Admin User
--- a/packages/app-admin-users-cognito/src/plugins/installation.tsx
+++ b/packages/app-admin-users-cognito/src/plugins/installation.tsx
@@ -126,15 +126,14 @@ const Install: React.FC<InstallProps> = ({ onInstalled }) => {
                             <View name={"adminUsers.installation.fields"} props={{ Bind, data }} />
                         </Grid>
 
-                        <Grid>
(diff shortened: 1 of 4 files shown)


adrians5j:
Deploying... hopefully I'll give you an update soon. :)

BTW one question, in the PR body it says:

> Added `isFirstInstall` boolean to `useInstaller()` hook

Is this really true? I don't see this change in the code. Only `packages/app-admin/src/components/AppInstaller/AppInstaller.tsx` was changed basically.

endymion1818:
Thank you Adrian! Yes, I added that in a previous PR.

adrians5j:
BTW I just tried the installer, and I didn't see anything different 🤔 

I cloned this branch, made sure to build the packages, and deployed a brand new project. This is what I saw (feel free to skip to 2nd half of the vide):

https://user-images.githubusercontent.com/5121148/166916158-1ef0fa69-19bd-4398-a130-33868c06ba97.mov

P.S. as stated in the GH PR template, when doing these kind of visual changes, it's useful to post a video, just so the rest of the team can see the result a bit quicker.

endymion1818:
> BTW I just tried the installer, and I didn't see anything different 🤔

I'm glad you didn't see anything different 😁 This puts back the last page of the installer so that you continue to your app instead of being redirected to a thank you page hosted on our website.

Good shout, I'll upload a video shortly.

packages/app-admin-users-cognito/src/plugins/installation.tsx
```diff
@@ -143,14 +142,22 @@ const Install: React.FC<InstallProps> = ({ onInstalled }) => {
...
-                            Create Admin User
-                        </ButtonPrimary>
+                        <Grid>
+                            <Cell span={8}>
+                                <p style={{ textAlign: "left" }}>By submitting the form, you agree to our Terms of Service and acknowledge our {privacyPolicyLink}.</p>
+                            </Cell>
+                            <Cell span={4}>
+                                <ButtonPrimary
+                                    data-testid="install-security-button"
+                                    onClick={ev => {
```
adrians5j: Nitpick / while we're at it..

Let's change the `onClick` prop to:

```ts
onClick={() => {
    submit();
}}
```

packages/app-admin/src/components/AppInstaller/AppInstaller.tsx
```diff
@@ -22,10 +22,15 @@ import { config as appConfig } from "@webiny/app/config";
...
     const tenantId = localStorage.get("webiny_tenant") || "root";
-
     const lsKey = `webiny_installation_${tenantId}`;
     const wbyVersion = appConfig.getKey("WEBINY_VERSION", process.env.REACT_APP_WEBINY_VERSION);
     const isRootTenant = tenantId === "root";
+    /* Cypress doesn't work with cross domains because of
+    * security issues. This setting allows us to avoid rendering
+    * the <iframe> without disabling chrome web security
+    * reference: https://docs.cypress.io/guides/guides/web-security#Insecure-Content 
+    */
```
adrians5j:
```suggestion

    /* 
    * This flag allows us to avoid rendering the <iframe> when the app is tested with Cypress
    * (Cypress doesn't work with cross domains because of security-related implications).
    * @see https://docs.cypress.io/guides/guides/web-security#Insecure-Content 
    */
```

adrians5j (changes requested):
Minor nitpicks, all in all good to go.

P.S. Just double check the CI/CD errors.

packages/app-admin-users-cognito/src/plugins/installation.tsx
Pavel910: That `submit` actually **does** take an event as an input. It's not required, but it is supported. And with that, you can simplify down to `onClick={submit}`.

adrians5j: Tried `onClick={submit}`, but it doesn't work because of the types in the background. Just so we don't increase the scope, I decided to go with the above solution.

Pavel910: Ok. I'll have a look at it later, and open a separate PR to fix it.

adrians5j: Double checked and since it is a quick fix.. we could do it maybe while we're at it.

```
packages/ui/src/Button/Button.tsx:16
```

So, change from:

```ts
onClick?: (event: React.MouseEvent<any, MouseEvent>) => void | null;
```

to:

```ts
onClick?: (event: React.MouseEvent<any, MouseEvent>) => void | null | Promise<void | null>;
```

adrians5j: The TS error that was originally thrown:

image

adrians5j: All in all... not sure, but maybe we can ignore this one for now, since maybe we'd want to come up with a smarter way to resolve this.

adrians5j (approved):
To me this looks good to merge. Unless @Pavel910 has any extra comments, let's do it.

Pavel910 (approved):
Let's merge this.

PR #2366, opened 2022-04-29
Allow change of locale during a single API call

Pavel910:
## Changes
This PR adds the ability to switch active locale during a request lifecycle, similar to how we can switch tenants.
I've also deprecated the `@webiny/api-i18n-content` package, as it was just a confusing package overall, and all of its functionality was moved to `@webiny/api-i18n`, and is collocated with the rest of I18N related functionality.

## How Has This Been Tested?
Jest tests.

## Documentation
Not required.


packages/api-apw/src/plugins/utils.ts
```diff
@@ -173,6 +170,6 @@ export const isInstallationPending = ({
     if (!tenant) {
         return true;
     }
-    const locale = i18nContent.locale;
-    return !locale;
+
+    return !i18n.getContentLocale();
```
endymion1818, line 174: I do prefer the structure of dropping it into a variable, and then returning that variable. Otherwise it's doing too much in my opinion.

packages/api-headless-cms/src/plugins/context.ts
```diff
@@ -1,17 +1,23 @@
 import { ContextPlugin } from "@webiny/handler";
 import { CmsContext } from "~/types";
 
+const DEFAULT_LOCALE_CODE = "en-US";
```
endymion1818, line 4: nice 👍

packages/api-i18n/src/graphql/context.ts
```diff
@@ -155,31 +168,58 @@ const createBaseContextPlugin = () => {
...
+            throw new NotAuthorizedError();
+        };
+
+        const setContentLocale: I18NContextObject["setContentLocale"] = locale => {
+            setCurrentLocale("content", locale);
+        };
+
+        const getContentLocale: I18NContextObject["getContentLocale"] = () => {
+            return getCurrentLocale("content");
+        };
```
endymion1818, line 209:
```suggestion
        const getContentLocale: I18NContextObject["getContentLocale"] = getCurrentLocale("content");
```

packages/api-page-builder-import-export/__tests__/pageImportExportTask/useHandler.ts
```diff
@@ -21,11 +21,16 @@ export default (params: Params = {}) => {
...
             apply: (context: PbContext) => {
-                if (context.i18nContent) {
+                if (context.i18n) {
                     return;
                 }
-                context.i18nContent = {
-                    ...(context.i18nContent || ({} as any)),
+
+                context.i18n = {
+                    // @ts-ignore
```
endymion1818, line 29: I think a little more reasoning around why this is tsignored would be helpful to others

Pavel910: Try removing the ignore 😂  Joking aside, this whole file needs proper restructuring, and we can open a dedicated cleanup issue for that. I'll open an issue right after we merge this PR, and will take care of it in the near future.

packages/api-i18n/src/graphql/context.ts
Pavel910: Your suggestion is not really the same. What I'm doing in the code is creating a new arrow function, which will return the current content locale when invoked. Your suggestion invokes it immediately, which defeats the purpose of it. We need the ability to get the content locale at any point in time, and that can also be after the current locale is changed. So in short - we need a function, not a value.

packages/api-apw/src/plugins/utils.ts
Pavel910: If you look closer, you'll notice that I'm replacing the old `i18nContent.locale` usage to `i18n.getContentLocale()`. That was the primary objective of this change. Converting it into one liner was just a natural next step. The resulting code is really common. Assigning to variable just to check if it's a `!locale` is a matter of preference. I usually try to avoid a new variable for obvious statements.

packages/api-i18n/src/graphql/context.ts
endymion1818: Ah yes my bad, I went a bit too far with my suggestion it seems.

PR #2375, opened 2022-05-04
Fix URI handling in the tenant router

Pavel910:
## Changes
This PR fixes URI construction in the tenant router Lambda@Edge function. The bug was discovered on a multi-tenant setup, when a requested page did not exist. That resulted in an error being thrown by the S3 bucket, and triggered CloudFront's `customErrorResponses` configuration, which invoked the tenant router again, with a `_NOT_FOUND_PAGE_/index.html` URI, which then ran into a bug in URI handling.

## How Has This Been Tested?
Manually.

## Documentation
User docs are not necessary, and the tenant router function now contains extensive code comments to cover the entire routing logic.


adrians5j (approved):
Looks good, loving all the comments. 👏🏻

endymion1818 (approved):

PR #2379, opened 2022-05-06
fix(scripts): filter package for those that include package.json file

endymion1818:
## Changes
- Filters out packages folders that don't have a `package.json` file

## Problem solved
- When developers were creating new packages, then switching branches, the `build`  script would fail because the package had been deleted, but the folder was still there (as it has a `/dist/` folder)
- Now the script only attempts to build if there is a package.json file in the package folder


Pavel910 (approved):
Looks good, thanks!

PR #2389, opened 2022-05-09
Caching traffic splitting config on CDN

kedrzu:
## Changes
Implementation according to the doc https://www.notion.so/webiny/How-traffic-config-is-cached-2c8db57ca2b547a2b2fb1adf378cd191#71ce40fedd9a4181af067ed9ed9333a2 

There is just an additional layer that is responsible for caching the underlying traffic splitting config endpoint.
For POC traffic splitting config sits inside an S3 bucket as a static file. Eventually, it will be a WCP endpoint.
Changing this should be rather straightforward - just follow TODOs I left.

## How Has This Been Tested?
Manually.


packages/aws-helpers/src/stagedRollouts/functions/configOriginRequest.ts
```diff
@@ -0,0 +1,68 @@
...
+    try {
+        const configJson = await loadConfigJson(event);
+        const response: CloudFrontResponse = {
+            status: "200",
+            statusDescription: "OK",
+            headers: {},
+            body: configJson
+        };
+
+        setHeader(response.headers, {
```
adrians5j, line 20: Just a note - looking at this file, it seems to us that this is the old code. For example the `setHeader` is already renamed to `setResponseHeader` in the previous PR.

So, there's gonna be some merging most probably.

adrians5j (approved):
@Pavel910 and I had a look - looks good. 

But you may want to merge the changes done in the previous PR. You're probably aware of this / you'll know the order of merging these PRs.

adrians5j (changes requested):
@kedrzu Please check CI/CD - there are some errors.

Thanks!

packages/aws-helpers/src/stagedRollouts/functions/configOriginRequest.ts
kedrzu: You can change headers for both response and request and they have the same structure in `Lambda@Edge` so I used a more generic approach, so that you can use the same function for request and response headers.

PR #2395, opened 2022-05-11
Moved VPC configuration to storage app

kedrzu:
## Changes
Turned out that VPC will be reused across multiple applications. For example, adding ElasticSearch to the stack requires us to have VPC available in the `storage` app. Also, because page prerendering is being moved from the `api` to the `website` app, we will probably need VPC there too.

So VPC should be inside the application that serves as a set of common resources for the entire system, that is `storage` app. 

BTW - this name slowly loses its sense because we no longer have persistent storages here, but also some other common resources, like VPC, event bus, and others. Maybe we should name it `core`?

## How Has This Been Tested?
Deployed and run webiny with VPC enabled and disabled.

## Documentation
There is a simple way of turning VPC on and off now and it can be easily dependent on environment, so maybe we should put into docs sometime.


adrians5j (approved):
@Pavel910 and I had a look - we can merge this.

adrians5j (changes requested):
BTW I'll change this to "Request changes" since there are still some merge conflicts that need to be resolved.

Pavel910 (approved):

PR #2400, opened 2022-05-13
feat(ui): upgrade RMWC

endymion1818:
## Changes
- Upgraded RMWC components to v7
- Upgrade @material components to versions used in v7 of RMWC
- Import SASS to use new syntax
- Update styles to resolve changes from the upgrade

## How can this be tested?
- https://d3mmrkoy170qpm.cloudfront.net/


apps/theme/pageBuilder/styles/medium.scss
```diff
@@ -1,8 +1,8 @@
 /**
  *  `Medium editor` styles
  */
-@import "~medium-editor/dist/css/medium-editor.css";
-@import "~medium-editor/dist/css/themes/mani.css";
+@import "medium-editor/dist/css/medium-editor.css";
```
Pavel910, line 4: Existing user projects will still have imports via `~`. Will this upgrade break this for those projects, or is this backwards compatible? This has to be tested with older projects.

apps/theme/pageBuilder/styles/medium.scss
```diff
@@ -1,8 +1,8 @@
...
+@import "medium-editor/dist/css/themes/mani.css";
```
Pavel910, line 5: `mani.css` ? I have a feeling this should be `main.css`. Please double-check.

apps/theme/styles.scss
```diff
@@ -1,3 +1,3 @@
 @use "./pageBuilder/styles/theme.scss";
 @use "./formBuilder/styles/theme.scss" as themeFormBuilder;
-@use "~@webiny/react-rich-text-renderer/styles.scss";
+@import "@webiny/react-rich-text-renderer/styles.scss";
```
Pavel910, line 3: What's the difference between `@use` and `@import`, and why is `import` only used  on the last scss file, but not the first two as well?

package.json
```diff
@@ -102,6 +102,7 @@
     "rxjs": "^6.5.5",
     "semver": "^6.3.0",
     "ts-jest": "^26.4.4",
+    "tslib": "2.4.0",
```
Pavel910: Why do we need this `tslib` package? We didn't need it thus far, I want to know what dependency made this a requirement, and if it has to sit in the root package.json.

package.json
```diff
@@ -211,20 +212,41 @@
...
+    "@material/form-field": "^5.1.0",
+    "@material/icon-button": "^5.1.0",
+    "@material/linear-ripple": "^5.1.0",
+    "@material/list": "^5.1.0",
+    "@material/radio": "^5.1.0",
+    "@material/select": "^5.1.0",
+    "@material/snackbar": "^5.1.0",
+    "@material/switch": "^5.1.0",
+    "@material/tab-scroller": "^5.1.0",
+    "@material/top-app-bar": "^5.1.0"
```
Pavel910: This is the wrong place for these dependencies. They should be moved to the package that's actually using these. If we're using RMWC, I'm not sure why we need to import these as dependencies at all. We can discuss this over a call if you're unsure what to do with these.

packages/app-admin/src/components/AppInstaller/styled.tsx
```diff
@@ -39,11 +39,12 @@ export const InstallContent = styled("div")({
...
-        height: "100vh",
-        ".webiny-split-view__right-panel-wrapper": {
-            height: "100vh"
-        }
+        height: "100vh"
+    },
+    ".webiny-split-view__right-panel-wrapper": {
+        height: "100vh"
     }
 });
```
Pavel910, line 49: If I remember this correctly, this `100vh` should apply to both left and right panels. With this new rule it looks like you're only applying that height to the right panel.

packages/app-form-builder/package.json
```diff
@@ -19,7 +19,7 @@
     "@editorjs/editorjs": "^2.20.1",
     "@emotion/core": "^10.0.17",
     "@emotion/styled": "^10.0.17",
-    "@rmwc/menu": "^5.6.0",
+    "@rmwc/menu": "7.0.3",
```
Pavel910: This bothers me because `@rmwc` packages should only be used in the `ui` package. Please make a note of these packages, and we should solve this in a separate PR. All our apps should be using the `ui` package exclusively, to maintain the abstraction in one place.

packages/app-page-builder/package.json
```diff
@@ -21,7 +21,7 @@
     "@fortawesome/free-regular-svg-icons": "^6.0.0",
     "@fortawesome/free-solid-svg-icons": "^6.0.0",
     "@fortawesome/react-fontawesome": "^0.1.17",
-    "@rmwc/menu": "^5.6.0",
+    "@rmwc/menu": "7.0.3",
```
Pavel910: Same as the previous file.

packages/cwp-template-aws/template/common/apps/theme/styles.scss
```diff
@@ -1,3 +1,3 @@
 @import "./pageBuilder/styles/theme.scss";
 @import "./formBuilder/styles/theme.scss";
-@import "~@webiny/react-rich-text-renderer/styles.scss";
+@import "Z@webiny/react-rich-text-renderer/styles.scss";
```
Pavel910: A typo: `Z@webiny` :)

packages/project-utils/bundling/app/config/webpack.config.js
```diff
@@ -40,9 +40,11 @@ const cssRegex = /\.css$/;
 const cssModuleRegex = /\.module\.css$/;
 const sassRegex = /\.(scss|sass)$/;
 const sassModuleRegex = /\.module\.(scss|sass)$/;
+
 const sassLoader = {
     loader: require.resolve("sass-loader"),
     options: {
+        // webpackImporter: false,
```
Pavel910: Leave a comment why this is commented out.

packages/ui/src/Chips/ChipIcon.tsx
```diff
@@ -1,5 +1,5 @@
 import React from "react";
-import { ChipIcon as RmwcChipIcon, ChipIconProps as RmwcChipIconProps } from "@rmwc/chip";
+import { Chip as RmwcChipIcon, ChipProps as RmwcChipIconProps } from "@rmwc/chip";
```
Pavel910: Previously, there were 2 separate components: a `Chip`, and a `ChipIcon`. `ChipIcon` was used as a child of the `Chip` component to add an icon to a chip, here's an example:

```
<Chip>
    <ChipIcon icon={<AnIcon/>}/>
    Chip Label
</Chip>
```

This was changed ages ago, but this `<ChipIcon>` component was not removed. We don't use it anywhere in Webiny, so feel free to delete this component and stories for it. It's meaningless because the `<Chip>` itself supports `icon` props.

Pavel910 (changes requested):
@endymion1818 did the first review so you can get some things sorted before jumping on a call. Let me know if any of my comments need clarification.

apps/theme/pageBuilder/styles/medium.scss
endymion1818: `mani` is in fact the name of the Medium editor theme we're using: https://github.com/yabwe/medium-editor/blob/master/src/sass/themes/mani.scss 😆

package.json
endymion1818: it's an implicit dependency of a few projects, and as a result of different versions being used you get this error on building:

`admin: Error: Attempted import error: '__spreadArray' is not exported from 'tslib' (imported as '__spreadArray').`

Unfortunately you still get this error if it's only listed in the `resolutions`.

apps/theme/styles.scss
endymion1818: ~We bumped up `sass` and `sass-loader` as part of this upgrade:~

- `@import` globally imports other sass files and [is being phased out](https://sass-lang.com/documentation/at-rules/import). It's appropriate here because we're importing `rich-text-renderer` styles from another package, whereas we're using the other two that come from this package. Also you get errors any other way around 🙃
- `@use` forwards mixins, variables etc _to the current sass file_, so it's more explicit as to what's happening
- `~` (tilde) is deprecated from `sass-loader` (https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules)

package.json
endymion1818: I moved these to `/packages/ui/package.json` and the build was successful. Sorry, my mistake I thought resolutions had to be at the top level `package.json` because the dependency `material-components-web` which also uses them is in this `package.json`.

packages/app-admin/src/components/AppInstaller/styled.tsx
endymion1818: This is just a syntax change, I've removed the nesting from the declaration as it wasn't getting evaluated.

packages/app-form-builder/package.json
endymion1818: Yep makes perfect sense

packages/app-page-builder/package.json
endymion1818: 👍🏻

packages/cwp-template-aws/template/common/apps/theme/styles.scss
endymion1818: 🙈

packages/project-utils/bundling/app/config/webpack.config.js
endymion1818: It should be removed, I was doing a test to see if it materially changed anything but it didn't

packages/ui/src/Chips/ChipIcon.tsx
endymion1818: Done, thanks

apps/theme/pageBuilder/styles/medium.scss
endymion1818: It's being deprecated but I couldn't find any notification as to when they might remove it. https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules

packages/ui/package.json
```diff
@@ -116,6 +116,30 @@
...
+    "@material/icon-button": "^5.1.0",
+    "@material/linear-ripple": "^5.1.0",
+    "@material/list": "^5.1.0",
+    "@material/radio": "^5.1.0",
+    "@material/select": "^5.1.0",
+    "@material/snackbar": "^5.1.0",
+    "@material/switch": "^5.1.0",
+    "@material/tab-scroller": "^5.1.0",
+    "@material/top-app-bar": "^5.1.0"
+  },
```
Pavel910: You cannot have `resolutions` per workspace. `resolutions` is only valid in the monorepo root, as their purpose is to resolve to one common dependency version for the entire project. And yarn tells you that when you try to install deps:

CleanShot 2022-05-23 at 12 58 57@2x

My main question is: why do you need to specify these at all? Aren't these packages internal dependencies of @rmwc packages?

From what I see, @rmwc has these as explicit dependencies in their packages:
CleanShot 2022-05-23 at 13 01 39@2x

This means that you don't need to specify any `@material/*` packages as Webiny dependencies at all.

endymion1818: Thanks for the convo about this, the issue is resolved by removing these dependencies. It turns out that updating the `material-components-web` package made it redundant.

Pavel910:
Found this weird behavior on Pages list filters:

https://user-images.githubusercontent.com/3920893/171034441-34a96d71-742e-4452-aab6-4bd7bbdea1bc.mp4

However, this works fine on Content Entries list filtering. Probably `select` options are not passed in the same way. RMWC 5 had some weird stuff going on with select boxes.

Pavel910:
Another thing to check, something related to Portals:

CleanShot 2022-05-30 at 19 32 33@2x

Add that `<Portal/>` component here: `packages/app-admin-rmwc/src/modules/Overlays/index.tsx:19`. There's a comment in that file describing what that stuff does.

Pavel910:
Secondary button styles are slightly off, this is how they currently are, and with the new RMWC, the border is different.

CleanShot 2022-05-30 at 22 57 22@2x

Pavel910:
@endymion1818 Adding more screenshots to resolve:

The following three screenshots show the same issue, the floating label is positioned a bit off. Seems like it's only an issue with values prefilled by the browser. RMWC is probably not detecting that there's a value in the field (just guessing here).

CleanShot 2022-06-05 at 16 01 35

CleanShot 2022-06-05 at 16 02 34

CleanShot 2022-06-05 at 16 04 28

Page Builder glitches:

CleanShot 2022-06-05 at 16 05 39

Menu doesn't seem to work at all (that's probably that thing with Portal):

https://user-images.githubusercontent.com/3920893/172054602-132654e5-8cad-431a-91af-e9da1a4d4fdb.mp4

These buttons are off:

CleanShot 2022-06-05 at 16 08 43

apps/admin/package.json
```diff
@@ -28,7 +28,8 @@
     "react": "16.14.0",
     "react-dom": "16.14.0",
     "regenerator-runtime": "^0.13.5",
-    "theme": "^0.1.0"
+    "theme": "^0.1.0",
+    "tslib": "^2.4.0"
```
Pavel910, line 32: @endymion1818 just to confirm: is `tslib` dependency still necessary? I know you mentioned that it's our rmwc forks that were causing issues, is that related?

packages/ui/src/Button/Button.styles.ts
```diff
@@ -0,0 +1,16 @@
...
+ * @type {string}
+ */
+const webinyButtonStyles = css(
+    {},
+    {
+        "&.webiny-ui-button--secondary:not(:disabled)": {
+            borderColor: "var(--mdc-theme-primary)"
+        }
+    }
+);
```
Pavel910: @endymion1818 This will look much nicer if you use the template literals; then you also don't need to use the camelCase for properties. You just write plain CSS instead.
```
const webinyButtonStyles = css`
    &.webiny-ui-button--secondary:not(:disabled) {
        border-color: var(--mdc-theme-primary)
    }
`
```

packages/ui/src/Input/Input.tsx
```diff
@@ -37,8 +39,24 @@ export type InputProps = FormComponentProps &
...
+ * @type {string}
+ */
+const webinyInputStyles = css(
+    {},
+    {
+        ".mdc-text-field__input:-webkit-autofill + .mdc-floating-label": {
+            transform: "translateY(-106%) scale(0.75)"
+        }
+    }
+);
```
Pavel910: Same as previous comment; try using template literals.

Pavel910 (changes requested):
@endymion1818 just a few minor things and we're good with the code review. The rest looks pretty solid. It's mostly adding of `data-testid` and some CSS tweaks across the board, so there isn't a whole lot of code changes that can break anything.

apps/admin/package.json
endymion1818: @Pavel910 unfortunately yes, Im afraid we still get that issue with the `spreadArray` function we identified before without explicitly adding this line.

packages/ui/src/Button/Button.styles.ts
endymion1818: I was following what I thought was a convention but very happy you said this, I prefer template literals too.

apps/admin/package.json
Pavel910: In that case, I assume we'll need to add this dependency as part of the project upgrade we're preparing for 5.30. Hopefully it's the only thing that needs changing in existing user projects. Have you tried upgrading from 5.29 ?

endymion1818: Ah, no haven't tested that. I'll run through it in the morning.

PR #2414, opened 2022-05-19
fix(api-headless-cms): storage operations generate pk out of model

brunozoric:
## Changes
Storage operations for the Headless CMS were taking locale and tenant out of entries - now it is switched to a model.
Fix for `Use as title` in app-headless-cms - bug happens when user adds new field and sets it as title and then tries to save the model.-

## How Has This Been Tested?
Jest and manually.


Pavel910 (approved):
Looks good. I really like how you now explicitly pass `id`, `locale` and `tenant` to `createPK` methods, instead of the entire entry/model. It's a lot more readable.

PR #2415, opened 2022-05-19
feat(app-apw): headless cms workflow

brunozoric:
## Changes
This PR adds Headless CMS section in the workflow creation (Content review page not included - will be another PR).
It is UI feature, but some things in the API had to be changed.

## How Has This Been Tested?
Jest and manually.


apps/admin/code/src/plugins/apw.ts
```diff
@@ -1,4 +1,4 @@
 /* Core apw (advanced publishing workflow app) plugins */
```
Pavel910: Let's register these internally, within the `<AdvancedPublishingWorkflow/>`, using `plugins.register(...)` directly, so  we don't clutter the project with this old way of adding plugins. 

See an example here: https://github.com/webiny/webiny-js/blob/next/packages/app-i18n/src/I18N.tsx#L22

packages/app-apw/package.json
```diff
@@ -18,6 +18,9 @@
     "@types/react": "^16.9.56",
     "@webiny/app": "^5.27.0",
     "@webiny/app-admin": "^5.27.0",
+    "@webiny/app-admin-core": "^5.27.0",
```
Pavel910: Use `@webiny/app-admin` here. `app-admin-core` is a very basic low level framework, it's used by `app-admin`, and no other packages need to use it directly.

packages/app-apw/src/hooks/useCmsModels.ts
```diff
@@ -0,0 +1,36 @@
...
+import get from "lodash/get";
+
+interface UseCmsModelsResult {
+    models: CmsModel[];
+    loading: boolean;
+}
+
+export const useCmsModels = (): UseCmsModelsResult => {
+    const { getCurrentLocale } = useI18N();
+    const locale = getCurrentLocale("content") || "en-US";
```
Pavel910: I wouldn't use any fallback here (`en-US`). Seeing that `en-US` is _not_ always the default locale for our clients, this can cause hard to debug issues. It's better to throw an error if the locale is not available.

packages/app-apw/src/plugins/Routes.tsx
```diff
@@ -1,25 +1,27 @@
...
+}
+const Loader: React.FC<LoaderProps> = ({ children, ...props }) => (
     <Suspense fallback={<CircularProgress />}>{React.cloneElement(children, props)}</Suspense>
 );
 
-export default [
-    new RoutePlugin({
-        route: (
-            <Route
+export const Routes: React.FC = () => {
```
Pavel910: Usually, I colocate Routes and Menus, because both can be shown/hidden based on permissions (and I'm sure there will be some permissions for this down the line). It's easier to group them if they're in the same file, because  1 permission usually affects both the route AND a menu. So I suggest you combine Menus and Routes into a single file. I personally call it a Module in other apps.

packages/app-apw/src/views/publishingWorkflows/PublishingWorkflowForm.tsx
```diff
@@ -121,12 +130,13 @@ const PublishingWorkflowForm = () => {
...
                                     </AccordionItem>
                                     <AccordionItem
+                                        open={true}
                                         icon={<WorkflowScopeIcon />}
                                         title={t`Scope`}
                                         description={t`Define the conditions when this workflow applies.`}
                                         iconClassName={accordionIconStyle}
                                     >
-                                        <WorkflowScope Bind={Bind} value={get(data, "scope")} />
+                                        <WorkflowScope Bind={Bind} workflow={data} />
```
Pavel910, line 139: You can also use `useBind({ name: "whatever" })` inside of `WorkflowScope` (or anywhere inside the `<Form>`) to avoid prop drilling. You could try using that here.

packages/app-apw/src/views/publishingWorkflows/components/pageBuilder/ApwPageBuilderWorkflowScope.tsx
```diff
@@ -0,0 +1,48 @@
...
+                            >{t`Specific categories and pages`}</option>
+                        </Select>
+                    </Bind>
+                </Box>
+                <Box>
+                    {type === ApwWorkflowScopeTypes.CUSTOM && (
+                        <PbScopeSettings Bind={Bind} runValidation={noPages && noCategories} />
+                    )}
+                </Box>
+            </Stack>
```
Pavel910: Please extract this into a separate component. It will be a lot easier to read if HOC and the actual component are decoupled.

Pavel910 (changes requested):
Several change requests here, thanks!

PR #2426, opened 2022-05-24
Tools Improvements

adrians5j:
#### 1. Introduced `delete-empty-package-folders`

When switching branches, I often get left with empty folders in `packages` folder. Then I have to manually delete them, because otherwise, different "invalid package" errors start to occur within various tools. For example:
image

This PR introduces a simple `delete-empty-package-folders` command that will quickly clear all folders that are empty (leftovers from the previous branch).

#### 2. `webiny workspaces` - Improve Package Validity Check

As per comments below, the thing that pushed me to create a cleanup script was an error that I received when running `webiny watch` command ([see](https://github.com/webiny/webiny-js/pull/2426#issuecomment-1136266299)). But after further investigation, I found out that [this simple fix](https://github.com/webiny/webiny-js/pull/2426/commits/8ea1e83f5a94ec8dcb0da36326a981884da18889) was only thing that was needed to resolve it. 

So, from now on, these scripts will work even when there are leftover folders in dev's project.

## How Has This Been Tested?
Manual testing.

## Documentation
None. Internal tooling. Script added to `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
     "watch:apps": "yarn webiny ws run watch --scope='@webiny/app*'",
     "watch:api": "yarn webiny ws run watch --scope='@webiny/api*'",
     "clear-dist": "yarn rimraf packages/*/dist",
+    "delete-empty-package-folders": "node scripts/deleteEmptyPackageFolders.js",
     "contributors:add": "contreebutors add --username",
     "contributors:generate": "contreebutors render",
     "commit": "node ./scripts/commitizen.js",
--- a/packages/project-utils/workspaces/index.js
+++ b/packages/project-utils/workspaces/index.js
@@ -1,12 +1,12 @@
 const fs = require("fs");
 const { linkWorkspaces } = require("./linkWorkspaces");
 
-const isFolder = p => fs.statSync(p).isDirectory();
+const hasPackageJson = p => fs.existsSync(p + "/package.json");
 
 const allWorkspaces = () => {
     const path = require("path");
     return require("get-yarn-workspaces")()
-        .filter(isFolder)
+        .filter(hasPackageJson)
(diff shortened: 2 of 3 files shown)


Pavel910:
@doitadrian I'm wondering why this is still an issue when we have this: https://github.com/webiny/webiny-js/pull/2379. Build scripts should ideally not care about folders without package.json, but I guess there are scenarios where it's not covered. And it's mostly `get-yarn-workspaces` that's not handling this properly. 

Feel free to add this script, but ideally we should not use `get-yarn-workspaces` directly, but use our own tool that filters out invalid workspaces.

adrians5j:
Yeah, actually I got this error while trying to start watch on `apps/admin` app today (luckily I still have this in my terminal history):

image

Since I had a feeling that I've also seen this issue somewhere else / when using a different script, I decided to add this script. But actually I didn't know about the PR you linked, so maybe my case is the only leftover here.

BTW I wanted to use our own tool in order to get a list of all projects, but the thing is...since our tool works correctly and it doesn't include invalid packages, I had to use directly `get-yarn-workspaces`. Only with it I could get both valid and invalid packages. :)

Not sure what to do now. On one side I'd leave the script, could still be helpful. On other side maybe It'd be smarter to check what's causing the issue when running the `watch` and potentially also the `build` command. 

Hm...I'll take a quick look, maybe it's an easy fix. There's a stack trace, super helpful.

Will check, will let you know.

adrians5j:
Ah it's because of this:

[`packages/project-utils/workspaces/index.js`](https://github.com/webiny/webiny-js/blob/next/packages/project-utils/workspaces/index.js).

Let's see...

endymion1818 (approved):

PR #2435, opened 2022-05-25
feat(api-headless-cms): entry meta field

brunozoric:
## Changes
PR adds a meta field to CMS entry object.

Meta field data is only visible in the Manage API, when using GraphQL, in `meta.data` property in the response and directly on the entry object under `meta` property.

Meta data is not meant to store complex data structures. When using DynamoDB + Elasticsearch, it is indexed so it is searchable with the right set of plugins.

## How Has This Been Tested?
Jest.


Pavel910 (approved):
For v5, I don't think we have a better option so this looks good.
For v6, we might think about restructuring the schema, to have a better separation of system fields and user-defined fields.

PR #2453, opened 2022-05-30
fix(app-headless-cms): ensure value is an array

### LINKED ISSUE #2452: Headless CMS : An object can not contain multiple files in a Content Model
### Version

5.27.0

### Operating System

OS

### Browser

Chrome

Hey Team, Thank you for your support. I have an issue regarding setting up Content Models. 

When I add files to the property, a list of files inside an object errors out in the preview and if I tried to save the model it errors out as well when trying to create a new Entity. 

Screenshot 2022-05-30 at 12 38 48

I want an object which contains all the files so that I am able to name the collection of files uploaded.

### What do you see instead?

_No response_

### Additional information

_No response_

### Possible solution

_No response_

Pavel910:
## Changes
This PR ensures that `file` field renderer value defaults to an array when configured as `multipleValues=true`.

Closes #2452 

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileFields.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileFields.tsx
@@ -38,7 +38,7 @@ const FieldRenderer: React.FC<FieldRendererProps> = ({ getBind, Label, field }) 
                 const { onChange } = bind;
 
                 // We need to make sure the value is an array, since this is a multi-value component.
-                const value = bind.value || [];
+                const value = (Array.isArray(bind.value) ? bind.value : [bind.value]).filter(Boolean);
 
                 return (
                     <FileUploadWrapper className={imageWrapperStyles}>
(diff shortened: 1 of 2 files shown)


packages/app-headless-cms/src/admin/plugins/fieldRenderers/file/fileFields.tsx
```diff
@@ -51,7 +55,7 @@ const FieldRenderer: React.FC<FieldRendererProps> = ({ getBind, Label, field })
 
                                         const urls = files.map(f => f.src);
                                         if (index === -1) {
-                                            onChange([...(value || []), ...urls]);
```
brunozoric, line 54: Maybe do a check with `Array.isArray(value)`?

Pavel910: @brunozoric The default value is handled on line `41`.  But, what should I do if the value is not an array for some reason? For example, user can switch from single to multipleValues after records already exist 🤔  Should I assume that the single value is a file URL, and just wrap it with an array?

brunozoric: Hm, thats good idea. If is array do nothing, if is string, make it string array and in the end filter with boolean to remove empty values in the array.
So something like this:
```typescript
const value = (Array.isArray(bind.value) ? bind.value : [bind.value]).filter(Boolean);
```

PR #2458, opened 2022-06-01
feat: add API for Page Builder Block creation

neatbyte-vnobis:
## Changes
Created new entity type Block. Add create GraphQL API for it.

## How Has This Been Tested?
Manual and integration tests.

## Documentation
Will be handled in separate pull request.


packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
```diff
@@ -0,0 +1,58 @@
...
+            id: {
+                type: "string"
+            },
+            name: {
+                type: "string"
+            },
+            type: {
+                type: "string"
+            },
+            blockCategory: {
```
adrians5j: Can we have it just as a `category`?

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
```diff
@@ -0,0 +1,58 @@
...
+            TYPE: {
+                type: "string"
+            },
+            id: {
+                type: "string"
+            },
+            name: {
+                type: "string"
+            },
+            type: {
```
adrians5j: What is `type` ? Can you remind me?

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
```diff
@@ -0,0 +1,58 @@
...
+                type: "string"
+            },
+            blockCategory: {
+                type: "string"
+            },
+            content: {
+                type: "map"
+            },
+            preview: {
```
adrians5j: What is `preview`? Does this need to be `map`?

packages/api-page-builder/src/types.ts
```diff
@@ -812,3 +813,36 @@ export interface BlockCategoryStorageOperations {
...
     delete(params: BlockCategoryStorageOperationsDeleteParams): Promise<BlockCategory>;
 }
+
+/**
+ * @category RecordModel
+ */
+export interface Block {
+    id: string;
+    name: string;
+    type: "element" | "block";
```
adrians5j: Do we really need this? 🤔

packages/api-page-builder/src/graphql/graphql/blocks.gql.ts
```diff
@@ -0,0 +1,50 @@
+import { GraphQLSchemaPlugin } from "@webiny/handler-graphql/types";
+import resolve from "./utils/resolve";
+import { PbContext } from "../types";
+
+export const createBlockGraphQL = (): GraphQLSchemaPlugin<PbContext> => {
+    return {
+        type: "graphql-schema",
```
adrians5j: Instead of returning an object, we can return an instance of a new `GraphQlSchemaPlugin`. For example: `packages/api-admin-users-cognito/src/graphql/user.gql.ts:18`.

packages/api-page-builder-so-ddb-es/src/operations/block/index.ts
```diff
@@ -0,0 +1,63 @@
...
+}
+export const createBlockStorageOperations = ({
+    entity
+}: CreateBlockStorageOperationsParams): BlockStorageOperations => {
+    const dataLoader = new BlockDataLoader({
+        entity
+    });
+
+    const create = async (params: BlockStorageOperationsCreateParams) => {
+        const { block } = params;
```
adrians5j: Quick question - this does contain the ID of the block (`block.id`), right?

adrians5j (changes requested):
I checked the majority of the PR, still need to check a couple of files.

All in all looks good. But I have a couple questions. 

Before continuing, most probably we'll want to see if we really need the `block.type` field here. I don't think that's going to be the case. Will check tomorrow and confirm.

P.S. 1: Please check errors reported in CI/CD.
P.S. 2: Just so you're not blocked, I believe you can start slowly working on PR2 stuff, because I don't see any big changes we'll want to perform here. It's up to you.

adrians5j (commented):
Did one additional review.

I don't have anything to add. Just please check the comments I left yday.

Thank you!

packages/api-page-builder-so-ddb-es/src/operations/block/fields.ts
```diff
@@ -0,0 +1,21 @@
+import { BlockDynamoDbFieldPlugin } from "~/plugins/definitions/BlockDynamoDbFieldPlugin";
```
adrians5j: BTW let's be more consistent with the rest of the system and name this folder `pageBlock`, the same way we have it with `pageElement`:

image

packages/api-page-builder-so-ddb/src/definitions/blockEntity.ts
```diff
@@ -0,0 +1,58 @@
+import { Entity, Table } from "dynamodb-toolbox";
```
adrians5j: Same here as in my previous comment. Let's be more consistent and name this `pageBlockEntity`. The same way we have it here:
image

adrians5j (commented):
Additional comment regarding the `type` field, which I mentioned in my initial review: we don't need it. This has existed on the `pageElement` because that entity encapsulated both blocks and page elements. Since we now have distinct entities for blocks, type is no longer needed.

Also, let's for now leave type on the page-elements side.  You can maybe just leave a comment here (`packages/api-page-builder/src/types.ts:12`):

```
// Not relevant since the introduction of page blocks. Left for backwards compatibility purposes.
```

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
neatbyte-vnobis: I took a chance to investigate this matter:
https://github.com/webiny/webiny-js/blob/809fcc5f4a4795d41e69d7fd1b07541e55408046/packages/app-page-builder/src/admin/views/Pages/Editor.tsx#L69
Look like this property is used by Page Builder Editor (Sorry, I cannot provide more deep details)
Possible values are "element" and "block" (but not same block as new entity I have created).

As new "Block" entity is mostly a "copy" of "pageElement" (as per requirements) and will be used in page builder too - same fields are present

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
neatbyte-vnobis: yes, but I suggest to keep it as is.
We have a "Category" entity in Page Builder app (it describes a "page category"), so it may be confusing to use "category" here (As it is a string value of slug, so nothing is pointing that it is a "blockCategory", not a "category" of the pages)

packages/api-page-builder-so-ddb-es/src/operations/block/fields.ts
neatbyte-vnobis: @adrians5j, should I change only folder name or entity name in all places?
I mean, apply this change only to ddb project or rename typescript types to "PageBlock" in api project too?

packages/api-page-builder/__tests__/graphql/lifecycleEvents.blocks.test.ts
```diff
@@ -0,0 +1,73 @@
...
+    let dummyBlock: Block;
+
+    beforeEach(async () => {
+        await createBlockCategory({
+            data: {
+                slug: blockCategory,
+                name: `name`
+            }
+        });
+        dummyBlock = await createDummyBlock();
```
neatbyte-vnobis: Regarding the warning: this variable will be used in update/delete test in next pull request.

packages/api-page-builder/src/types.ts
neatbyte-vnobis: yes, reply above (used in page builder app - https://github.com/webiny/webiny-js/blob/809fcc5f4a4795d41e69d7fd1b07541e55408046/packages/app-page-builder/src/admin/views/Pages/Editor.tsx#L69)

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
neatbyte-vnobis: according to code - it is an image data
https://github.com/webiny/webiny-js/blob/809fcc5f4a4795d41e69d7fd1b07541e55408046/packages/app-page-builder/src/admin/utils/createBlockPlugin.tsx#L13

looks like "map" is used to describe an object here (with image data, like src, etc.)?

packages/api-page-builder-so-ddb-es/src/operations/block/index.ts
neatbyte-vnobis: Yes, ID is created inside `api` create method:
https://github.com/webiny/webiny-js/blob/809fcc5f4a4795d41e69d7fd1b07541e55408046/packages/api-page-builder/src/graphql/crud/blocks.crud.ts#L74

It is used inside "createSortKey"

packages/api-page-builder/src/graphql/graphql/blocks.gql.ts
neatbyte-vnobis: @adrians5j 
Do you mean to remove arrow function and export just an object (anyway it is always same)?
I will still use a named export (so it will be same as other api-page-builder files).

Also, other "api-page-builder" graphql files have an arrow function too - change them or leave as is?

packages/api-page-builder-so-ddb-es/src/operations/block/fields.ts
adrians5j: Let's do both. So, rename the folder and also make sure we have the full `pageBlock` name in consts / types / etc. For example, `createBlockEntity` should be `createPageBlockEntity`.

packages/api-page-builder/__tests__/graphql/lifecycleEvents.blocks.test.ts
adrians5j: Cool.

Still, let's then use `// eslint-disable-next-line` in order for the CI checks to all pass.

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
adrians5j: Alright, let's keep it as is. 

Thanks for clarification!

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
adrians5j: Yeah, but if we just look at the `type` field, it doesn't serve any purpose. That's why I'd just remove it.

We'll see how we'll deal with this on the frontend later.

packages/api-page-builder-so-ddb-es/src/definitions/blockEntity.ts
adrians5j: OK, let's leave it as is.

Thanks!

packages/api-page-builder-so-ddb-es/src/operations/block/index.ts
adrians5j: Thanks!

adrians5j (changes requested):
Almost all good. Just a few minor pending things to look at and we merge.

Thanks for the clarifications you put here, helped!

packages/api-page-builder/src/graphql/graphql/blocks.gql.ts
neatbyte-vnobis: @adrians5j 
have you missed this comment? or do nothing?

packages/api-page-builder-so-ddb/src/definitions/blockEntity.ts
neatbyte-vnobis: answered in issue above
https://github.com/webiny/webiny-js/pull/2458/files#r891042674

packages/api-page-builder-so-ddb-es/src/operations/block/index.ts
neatbyte-vnobis: resolving issues

packages/api-page-builder/src/types.ts
neatbyte-vnobis: according to decision to remove `type` field at all - it will be removed

packages/api-page-builder-so-ddb-es/src/operations/block/fields.ts
neatbyte-vnobis: @adrians5j @SvenAlHamad 
Just in case, I want to ask - what to do with block categories code? (it is already merged into umbrella branch)
it is not starting with `page*` prefix (not it is "blockCategory", not a "pageBlockCategory")

packages/api-page-builder/src/graphql/graphql/blocks.gql.ts
adrians5j: I answered this via Slack @neatbyte-vnobis.

Sorry for the delay on this one, I actually missed your comment.

adrians5j (approved):
Looks good! Thanks for going the extra mile on the naming consistency.

Let's move forward!

PR #2460, opened 2022-06-01
Extract the Composition API into a standalone package

Pavel910:
## Changes
This PR extracts the Composition APi from `@webiny/app-admin-core` into a dedicated package, `@webiny/react-composition`, so it can be used outside of the Admin app, and even outside of Webiny. It's just a cool little composition framework.

This PR also makes `@webiny/react-router` `<Link>` component composable, using this composition package. It's a very useful addition, that can help us with implementation of, for example, analytics, data preloading (on the website), and even prefixing all Admin links with tenant IDs in a multi-tenant setup without thinking about it. It simply makes it super easy to compose all links in your app (website and admin).

## How Has This Been Tested?
Manually. Jest tests pending for `react-composition` package.

## Documentation
Not necessary at the moment; this package is used internally, and it's re-exported from `@webiny/app-admin`, so it's already covered by our existing docs.

### DIFF
--- a/packages/app-admin-core/src/admin.tsx
+++ b/packages/app-admin-core/src/admin.tsx
@@ -4,30 +4,18 @@ import React, {
     useMemo,
     useState,
     useCallback,
-    FC,
     FunctionComponentElement,
-    ComponentType,
     ReactElement
 } from "react";
 import { BrowserRouter, RouteProps } from "@webiny/react-router";
+import { compose, HigherOrderComponent, CompositionProvider } from "@webiny/react-composition";
 import { Routes as SortRoutes } from "./components/utils/Routes";
 import { DebounceRender } from "./components/utils/DebounceRender";
 import { PluginsProvider } from "./components/core/Plugins";
 
-function compose(...fns: HigherOrderComponent[]) {
-    return function ComposedComponent(Base: FC<unknown>): FC<unknown> {
-        return fns.reduceRight((Component, hoc) => hoc(Component), Base);
-    };
-}
-
-interface Wrapper {
-    component: ComponentType<unknown>;
-    wrappers: HigherOrderComponent[];
-}
 interface State {
     routes: Record<string, ReactElement<RouteProps>>;
     plugins: JSX.Element[];
-    wrappers: Map<ComponentType<unknown>, Wrapper>;
     providers: HigherOrderComponent[];
 }
 
@@ -35,20 +23,12 @@ interface AdminContext extends State {
     addRoute(route: JSX.Element): void;
     addProvider(hoc: HigherOrderComponent): void;
     addPlugin(plugin: React.ReactNode): void;
-    addComponentWrappers(
-        component: React.ComponentType<unknown>,
-        hocs: HigherOrderComponent[]
-    ): void;
 }
 
 const AdminContext = createContext<AdminContext>({
     routes: {},
     plugins: [],
-    wrappers: new Map(),
     providers: [],
-    addComponentWrappers: () => {
-        return void 0;
-    },
     addPlugin: () => {
         return void 0;
     },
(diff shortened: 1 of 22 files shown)


packages/app-admin-core/src/admin.tsx
```diff
@@ -179,15 +115,17 @@ export const Admin = ({ children }: AdminProps) => {
...
+                {children}
+                <BrowserRouter>
+                    <Providers>
+                        <PluginsProvider>{state.plugins}</PluginsProvider>
+                        <DebounceRender>
+                            <AdminRouter />
+                        </DebounceRender>
+                    </Providers>
+                </BrowserRouter>
+            </CompositionProvider>
```
endymion1818, line 128: Such a nice api.

packages/react-composition/package.json
```diff
@@ -0,0 +1,44 @@
...
+  "devDependencies": {
+    "@babel/cli": "^7.16.0",
+    "@babel/core": "^7.16.0",
+    "@babel/preset-env": "^7.16.4",
+    "@babel/preset-typescript": "^7.16.0",
+    "@webiny/cli": "^5.27.0",
+    "@webiny/project-utils": "^5.27.0",
+    "jest": "^26.6.3",
+    "merge": "^1.2.1",
+    "rimraf": "^3.0.2",
```
endymion1818: Necessary as a dependency?

endymion1818: `merge` too?

packages/react-composition/src/Context.tsx
```diff
@@ -0,0 +1,139 @@
...
+    /**
+     * HOCs used to compose the original component.
+     */
+    hocs: HigherOrderComponent[];
+}
+
+/**
+ * IMPORTANT: TInputProps default type is `any` because this interface is use as a prop type in the `Compose` component.
+ * You can pass any HigherOrderComponent as a prop, regardless of its TInputProps type. The only way to allow that is
+ * to let it be `any` in this interface.
```
endymion1818, line 31: Good explanation 👍🏻

packages/react-composition/src/Context.tsx
```diff
@@ -0,0 +1,139 @@
...
+    (Component: FC<TInputProps>): FC<TOutputProps>;
+}
+
+type ComposedComponents = Map<ComponentType<unknown>, ComposedComponent>;
+
+interface CompositionContext {
+    components: ComposedComponents;
+    getComponent(component: ComponentType<unknown>): ComponentType<unknown> | undefined;
+    composeComponent(component: ComponentType<unknown>, hocs: HigherOrderComponent[]): void;
+}
```
endymion1818, line 43: Quite a lot of typescript interfaces & types here and above, worth abstracting into their own file?

packages/react-composition/src/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from "./Context";
+// export type { HigherOrderComponent } from "./Context";
```
endymion1818: 🙈

packages/react-composition/src/index.ts
```diff
@@ -0,0 +1,5 @@
...
+export * from "./Compose";
+// export type { ComposeProps, ComposableFC } from "./Compose";
```
endymion1818: 🙈

endymion1818 (commented):
Great to have this a separate package, you have a few lines commented out hence the comment instead of approval.

packages/react-composition/package.json
Pavel910: `jest` and `merge` are probably not necessary. `rimraf` I have to check, maybe we can remove it from all package deps (it's present everywhere), because it used to be in the package scripts, but now it's no longer a direct dep.

packages/react-composition/src/Context.tsx
Pavel910: I like to keep internal types colocated with code, especially because they're not exported for public usage. It's also easier to maintain them this way. Not all types need to be exported from the package root, so I kept these here.

packages/react-composition/src/index.ts
Pavel910: Ah good catch, forgot to double-check these.

PR #2464, opened 2022-06-03
Fix not-found page prerendering

Pavel910:
## Changes
This PR fixes issues with prerendering of "not found" pages. Until now, `/not-found` path was a system path that we assumed will always be used for rendering of a not-found page. However, if you change the website settings in the Admin app, and change that page for a custom one, not found page rendering will keep returning the old page.

Notable changes:
- removed `home`, `notFound`, and `visibility` settings from the Page 
- added missing meta data to `not found` page, when website settings are updated
- `__PS_NOT_FOUND_PAGE__` now contains the path of current "not found" page, instead of being just a `boolean`
- `Page/index.tsx` in the Website app is refactored to use react-router location, instead of `window.location` (this was a bug of its own; changes in the URL would not trigger component re-render, and a new page would never load)
- implemented handling of `returnNotFoundPage` (which never did anything)
- fixed Page Editor loading when `createPageFrom` was triggered (missing dependency in `useMemo`)

## How Has This Been Tested?
Manually.


packages/api-prerendering-service/src/render/renderUrl.ts
```diff
@@ -160,7 +160,7 @@ export const defaultRenderUrlFunction = async (
     const notFoundPage = params?.args?.configuration?.meta?.notFoundPage;
     if (notFoundPage) {
         console.log("Setting locale (__PS_NOT_FOUND_PAGE__) to window object....");
-        windowSet(browserPage, "__PS_NOT_FOUND_PAGE__", true);
+        windowSet(browserPage, "__PS_NOT_FOUND_PAGE__", params.args.path || "");
```
Pavel910: This might not even be necessary, double-check.

packages/api-page-builder/src/prerendering/hooks/afterSettingsUpdate.ts
```diff
@@ -39,6 +39,9 @@ export default () => {
                                 {
                                     path: page.path,
                                     configuration: {
+                                        meta: {
+                                            notFoundPage: true
+                                        },
                                         storage: { folder: NOT_FOUND_FOLDER }
                                     }
```
Pavel910: Aadd another entry for previous not-found page, to remove `meta`.

adrians5j (dismissed):
Marking this as "changes requested", because of the additional call notes.

packages/api-prerendering-service/src/render/renderUrl.ts
Pavel910: Just to confirm @adrians5j, this is definitely _not_ used. I removed it in my last commit.

PR #2469, opened 2022-06-09
fix(prerendering): update to work with event-driven prerendering

Pavel910:
## Changes
This PR adds the missing pieces to the new setup of Pulumi to make the Prerendering Service work. One of the main things that are addressed with this PR is making prerendering work in the new event-driven environment, using EventBus and SQS, where jobs can be processed in parallel. There were issues with parallel jobs working on the same DDB record, randomly overwriting each other. Previously, this issue was not visible, because all jobs were processed sequentially, and it never came to light.

* added a bunch of comments to clarify certain piece of code
* some code was slightly refactored to improve readability
* dropped redundant code (like passing `context` all over the place)
* added the ability to `exclude` paths from rerendering (when issuing full rerenders, we sometimes need to exclude certain paths, to avoid concurrent double-processing)
* `notFoundPage` flag is now handled during HTML generation, and a copy of this page is now stored to `_NOT_FOUND_PAGE_` folder in S3 bucket (this removes the need to issue 2 render jobs, and solves the concurrent rendering problem)

## How Has This Been Tested?
Manually.

## Related PRs
- [x] #2478
- [x] #2484
- [x] #2497


adrians5j (dismissed):
Marking this as "Request changes".

PR #2486, opened 2022-06-15
Fix validation for empty string slug value

neatbyte-vnobis:
## Changes
Fix for Page Block Categories API: usage of empty string as a slug value caused and error about invalid DynamoDB query

## How Has This Been Tested?
Manual testing

## Documentation
No updates in documentation


adrians5j (approved):
Looks great. Thank you!

PR #2495, opened 2022-06-18
fix: check for existing Page Blocks before Block Category delete operation

neatbyte-vnobis:
## Changes
Add check for existing Page Blocks before deleting a Block Category entity (so Page Blocks will always reference an existing Block Category)

## How Has This Been Tested?
Manual testing

## Documentation
None


adrians5j (approved):
Also good! Thanks for implementing this check!

PR #2496, opened 2022-06-19
Page builder blocks v2 UI list page blocks

neatbyte-vnobis:
## Changes
Add Page Blocks page to Page Builder UI app

## How Has This Been Tested?
Manual testing

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/PageBuilder.tsx
+++ b/packages/app-page-builder/src/PageBuilder.tsx
@@ -57,7 +57,7 @@ const PageBuilderMenu: React.FC = () => {
                         />
                     </HasPermission>
                 </Menu>
-                <Menu name="pageBuilder.blocks" label={"Page Blocks"}>
+                <Menu name="pageBuilder.blocks" label={"Blocks"}>
                     <HasPermission name={"pb.block"}>
                         <Menu
                             name="pageBuilder.blocks.categories"
--- a/packages/app-page-builder/src/PageBuilder.tsx
+++ b/packages/app-page-builder/src/PageBuilder.tsx
@@ -57,12 +57,17 @@ const PageBuilderMenu: React.FC = () => {
                         />
                     </HasPermission>
                 </Menu>
-                <Menu name="pageBuilder.blocks" label={"Blocks"}>
+                <Menu name="pageBuilder.blocks" label={"Page Blocks"}>
                     <HasPermission name={"pb.block"}>
                         <Menu
                             name="pageBuilder.blocks.categories"
                             label={"Categories"}
                             path="/page-builder/block-categories"
+                        />
+                        <Menu
+                            name="pageBuilder.blocks.pageBlocks"
+                            label={"Blocks"}
+                            path="/page-builder/page-blocks"
                         />
                     </HasPermission>
                 </Menu>
--- a/packages/app-page-builder/src/admin/plugins/routes.tsx
+++ b/packages/app-page-builder/src/admin/plugins/routes.tsx
@@ -123,7 +123,7 @@ const plugins: RoutePlugin[] = [
                 render={() => (
                     <SecureRoute permission={ROLE_PB_BLOCK}>
                         <AdminLayout>
-                            <Helmet title={"Page Builder - Page Blocks"} />
+                            <Helmet title={"Blocks"} />
                             <PageBlocks />
                         </AdminLayout>
                     </SecureRoute>
--- a/packages/app-page-builder/src/admin/plugins/routes.tsx
+++ b/packages/app-page-builder/src/admin/plugins/routes.tsx
@@ -11,6 +11,7 @@ import Menus from "../views/Menus/Menus";
 import Pages from "../views/Pages/Pages";
 import Editor from "../views/Pages/Editor";
 import BlockCategories from "../views/BlockCategories/BlockCategories";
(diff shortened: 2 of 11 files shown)


packages/app-page-builder/src/admin/plugins/routes.tsx
```diff
@@ -111,6 +112,24 @@ const plugins: RoutePlugin[] = [
...
+        name: "route-pb-page-blocks",
+        type: "route",
+        route: (
+            <Route
+                exact
+                path="/page-builder/page-blocks"
+                render={() => (
+                    <SecureRoute permission={ROLE_PB_BLOCK}>
+                        <AdminLayout>
+                            <Helmet title={"Page Builder - Page Blocks"} />
```
adrians5j: On the frontend / where we're just printing labels/titles, we can use just `Blocks` label (so, `Page Builder - Blocks`.

packages/app-page-builder/src/PageBuilder.tsx
```diff
@@ -57,13 +57,18 @@ const PageBuilderMenu: React.FC = () => {
                         />
                     </HasPermission>
                 </Menu>
-                <Menu name="pageBuilder.blocks" label={"Blocks"}>
+                <Menu name="pageBuilder.blocks" label={"Page Blocks"}>
```
adrians5j: Let's do just `Blocks` here as well.

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
```diff
@@ -0,0 +1,168 @@
+import React, { useCallback, useMemo, useState } from "react";
+import { i18n } from "@webiny/app/i18n";
+import { useRouter } from "@webiny/react-router";
+import { useQuery } from "@apollo/react-hooks";
```
adrians5j, line 4: Can you clarify please, what is this view? Can you provide a screenshot?

adrians5j (changes requested):
Just a couple of comments.

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
neatbyte-vnobis: image (12)
Here is a screenshot

Design link - https://miro.com/app/board/uXjVO1XGqwg=/?share_link_id=951919244316

adrians5j:
Was about to merget this, but I noticed one thing.

@neatbyte-vnobis , should this be "Search categories" here? Maybe "Search blocks" ?

image

neatbyte-vnobis:
@adrians5j 
There we are searching not blocks but categories (left panel list)
But here I may be wrong...
@SvenAlHamad can you clarify?

Design (https://miro.com/app/board/uXjVO1XGqwg=/):
image

SvenAlHamad:
We are searching for blocks, that’s correct.

neatbyte-vnobis:
@SvenAlHamad @adrians5j 
In such case left panel should be affected too, am I right?
So if we search for "Heading" and there are 3 categories with such blocks - we show 3 categories and only blocks containing "Heading" word.
In such case:
1) other categories are not shown
2) counter bellow categories will show number of blocks that are matching search query

I will have to rewrite quite a few things to do it:
now left and right parts of the page are more or less independent (as other similar pages aka categories)
but with such change I need to move all GraphQL queries to parent component (so left and right parts will be using a data from it)
@adrians5j maybe I am missing an easy solution here? Feel free to give me pointer.

adrians5j:
@neatbyte-vnobis In all of our views, these search inputs are doing a search over the entity the user's currently looking at.

For example, on the pages view, pages are being searched. On the page categories view, page categories are being searched.

So, if we were to follow the above examples and wanted to stay consistent, typing into the search field would do a search against existing page blocks.

But!

Now when I look at it, I can see that this view **IS** a bit different than the other ones we have. Instead of just showing blocks in the list, we're showing block categories. Once clicked, the blocks are shown on the right.

And now I do think we need to think about this.

I do agree that this should search for blocks because the user will most probably want to do that (instead of searching for block categories). But I can also see that, in the provided designs, the label said `SEARCH CATEGORIES`, so I can understand that this might require some changes on your side @neatbyte-vnobis.

So.... what I'd do here now is....... 

@SvenAlHamad - please confirm the direction into which we want to move. Do we want to leave it as-is (and save potentially some time), or do we want to revisit this now, and make the search bar do a search over blocks?

Hopefully, I didn't miss anything (like some docs), but yeah, let's first sync on what exactly we want to have here.

SvenAlHamad:
Ok, let's leave it so it's searching blocks if that's what's currently implemented. Once we start using it a bit I think we'll have then one milestone addressing the initial feedback and some additional polishes.

adrians5j:
@SvenAlHamad 

Actually, the way it's currently implemented is that it's searching for block categories, not blocks.

Miro:
image

Real app:
image

I'd still leave it as is, so we can address it then in the MS you mentioned.

SvenAlHamad:
Sorry, I miss typed in my previous comment. I DO understand it's searching "block categories" and I'm happy to leave it like that for now.

adrians5j:
@neatbyte-vnobis feel free to merge this one. 

Thank you!

adrians5j (approved):

PR #2500, opened 2022-06-22
Enable APW tests and disable APW in Pulumi

Pavel910:
## Changes
This PR extracts APW Pulumi config into `apps/api/pulumi` while it's still in development. Also, APW is removed from `app-serverless-aws` package until it's ready to be released. This way the app is not enabled even if WCP license allows you to access it.

For development, the app can still be activated using the standard admin app plugin, and pulumi app configs.

## How Has This Been Tested?
Manually.


adrians5j (approved):
Checked it - looking good!

PR #2515, opened 2022-07-04
Fixing getFileTypePlugin for better finding of the correct plugin

DasRed:
Insure loop breaks after first match.

### DIFF
--- a/packages/app-admin/src/components/FileManager/getFileTypePlugin.ts
+++ b/packages/app-admin/src/components/FileManager/getFileTypePlugin.ts
@@ -30,6 +30,7 @@ export default function getFileTypePlugin(
         }
         if (types.find(t => minimatch(file.type, t))) {
             plugin = fileTypePlugins[i];
+            break;
         }
     }
 
--- a/packages/app-admin/src/components/FileManager/getFileTypePlugin.ts
+++ b/packages/app-admin/src/components/FileManager/getFileTypePlugin.ts
@@ -15,26 +15,37 @@ export default function getFileTypePlugin(
     const fileTypePlugins = [
         ...plugins.byType<AdminFileManagerFileTypePlugin>("admin-file-manager-file-type"),
         ...plugins.byType<FileManagerFileTypePlugin>(FileManagerFileTypePlugin.type)
-    ];
+    ]
+    // make it unique
+    .filter((value, index, self) => self.indexOf(value) === index);
 
-    /**
-     * TODO: if we are searching last available plugin, we can refactor this.
-     * TODO: check out @pavel
-     */
-    let plugin = null;
-    for (let i = 0; i < fileTypePlugins.length; i++) {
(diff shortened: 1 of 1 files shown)


DasRed:
The previous version of the method getFileTypePlugin can not handle the plugins correctly.

The following plugin array can be:
```javascript
[
    {types: ['image/*'], name: 'A'},
    {types: ['*/*'], name: 'B'},
]
```
In that case, the second entry will be selected, because the iteration always iterate over all entries. If an entry matches, the previous found entry will be overwritten.

If you add a plugin with video/*, this plugin will be added at the beginning of the array. The new plugin will never be taken.

The new version makes the plugin array unique, reduces the amount of plugins per type to 1 (more than one plugin per type can not be handled by the current ui) and sorts the types by amount of '*'. Lower count of '*' will be more at the beginning of the result.

**Example**
This could be the "raw" plugin array
```javascript
[
    {types: ['video/*'], name: 'C'},
    {types: ['video/mp4'], name: 'D'},
    {types: ['image/*'], name: 'A'},
    {types: ['video/mp4'], name: 'E'},
    {types: ['*/*'], name: 'B'},
]
```
Previous version would choose the plugin B.

After sorting
```javascript
[
    {types: ['video/mp4'], name: 'D'},
    {types: ['video/*'], name: 'C'},
    {types: ['image/*'], name: 'A'},
    {types: ['*/*'], name: 'B'},
]
```
New version would choose the plugin D.

DasRed:
I did not write a unit test, because I can't find the unit test structure. Where can I implement the unit test for this function?

Pavel910:
@DasRed Sorry for late reply. Here we go:

> In that case, the second entry will be selected, because the iteration always iterate over all entries. If an entry matches, the previous found entry will be overwritten.

This was intentional, to allow people to register their own plugins and overwrite existing once. If we just stop at first match, you will never be able to override a plugin for any type that is already registered.

I like the sorting approach you introduced though; I think we just need to make it `last in wins` when grouping by type, to allow overrides. Say Webiny provides a `video/mp4` default plugin, if we do `first in wins`, you willn ever be able to override it. If we do `last in wins`, you'll just register your own `video/mp4` plugin and it will override the default one in the process of grouping.


For unit tests, please have a look at `packages/utils`, or `packages/plugins` to find examples of both the Jest setup, as well as PluginsContainer testing. For testing this function, you can either register plugins into the global plugins container, or modify the function to accept plugins container as a second parameter.

Let us know if you need help with this.

Pavel910:
@DasRed I'm closing this PR in favor of https://github.com/webiny/webiny-js/pull/2605, since we, unfortunately, lost touch and we really want to fix this. Your code was used in the new PR, so all the credit goes to you (I described it in the other PR), I just added some internal tweaks necessary to cover other use-cases that we have in Webiny.

Thanks for your contribution! 🍻

PR #2561, opened 2022-08-02
chore: upgrade to react 17

leopuleo:
## Changes
This PR upgrades React from ~16 to ~17.

## How Has This Been Tested?

Manually and CI/CD


Pavel910 (approved):
This looks great, the changes to the code are minimal, it's mostly package.json updates.
We need to test what happens to projects that are created with React 16, and then get upgraded. But let's do it in a separate issue. I expect there will be some project upgrades to update resolutions and app dependencies in the user's projects.

PR #2568, opened 2022-08-08
feat: add Block Editor integration

neatbyte-vnobis:
## Changes
Add Block Editor page integration

## How Has This Been Tested?
Manual

## Documentation
None


Pavel910:
Add a message saying that there are no block categories, and a link to "Create a new category", like we have on the Pages view:
CleanShot 2022-08-10 at 10 08 45

CleanShot 2022-08-10 at 10 07 59


Add some padding to create block dialog, same as in the pages view. Feel free to use identical styles that are present in the Pages view, and render category title and a slug underneath:
CleanShot 2022-08-10 at 10 10 24

Change the `blockCategory` property to `category`, the `block` prefix is redundant:
CleanShot 2022-08-10 at 10 14 38

In Page Builder blocks list, it would be nice if categories could have icons and descriptions. For that we need to expand the Block Categories module to have an icon and a description. There already is an Icon Picker component, see here: `packages/app-headless-cms/src/admin/views/contentModelGroups/IconPicker.tsx`. You can just C/P the component, and we'll worry about reusability later. For a visual example, see `http://localhost:3001/cms/content-model-groups?new=true`

CleanShot 2022-08-10 at 12 20 02

For `blockCategories`, which you load and generate plugins in the `Editor.tsx`, you really don't need to complicate things with those plugins; you can simply have a Recoil Atom for that. It will be a lot easier to use that data anywhere you need, especially if you create a `useBlockCategories` hook.
CleanShot 2022-08-10 at 12 25 39@2x

Regarding the dotted element outline on the block preview img, don't worry about that for now. It's not related to element highlight, that's just how we styled elements in the editor. it's not vital at this point so don't bother with that.

neatbyte-vnobis:
@SvenAlHamad Please confirm part about "categories could have icons and descriptions"
it requires API changes (2 new fields), so we need to extend scope for these changes.
(or we can create an "improvement" task about it)

SvenAlHamad:
Hey @neatbyte-vnobis happy for your to invest the extra time to add the two fields. Please just confirm the estimate for this so I can update the milestone on upwork.

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
```diff
@@ -247,17 +255,45 @@ const BlocksByCategoriesDataList = ({ canCreate }: PageBuilderBlocksByCategories
...
+                                ))}
+                            </List>
+                        )}
+                    </React.Fragment>
+                </DialogContent>
+                <DialogActions>
+                    <ButtonDefault
+                        onClick={() => history.push("/page-builder/block-categories?new=true")}
+                    >
+                        + Create new block category
```
Pavel910: Grammar fix: `Create a new block category`

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
```diff
@@ -247,17 +255,45 @@ const BlocksByCategoriesDataList = ({ canCreate }: PageBuilderBlocksByCategories
             <Dialog
                 open={isNewPageBlockDialogOpen}
                 onClose={() => setIsNewPageBlockDialogOpen(false)}
+                className={narrowDialog}
             >
                 <DialogTitle>
                     <Typography use="headline5">Please select block category</Typography>
```
Pavel910: Grammar fix: `Please select a block category`

packages/app-page-builder/src/blockEditor/Editor.tsx
```diff
@@ -87,7 +84,10 @@ export const BlockEditor: React.FC = () => {
             <BlockEditorConfig />
             <LoadData>
                 <PbEditor
-                    stateInitializerFactory={createStateInitializer(block as BlockWithContent)}
+                    stateInitializerFactory={createStateInitializer(
+                        block as BlockWithContent,
+                        blockCategories as BlockCategoriesAtomType
```
Pavel910: As noted in my previous comment, you won't need this, if you move the whole logic of loading block categories into the hook.

packages/app-page-builder/src/blockEditor/Editor.tsx
```diff
@@ -48,11 +48,8 @@ export const BlockEditor: React.FC = () => {
             .then(({ data }) => {
                 const blockCategoriesData: PbBlockCategory[] =
                     get(data, "pageBuilder.listBlockCategories.data") || [];
-                blockCategoriesData.forEach(element => {
-                    createBlockCategoryPlugin({
-                        ...element
-                    });
-                });
+
```
Pavel910: To make this part reusable, I would move this whole section of `const blockCategories = client.query(...)` into the `useBlockCategories` hook, and then you also don't need to pass blocks to editor state initializer. So you would reuse that hook from both the block editor, as well as the page editor.

It just crossed my mind that you could also just use Apollo cache as state management for block categories, so you won't even need a recoil atom 🤦 (sorry I pushed you into that direction) . You basically just have the `useQuery` hook wrapped into a `useBlockCategories` hook, and all of a sudden it's reusable everywhere without any state management.

Something like this:
```ts
export const useBlockCategories() {
    const { data } = useQuery(LIST_BLOCK_CATEGORIES);

    return get(data, "pageBuilder.listBlockCategories.data") || [];
}
```

What do you think?

Pavel910 (changes requested):
@neatbyte-vnobis we're almost there. I'll just bother you with a few small things, and hopefully this is the last change request :)

Pavel910 (approved):
Thanks @neatbyte-vnobis! Feel free to merge this.

PR #2582, opened 2022-08-14
[TEST] Cypress tests for categories - MS3

razvanvancea:
## Changes
This PR contains new cypress tests on the categories (milestone 3). It also includes new data-test-ids on elements that the tests interact with.

## How Has This Been Tested?
All the new tests passed on local environment.


packages/app-page-builder/src/editor/ui/views/GeneralSettingsView.ts
```diff
@@ -40,6 +40,7 @@ export class GeneralSettingsView extends PageSettingsFormView {
                 name: "path",
                 label: "Path",
                 description: `For example: "/about-us". Must contain at least two characters.`,
+                dataTestId: "data-test-path",
```
adrians5j: The name of this test iD sounds a bit weird. 

Maybe something more specific, like `general-settings-path-input` ?

packages/app-page-builder/src/editor/plugins/elementSettings/text/TextSettings.tsx
```diff
@@ -208,6 +209,7 @@ const TextSettings: React.FC<TextSettingsProps> = ({ defaultAccordionValue, opti
                     rightCellSpan={7}
                 >
                     <SelectField
+                        data-testid={"data-test-typography"}
```
adrians5j: Let's be more explicit here as well and maybe use something like `text-settings-typography`.

BTW no need to use `data-test-` prefix. Doesn't add any value, just makes the ID longer ultimately.

packages/app-page-builder/src/editor/plugins/elementSettings/text/TextSettings.tsx
```diff
@@ -180,6 +180,7 @@ const TextSettings: React.FC<TextSettingsProps> = ({ defaultAccordionValue, opti
             <>
                 <Wrapper containerClassName={classes.grid} label={"Color"}>
                     <BaseColorPicker
+                        data-testid={"text-color-picker"}
```
adrians5j: To be more in-sync with the IDs below, let's also prefix this test ID with `text-settings` instead of `text`.

packages/app-page-builder/src/editor/plugins/elementSettings/delete/index.tsx
```diff
@@ -14,6 +14,7 @@ export default {
                     tooltip={"Delete element"}
                     shortcut={["Backspace", "Delete"]}
                     icon={<DeleteIcon />}
+                    data-testid={"delete-element"}
```
adrians5j: Let's rename to `element-settings-delete-element-button`.

Just so we're a little bit more explicit.

packages/app-page-builder/src/editor/plugins/elementSettings/components/ColorPicker.tsx
```diff
@@ -5,6 +5,7 @@ import { activeElementAtom, elementByIdSelector } from "../../../recoil/modules"
 import { Typography } from "@webiny/ui/Typography";
 import { Grid, Cell } from "@webiny/ui/Grid";
 import { useRecoilValue } from "recoil";
+import { string } from "yargs";
```
adrians5j: How did this end up in here? 🙂

packages/app-page-builder/src/editor/plugins/elementSettings/clone/index.tsx
```diff
@@ -10,7 +10,11 @@ export default {
     renderAction() {
         return (
             <CloneAction>
-                <Action tooltip={"Clone element"} icon={<CloneIcon />} />
+                <Action
+                    tooltip={"Clone element"}
+                    icon={<CloneIcon />}
+                    data-testid={"clone-element"}
```
adrians5j, line 16:
```suggestion
                    data-testid={"element-settings-clone-element-button"}
```

packages/app-page-builder/src/pageEditor/config/editorBar/PageSettings/PageSettingsButton.tsx
```diff
@@ -12,7 +12,13 @@ const PageSettingsButton: React.FC = () => {
         setState(true);
     }, []);
 
-    return <IconButton onClick={onClickHandler} icon={<SettingsIcon />} />;
+    return (
+        <IconButton
+            onClick={onClickHandler}
+            icon={<SettingsIcon />}
+            data-testid={"page-settings-btn"}
```
adrians5j, line 19:
```suggestion
            data-testid={"page-settings-button"}
```

packages/app-page-builder/src/editor/plugins/toolbar/addElement/index.tsx
```diff
@@ -14,6 +14,7 @@ export default {
                 plugin={this.name}
                 icon={<AddIcon />}
                 closeOtherInGroup={true}
+                data-testid={"add-element"}
```
adrians5j, line 17:
```suggestion
                data-testid={"add-element-button"}
```

packages/app-page-builder/src/editor/plugins/elementSettings/text/TextSettings.tsx
```diff
@@ -192,9 +193,9 @@ const TextSettings: React.FC<TextSettingsProps> = ({ defaultAccordionValue, opti
                         leftCellSpan={5}
                         rightCellSpan={7}
                     >
-                        <SelectField value={text.tag} onChange={updateTag}>
+                        <SelectField value={text.tag} onChange={updateTag} data-testid={"data-test-heading-type"}>
```
adrians5j: To be more in-sync with the IDs below, let's also prefix this test ID with `text-settings` instead of `data-test`.

adrians5j: Also, let's rename `heading-type` to `heading-select`.

packages/app-page-builder/src/editor/components/Editor/EditorSidebar.tsx
```diff
@@ -63,10 +63,10 @@ export const EditorSidebar: React.FC = React.memo(() => {
     return (
         <Elevation z={1} className={rightSideBar}>
             <Tabs value={sidebar.activeTabIndex} updateValue={setActiveTabIndex}>
-                <EditorSidebarTab label={"Style"}>
+                <EditorSidebarTab label={"Style"} data-testid={"style-data-test"}>
```
adrians5j, line 66:
```suggestion
                <EditorSidebarTab label={"Style"} data-testid={"editor-sidebar-style-tab"}>
```

packages/app-page-builder/src/editor/components/Editor/EditorSidebar.tsx
```diff
@@ -63,10 +63,10 @@ export const EditorSidebar: React.FC = React.memo(() => {
...
                     <StyleSettingsTabContent />
                 </EditorSidebarTab>
-                <EditorSidebarTab label={"Element"} disabled={!element}>
+                <EditorSidebarTab label={"Element"} disabled={!element} data-testid={"element-data-test"}>
```
adrians5j:
```suggestion
                <EditorSidebarTab label={"Element"} disabled={!element} data-testid={"editor-sidebar-element-tab"}>
```

packages/app-page-builder/src/editor/components/Element.tsx
```diff
@@ -139,6 +139,7 @@ const ElementComponent: React.FC<ElementPropsType> = ({
                     active={isActive || false}
                     style={{ ...defaultStyle, ...transitionStyles[state] }}
                     className={"webiny-pb-page-element-container"}
+                    data-testid={"pb-page-element"}
```
adrians5j, line 142:
```suggestion
                    data-testid={"page-element"}
```

I'd either loose the `pb-` prefix or update other test IDs and and it there as well.

packages/app-page-builder/src/editor/components/DropZone/Horizontal.tsx
```diff
@@ -46,13 +47,21 @@ export interface HorizontalPropsType {
...
 }
 
-const HorizontalComponent: React.FC<HorizontalPropsType> = ({ below, onDrop, isVisible, type }) => {
+const HorizontalComponent: React.FC<HorizontalPropsType> = props => {
+    const { below, onDrop, isVisible, type } = props;
     return (
         <Droppable type={type} isVisible={isVisible} onDrop={onDrop}>
             {({ isOver, drop }) => (
                 <div
+                    className={props["data-testid"]}
```
adrians5j: Do we need to pass the test ID to `className`? I don't think so, but correct me if I'm wrong.

packages/app-page-builder/src/editor/components/DropZone/Horizontal.tsx
```diff
@@ -1,6 +1,7 @@
 import React from "react";
 import Droppable, { DroppableIsVisiblePropType, DroppableOnDropPropType } from "../Droppable";
 import styled from "@emotion/styled";
+import { string } from "yargs";
```
adrians5j: Let's remove this.

cypress/support/commands.js
```diff
@@ -34,6 +34,7 @@ import "./fileManager/fmDeleteFile";
 import "./fileManager/fmDeleteAllFiles";
 import "./fileManager/fmListTags";
 import "cypress-mailosaur";
+import "@4tw/cypress-drag-drop";
```
adrians5j: Just dbl-checking, do we need this?

cypress/integration/admin/pageBuilder/page/importExportPages.spec.js
```diff
@@ -1,129 +1,131 @@
-// This test fails on import. It either times out or is unable to input the ZIP file URL. Invetigation required.
+Cypress.on('uncaught:exception', (err, runnable) => {
```
adrians5j: The goal here was to improve the stability of this test (reduce the flakiness).

Adding `uncaught:exception` handler and just hiding/ignoring the error is definitely not the way to go.

We need to analyze the test. For example, run it multiple times, and check if there are points in the test that sometimes pass / sometimes fail. Then, we need to figure out why this is happening.

cypress/integration/admin/pageBuilder/category/categories.spec.js
```diff
@@ -1,3 +1,4 @@
+import { divide } from "lodash";
```
adrians5j: Let's move tests that are not related to categories into a separate dedicate file, for example, `pageBuilderEditor.spec.js`.

adrians5j (changes requested):
The first iteration done.

Please check the comments when you get a chance. Also, feel free to ping me for any further questions.

Finally, please check CI/CD issues and the results of `/cypress` tests. Cypress tests did not run because of some initial code issues (nothing spectacular, but it does block the execution).

cypress/integration/admin/pageBuilder/page/importExportPages.spec.js
razvanvancea: from what I've seen so far the test fails because of a random console error thrown by the application itself - I thought instead of having this test skipped, it would add value to our execution by running and bypassing that error (I am assuming that random console error does not have a major impact as long as the test gets executed and the scenario passes all the asserts). However I will run the test again and send you that error

razvanvancea: image

adrians5j:
Checked the Cypress run, a couple of errors I spotted:

image

BTW, a bit of a background...

We run Cypress test suites twice, on two different Webiny setups: one with just DynamoDB database, the other one DynamoDB + ElasticSearch. Ultimately, both test suites need to pass.

Also, usually, when we see the same tests failing on each setup, that probably means it's not flakiness, it's actually something not working as expected (be it on the app code level or Cypress test code level). If a test is failing just on one setup (e.g. on DynamoDB + ElasticSearch setup), that usually is a hint that a test is flaky.

In the above sshot, we can see `categories.spec.js` failing only on the left side. That could be flakiness.

On the other hand, `searchAndSortCategories.spec.js` and `importExportPages.spec.js` fails on both setups. As mentioned, a good indicator that something is actually wrong.

In any case, we'll need to inspect these before we merge. Let me know if you'll need any extra help here.

@razvanvancea

PR #2590, opened 2022-08-23
feat: add icon and description fields to Block Categories

neatbyte-vnobis:
## Changes
Add two fields to Block Category: icon and description

## How Has This Been Tested?
Manual testing 

## Documentation
None


packages/app-page-builder/src/admin/plugins/icons/index.tsx
```diff
@@ -0,0 +1,48 @@
+import React from "react";
```
Pavel910, line 2: @neatbyte-vnobis please add a TODO comment at the top of the file:

```
TODO: find a way to avoid copying and registering icons twice (Headless CMS has an identical plugin)
```

This is for the core team to handle in the future.

packages/app-page-builder/src/admin/views/BlockCategories/DelayedOnChange.ts
```diff
@@ -0,0 +1,107 @@
+import * as React from "react";
```
Pavel910: Let's move this file to the `ui` package, because I see it also exists in the editor, and in the `app-headless-cms` package, so this is already the 3rd copy of it. You don't need to create a storybook file for it, just put in its own folder, like the other components, and use that in the `app-page-builder`.

CleanShot 2022-08-24 at 10 56 49

packages/app-page-builder/src/admin/views/BlockCategories/IconPicker.tsx
```diff
@@ -0,0 +1,230 @@
+import * as React from "react";
```
Pavel910, line 2: Add a TODO here as well please:

```
TODO: find a better way to share IconPicker with icons across apps.
```

Pavel910 (changes requested):
Just a few minor changes and this is ready to merge. Thanks!

packages/app-page-builder/src/admin/views/BlockCategories/DelayedOnChange.ts
neatbyte-vnobis: @Pavel910 there are few places where such component is used:
image
But there are 2 variants: as a class component and as a functional component.
I am not 100% sure they are completely same (but they looks similar).

Which variant is preferable for extraction? Should I replace other usages too?

Pavel910: @neatbyte-vnobis please use the functional component variant, and only do it for the page builder package; we'll take care of other packages, just to avoid running into rabbit holes with this PR :) Thanks!

Pavel910 (approved):
Looks good! Thanks! 👍

PR #2591, opened 2022-08-26
Improve Error Logging (WCP / User Seats Management Functionality)

adrians5j:
## Changes
This PR improves error handling done in the "WCP increment/decrement user seats" code section.

## How Has This Been Tested?
Manually.

## Documentation
N/A (a small improvement)

### DIFF
--- a/packages/api-wcp/src/createWcp.ts
+++ b/packages/api-wcp/src/createWcp.ts
@@ -50,11 +50,10 @@ export const createWcp = async (): Promise<WcpContextObject> => {
             return;
         }
 
-        // If the `.json()` call throws an error, we don't want to disrupt the flow.
         let jsonParseError, json;
 
         try {
-            json = await response.json().catch();
+            json = await response.json();
         } catch (e) {
             jsonParseError = e;
         }
--- a/packages/api-wcp/src/createWcp.ts
+++ b/packages/api-wcp/src/createWcp.ts
@@ -58,7 +58,7 @@ export const createWcp = async (): Promise<WcpContextObject> => {
             jsonParseError = e;
         }
 
-        let message = `An error occurred while trying to ${operation} user seats.`;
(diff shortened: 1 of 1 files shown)


packages/api-wcp/src/createWcp.ts
```diff
@@ -50,25 +50,22 @@ export const createWcp = async (): Promise<WcpContextObject> => {
...
+        }
+
+        let message = `An error occurred while trying to ${operation} user seats.`;
+        if (jsonParseError) {
+            message += " Could not JSON-parse received HTTP response.";
         }
+
+        console.error(message, response.status, response.statusText, jsonParseError || json);
+
+        throw new Error(message);
```
Pavel910: I would throw a `WError` (you're already importing it at the top of the file), so you can set a proper error code. Error message on its own is not enough.

packages/api-wcp/src/createWcp.ts
```diff
@@ -50,25 +50,22 @@ export const createWcp = async (): Promise<WcpContextObject> => {
...
-                response.status,
-                response.statusText
-            );
-            throw new Error(
-                `An error occurred while trying to ${operation} user seats. Please check logs for more info.`
-            );
+            jsonParseError = e;
+        }
+
+        let message = `An error occurred while trying to ${operation} user seats.`;
```
Pavel910: Maybe let's try this form of message:
Instead of `An error occurred while trying to`, we can say `Failed to`, which essentially means the same thing, but is 5 times shorter. It's better from the user perspective, and us typing. You just get the message across faster.

I'd make it `Failed to ${operation} user seats.`. And if you agree, I'd generally make this a standard way we write these messages.

Pavel910 (changes requested):

Pavel910 (approved):

PR #2592, opened 2022-08-29
feat: folder CRUD module

leopuleo:
## Changes
With this PR we AIM to add a new folder CRUD module to handle folder-related operations. This is part of the wider context ACO -> Advanced Content Organisation.

## How Has This Been Tested?

Jest and manually

## Documentation

Not needed at the moment.


packages/api-folders-so-ddb/src/definitions/entities.ts
```diff
@@ -0,0 +1,90 @@
...
+        },
+        createdBy: {
+            type: "map"
+        },
+        tenant: {
+            type: "string"
+        },
+        locale: {
+            type: "string"
+        },
```
Pavel910: Let's store these into a `data` field, so it's easy to extend via plugins. We made this mistake of having individual fields several times, and it's super difficult to extend. Let's go with the approach which makes it easy to extend, so just put all folder related properties into a `data` map. An example: `packages/api-prerendering-service-so-ddb/src/definitions/render.ts:32`

packages/api-folders-so-ddb/src/definitions/entities.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const createEntryEntity = (table: Table, attributes: Attributes = {}) => {
+    return createEntity(ENTITIES.ENTRY, table, {
+        id: {
+            type: "string"
+        },
+        folderId: {
+            type: "string"
+        },
+        eId: {
```
Pavel910: Let's use full words, like `entryId`, to make it clear what this attribute is.

And also, same as with the `folder` entity, replace all attributes with a simple `data` map.

packages/api-folders-so-ddb/src/index.ts
```diff
@@ -0,0 +1,303 @@
...
+                    const response = await entities.folders.get(
+                        createFolderKeys({ id, tenant, locale })
+                    );
+                    if (response.Item) {
+                        result = response.Item;
+                    }
+                } else if (slug) {
+                    result = await queryOne({
+                        entity: entities.folders,
+                        partitionKey: `T#${tenant}#L#${locale}#CATEGORY#${category}#FOLDERS`,
```
Pavel910: Maybe extract this into a utility function? This string is constructed 3 times in this single file. And for Entries as well.

packages/api-folders/package.json
```diff
@@ -0,0 +1,45 @@
+{
```
Pavel910, line 1: Set `private: true`

packages/api-folders-so-ddb/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@webiny/api-folders-so-ddb",
```
Pavel910, line 2: Set `private: true`

packages/api-folders/src/context/entries.ts
```diff
@@ -0,0 +1,166 @@
...
+import { validation } from "@webiny/validation";
+
+import { FoldersConfig, Entry, ListEntriesParams, EntryInput, IEntries } from "~/types";
+
+const CreateDataModel = withFields({
+    id: string({ validation: validation.create("required") }),
+    folderId: string({ validation: validation.create("required") }),
+    tenant: string({ validation: validation.create("required") }),
+    locale: string({ validation: validation.create("required") })
+})();
```
Pavel910: We want to retire `commodo`, so let's use this opportunity and switch to Joi. You can find an example here: https://github.com/webiny/webiny-js/blob/next/packages/api-mailer/src/crud/mailer/onBeforeSend.ts#L9

packages/api-folders/src/context/folders.ts
```diff
@@ -0,0 +1,174 @@
...
+    slug: string({ validation: validation.create("required,minLength:3") }),
+    category: string({ validation: validation.create("required,in:page:cms:file") }),
+    tenant: string({ validation: validation.create("required") }),
+    locale: string({ validation: validation.create("required") })
+})();
+
+const UpdateDataModel = withFields({
+    name: string({ validation: validation.create("minLength:3") }),
+    slug: string({ validation: validation.create("minLength:3") })
+})();
```
Pavel910: Same here, let's switch to Joi.

packages/api-folders/src/context/index.ts
```diff
@@ -0,0 +1,14 @@
+import createEntriesContext from "./entries";
+import createFoldersContext from "./folders";
+
+import { Folders, FoldersConfig } from "~/types";
+
+export default async (config: FoldersConfig): Promise<Folders> => {
```
Pavel910: Let's use a named export here, `createContext`, because we're exporting a function, and IDE can find usages reliably, and it also helps with refactors. In general, we're trying to avoid default exports/imports on functions/components, and when you see it in the existing code, it was a bad decision made in the past years.

packages/api-folders/src/graphql/entries.gql.ts
```diff
@@ -0,0 +1,96 @@
+import { Response, ErrorResponse } from "@webiny/handler-graphql/responses";
+import { GraphQLSchemaPlugin } from "@webiny/handler-graphql/plugins/GraphQLSchemaPlugin";
+import { FoldersContext } from "~/types";
+
+export default new GraphQLSchemaPlugin<FoldersContext>({
+    typeDefs: /* GraphQL */ `
+        type Entry {
+            id: ID!
+            eId: ID!
```
Pavel910: Mke this `entryId`

packages/api-folders/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+            getLocaleCode,
+            getIdentity
+        });
+    });
+};
+
+export const createFoldersGraphQL = () => {
+    return new ContextPlugin<FoldersContext>(context => {
+        context.plugins.register(graphqlPlugins);
+    });
```
Pavel910: You don't need to have a ContextPlugin to register GraphQL plugins. Simply return `graphqlPlugins`, like this:

```
export const createFoldersGraphQL = () => {
    return graphqlPlugins;
}
```

packages/api-folders/src/types.ts
```diff
@@ -0,0 +1,173 @@
...
+    sort: string[];
+}
+
+export type StorageOperationsCreateFolderParams = CreateFolderParams;
+export type StorageOperationsUpdateFolderParams = UpdateFolderParams;
+export type StorageOperationsDeleteFolderParams = DeleteFolderParams;
+
+export interface Entry {
+    id: string;
+    eId: string;
```
Pavel910: Replace all `eId` with `entryId`.

Pavel910 (changes requested):

packages/api-folders/src/context/folders.ts
```diff
@@ -0,0 +1,167 @@
...
+});
+
+export const createFoldersContext = async ({
+    getTenantId,
+    getLocaleCode,
+    getIdentity,
+    storageOperations
+}: FoldersConfig): Promise<IFolders> => {
+    const tenant = getTenantId();
+    const locale = getLocaleCode();
```
Pavel910: Getting these values here will cause issues if you switch tenants during a single request lifecycle, because they will never get updated, and data will be stored to the wrong tenant. These functions should be called within each CRUD methods individually.

packages/api-folders/src/context/links.ts
```diff
@@ -0,0 +1,155 @@
...
+});
+
+export const createLinksContext = async ({
+    getTenantId,
+    getLocaleCode,
+    getIdentity,
+    storageOperations
+}: FoldersConfig): Promise<ILinks> => {
+    const tenant = getTenantId();
+    const locale = getLocaleCode();
```
Pavel910: Same as in the previous comment; move these into every CRUD method.

packages/api-folders/src/types.ts
```diff
@@ -0,0 +1,178 @@
...
+    id: string;
+    type: string;
+    displayName: string | null;
+}
+
+export enum Type {
+    PAGE = "page",
+    CMS = "cms",
+    FILE = "file"
+}
```
Pavel910: The issue with this is that you're hardcoding these 3 values, which implicitly makes this package aware of _where_ it's going to be used. Let's not use enums for this; if we do, we won't be able to use this `api-folders` with any other app, unless we add a new enum value.

packages/api-folders/src/types.ts
```diff
@@ -0,0 +1,178 @@
...
+
+export interface Folder {
+    id: string;
+    name: string;
+    slug: string;
+    type: Type;
```
Pavel910: Make this be a simple `string`.

packages/api-folders/src/types.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+export type FolderInput = Pick<Folder, "name" | "slug" | "type" | "parentId">;
+
+export interface GetFolderParams {
+    id: string;
+}
+
+export interface ListFoldersWhere {
+    type: Type;
```
Pavel910: `string`

packages/api-folders/src/types.ts
```diff
@@ -0,0 +1,178 @@
...
+}
+
+export interface DeleteFolderParams {
+    folder: Folder;
+}
+
+export interface StorageOperationsGetFolderParams {
+    id?: string;
+    slug?: string;
+    type?: Type;
```
Pavel910: `string`

packages/api-folders/src/graphql/base.gql.ts
```diff
@@ -0,0 +1,54 @@
...
+        type FolderBooleanResponse {
+            data: Boolean
+            error: FolderError
+        }
+
+        enum FolderType {
+            page
+            cms
+            file
+        }
```
Pavel910: Remove this type and use `String`; I explained the reason in the types file below.

packages/api-folders/src/types.ts
```diff
@@ -0,0 +1,178 @@
...
+    link: Link;
+}
+
+export interface DeleteLinkParams {
+    link: Link;
+}
+
+export interface StorageOperationsGetLinkParams {
+    id?: string;
+    eId?: string;
```
Pavel910: `entryId`

Pavel910 (changes requested):

packages/api-folders-so-ddb/src/operations/folders.ts
```diff
@@ -0,0 +1,174 @@
...
+
+export const createFoldersStorageOperations = (
+    entity: Entity<any>,
+    table: Table
+): FoldersStorageOperations => {
+    const createFolderGsiPartitionKey = ({
+        tenant,
+        locale,
+        type
+    }: Pick<Folder, "tenant" | "locale" | "type">) => `T#${tenant}#L#${locale}#FOLDERS#${type}`;
```
Pavel910: Convert this into a proper `return` as well :)

packages/api-folders-so-ddb/src/operations/folders.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    const createFolderGsiSearchKey = ({ slug, parentId }: Pick<Folder, "slug" | "parentId">) => {
+        return ((!!parentId && `${parentId}#`) || "") + slug;
```
Pavel910: This is quite unreadable; I suggest we rewrite it like this:

```
return [parentId, slug].filter(Boolean).join("#");
```

Pavel910 (changes requested):
@leopuleo 2 tiny changes and we're done with this! 👍

Pavel910 (approved):

PR #2595, opened 2022-08-30
feat: use Blocks as references

neatbyte-vnobis:
## Changes
Change Page Block usage to be used as a "reference". Add unlink functionality.

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-page-builder/src/types.ts
+++ b/packages/api-page-builder/src/types.ts
@@ -11,7 +11,7 @@ export interface PageElement {
     name: string;
     type: "element" | "block";
     category: string;
-    content: File;
+    content: any;
     preview: File;
     createdOn: string;
     createdBy: CreatedBy;
@@ -818,7 +818,7 @@ export interface PageBlock {
     id: string;
     name: string;
     blockCategory: string;
-    content: File;
+    content: any;
     preview: File;
     createdOn: string;
     createdBy: CreatedBy;
--- a/packages/app-page-builder/src/editor/components/Editor/Sidebar/StyleSettingsTabContent.tsx
+++ b/packages/app-page-builder/src/editor/components/Editor/Sidebar/StyleSettingsTabContent.tsx
@@ -5,6 +5,9 @@ import { ReactComponent as TouchIcon } from "./icons/touch_app.svg";
 import { ReactComponent as WarningIcon } from "./icons/warning-black.svg";
 import { useActiveElement } from "~/editor/hooks/useActiveElement";
 import useElementStyleSettings from "~/editor/plugins/elementSettings/hooks/useElementStyleSettings";
+import { ButtonPrimary } from "@webiny/ui/Button";
+import UnlinkBlockAction from "~/editor/plugins/elementSettings/unlinkBlock/UnlinkBlockAction";
+import { ReactComponent as InfoIcon } from "@webiny/app-admin/assets/icons/info.svg";
 
 const RootElement = styled("div")({
     height: "calc(100vh - 65px - 48px)", // Subtract top-bar and tab-header height
@@ -22,6 +25,28 @@ const RootElement = styled("div")({
     }
 });
 
+const UnlinkBlockWrapper = styled("div")({
+    padding: "16px",
+    display: "grid",
+    rowGap: "16px",
+    justifyContent: "center",
+    alignItems: "center",
+    margin: "16px",
+    textAlign: "center",
+    backgroundColor: "var(--mdc-theme-background)",
+    border: "3px dashed var(--webiny-theme-color-border)",
+    borderRadius: "5px",
+    "& .info-wrapper": {
+        display: "flex",
+        alignItems: "center",
+        fontSize: "10px",
+        "& svg": {
+            width: "18px",
+            marginRight: "5px"
+        }
+    }
+});
+
 const StyleSettingsTabContent: React.FC = () => {
     const [element] = useActiveElement();
     const elementStyleSettings = useElementStyleSettings();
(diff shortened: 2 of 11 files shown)


packages/api-page-builder/src/types.ts
```diff
@@ -11,7 +11,7 @@ export interface PageElement {
     name: string;
     type: "element" | "block";
     category: string;
-    content: File;
+    content: any;
```
Pavel910, line 14: Good catch! 😅

packages/app-page-builder/src/editor/components/Editor/Sidebar/StyleSettingsTabContent.tsx
```diff
@@ -35,6 +60,23 @@ const StyleSettingsTabContent: React.FC = () => {
...
+                        <ButtonPrimary>Unlink block</ButtonPrimary>
+                    </UnlinkBlockAction>
+                </div>
+                <div className="info-wrapper">
+                    <InfoIcon /> Click here to learn more about how block work
+                </div>
+            </UnlinkBlockWrapper>
+        );
+    }
+
```
Pavel910, line 79: This part is hardcoded into the core editor, which is not good because it only really applies to the Page Editor. So let's move this into the `pageEditor` plugins:
```packages/app-page-builder/src/pageEditor/config/PageEditorConfig.tsx```
and use this as a reference for implementation 
```packages/app-page-builder/src/blockEditor/config/BlockElementSidebar.tsx:6```

File: packages/app-page-builder/src/editor/contexts/EventActionHandlerProvider.tsx [hunk 3]
```diff
@@ -127,13 +127,12 @@ export const EventActionHandlerProvider = makeComposable<
...
-        pluginsAtomValueRef.current = pluginsAtomValue;
-        uiAtomValueRef.current = uiAtomValue;
-        snapshotRef.current = snapshot;
-    }, [sidebarAtomValue, rootElementAtomValue, pluginsAtomValue, uiAtomValue]);
+    // reinit ref values
+    sidebarAtomValueRef.current = sidebarAtomValue;
+    rootElementAtomValueRef.current = rootElementAtomValue;
+    pluginsAtomValueRef.current = pluginsAtomValue;
+    uiAtomValueRef.current = uiAtomValue;
+    snapshotRef.current = snapshot;
```
Pavel910, line 135: Just curios, which part of your code was not working because of the `useEffect`?

packages/app-page-builder/src/editor/helpers.ts
```diff
@@ -165,6 +165,26 @@ export const createBlockElements = (name: string): PbEditorElement => {
...
+        // @ts-ignore
+        id: getNanoid(),
+        // @ts-ignore
+        elements: [],
+        ...blockElement,
+        // @ts-ignore
+        data: { ...blockElement.data, blockId: plugin.id }
+    };
+};
+
```
Pavel910, line 187: Similar to the tabs, please move this code into the `pageEditor` section of the package. Somewhere near `SearchBlocks` component.

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveAction.tsx
```diff
@@ -94,7 +94,13 @@ const SaveAction: React.FC = ({ children }) => {
...
 
     const onSubmit = async (formData: PbDocumentElement) => {
-        formData.content = pluginOnSave(removeIds((await getElementTree(element)) as PbElement));
+        const pbElement = (await getElementTree(element)) as PbElement;
+        if (formData.type === "block") {
+            // We need ids for block editor
+            formData.content = pluginOnSave(pbElement);
+        } else {
+            formData.content = pluginOnSave(removeIds(pbElement));
+        }
```
Pavel910, line 103: Please add a TODO comment here, I need to find a better way to handle this, because again, this is hardcoding the logic into the core editor, which I want to avoid.

```
// TODO: find a way to decouple this.
// One option is to have a `save` plugin in `blockEditor` AND in the `pageEditor` where each can be implemented differently, but then we need to extract the image creation logic somewhere.
```

packages/app-page-builder/src/editor/plugins/elementSettings/unlinkBlock/UnlinkBlockAction.ts
```diff
@@ -0,0 +1,38 @@
...
+                },
+                history: true
+            })
+        );
+    }, [activeElementId]);
+
+    return React.cloneElement(children, { onClick });
+};
+
+export default React.memo(UnlinkBlockAction);
```
Pavel910, line 38: Colocate this with the StyleSettingsTab plugin which we discussed earlier.

packages/app-page-builder/src/editor/plugins/elements/block/Block.tsx
```diff
@@ -16,22 +18,40 @@ interface BlockType {
...
+                                elementId={id}
+                                elementStyle={elementStyle}
+                                elementAttributes={elementAttributes}
+                                customClasses={[
+                                    "webiny-pb-layout-block webiny-pb-base-page-element-style",
+                                    ...customClasses
+                                ]}
+                                combineClassNames={combineClassNames}
+                            />
+                        )
```
Pavel910, line 54: In [this branch](https://github.com/webiny/webiny-js/blob/page-builder-blocks-v2-disable-block-element/packages/app-page-builder/src/render/components/ElementRoot.tsx#L115), I made `ElementRoot` composable, and using that you can nicely solve this via a plugin specific to the `pageEditor`.

See an example here: https://github.com/webiny/webiny-js/blob/page-builder-blocks-v2-disable-block-element/packages/app-page-builder/src/render/components/ElementRoot.tsx#L115

It was related to `Disable editing of blocks inside Page Builder` task, and I wrote a full description with links in Notion, but now I'm not sure you have access to that 🤔 In any case, please merge that branch into your working branch and you'll get these mechanisms of composing the `ElementRoot`.

Pavel910 (changes requested):
@neatbyte-vnobis Left a few comments and changes to be done. Please let me know if you need help with that composable `ElementRoot` component, and make sure you merge the branch I references in the comments. It solves the first task of the milestone + exposes a mechanisms for ElementRoot composition.

PR #2602, opened 2022-09-05
Clear Apollo Cache when content model is deleted

Pavel910:
## Changes
This PR adds logic to remove a content model from Apollo Cache when a model is deleted. Since we use `modelId` as a cache id, if a model is _not_ removed from cache, and is then recreated with the same modelId, there are problems with the new model because Apollo thinks that the data in the cache (which remains after model is deleted) belongs to this new model.

In the process I also switched all usages of `cache-get-object-id` plugin to use the proper class plugin, and added a re-export of useful plugins from `@webiny/app` package root.

## How Has This Been Tested?
Manually.


brunozoric (approved):
Everything looks fine.

PR #2609, opened 2022-09-06
Ensure WCP-related Environment Variables Are Correctly Assigned With the Watch Command

adrians5j:
## Changes

This PR fixes the following bug.

If you deploy your WCP-enabled project, all of the WCP-related env variables are correctly assigned. 

On the other hand, the moment you start development with the watch command, all of the WCP-related env variables disappear. This is because, when running the watch command, the code that assigns these variables does not get executed. It only gets executed with the deploy command.

This PR ensures all of the WCP-related env vars are correctly assigned both with the deploy and watch commands.

## How Has This Been Tested?
Manual.

## Documentation
Changelog.


packages/cli/commands/wcp/hooks.js
```diff
@@ -21,100 +21,113 @@ const { getUser, getProjectEnvironment, updateUserLastActiveOn } = require("./ut
...
-                id: projectEnvironment.id,
-                apiKey: projectEnvironment.apiKey,
-                org: { id: projectEnvironment.org.id },
-                project: { id: projectEnvironment.project.id }
-            };
-
-            process.env.WCP_PROJECT_ENVIRONMENT = encrypt(wcpProjectEnvironment);
-            process.env.WCP_PROJECT_ENVIRONMENT_API_KEY = projectEnvironment.apiKey;
-        }
+        hook: getEnvironmentHookHandler
```
Pavel910, line 119: It's not a `getter` if you're not executing it :) Just rename it to `environmentHookHandler`, then it can work without being called.

Pavel910: Ah just realized, `get` is not a "factory" for a hook, it's the name of the hook logic. The hook is _getting_ the environment. My bad.

PR #2613, opened 2022-09-08
feat(app-headless-cms): make fieldId changeable

brunozoric:
## Changes
UI changes to support the field alias. Basically we will make fieldId changeable and it will serve as the field alias (graphql identifier).

The `storageId` property cannot be customly created or updated - it is an autogenerated value.

## How Has This Been Tested?
Manually and cypress.


Pavel910 (approved):
I particularly liked the section where you moved layout validation into a hook, instead of hardcoding it in the CRUD 👍

PR #2618, opened 2022-09-12
feat: blocks as refs

neatbyte-vnobis:
## Changes
Change block handling to use refs, add unlink button.

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/pageEditor/plugins/elementSettings/UnlinkBlockAction.ts
+++ b/packages/app-page-builder/src/pageEditor/plugins/elementSettings/UnlinkBlockAction.ts
@@ -1,0 +1,36 @@
+import React, { useCallback } from "react";
+import { useEventActionHandler } from "~/editor/hooks/useEventActionHandler";
+import { UpdateElementActionEvent } from "~/editor/recoil/actions";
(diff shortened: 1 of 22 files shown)


packages/app-page-builder/src/pageEditor/plugins/elementSettings/UnlinkBlockAction.ts
```diff
@@ -0,0 +1,40 @@
...
+                new UpdateElementActionEvent({
+                    element: {
+                        ...element,
+                        data: newData
+                    },
+                    history: true
+                })
+            );
+        }
+    }, [activeElementId]);
```
Pavel910: This whole section can be simplified; I added lots of hooks for this in https://github.com/webiny/webiny-js/tree/page-builder-blocks-v2-disable-block-element-updated/packages/app-page-builder/src/editor/hooks.

What you need are 2 hooks: `useActiveElement` and `useUpdateElement`, and this whole mess will go away.

Pavel910 (changes requested):
@neatbyte-vnobis That `inert` is a nice trick 👍 Definitely better than absolutely positioned divs 😅 

Please look at the one comment I left, and we're good with this PR. Thanks!

Pavel910 (approved):

PR #2620, opened 2022-09-13
feat(api-headless-cms-ddb): implement conversion of field paths

brunozoric:
## Changes
Implement field path transformations into api-headless-cms-ddb package.

## How Has This Been Tested?
Jest.

### DIFF
--- a/packages/api-headless-cms-ddb/src/operations/entry/index.ts
+++ b/packages/api-headless-cms-ddb/src/operations/entry/index.ts
@@ -22,8 +22,9 @@ import {
     CmsEntryStorageOperationsRequestReviewParams,
     CmsEntryStorageOperationsUnpublishParams,
     CmsEntryStorageOperationsUpdateParams,
-    CmsModel,
-    CONTENT_ENTRY_STATUS
+    StorageOperationsCmsModel,
+    CONTENT_ENTRY_STATUS,
+    CmsStorageEntry
 } from "@webiny/api-headless-cms/types";
 import { Entity } from "dynamodb-toolbox";
 import { filterItems, buildModelFields, sortEntryItems, FilterItemFromStorage } from "./utils";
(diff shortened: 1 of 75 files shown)


packages/api-headless-cms-ddb/src/operations/entry/index.ts
```diff
@@ -96,8 +129,11 @@ export const createEntriesStorageOperations = (
         };
     };
 
-    const create = async (model: CmsModel, args: CmsEntryStorageOperationsCreateParams) => {
-        const { entry, storageEntry } = args;
+    const create = async (
+        model: StorageOperationsCmsModel,
+        args: CmsEntryStorageOperationsCreateParams
```
Pavel910: This here still has `args`, but all other methods have `params`. Make this `params` as well.

Pavel910 (changes requested):
@brunozoric just one nitpick here 😂 

Overall, I really like how the `packages/api-headless-cms-ddb/src/operations/entry/index.ts` file is evolving, it looks better and better, neatly organized and readable 👏 

Can't wait for the day when we won't have to pass `storageEntry` but just entry instead.

packages/api-headless-cms/src/fieldConverters/CmsModelDefaultFieldConverterPlugin.ts
```diff
@@ -11,6 +11,13 @@ export class CmsModelDefaultFieldConverterPlugin extends CmsModelFieldConverterP
     }
 
     public override convertToStorage({ field, value }: ConvertParams): any {
+        /**
+         * Do not convert if no value was passed.
+         * Note that we only string undefined.
```
Pavel910: `Note that we only string undefined.` ?

packages/api-headless-cms/src/fieldConverters/CmsModelDefaultFieldConverterPlugin.ts
```diff
@@ -20,6 +27,13 @@ export class CmsModelDefaultFieldConverterPlugin extends CmsModelFieldConverterP
     }
 
     public override convertFromStorage({ field, value }: ConvertParams): any {
+        /**
+         * Do not convert if no value was passed.
+         * Note that we only string undefined.
```
Pavel910: `Note that we only string undefined.` ?

Pavel910 (approved):
Update those 2 comment lines and feel free to merge.

PR #2621, opened 2022-09-13
Headless Pages

Pavel910:
## Changes
This PR introduces tagging of content models and model fields, and adds a "page template" list view. This serves as a starting point to develop a dedicated "Page Template Editor" and Headless Pages functionality. Since page templates are still content models at its core, tags are the simplest way to differentiate between various types of content models. 

From the API perspective, we're still using the same `listContentModels` query, but we're filtering models on the client side, to render type specific UI. However, from the GraphQL schema standpoint, all models are processed the same way, and their respective schemas are generated using the same underlying logic.

I'll use this PR as an umbrella PR for all further tasks.

1. https://github.com/webiny/webiny-js/pull/2623
2. https://github.com/webiny/webiny-js/pull/2627


packages/api-headless-cms/src/graphql/schema/contentModels.ts
```diff
@@ -31,7 +31,19 @@ export const createModelsSchema = (context: CmsContext): GraphQLSchemaPlugin<Cms
...
+        CmsContentModelField: {
+            tags(field) {
+                // Make sure `tags` are always returned as an array.
+                return Array.isArray(field.tags) ? field.tags : [];
+            }
+        },
         CmsContentModel: {
+            tags(model: CmsModel) {
+                // Make sure `tags` always contain a `type` tag, to differentiate between models.
+                const hasType = (model.tags || []).find(tag => tag.startsWith("type:"));
```
brunozoric, line 43: To be correct, use the .some() on array

packages/api-headless-cms/src/graphql/schema/contentModels.ts
```diff
@@ -123,6 +136,7 @@ export const createModelsSchema = (context: CmsContext): GraphQLSchemaPlugin<Cms
                 layout: [[ID!]!]
                 fields: [CmsContentModelFieldInput!]
                 titleFieldId: String
+                tags: [String]
```
brunozoric: Add ! into the array. tag should not be null

brunozoric: And now I see the create model mutation and it looks like tags must be an array and items in it must be string. So just add ! and make sure that null does not go out of the resolvers in any case.

packages/api-headless-cms/src/types.ts
```diff
@@ -169,6 +169,10 @@ export interface CmsModelField {
      *
      */
     multipleValues?: boolean;
+    /**
+     * Fields can be tagged to give them contextual meaning.
+     */
+    tags?: string[];
```
brunozoric, line 175: If we use the mutation as the source, this should never be optional.

packages/api-headless-cms/src/types.ts
```diff
@@ -396,6 +400,10 @@ export interface CmsModel {
      * ```
      */
     layout: string[][];
+    /**
+     * Models can be tagged to give them contextual meaning.
+     */
+    tags?: string[];
```
brunozoric, line 406: Not optional.

packages/app-headless-cms/src/modules/pageTemplates/views/editor/EditorPlugins/SidebarPlugin.tsx
```diff
@@ -0,0 +1,41 @@
...
+    overflow: scroll;
+    height: calc(100vh - 70px);
+`;
+
+export const SidebarPlugin = createComponentPlugin(Editor.Sidebar, OriginalSidebar => {
+    return function Sidebar({ children, ...props }) {
+        const { tabsRef } = useContentModelEditor();
+
+        const onFieldDragStart = useCallback(() => {
+            if (tabsRef.current!.getActiveIndex() > 1) {
```
brunozoric: I would rather do a check if tabsRef.current exists than coerce ... If something goes wrong in the parent component and nothing is set into the `.current`, this will break.

packages/app-headless-cms/src/modules/pageTemplates/views/editor/EditorPlugins/index.tsx
```diff
@@ -0,0 +1,12 @@
+import React from "react";
+import { SidebarPlugin } from "./SidebarPlugin";
+import { ContentTabsPlugin } from "./ContentTabsPlugin";
+
+export const EditorPlugins = () => {
```
brunozoric: Add the `React.FC` type

packages/app-headless-cms/src/modules/pageTemplates/views/editor/ModelFieldPlugins.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useState } from "react";
+import { plugins } from "@webiny/plugins";
+import { CircularProgress } from "@webiny/ui/Progress";
+import { useQuery } from "~/admin/hooks";
+import { LIST_CONTENT_MODELS, ListCmsModelsQueryResponse } from "~/admin/viewsGraphql";
+import { CmsEditorFieldTypePlugin, CmsModel } from "~/types";
+import { ReactComponent as ObjectIcon } from "~/admin/icons/ballot_black_24dp.svg";
+
+const isContentModel = (model: CmsModel) => {
+    return Boolean(model.tags.find(tag => tag === "type:contentModel"));
```
brunozoric: Use `.some()` method and no need to do the `Boolean`.

packages/app-headless-cms/src/modules/pageTemplates/views/editor/ModelFieldPlugins.tsx
```diff
@@ -0,0 +1,51 @@
...
+                            multipleValuesLabel: "",
+                            createField: () => ({
+                                type: `model-${model.modelId}`,
+                                validation: [],
+                                renderer: {
+                                    name: "model-renderer"
+                                }
+                            })
+                        },
+                        tags: ["content-model", ...(model.tags || [])]
```
brunozoric, line 41: Tag here is `content-model` and I saw in some other places `type:contentModel`. I assume the difference is due to one being model and other a field...? Maybe use the camel case in both cases?

packages/app-headless-cms/src/modules/pageTemplates/views/editor/PageTemplateEditor.tsx
```diff
@@ -0,0 +1,27 @@
...
+    }
+    return (
+        <ModelFieldPlugins>
+            <EditorPlugins />
+            <Editor modelId={modelId} apolloClient={apolloClient} />
+        </ModelFieldPlugins>
+    );
+};
+
+export default PageTemplateEditor;
```
brunozoric, line 27: Didn't we say we will avoid `export default`?

packages/app-headless-cms/src/modules/pageTemplates/views/list/CloneContentModelDialog.tsx
```diff
@@ -0,0 +1,283 @@
...
+        width: 600px;
+        min-width: 600px;
+    }
+`;
+
+const noPadding = css`
+    padding: 5px !important;
+`;
+
+export interface Props {
```
brunozoric, line 42: If this file is going to be re-exported at some point, and there are other files that export `Props` interface, it will cause problems. Maybe use `CloneContentModelDialogProps`.

packages/app-headless-cms/src/modules/pageTemplates/views/list/CloneContentModelDialog.tsx
```diff
@@ -0,0 +1,283 @@
...
+                            </UID.DialogActions>
+                        </>
+                    )}
+                </Form>
+            )}
+        </UID.Dialog>
+    );
+};
+
+export default CloneContentModelDialog;
```
brunozoric, line 283: Avoid `export default`.

brunozoric (changes requested):
I left some comments which we can discuss. I didn't comment on all the `export default`, but try to avoid those.

packages/app-headless-cms/src/modules/pageTemplates/views/list/CloneContentModelDialog.tsx
Pavel910: @brunozoric It's here because `React.lazy` expects a default export, and we code-split these views.

packages/app-headless-cms/src/modules/pageTemplates/views/editor/PageTemplateEditor.tsx
Pavel910: This is loaded using `React.lazy`, so it expects a default export, I'll see if I can do some voodoo on the import statement to allow a named export here.

packages/app-headless-cms/src/modules/pageTemplates/views/editor/ModelFieldPlugins.tsx
Pavel910: Yeah good catch. That `content-model` is a left-over. Tags will be pulled in from `model.tags` exclusively, so I'll remove this.

packages/api-headless-cms/src/graphql/schema/contentModels.ts
Pavel910: I'm setting this as `tags: [String!]`, because you don't have to send tags at all. If you don't send them, the API will still assign default tags for you, so from the GraphQL perspective, it's not required.

PR #2623, opened 2022-09-14
feat(ui): add protectedTags prop

Pavel910:
## Changes
This PR converts the Tags component into a functional component (from the old class component), and adds support for `protectedTags`. Protected tags will not be removable by the user. This allows us to have predefined system tags that a user can't remove.

## How Has This Been Tested?
Manually.


Pavel910:
This PR was abandoned. Protected tags are implemented as part of #2862

PR #2627, opened 2022-09-14
Extract base content model editor

Pavel910:
## Changes
This PR extracts a base content model editor "framework" (like we did with the Page Builder Editor), and makes the base editor composable. This allows us to easily create editor variations with different functionality, sharing the same core (field plugins, d&d mechanics, field layouts).

The existing content model editor was also refactored to use this new base editor, as a test to prove that the "framework" works. The entire concept revolves around composable components.

## How Has This Been Tested?
Manually.


Pavel910:
Closing this PR, we won't go forward with this.

PR #2633, opened 2022-09-18
feat: make getElementTree composable

neatbyte-vnobis:
## Changes
Make getElementTree  composable. Filter out ref blocks from Navigator tree.

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/types.ts
+++ b/packages/app-page-builder/src/types.ts
@@ -688,11 +688,16 @@ export type PbRenderElementPluginArgs = {
     elementType?: string;
 };
 
+export type GetElementTreeProps = {
+    element?: PbEditorElement;
+    path?: string[];
+} | void;
+
 // ============== EVENT ACTION HANDLER ================= //
 // TODO: at some point, convert this into an interface, and use module augmentation to add new properties.
 export type EventActionHandlerCallableState<TState = PbState> = PbState<TState> & {
     getElementById(id: string): Promise<PbEditorElement>;
-    getElementTree(props?: { element?: PbEditorElement }): Promise<PbEditorElement>;
+    getElementTree(props: GetElementTreeProps): Promise<PbEditorElement>;
 };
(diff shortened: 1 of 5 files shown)


packages/app-page-builder/src/types.ts
```diff
@@ -692,7 +692,7 @@ export type PbRenderElementPluginArgs = {
 // TODO: at some point, convert this into an interface, and use module augmentation to add new properties.
 export type EventActionHandlerCallableState<TState = PbState> = PbState<TState> & {
     getElementById(id: string): Promise<PbEditorElement>;
-    getElementTree(element?: PbEditorElement): Promise<any>;
+    getElementTree(props?: { element?: PbEditorElement }): Promise<any>;
```
Pavel910: @neatbyte-vnobis is there a reason to convert this to a `props` object? I must be missing something, but it doesn't look like you're using anything other than the `element`, so not sure you need this 🤔 

And secondary, why `Promise<any>`? If you have a different form of `element` coming back from `getElementTree`, maybe it's better to add support for generic type on that function.

Pavel910 (commented):
@neatbyte-vnobis I left a comment, want to clarify something before approval or requesting changes. Thanks!

packages/app-page-builder/src/types.ts
neatbyte-vnobis: @Pavel910 I have updated the types

Not only `element` is used (in recursion also `path` is passed).
Due to composeAsync inner working only one parameter can be passed, so `getElementTree(element, path)` was replaced with `getElementTree({ element, path })`.

I have created a new type inside types (as there were multiple definitions with different interfaces in types and EventActionHandlerProvider.tsx)

And yes, no need for `any` there

Pavel910 (approved):
@neatbyte-vnobis thanks for clarification, makes sense.

Issue #2629, opened 2022-09-16
Don't use utf8 encoding when reading files for upload after deploy


Pavel910:
### Version

5.31.0

### What are the steps to reproduce this bug?

Create a 5.31. project, or upgrade an older project to that version, and run a deploy.

### What is the expected behavior?

Non-text files are uploaded to S3 bucket correctly, and are accessible.

### What do you see instead?

Static non-text assets (images, videos), are broken/inaccessible. Also, if you check the S3 bucket, these files have a wrong file size.

### Possible solution

The solution is to remove the `utf8` encoding parameter from the call to `fs.readFileSync`:
https://github.com/webiny/webiny-js/blob/next/packages/pulumi-aws/src/utils/uploadFolderToS3.ts#L142

By simply removing that argument, `readFileSync` will read a raw buffer, which will result in a correct file content uploaded to the S3 bucket.

chiizzy1:
Hi Pavel, I have made the necessary edit to the code and I'll go ahead to open a pull request.

adrians5j:
Closed by @chiizzy1, thanks again 🍻 !

PR #2635, opened 2022-09-20
fix: remove the utf8 encoding parameter from the call to fs.readFile

chiizzy1:
## Related Issue
#2629

## Changes
I remove the utf8 encoding parameter from the call to fs.readFileSync which caused the static non-text assets (images, videos) to be broken/inaccessible.

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/pulumi-aws/src/utils/uploadFolderToS3.ts
+++ b/packages/pulumi-aws/src/utils/uploadFolderToS3.ts
@@ -139,7 +139,7 @@ export const uploadFolderToS3 = async ({
                                 ...fields,
                                 "Content-Type": contentType || "",
                                 "X-Amz-Meta-Checksum": checksum,
-                                file: fs.readFileSync(path, "utf8")
+                                file: fs.readFileSync(path)
                             };
 
                             if (cacheControl) {
--- a/packages/pulumi-aws/src/utils/uploadFolderToS3.ts
+++ b/packages/pulumi-aws/src/utils/uploadFolderToS3.ts
@@ -135,7 +135,7 @@ export const uploadFolderToS3 = async ({
                                 cacheControl: cacheControl ? cacheControl.value : undefined
                             });
 
-                            const data: Record<string, string> = {
+                            const data: Record<string, string | Buffer> = {
                                 ...fields,
                                 "Content-Type": contentType || "",
                                 "X-Amz-Meta-Checksum": checksum,


Pavel910 (approved):
~Looks good, thank you!~

EDIT: just realized CI is failing, please see my next comment.

packages/pulumi-aws/src/utils/uploadFolderToS3.ts
```diff
@@ -139,7 +139,7 @@ export const uploadFolderToS3 = async ({
                                 ...fields,
                                 "Content-Type": contentType || "",
                                 "X-Amz-Meta-Checksum": checksum,
-                                file: fs.readFileSync(path, "utf8")
+                                file: fs.readFileSync(path)
```
Pavel910, line 142: @chiizzy1 you'll need to adapt the TS type for `data`, because of this error here:
https://github.com/webiny/webiny-js/pull/2635/checks#step:12:27

Previously, all values were simple strings, but with this change, `file` is now a Buffer, and not a string.

Pavel910 (changes requested):
We need to modify the type of `data` as well, since not everything is a string with this change.

packages/pulumi-aws/src/utils/uploadFolderToS3.ts
chiizzy1: Hi @Pavel910, I don't know much about typeScript, but do you think either of these could work?
1 - file: JSON.parse(fs.readFileSync(path, 'utf-8'))
or this
2 -  file: fs.readFileSync(path.toString())

adrians5j: @chiizzy1 

Just add the `| Buffer` to the `data` const's type:

```ts
const data: Record<string, string | Buffer> = { ...
```

This could probably be done a bit smarter / cleaner, but in order to simplify things and since this is an internal / narrow-scoped thing, this will be fine for now.

chiizzy1: @adrians5j Thanks for the help.

chiizzy1: @Pavel910 please review my PR.

adrians5j:
This looks good @chiizzy1!

Thanks a lot for your time, we'll make sure to include your name in the changelog for the 5.33.0 release and in the list of contributors. 🙂

PR #2640, opened 2022-09-26
feat: add folder tree component

leopuleo:
## Changes

Related [WEB-1592](https://webiny.atlassian.net/browse/WEB-1592) 

With this PR we create the folder tree component to enable users to list and create folders for a specific `type`. This is part of the wider context ACO -> Advanced Content Organisation.

## How Has This Been Tested?

TBD

## Documentation

README.md inside package


packages/app-folders/package.json
```diff
@@ -0,0 +1,54 @@
...
+    "@material-design-icons/svg": "^0.12.1",
+    "@minoru/react-dnd-treeview": "^3.1.0",
+    "@webiny/app": "^5.32.0",
+    "@webiny/app-admin": "^5.32.0",
+    "@webiny/form": "^5.32.0",
+    "@webiny/ui": "^5.32.0",
+    "@webiny/validation": "^5.32.0",
+    "graphql": "^15.7.2",
+    "graphql-tag": "^2.12.6",
+    "lodash": "^4.17.11",
```
Pavel910: `lodash.get`

packages/app-folders/src/components/Tree/ButtonCreate/index.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+type Props = {
+    onClick: () => void;
+};
+
+const t = i18n.ns("app-folders/components/tree/button-create");
+
+export const CreateButton: React.FC<Props> = ({ onClick }) => {
+    return (
+        <Button onClick={() => onClick()}>
```
Pavel910: Check this, and if possible, pass `onClick` directly as a prop, without the extra arrow function.

packages/app-folders/src/components/Tree/DialogCreate/index.tsx
```diff
@@ -0,0 +1,120 @@
...
+        throw new Error(t`Slug can contain only letters, numbers, dashes and underscores`);
+    };
+
+    useEffect(() => {
+        setDialogOpen(open);
+    }, [open]);
+
+    return (
+        <CreateDialogContainer open={dialogOpen} onClose={onClose}>
+            {dialogOpen && (
```
Pavel910, line 63: No need for `dialogOpen` check.

packages/app-folders/src/components/Tree/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+            setTreeData(data);
+        }
+    }, [folders]);
+
+    useEffect(() => {
+        if (folders) {
+            const list = createInitialOpenList(folders, focusedFolderId);
+            setInitialOpenList(list);
+        }
+    }, [folders]);
```
Pavel910: Maybe combine this into 1 hook, and also drop the `const` assignment.

packages/app-folders/src/contexts/Folders.tsx
```diff
@@ -0,0 +1,100 @@
...
+        const { data, loading: queryLoading } = await client.query<
+            ListFoldersResponse,
+            ListFoldersQueryVariables
+        >({
+            query: LIST_FOLDERS,
+            variables: { type },
+            fetchPolicy: "network-only"
+        });
+
+        setListLoading(queryLoading);
```
Pavel910: Check the loading state; `true` is never set 🤔

packages/app-folders/src/contexts/Folders.tsx
```diff
@@ -0,0 +1,100 @@
...
+            throw new Error("Folder `type` is mandatory");
+        }
+
+        const { data, loading: queryLoading } = await client.query<
+            ListFoldersResponse,
+            ListFoldersQueryVariables
+        >({
+            query: LIST_FOLDERS,
+            variables: { type },
+            fetchPolicy: "network-only"
```
Pavel910: Avoid `network-only`.

Pavel910 (changes requested):
Several minor things to tweak and double-check 🍻

packages/app-folders/src/hooks/useCreateFolder.ts
```diff
@@ -0,0 +1,33 @@
...
+import { useListFolders } from "./useListFolders";
+
+const t = i18n.ns("app-folders/hooks/use-create-folder");
+
+export const useCreateFolder = () => {
+    const { createFolder: create, loading } = useFolders();
+    const { listFolders } = useListFolders();
+    const { showSnackbar } = useSnackbar();
+
+    const createFolder = async (data: Partial<FolderItem>, type: string) => {
```
Pavel910: `Partial` is too permissive, use either Pick or Omit.

Pavel910 (approved):

packages/api-folders-so-ddb/src/operations/folders.ts
```diff
@@ -151,7 +151,7 @@ export const createFoldersStorageOperations = (
                     partitionKey: createFolderGsiPartitionKey({ tenant, locale, type }),
                     options: {
                         index: "GSI1",
-                        beginsWith: folder.parentId
+                        beginsWith: folder.id
```
Pavel910: I would suggest adding a `#` to this value, so it looks like `${folder.id}#`, to be 100% sure that it never accidentally matches another `id`. It's usually a good practice to do it when using `beginsWith`.

packages/app-folders/src/contexts/folders.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+        const response = await callback();
+
+        setLoading({
+            ...loading,
+            [type]: false
+        });
+
+        return response;
+    };
```
Pavel910: Wrap this into a `useCallback`, because we don't want to recreate a function on every change of provider state. `setLoading` will maintain the reference, so you don't need to specify any dependencies in the `useCallback`. Then, use `setLoading` with a callback:

```ts
const wrapper = useCallback(() => {
    setLoading(loading => ({ ...loading, [type]: true }));
}, []);
```

packages/app-folders/src/contexts/folders.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+    const context: FoldersContext = {
+        folders,
+        loading,
+        async listFolders(type: string) {
```
Pavel910, line 54: All of the context methods can be memoized using the `useCallback`, as they only use the `setFolders`, so their dependencies never change. Just declare them outside of this object. This will help the React components that are using these methods with re-render management.

packages/app-folders/src/contexts/links.tsx
```diff
@@ -0,0 +1,197 @@
...
+    deleteLink(link: LinkItem): Promise<true>;
+}
+
+export const LinksContext = React.createContext<LinksContext | undefined>(undefined);
+
+interface Props {
+    children: ReactNode;
+}
+
+export const LinksProvider = ({ children }: Props) => {
```
Pavel910, line 40: This provider is identical to the previous one, so apply all the previous comments to this file as well.

packages/app-folders/src/contexts/links.tsx
```diff
@@ -0,0 +1,197 @@
...
+
+        const response = await callback();
+
+        setLoading({
+            ...loading,
+            [type]: false
+        });
+
+        return response;
+    };
```
Pavel910: You could extract this function and make it reusable from both providers, simply by creating a factory which receives a `setLoading` parameter, and returns the wrapper.

packages/app-folders/src/hooks/useFolders.ts
```diff
@@ -0,0 +1,40 @@
...
+    }, []);
+
+    return {
+        /**
+         * NOTE: do NOT expose listFolders from this hook, because you already have folders in the `folders` property.
+         * You'll never need to call `listFolders` from any component. As soon as you call `useFolders()`, you'll initiate
+         * fetching of `folders`, which is managed by the FoldersContext.
+         */
+        loading: context.loading,
+        folders: context.folders[type],
```
Pavel910: Just crossed my mind: if folders are grouped by type, how do you know which type of folders is actually changing the `loading` flags? :) I think loading state also needs to be per type. Correct me if I'm wrong.

Pavel910 (changes requested):
@leopuleo Left some comments; please have a look, and we can sync up additionally if something needs to be clarified.

Pavel910 (approved):

PR #2644, opened 2022-09-28
fix: fix not refreshed Navigator tree for editors

neatbyte-vnobis:
## Changes
Fix useEffect dependency, so now Navigator side panel is refreshed properly.
But as a side effect this effect is called much more often (on element on hover for example)
But if I understand it properly we are using refs here exactly to keep performance impact minimal

PS: we may split this useEffect in separate... but I don't thing it will improve performance much

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-vnobis:
@Pavel910 please review this pull request

Pavel910 (approved):

PR #2665, opened 2022-10-06
Fix PB import/export private file handling

Pavel910:
## Changes
This  PR is related to #2664 in a sense that we no longer have a `file.key` property, so we can't use S3 bucket keys in the export logic. The process was simplified to always use ZIP file URLs, instead of S3 bucket keys.

Additionally, interaction with S3 was fixed, and `public-read` ACL was replaced with a `private` one, which also required changing the URL from simple S3 file URL to a presigned URL.

## How Has This Been Tested?
Manually.

## Documentation
Not necessary, these are internal changes.


adrians5j (approved):
lgtm

PR #2670, opened 2022-10-07
Website App Config - Introduce `previewDomains` Config Param

### LINKED ISSUE #2616: Expose configuration for website app preview Cloudfront

We are currently configuring our website domain and would like the preview domain to also be configurable, but it's not exposed in the app.resources during the createWebsiteApp function.

When a user is working with the page builder, and they want to see work in progress, they get taken to the prerendering service and it's on an internal domain.  We'd like to expose that so that our users would be taken to something like web-preview.mydomain.com

Leaving the configuration free and manually managing it.  It is working at the moment because the custom headers are not applied to it - so it works if I add our wild card cert and alternate domain name to cloud front and configure the front end.  Longer term, it'd be better to have this be something that's scripted :)

adrians5j:
## Changes

Closes #2616

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


Pavel910 (approved):
Nice and clean! 👍

adrians5j:
Thank you 🙂

PR #2672, opened 2022-10-07
feat: page block variables feature

neatbyte-vnobis:
## Changes
Block variables feature

## How Has This Been Tested?
Manual

## Documentation
None


Pavel910:
To start the review, I pulled the branch and just started the app and clicked through. It's looking really good 😃 👏 

Things I noticed that I'd like us to address:

1) On the code level, let's make this a bit more consistent: we have a `blockId`, but then `varRef`; `varRef` is effectively an ID, right? let's just call it that, and use `id` key instead of `varRef`. Especially since it's inside of a `variables` array so we know it's a variable.

CleanShot 2022-10-11 at 11 06 41

2) Empty grid block got lost, we still need to have it for cases when users want to just drop stuff into the page sporadically.

CleanShot 2022-10-11 at 11 15 55

3) There's an `updatePage` mutation being issued for every character that I type into the variable input:

https://user-images.githubusercontent.com/3920893/195054489-6e44bccb-b2aa-4ca3-a6ed-c4f6a291b2b1.mp4

I'll submit the second part of the review through actual code review, this feedback was easier to post this way.

packages/api-page-builder/src/graphql/graphql/pages.gql.ts
```diff
@@ -285,17 +285,41 @@ const createBasePageGraphQL = (): GraphQLSchemaPlugin<PbContext> => {
...
+                                    data: {
+                                        blockId,
+                                        ...blockData?.content?.data,
+                                        variables
+                                    },
                                     elements: blockData?.content?.elements || []
                                 });
                             } else {
-                                blocks.push(block);
+                                blocks.push(pageBlock);
```
Pavel910: Please extract this logic into a separate file, just so we keep the resolvers shorter and easier to read. Besides, you'll be able to unit test this mapping logic if it lives in a standalone function.

packages/app-page-builder/src/blockEditor/components/elementSettingsTab/VariableSettings.tsx
```diff
@@ -0,0 +1,109 @@
...
+        );
+
+        return { label: variable?.label };
+    }, [block, element]);
+
+    const onSubmit = useCallback(
+        formData => {
+            if (block && block.id) {
+                const newVariables = block.data?.variables?.map((variable: PbBlockVariable) => {
+                    if (variable?.varRef === element?.data?.varRef) {
```
Pavel910: As I mentioned in the PR comment, let's use `variableId` within individual elements, instead of `varRef`.

packages/app-page-builder/src/blockEditor/config/editorBar/SaveBlockButton/SaveBlockButton.tsx
```diff
@@ -19,11 +19,10 @@ const DefaultSaveBlockButton: React.FC = () => {
...
                 debounce: false,
                 onFinish() {
-                    history.push(`/page-builder/page-blocks`);
-
                     // Let's wait a bit, because we are also redirecting the user.
                     setTimeout(() => {
-                        showSnackbar(`Block ${block.name} saved successfully!`);
+                        showSnackbar(`Block "${block.name}" saved successfully!`);
+                        history.push(`/page-builder/page-blocks`);
                     }, 500);
```
Pavel910: Just curious, why do we need a timeout here? This is not a change request, just want to know, because snackbar should be visible even after router change.

packages/app-page-builder/src/editor/plugins/elementSettings/clone/CloneAction.ts
```diff
@@ -1,11 +1,23 @@
...
 
+const removeVarRef = (element: PbEditorElement) => {
+    if (element?.data?.varRef) {
+        const elementCopy = cloneDeep(element);
+        delete elementCopy.data.varRef;
+
+        return elementCopy;
+    } else {
+        return element;
+    }
```
Pavel910: Nitpicking here, but you don't need the `else` part, since you're returning within the `if`.

Pavel910 (changes requested):
Overall, this looks good. I didn't want to nitpick too much, but I just want us to address the things from my first PR comment, and that's it.

packages/app-page-builder/src/blockEditor/config/editorBar/SaveBlockButton/SaveBlockButton.tsx
neatbyte-vnobis: @Pavel910 I have changed it, so setTimeout is not required anymore

but there are few more places where such approach is used:
`packages/app-page-builder/src/pageEditor/config/editorBar/PublishPageButton/PublishPageButton.tsx line 45`
`packages/app-page-builder/src/pageEditor/config/editorBar/PageOptionsMenu/SetAsHomepageButton.tsx line 56`

should I change them too?

Pavel910: @neatbyte-vnobis To be honest, I think this is a leftover from a long time ago. If you can, please try removing it and see if it works. If not, leave it as is. Thanks a lot!

packages/app-page-builder/src/editor/plugins/elementSettings/clone/CloneAction.ts
neatbyte-vnobis: inside if func is returning the **cloned** element and in the else - original element is returned
it is not required to clone the element if it is not using a variable (it is pointless, just more cpu work)

The only reason I am cloning the element is to remove varRef (aka id) without affecting the element we are cloning now.
It is a case for element with a variable.
If element is without variables - original element can be returned

neatbyte-vnobis:
@Pavel910 
regarding issue 3) about a lot of updatePage API calls: I have fixed this with sage of `history: false` param in update event
so now only onBlur will trigger an API call (like other editor inputs, aka "heading", etc.)

packages/app-page-builder/src/editor/plugins/elementSettings/clone/CloneAction.ts
Pavel910: @neatbyte-vnobis I didn't mean the logic, but the actual `else` statement. So just do the return like this:

```ts
if (...) {
    // your logic
    return whatever;
}

return element;
```

neatbyte-vnobis: got it

neatbyte-vnobis: fixed

Pavel910 (approved):
@neatbyte-vnobis great, thanks!

PR #2676, opened 2022-10-11
fix: folders-api update operation

leopuleo:
## Changes
With this PR we aim to fix the following: 
- `updateFolder` operation: check if a folder with the same slug exists at the same level of the tree (parentId)
- Dependencies optimization
- Use `get` util to retrieve the folder from ddb

## How Has This Been Tested?
Jest

## Documentation
Not needed


packages/api-folders-so-ddb/src/operations/folders.ts
```diff
@@ -110,7 +111,9 @@ export const createFoldersStorageOperations = (
                     items,
                     sort,
                     fields: []
-                }).map(item => item?.data);
+                })
+                    .map(item => item?.data)
+                    .filter(item => Boolean(item));
```
Pavel910: This can be simplified to just `.filter(Boolean)`.

packages/api-folders-so-ddb/src/operations/links.ts
```diff
@@ -101,7 +102,9 @@ export const createLinksStorageOperations = (entity: Entity<any>): LinksStorageO
                     items,
                     sort,
                     fields: []
-                }).map(item => item?.data);
+                })
+                    .map(item => item?.data)
+                    .filter(item => Boolean(item));
```
Pavel910: `.filter(Boolean)`

Pavel910 (approved):

PR #2690, opened 2022-10-17
Page builder blocks v2: variables for other elements

neatbyte-vnobis:
## Changes
Add support of other variable types

Notes:
1) [MediumEditor](https://github.com/webiny/webiny-js/compare/page-builder-blocks-v2...page-builder-blocks-v2-variables-for-other-elements?expand=1#diff-31cbe1d931f4cbfca68a1338baf8fe2ce156bff1c43d5b224291954ff1957949) changes are due to "cached" initial state of the rich text editor (it was stored in scope on initial subscribe, and was not updated after that)
2) I have tried to use a makeComposable for image/imageList components (but I not sure if it worth it, please take a look)
3) I have added a "type" property to the variables (as different elements are storing their element value in different places). And now value is "any" (not "string")
4) I have wrapped right panel inputs with local state (otherwise, on edit in the middle of the string a cursor was jumping to the end of it... do not why, but I have not investigated it deeply)

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/blockEditor/config/ElementSettingsTabContentPlugin.tsx
+++ b/packages/app-page-builder/src/blockEditor/config/ElementSettingsTabContentPlugin.tsx
@@ -6,12 +6,14 @@ import ElementNotLinked from "~/blockEditor/components/elementSettingsTab/Elemen
 import VariableSettings from "~/blockEditor/components/elementSettingsTab/VariableSettings";
 import VariablesList from "~/blockEditor/components/elementSettingsTab/VariablesList";
 
+const elementTypes = ["heading", "paragraph", "image", "images-list", "quote", "list", "button"];
+
 export const ElementSettingsTabContentPlugin = createComponentPlugin(
     SidebarActions,
     SidebarActionsWrapper => {
         return function SettingsTabContent({ children, ...props }) {
             const [element] = useActiveElement();
-            const canHaveVariable = element && element.type === "heading";
+            const canHaveVariable = element && elementTypes.some(type => type === element.type);
             const hasVariable = element && element.data?.variableId;
             const isBlock = element && element.type === "block";
 
--- a/packages/app-page-builder/src/blockEditor/config/eventActions/saveBlock/saveBlockAction.ts
+++ b/packages/app-page-builder/src/blockEditor/config/eventActions/saveBlock/saveBlockAction.ts
@@ -30,10 +30,28 @@ const syncBlockVariables = (block: PbElement) => {
         variable: PbBlockVariable
     ) {
         const dataObject = findNestedObj(block.elements, "variableId", variable.id);
+        const isTextVariable =
+            variable.type === "heading" ||
+            variable.type === "paragraph" ||
+            variable.type === "quote" ||
+            variable.type === "list";
 
-        if (dataObject) {
+        if (dataObject && isTextVariable) {
             result.push({ ...variable, value: dataObject?.text?.data?.text });
         }
+        if (dataObject && variable.type === "image") {
+            result.push({ ...variable, value: dataObject?.image?.file });
+        }
+        if (dataObject && variable.type === "images-list") {
+            result.push({ ...variable, value: dataObject?.images });
+        }
+        if (dataObject && variable.type === "button") {
+            result.push({
+                ...variable,
+                value: { label: dataObject?.buttonText, url: dataObject?.action?.href }
+            });
+        }
+
         return result;
     },
     []);
--- a/packages/app-page-builder/src/blockEditor/plugins/elementSettings/CreateVariableAction.ts
+++ b/packages/app-page-builder/src/blockEditor/plugins/elementSettings/CreateVariableAction.ts
@@ -28,6 +28,7 @@ const CreateVariableAction: React.FC<CreateVariableActionPropsType> = ({ childre
                         ...(block.data?.variables || []),
                         {
                             id: element.id,
+                            type: element.type,
                             label: startCase(camelCase(element.type))
                         }
                     ]
--- a/packages/app-page-builder/src/editor/plugins/elementSettings/variable/RichInput.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elementSettings/variable/RichInput.tsx
@@ -1,0 +1,54 @@
+import React, { useState } from "react";
+import styled from "@emotion/styled";
+import { CoreOptions } from "medium-editor";
+import ReactMediumEditor from "~/editor/components/MediumEditor";
+
+const InputWrapper = styled("div")({
+    padding: "20px 16px",
+    backgroundColor: "rgba(212, 212, 212, 0.5)",
+    borderBottom: "1px solid",
+
+    "&:hover": {
+        backgroundColor: "rgba(212, 212, 212, 0.7)"
+    },
+
+    "&>p": {
+        minHeight: "auto",
+        lineHeight: "normal"
+    }
+});
+
+const DEFAULT_EDITOR_OPTIONS: CoreOptions = {
+    toolbar: {
+        buttons: ["bold", "italic", "underline", "anchor"]
+    },
+    anchor: {
+        targetCheckbox: true,
+        targetCheckboxText: "Open in a new tab"
+    }
+};
+
+interface RichInputProps {
+    value: string;
+    onChange: (value: string) => void;
+    onSelect: (value: string) => void;
+    [key: string]: any;
+}
+
+const RichInput: React.FC<RichInputProps> = ({ value, onChange, onSelect }) => {
+    const [initialValue] = useState(value);
+
+    return (
+        <InputWrapper className="webiny-pb-page-element-text">
+            <ReactMediumEditor
+                tag="p"
+                value={initialValue}
+                onChange={onChange}
+                onSelect={onSelect}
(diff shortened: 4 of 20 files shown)


packages/app-page-builder/src/blockEditor/config/ElementSettingsTabContentPlugin.tsx
```diff
@@ -6,12 +6,14 @@ import ElementNotLinked from "~/blockEditor/components/elementSettingsTab/Elemen
 import VariableSettings from "~/blockEditor/components/elementSettingsTab/VariableSettings";
 import VariablesList from "~/blockEditor/components/elementSettingsTab/VariablesList";
 
+const elementTypes = ["heading", "paragraph", "image", "images-list", "quote", "list", "button"];
```
Pavel910, line 9: Once you implement my instructions from the PR comment, with new plugins, you'll be able to fetch this dynamically, from plugins.

packages/app-page-builder/src/blockEditor/config/eventActions/saveBlock/saveBlockAction.ts
```diff
@@ -30,10 +30,28 @@ const syncBlockVariables = (block: PbElement) => {
         variable: PbBlockVariable
     ) {
         const dataObject = findNestedObj(block.elements, "variableId", variable.id);
+        const isTextVariable =
```
Pavel910, line 33: Same as the previous comment, once we have individual plugins in place, you'll be able to call those plugins to return a value you need from them, without hardcoding this logic here.

packages/app-page-builder/src/blockEditor/plugins/elementSettings/CreateVariableAction.ts
```diff
@@ -28,6 +28,7 @@ const CreateVariableAction: React.FC<CreateVariableActionPropsType> = ({ childre
                         ...(block.data?.variables || []),
                         {
                             id: element.id,
+                            type: element.type,
```
Pavel910, line 31: Continuing the theme here :)
This entire section will be used via `variablePlugin.createVariables({ element })`.

packages/app-page-builder/src/editor/plugins/elementSettings/variable/RichInput.tsx
```diff
@@ -0,0 +1,54 @@
...
+                value={initialValue}
+                onChange={onChange}
+                onSelect={onSelect}
+                options={DEFAULT_EDITOR_OPTIONS}
+            />
+        </InputWrapper>
+    );
+};
+
+export default RichInput;
```
Pavel910, line 54: There's a better mechanism for this, but let's leave this like this for the time being, until we sort out individual element variable plugins.

packages/app-page-builder/src/editor/plugins/elementSettings/variable/TextInput.tsx
```diff
@@ -0,0 +1,33 @@
...
+            onChange={value => {
+                onChange(value);
+                setLocalValue(value);
+            }}
+            onBlur={onBlur}
+        />
+    );
+};
+
+export default TextInput;
```
Pavel910, line 33: Same as the previous comment; we'll improve this down the road, for now let's leave it as is.

packages/app-page-builder/src/editor/plugins/elementSettings/variable/VariableSettings.tsx
```diff
@@ -1,32 +1,56 @@
...
+});
+
+const textToBlockQuote = (text: string) => {
+    return `<blockquote><q>${text}</q></blockquote>`;
+};
+
+const blockQuoteToText = (text: string) => {
+    return text.replace("<blockquote><q>", "").replace("</q></blockquote>", "");
+};
+
```
Pavel910, line 37: This entire file needs to be refactored to use the new plugins that I described in the PR comment section. That way we'll achieve a complete separation of concerns here, and won't need to mix different element handling logic in this component.

packages/app-page-builder/src/editor/plugins/elementSettings/variable/VariableSettings.tsx
```diff
@@ -52,17 +76,68 @@ const VariableSettings: React.FC = () => {
...
+                                label="URL"
+                                value={variable?.value?.url}
+                                onChange={value =>
+                                    onChange({ ...variable.value, url: value }, variable.id)
+                                }
+                                onBlur={onBlur}
+                            />
+                        </div>
+                    )}
+                </div>
```
Pavel910, line 138: Refactor this to look up the appropriate variable plugin using the `element.type`, and call `variablePlugin.renderVariables()` to handle individual element independently.

packages/app-page-builder/src/editor/plugins/elementSettings/variable/VariableSettings.tsx
```diff
@@ -1,32 +1,56 @@
...
+    return text.replace("<blockquote><q>", "").replace("</q></blockquote>", "");
+};
+
 const VariableSettings: React.FC = () => {
     const [element] = useActiveElement();
     const updateElement = useUpdateElement();
 
     const onChange = useCallback(
-        (value: string, variableId: string) => {
+        (value: any, variableId: string, history = false) => {
```
Pavel910, line 43: Extract this into a hook, like `useVariableOnChange` which you can then use in each individual element variable plugin.

packages/app-page-builder/src/editor/plugins/elements/imagesList/ImagesList.tsx
```diff
@@ -30,4 +27,9 @@ const ImagesList: React.FC<ImagesListProps> = ({ data }) => {
     return <ListComponent data={images} theme={theme} />;
 };
 
-export default React.memo(ImagesList);
+export default makeComposable(
```
Pavel910, line 30: To answer your question about `makeComposable`, it doesn't hurt. It's great to see you're utilizing the built-in tools 👍 Let's keep this and see if this is the right approach in this situation.

packages/app-page-builder/src/types.ts
```diff
@@ -208,8 +208,9 @@ export interface PbElement {
 
 export interface PbBlockVariable {
     id: string;
+    type: string;
     label: string;
-    value?: string;
+    value?: any;
 }
```
Pavel910, line 214: It's probably a good idea to make this interface generic, so you can use the default interface where the variable type is unknown, but use a concrete data type within the rendering logic:

```ts
export interface PbBlockVariable<TValue = any> {
    id: string;
    type: string;
    label: string;
    value?: TValue;
}
```

Then you can use it like this:

```ts
type ButtonVariable = {
  label: string;
  url: string;
}

const variable: PbBlockVariable<ButtonVariable> = element.whatever;

// variable.value.label & variable.value.url are now properly typed and provide autocomplete.
```

Pavel910 (changes requested):
@neatbyte-vnobis Did a thorough review of the entire PR, and we'll not rush with this one, as there are things to address on an architectural level here, which will make the code nicer, and the logic more decoupled and pluginable.

---

My thinking goes like this. We need to create 2 plugin types:

```ts
/**
 * This plugin is responsible for handling the "create variable" event, when a user wants to create a
 * variable for the selected element of type `elementType`. Elements can support multiple variables, depending on the
 * complexity of their settings, so we need to return an array of PbBlockVariable from the `create` callback.
 *
 * Example: `button` element has a `label` and a `URL`, so it will create 2 variables.
 *
 * Hint: to group variables per element, we can use a dot notation for variable id: `hoEX06c7Xn.label` and `hoEX06c7Xn.url`
 */
export type PbBlockEditorCreateVariablePlugin = Plugin & {
    type: "pb-block-editor-create-variable";
    elementType: string;
    createVariables: (params: { element: PbEditorElement }) => PbBlockVariable[];
};

/**
 * This plugin is responsible for rendering of element variables  of type `elementType`.
 * Example: `button` element will result in 2 inputs being rendered by the `render` callback (label and URL).
 * No parameters are necessary here, as you can always access the active element using hooks.
 */
export type PbEditorPageElementVariableRendererPlugin = Plugin & {
    type: "pb-editor-page-element-variable-renderer";
    elementType: string;
    // If needed for value transformation before the API call.
    getVariableValue(): any; 
    renderVariables: () => ReactNode;
};
```

Please read the code comments before continuing reading.

I intentionally made `createVariables` return an array of variables, since you mentioned that a user might be able to  delete default presets of values when you create a variable set for an element. If you don't need this, convert it to a single variable instead.

The idea is to have a variable plugin for every type of element that needs to support variables. This way we avoid hardcoding of element types, conditional logic, etc. everywhere in the system. Also, this makes it possible for developers to create their own variable plugins for their own custom elements.

I also left several related comments in the actual PR review, so this description should make it all pretty clear. If not, feel free to get in touch.

The location of `PbEditorPageElementVariableRendererPlugin` plugins:
CleanShot 2022-10-17 at 18 38 34

For `PbBlockEditorCreateVariablePlugin`, create a folder in the `blockEditor/plugins` folder.

To import/register those plugins, use the approach used in the `app-i18n` package: `packages/app-i18n/src/I18N.tsx:22`

neatbyte-vnobis:
Refactoring was done in pull https://github.com/webiny/webiny-js/pull/2734

PR #2734, opened 2022-10-25
feat: refactored Block Variables with plugins usage

neatbyte-vnobis:
## Changes
Refactor Block Variables to use plugin approach

Note:
1. Maybe plugins related to variables should be loaded like here `packages/app-page-builder/src/admin/components/EditorPluginsLoader.tsx`?

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elementSettings/variable/MultipleImageVariableInput.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elementSettings/variable/MultipleImageVariableInput.tsx
@@ -1,0 +1,89 @@
+import React from "react";
+import { css } from "emotion";
+/**
+ * Package react-sortable does not have types.
+ */
+// @ts-ignore
+import { sortable } from "react-sortable";
+import { cloneDeep } from "lodash";
+import { FileManager } from "@webiny/app-admin/components";
+import File from "~/editor/plugins/elements/imagesList/File";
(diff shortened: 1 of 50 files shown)


packages/app-page-builder/src/editor/plugins/elementSettings/variable/MultipleImageVariableInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+const SortableItem = sortable(Item);
+
+interface MultipleImageVariableInputProps {
+    variableId: string;
+}
+
+const MultipleImageVariableInput: React.FC<MultipleImageVariableInputProps> = ({ variableId }) => {
+    const { value, onChange } = useVariable(variableId);
+
+    return (
```
Pavel910, line 44: Please add a TODO comment here, saying:
```
Update to use the new `render` prop, implemented in 5.33.0.
```

We'll take care of it later.

Pavel910 (approved):
@neatbyte-vnobis Love it! Thanks a bunch!

Please note the one comment I left, just add that TODO for us, and we can merge this.
Also, don't worry about those plugins being registered from React, that's how I want them to be, at some point I'll figure out a way to remove those editor plugins from user's projects, but for now this is perfectly fine.

Thank you!

PR #2742, opened 2022-11-01
fix: fix pasting into paragraph/text editor component

### LINKED ISSUE #2346: [Page Builder] Pasting function pastes text at the beginning of the paragraph instead of new row
### Version

5.24.0

### Operating System

MacOS 12.2.1

### Browser

Chrome

1. In Page Builder, create a new page and add an element to it. 
(I've had this issue with Heading, Paragraph, List and Block quote elements).

2. Within the element, write some text then try to paste more text at the end of the sentence or in a new row. 

The pasted text should be pasted in the current cursor position. 

### What do you see instead?

At the moment, the pasted text will be added at the very beginning of the text in the element. 

I've tried several options and found that: 

1. if the text is pasted after a blank space or a new row, the error will occur and the text will be added at the beginning of the paragraph. 

Situation A - one space added at the end of the sentence:
image

Result:
image

Situation B - new row added
image

Result:
image

2. If the text is pasted right after another character (without space or row break) it will be added correctly. 

Situation C - pasted without space or rowbreak
image

Result: 
image

### Additional information

_No response_

### Possible solution

_No response_

neatbyte-vnobis:
## Changes
Fix "Pasting function pastes text at the beginning of the paragraph instead of new row" bug.
Closes #2346

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/components/Text/PbText.tsx
+++ b/packages/app-page-builder/src/editor/components/Text/PbText.tsx
@@ -59,7 +59,10 @@ const PbText: React.FC<TextElementProps> = ({ elementId, mediumEditorOptions, ro
         return null;
     }
 
-    const textContent = variableValue || get(element, `${DATA_NAMESPACE}.data.text`);
+    // only used to set initial value, so no reason to update it every time
+    const textContent = useMemo(() => {
+        return variableValue || get(element, `${DATA_NAMESPACE}.data.text`);
+    }, []);
     const tag = get(value, "tag");
     const typography = get(value, "typography");
 
--- a/packages/app-page-builder/src/editor/components/Text/PbText.tsx
+++ b/packages/app-page-builder/src/editor/components/Text/PbText.tsx
@@ -38,6 +38,11 @@ const PbText: React.FC<TextElementProps> = ({ elementId, mediumEditorOptions, ro
         [displayMode]
     );
 
+    const initialText = useMemo(
+        () => variableValue || get(element, `${DATA_NAMESPACE}.data.text`),
+        []
+    );
+
     const value = get(element, `${DATA_NAMESPACE}.${displayMode}`, fallbackValue);
 
     const onChange = useCallback(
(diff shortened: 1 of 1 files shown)


SvenAlHamad:
This issue still seems to be present. The text is pasted always at the start of the paragraph, instead of where the current cursor is positioned. 

https://user-images.githubusercontent.com/3808420/199535373-a59b4db4-02b4-43ed-9545-8f7fb4690d65.mov

SvenAlHamad:
Re-tested this one, and it's working fine now. I had an issue with my build command previously.

packages/app-page-builder/src/editor/components/Text/PbText.tsx
```diff
@@ -59,7 +59,10 @@ const PbText: React.FC<TextElementProps> = ({ elementId, mediumEditorOptions, ro
         return null;
     }
 
-    const textContent = variableValue || get(element, `${DATA_NAMESPACE}.data.text`);
+    // only used to set initial value, so no reason to update it every time
```
adrians5j:
```suggestion
    // Stores only the initial content, so no reason to update on each render.
```

adrians5j (changes requested):
Thanks for resolving this.

Although I had one nitpick before we merge. 🙂

packages/app-page-builder/src/editor/components/Text/PbText.tsx
```diff
@@ -59,7 +59,10 @@ const PbText: React.FC<TextElementProps> = ({ elementId, mediumEditorOptions, ro
...
+    const textContent = useMemo(() => {
```
adrians5j: BTW, can we rename this variable then to `initialText`? That way even the comment would not be needed.

adrians5j (changes requested):
Added one quick question.

packages/app-page-builder/src/editor/components/Text/PbText.tsx
adrians5j: Also, this `useMemo` hook is used after the if statement that can potentially return `null` (check L59).

This hook usage should be relocated above the if statement.

neatbyte-vnobis: @adrians5j I have changed it according to a comment.
Also I have dropped `return` statement (to be in same style as `fallbackValue`)

adrians5j (approved):
Nice.

PR #2746, opened 2022-11-01
fix: remove padding from selected grid cells

### LINKED ISSUE #2696: Page Builder - adding a 5px padding in the editor for every PB cell
### Version

5.33

### Operating System

macOS

### Browser

Chrome

1. Add a block to a page and set a background image
2. Add a grid inside that block and define a background colour with some transparency

You'll get an effect such as this
image

Whilst editing the page you will see there is a 5px padding added around every cell in the tree. Note: that is an artefact introduced by PB when in edit mode. When previewing the page or seeing the published page that 5px is not rendered.

### What do you see instead?

A 5px padding added around every PB cell to help visualise the hierarchy of elements. Unfortunately, that artefact causes confusion as the user may be led to believe that there is a true padding or margin somewhere in their design, when there isn't.

### Additional information

The padding seems to be introduced by the [ElementType]ContainerStyle class

image

### Possible solution

Remove the padding from the [ElementType]ContainerStyle class

neatbyte-vnobis:
## Changes
Fix issue "Let’s remove the padding the element selector adds".
Closes #2696

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elements/block/Block.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elements/block/Block.tsx
@@ -8,7 +8,6 @@ import { ElementRoot } from "~/render/components/ElementRoot";
 const BlockStyle = styled("div")({
     position: "relative",
     color: "#666",
-    padding: 5,
     boxSizing: "border-box"
 });
 interface BlockType {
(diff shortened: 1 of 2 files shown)


SvenAlHamad:
This is partially fixed. An adjustment is also needed inside `packages/app-page-builder/src/editor/plugins/elements/block/Block.tsx`

image

SvenAlHamad:
Looks good

PR #2747, opened 2022-11-01
fix: fix issue with px/% changes and cursor position

### LINKED ISSUE #2699: Page Builder - margin and padding units
### Version

5.33.0

### Operating System

macOS

### Browser

Chrome

Create any page and add a block to it. Try to change the margin (same applies to padding in that page).

There are 2 issues with the margins/padding units:

1. The units change from px to % - when trying to edit the margins - without any user input 
2. The cursor is positioned on the rightmost side of the field, which is confusing from an editing experience. It should be positioned in the position where its clicked or just highlight the entire selected value so the user can enter a new value

### What do you see instead?

See video where 2 issues are shown

https://user-images.githubusercontent.com/8825508/196694195-f78e0ef3-915b-4b96-ad72-c1649bf86af4.mov

### Additional information

_No response_

### Possible solution

_No response_

neatbyte-vnobis:
## Changes
Fix "Issue with px/% changes and cursor position"
Closes #2699

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elementSettings/components/SpacingPicker.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elementSettings/components/SpacingPicker.tsx
@@ -78,8 +78,8 @@ const SpacingPicker: React.FC<SpacingPickerProps> = ({
     useDefaultStyle = true
 }) => {
     const formData = useMemo(() => {
-        const parsedValue = parseInt(value);
-        const regx = new RegExp(`${parsedValue}`, "g");
+        const parsedValue = parseFloat(value);
+        const regx = new RegExp(`[0-9.+-]+`, "g");
         const unit = value.replace(regx, "");
 
         if (Number.isNaN(parsedValue) && unit === "auto") {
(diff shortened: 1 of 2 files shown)


SvenAlHamad:
This issue seem to be partially fixed. The switching from px to % doesn't happen that often any more, but if you delete all the numbers from the number field, while you have `px` selected as the unit, in that case it will switch back to `%`.

Regarding the cursor, now the cursor is positioned to the exact place where you clicked, which means that if you click on the end, it's still be positioned at the end of the field. I would suggest maybe a better behavior, and that's when you click on the field that all the text in the field is selected. I found this to have a better UX. 

https://user-images.githubusercontent.com/3808420/199544341-3aedd04a-4ff0-45f2-bfd2-60bda48c3474.mov

SvenAlHamad:
I tested this one again. The flicking of the px/% is now sorted, but another odd behavior was introduced in this fix. When you click on the field, the current value is selected, and if you then just type in a new value, it works fine, but everytime you use backspace to remove all the numbers from the field, it then adds `0` into the field, but positions the cursor behind it, which then the user needs to again delete for in order to type new values. Can we add a behavior so that when I delete the value in the field, the field is left empty while I'm still focused on it. When I remove the focus, then you can restart it to zero, if I left it empty.


https://user-images.githubusercontent.com/3808420/200059982-fde33723-b020-4143-89b2-fca9ec22af95.mov

neatbyte-vnobis:
@SvenAlHamad pull was updated.
validation for margin now is a copy of logic used for a padding

Pavel910 (approved):
Code wise, looks good. I'll let you guys test the user interaction.

SvenAlHamad:
Works great - thanks for sorting it out @neatbyte-vnobis

PR #2748, opened 2022-11-01
fix: fix incorrect error thrown by the createCategory Mutation

### LINKED ISSUE #2471: GraphQL - `createCategory` Mutation Throwing an Incorrect Error Message
### Version

5.29.0

### Operating System

/

### Browser

/

Just try issuing the following mutation, and you should be able to see the error message on the right:

Snipaste_2022-06-09_11-50-28

We should not get this system-level error. We should get the error message thrown from the `CreateDataModel` model (`packages/api-page-builder/src/graphql/crud/categories.crud.ts:195`).

### What do you see instead?

The error is shown in the screenshot.

### Additional information

_No response_

### Possible solution

This is basically happening because of the incorrect order of operations here.

image

The fix needs to be applied [here](https://github.com/webiny/webiny-js/blob/next/packages/api-page-builder/src/graphql/crud/categories.crud.ts#L187-L195). We need to switch the order of operations.

neatbyte-vnobis:
## Changes
Fix issue "Incorrect error thrown by the createCategory Mutation"
Closes #2471

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-page-builder/src/graphql/crud/categories.crud.ts
+++ b/packages/api-page-builder/src/graphql/crud/categories.crud.ts
@@ -184,15 +184,15 @@ export const createCategoriesCrud = (params: CreateCategoriesCrudParams): Catego
         async createCategory(this: PageBuilderContextObject, input) {
             await checkBasePermissions(context, PERMISSION_NAME, { rwd: "w" });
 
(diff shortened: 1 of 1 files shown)


adrians5j (approved):
Looks good!

PR #2750, opened 2022-11-01
fix: fix incorrect label on Pages export

neatbyte-vnobis:
## Changes
Fix issue "Exporting Pages - Incorrect Label"

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
I don't think the issue was fixed. I've tested it and the behavior is still the same as before. 

https://user-images.githubusercontent.com/3808420/199547303-9f633d63-60c7-4034-a5a9-1355ca991c82.mov

SvenAlHamad:
This one is also good.

PR #2756, opened 2022-11-04
fix: fix too strict Iframe component validation

### LINKED ISSUE #2411: Pagebuilder: Iframe component's validation is too restrictive.
### Version

5.26.0

### Operating System

macos 12.2.1

### Browser

Chrome 101.0.4951.64

1. Create an iframe on a page and start typing a url in the field. The admin app will crash when validation fails on https://w
2. Create an iframe on a page, and try pasting the src with get parameters (e.g. google maps from this site: https://www.maps.ie/create-google-map/). The admin app will also crash.
3

Validation shouldn't cause the app to crash, and any url including get parameters should work.

### What do you see instead?

Crashed app.

### Additional information

_No response_

### Possible solution

_No response_

neatbyte-vnobis:
## Changes
Fixes issue "Iframe component's validation is too restrictive."
Closes #2411

PS: after some thoughts, maybe just "url:base" is enough here... but it is still have some limits compared to code below

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elements/media/iframe/IFrameSettings.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elements/media/iframe/IFrameSettings.tsx
@@ -2,7 +2,6 @@ import React from "react";
 import { css } from "emotion";
 import { merge } from "dot-prop-immutable";
 import { Form } from "@webiny/form";
-import { validation } from "@webiny/validation";
 import { withActiveElement } from "~/editor/components";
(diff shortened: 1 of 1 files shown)


SvenAlHamad:
I've tested and it looks good from my end. It could be improved from the UX side a bit, but we can do that another time. @adrians5j or @Pavel910 please review the code implementation before we merge this one.

Pavel910 (approved):

PR #2757, opened 2022-11-04
fix: fix image default mobile/table alignment

neatbyte-vnobis:
## Changes
Fix issue "Mobile/table default alignment on image".

Note: `applyFallbackDisplayMode` was giving wrong fallback value if value type equals string (It was giving first defined and not the closest as it is for non string values)

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
I've tested this one, and it works fine now. @Pavel910 @adrians5j - please review the code implementation.

Pavel910 (approved):

PR #2758, opened 2022-11-04
fix: fix corrupted undo state after deleting a block

neatbyte-vnobis:
## Changes
Fix issue "The undo state gets corrupted when deleting a block"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/contexts/EventActionHandlerProvider.tsx
+++ b/packages/app-page-builder/src/editor/contexts/EventActionHandlerProvider.tsx
@@ -43,7 +43,7 @@ import {
     GetElementTreeProps
 } from "~/types";
 import { composeAsync, composeSync, AsyncProcessor, SyncProcessor } from "@webiny/utils/compose";
-import { UpdateElementTreeActionEvent } from "~/editor/recoil/actions";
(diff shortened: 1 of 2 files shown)


SvenAlHamad:
I tested this one, and now when you delete a block and do a undo, the block is again visible in the navigator, but there's still something wrong with the state, as if you reload the page, the page is empty again. 


https://user-images.githubusercontent.com/3808420/200064255-738e05db-1a60-425f-bd19-2235c96c79f1.mov

neatbyte-vnobis:
@SvenAlHamad I have not tried to reload page...
But I have an idea about why it is not there, I will fix it on Monday

neatbyte-vnobis:
@SvenAlHamad I have updates on this task:
Indeed there is an issue with saving - it happens with 2 second delay after undo (`debounce: true` is used).
But it is required thing to prevent spamming calls to the API (if few undo are happening one after one, so only last one will do the API call).

On video there were definitely more then 2 seconds... and I was not able to reproduce such issue (all was saved).

SvenAlHamad:
I've retested, and yes, if I wait for over 2s it works fine. I think at this stage, that should be ok. The bug is fixed. 
@adrians5j @Pavel910 - can one of you please review the code if it's good to be merged.

Pavel910 (approved):

PR #2762, opened 2022-11-09
fix: fix Image component shadow

### LINKED ISSUE #2466: Pagebuilder: Shadow not correctly applied to image.
### Version

5.26

### Operating System

MacOS

### Browser

Chrome

Create a page with an image in a grid.
Set the image width to 200px.
Add a shadow to the image.
The shadow is incorrectly applied to the image container (webiny-pb-base-page-element-style webiny-pb-page-element-image), not the image itself.

The image should have the shadow.

### What do you see instead?

The image's container has the shadow, not the image.

### Additional information

_No response_

### Possible solution

_No response_

neatbyte-vnobis:
## Changes
Fix issue "Shadow not correctly applied to image"
Closes #2466

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/apps/theme/pageBuilder/styles/elements/image.scss
+++ b/apps/theme/pageBuilder/styles/elements/image.scss
@@ -9,7 +9,6 @@ img {
 }
 
 .webiny-pb-page-element-image {
-  overflow: hidden;
   box-sizing: border-box;
   box-shadow: none !important;
   //display: flex; removed because of testimonial #2
--- a/apps/theme/pageBuilder/styles/elements/image.scss
+++ b/apps/theme/pageBuilder/styles/elements/image.scss
@@ -11,9 +11,11 @@ img {
 .webiny-pb-page-element-image {
   overflow: hidden;
   box-sizing: border-box;
+  box-shadow: none !important;
   //display: flex; removed because of testimonial #2
   img {
     max-width: 100%;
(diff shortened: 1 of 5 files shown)


SvenAlHamad:
@neatbyte-vnobis I've tested this one, and the shadow is fixed inside the Page Editor, but if you use the preview option, there it's still the same as before. This tells me that the editor plugin was updated with the fix, but the render (used in the preview) was not.

Pavel910 (approved):

neatbyte-vnobis:
@SvenAlHamad I was trying to reproduce this... but I see a shadow on a preview
If you see the issue please do a video recording (so I can follow exact steps)

apps/theme/pageBuilder/styles/elements/image.scss
```diff
@@ -11,9 +11,11 @@ img {
 .webiny-pb-page-element-image {
   overflow: hidden;
```
SvenAlHamad: I think this has to be removed as well. Try inserting an image 100x100px in the editor, and assign a shadow with a spread of 5px to the image. You'll see the shadow is only visible on the sides, and gets hidden on the top and bottom. 
image

SvenAlHamad (changes requested):

Issue #2413, opened 2022-05-19
Pagebuilder: Grid enhancement: # columns by viewport control


jimjoes:
### Is your feature request related to a problem? Please describe.

It would be useful to be able to specify how the grid behaves at different viewports. At the moment, it's 1 column wide < 768  and an arbitrary number of columns wide > 768. This is quite inflexible and makes it difficult to make fluidly responsive layouts.

### Describe the solution you'd like.

Be able to specify how the grid behaves at different viewports. Maybe a min property? Or maybe a viewport specific number of columns?

### Describe alternatives you've considered.

Tried using the grid columns property, but it isn't speicifc to viewport like other settings, and using it removes any content in columns that are 'descoped' when the property changes.

autoexec-midi:
Also important: ordering columns (reverse) by viewports

SvenAlHamad:
Thanks @autoexec-midi - I've noted this down. We'll be addressing this feature this quarter.

Issue #2412, opened 2022-05-19
Pagebuilder: Grid component enhancement: grid order


jimjoes:
### Is your feature request related to a problem? Please describe.

Many layouts have a section with a bunch of image/text rows (e.g. feature blocks) that go in sequence, left/right then right/left (e.g. https://www.bookdapp.io/). While this is achievable on desktop with Webiny, when the grid is rendered on mobile, the order of the opposing grids needs to be 'reversible' so that the order on mobile is image/text image/text, not image/text text/image as it is currently. 

### Describe the solution you'd like.

Ability to 'reverse' the ordering at specific viewports, perhaps using flex-direction?

Issue #1583, opened 2021-04-30
Page Builder: Ability to modify a grid's configuration in different responsive views.


jimjoes:
## This is:

- Feature request

## Detailed Description
It would be useful to be able to add/remove cells in a grid in different responsive viewports.

## Possible Solution
Make the grid settings persist per viewport, in the same way other properties such as width etc can be set differently for the same element in different viewports.

jimjoes:
Bumping this

SvenAlHamad:
Closing this -> this is possible by navigating to the required device view and just activating the "hide" option on the selected cell that you wish to hide on that particular device.

PR #2763, opened 2022-11-09
feat: allow Grid customization according to a viewport size

neatbyte-vnobis:
## Changes
Implementation of feature "Add the ability to customize the grid depending on the device viewport".
Closes:
- #1583
- #2413
- #2412

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
@neatbyte-vnobis - can you please double check this one, I'm not able to build it: 
image

Pavel910:
@SvenAlHamad this usually happens when a new file is added to the package, but the package is not built. `watch` doesn't build TS declaration files, only transpiles with `babel`, so you need to `yarn build` manually. Try building the branch with `yarn build` before starting the admin app.

SvenAlHamad:
Alright, after I did a build, it's working. Thanks @Pavel910 
@neatbyte-vnobis The solution is working as expected. I would just add a very small tweak: 
image
Can you please rename the label "Grid Settings" into "Grid Size", and the the new "Grid Options" should become "Grid Settings". So we would have "Grid Size" and "Grid Settings".

neatbyte-vnobis:
@SvenAlHamad pull was updated according the comment

packages/app-page-builder/src/render/plugins/elementSettings/grid/index.ts
```diff
@@ -7,7 +7,7 @@ export default {
     name: "pb-render-page-element-style-grid",
     type: "pb-render-page-element-style",
     renderStyle({ element, style }) {
-        const gridOptions = get(element, "data.settings.gridOptions");
+        const gridSettings = get(element, "data.settings.gridSettings");
```
Pavel910, line 10: To make this backwards-compatible, let's `get` from both the new and the old paths:

- `data.settings.gridSettings`
- `data.settings.gridOptions`

`get` value from the first one, and if it doesn't exist, fall back to the second (old) one.


In other plugins, editor plugins, make sure you support the old data as well, otherwise we'll break existing pages.

Pavel910 (changes requested):
Please have a look at the comment and double check your changes so that existing pages don't break after we deploy this upgrade. So essentially, both `render` and `editor` plugins must support both old and new data paths.

packages/app-page-builder/src/render/plugins/elementSettings/grid/index.ts
neatbyte-vnobis: @Pavel910 both properties (old `gridOptions` and new `gridSettings`) were added in this pull, so looks like no need to do this.
`grip` property used for size was left as is (it was hardcoded in one of the initial pages, so I have left it unchanged).

But agree that is important thing and will try to keep an eye on such changes in future tasks.

Pavel910: @neatbyte-vnobis ah ok, sorry. Too many properties to keep in mind 😅 All good then, thanks!

Pavel910 (approved):

PR #2764, opened 2022-11-10
feat: new Skeleton UI component

leopuleo:
## Changes
With this PR we introduce a new UI component: `<Skeleton>` for loading states. In details: 

- Upgrade [react-loading-skeleton](https://www.npmjs.com/package/react-loading-skeleton) to latest version
- Create new `<Skeleton>` component
- Refactor loading state for `<DataList>` component

### DataList Before
Screenshot 2022-11-10 at 11 55 15

### DataList After
Screenshot 2022-11-10 at 11 29 57

### Example with DataTable
Screenshot 2022-11-10 at 11 31 42

## How Has This Been Tested?
Cypress

## Documentation
Inside UI package

### DIFF
--- a/packages/ui/src/List/DataList/Loader.tsx
+++ b/packages/ui/src/List/DataList/Loader.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { ReactElement } from "react";
 import styled from "@emotion/styled";
 
 import { Skeleton } from "~/Skeleton";
@@ -49,12 +49,13 @@ const Actions = styled("div")`
     }
 `;
 
-const Loader: React.FC = () => (
-    <LoaderUl data-testid={"default-data-list.loading"}>
-        {Array(5)
-            .fill("")
-            .map(item => (
-                <li key={"list-" + item}>
+const Loader = (): ReactElement => {
+    const lines = Array.from(Array(5).keys());
+
+    return (
+        <LoaderUl data-testid={"default-data-list.loading"}>
+            {lines.map(line => (
+                <li key={"list-" + line}>
                     <LoaderWrapper>
                         <Graphic>
                             <Skeleton height={36} />
@@ -77,7 +78,8 @@ const Loader: React.FC = () => (
                     </LoaderWrapper>
                 </li>
             ))}
-    </LoaderUl>
-);
+        </LoaderUl>
+    );
+};
 
 export default Loader;
--- a/packages/ui/src/List/DataList/Loader.tsx
+++ b/packages/ui/src/List/DataList/Loader.tsx
@@ -1,64 +1,82 @@
 import React from "react";
 import styled from "@emotion/styled";
-/**
- * Package react-loading-skeleton does not have types.
- */
-// @ts-ignore
-import Skeleton from "react-loading-skeleton";
 
-const LoaderUl = styled("ul")({
-    listStyle: "none",
-    padding: "10px 20px",
-    "li > div": {
-        display: "inline-block",
-        verticalAlign: "middle",
-        ".react-loading-skeleton": {
-            backgroundColor: "var(--mdc-theme-background)",
-            backgroundImage:
-                "linear-gradient(90deg, var(--mdc-theme-background), var(--mdc-theme-surface), var(--mdc-theme-background))"
-        }
-    },
-    ".graphic": {
-        fontSize: 36,
-        width: 36
-    },
-    ".data": {
-        width: "calc(-36px + 75%)",
-        marginLeft: 10
-    },
-    ".actions": {
-        width: "calc(-36px + 25%)",
-        marginLeft: 10,
-        textAlign: "right",
-        "> div": {
-            display: "inline-block",
-            fontSize: 24,
-            marginLeft: 10,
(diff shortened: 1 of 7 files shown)


packages/ui/src/Skeleton/Skeleton.tsx
```diff
@@ -0,0 +1,15 @@
+import React, { ReactElement } from "react";
+import ReactLoadingSkeleton, { SkeletonProps, SkeletonTheme } from "react-loading-skeleton";
+
+import "react-loading-skeleton/dist/skeleton.css";
+
+export const Skeleton = (props: SkeletonProps): ReactElement => {
```
adrians5j, line 6: Can we rather use `Skeleton: React.FC<SkeletonProps>` here?

packages/ui/src/Skeleton/README.md
```diff
@@ -0,0 +1,21 @@
+# Skeleton
+
+### Additional information
+
+[https://www.npmjs.com/package/react-loading-skeleton](https://www.npmjs.com/package/react-loading-skeleton)
+
+### Description
+
+`Skeleton` component is used to animated loading skeletons within Webiny.
```
adrians5j:
```suggestion
`Skeleton` component is used to render animated loading skeletons within Webiny.
```

packages/ui/src/Skeleton/Skeleton.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+
+const story = storiesOf("Components/Skeleton", module);
+
+story.add(
+    "usage",
+    () => {
+        return (
+            <Story>
+                <StoryReadme>{readme}</StoryReadme>
+                <StorySandbox title={"A simple loading Skeleton, single line"}>
```
adrians5j, line 15: Does Storybook even work? 🤔 
I remember it not working, but it was super long time ago when I last tried it.

packages/ui/src/List/DataList/Loader.tsx
```diff
@@ -1,64 +1,82 @@
...
-                    <div>
-                        <Skeleton />
-                    </div>
-                </div>
-            </li>
-        ))}
+        {Array(5)
+            .fill("")
+            .map(item => (
+                <li key={"list-" + item}>
```
adrians5j: Because of `fill("")`, isn't every item here `""`? If so, then the `key` prop on `li` won't work (all keys will be the same - `"list-"`).

You probably wanted to use index here?

Also, on a 2nd thought, maybe we could just, somewhere around L51, do:

```ts
// Not sure about the const name "items". Maybe "lines"? if you come up with a better name, feel free to use.
const items = Array.from(Array(5).keys());
```

... and then we'd have a bit less verbose code in the actual JSX. So, just `{items.map(item => { ...`.

packages/ui/src/List/DataList/Loader.tsx
```diff
@@ -1,64 +1,82 @@
...
+
+const LoaderWrapper = styled("div")`
+    margin-bottom: 16px;
+    display: flex;
+    width: 100%;
+    align-items: center;
+    justify-content: space-around;
+`;
+
+const Graphic = styled("div")`
```
adrians5j, line 19: The fact the we need so much CSS is unfortunate, but let's leave if for now as is.

adrians5j (changes requested):
A couple of minor things, but all in all, we should be good to go soon.

packages/ui/src/Skeleton/Skeleton.tsx
leopuleo: Yes we can, I just take [this PR](https://github.com/facebook/create-react-app/pull/8177) as reference while I create new components

packages/ui/src/Skeleton/Skeleton.stories.tsx
leopuleo: Not working 😢 
I created the story to document the component and so it's ready when we fix Storybook.

packages/ui/src/List/DataList/Loader.tsx
```diff
@@ -1,65 +1,85 @@
...
+
+    .actions-skeleton {
+        width: 24px;
+        height: 24px;
+        margin-left: 16px;
+    }
+`;
+
+const Loader = (): ReactElement => {
+    const lines = Array.from(Array(5).keys());
```
adrians5j, line 53: P.S. Normally I'd also use `useMemo` here, but since the component doesn't have any props, I think we're good.

packages/ui/src/Skeleton/Skeleton.stories.tsx
adrians5j: Cool. :)

packages/ui/src/Skeleton/Skeleton.tsx
adrians5j: Good point @leopuleo , let's leave it as is then. BTW I like the **Provides an implicit definition of children** point.

adrians5j (approved):
Looks good, thank you @leopuleo. 😊

PR #2770, opened 2022-11-11
fix: fix root cause of failing test

neatbyte-vnobis:
## Changes
Fix empty array case (caused failing in the test)

## How Has This Been Tested?
None

## Documentation
None


Pavel910 (approved):
Good catch, thank you!

PR #2794, opened 2022-11-22
feat: add id and css class to elements

neatbyte-vnobis:
## Changes
Implements issue "Ability to assign a css class and ID to elements"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/toolbar/navigator/TreeView.tsx
+++ b/packages/app-page-builder/src/editor/plugins/toolbar/navigator/TreeView.tsx
@@ -156,7 +156,7 @@ const TreeViewItem: React.FC<TreeViewItemProps> = ({ element, level, children, i
                     <Typography use={"body2"} className={"title"}>
                         {element.type}
                         {elementIdAttribute && (
-                            <p className={elementIdStyle}>{` (${elementIdAttribute})`}</p>
+                            <p className={elementIdStyle}>{`#${elementIdAttribute}`}</p>
                         )}
                     </Typography>
                     {hidden ? <ElementVisibilityAction elementId={elementId} /> : null}
--- a/packages/app-page-builder/src/editor/plugins/toolbar/navigator/TreeView.tsx
+++ b/packages/app-page-builder/src/editor/plugins/toolbar/navigator/TreeView.tsx
@@ -15,7 +15,8 @@ import { NavigatorContext } from "./Navigator";
 import { PbEditorElement } from "~/types";
 
 const elementIdStyle = css({
-    textTransform: "none"
+    textTransform: "none",
+    fontSize: "80%"
 });
 
 const ElementVisibilityAction = ({ elementId }: { elementId: string }) => {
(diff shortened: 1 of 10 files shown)


neatbyte-vnobis:
@SvenAlHamad @Pavel910 
FYI: for long element ids Navigator UI will become not very nice:
image

Maybe it worth to disable line wrap and add ... at the end of the line?
But in such way part of id value will be hidden, maybe label (browser build in onhovel tooltip) with full value should be added?

SvenAlHamad:
Yeah, that's a good point. 
Let's place the element id into a new row, and set font-size to 80%, this way longer element IDs will still fit. 
image

neatbyte-vnobis:
@SvenAlHamad change is done, but I propose to replace `()` with `#` prefix (it is same as CSS selector and more self descriptive in my opinion)
See image bellow (maybe bolder font will be better):
image (30)

SvenAlHamad:
Looks good, thanks for sorting it out @neatbyte-vnobis

SvenAlHamad:
@Pavel910 - this one's good to be reviewed.

neatbyte-vnobis:

packages/app-page-builder/src/editor/plugins/toolbar/navigator/TreeView.tsx
```diff
@@ -1,18 +1,24 @@
...
-import { useVisibilitySetting } from "~/editor/plugins/elementSettings/visibility/VisibilitySettings";
+import { useVisibilitySetting } from "~/editor/plugins/elementSettings/property/PropertySettings";
 import { ElementTypeContainer } from "./StyledComponents";
 import CollapsableList from "./CollapsableList";
 import DragBlockIndicator from "./DragBlockIndicator";
 import { BLOCK, useMoveBlock, useSortableList } from "./navigatorHooks";
 import { NavigatorContext } from "./Navigator";
 import { PbEditorElement } from "~/types";
 
+const elementIdStyle = css({
```
Pavel910: Just a note, feel free to use the template literal, and use plain CSS, instead of the object notation. And I mean not just this PR, but in general.

```
const myClass = css`
  color: red;
`

const StyledComponent = styled.div`
  border: 1px solid black;
`
```

And so on. It's easier, and you get very nice autocomplete on CSS within the string literals.

Pavel910 (approved):

Pavel910 (approved):

neatbyte-vnobis:
@Pavel910 I was planning to merge this change into branch `page-builder-blocks-v2.1`...
But I see that you have merged it into `next` branch.
Should I merge `next` into it? Is it safe action now?
Or will it be better to create separate pull into `page-builder-blocks-v2.1`?

Pavel910:
@neatbyte-vnobis if you need it in your umbrella branch, feel free to merge the branch directly.
But if you don't need it, then you can just leave it as is. We needed this feature in one of our internal projects, so I merged it directly into `next`, it all worked fine. You can consider this task "done", it's already incorporated.

And yes, you _can_ merge `next` into your umbrella branch, it shouldn't cause any issues.

neatbyte-vnobis:
Merged to `page-builder-blocks-v2.1` branch in pull https://github.com/webiny/webiny-js/pull/2815

PR #2795, opened 2022-11-22
feat: add site-wide custom scripts

neatbyte-vnobis:
## Changes
Implement issue "Ability to insert custom scripts"

## How Has This Been Tested?
Manual

## Documentation
None?

### DIFF
--- a/packages/app-page-builder/src/modules/WebsiteSettings/settingsGroups/HtmlTags.tsx
+++ b/packages/app-page-builder/src/modules/WebsiteSettings/settingsGroups/HtmlTags.tsx
@@ -1,0 +1,61 @@
+import React from "react";
+import { css } from "emotion";
+import { Bind } from "@webiny/form";
+import { AddPbWebsiteSettings } from "../AddPbWebsiteSettings";
+import { CodeEditor } from "@webiny/ui/CodeEditor";
+import { Typography } from "@webiny/ui/Typography";
+
+const inputStyle = css({
+    "& span": {
+        color: "var(--mdc-theme-text-primary-on-background)"
+    },
+    "& .ace_editor": {
+        marginTop: "8px",
+        maxHeight: "200px",
+        fontSize: 16,
+
+        ".ace_info": {
+            background: "none"
+        }
+    }
+});
+
+type CodeEditorInputProps = {
+    name: string;
+    label: string;
+    description: string;
+};
+
+const CodeEditorInput = ({ name, label, description }: CodeEditorInputProps) => {
+    return (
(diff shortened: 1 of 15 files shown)


SvenAlHamad:
@neatbyte-vnobis Just checking if you saw my comment on this one on the previous branch. I'm copy/pasting it here so we have it in the current branch as well. 

---

Can you please double check this one. When I insert a meta tag into the tags field, it shows up, but if I insert a simple script, like this one, it's not inserted into the page:

<script>
    alert('Hello from header');
</script>

adrians5j:
Just a heads up...

When inserting scripts into HTML, we should make sure this doesn't happen as well:
https://webiny-community.slack.com/archives/C014Y0HGJ0Y/p1663937553970529

neatbyte-vnobis:
I have fixed described issues
@SvenAlHamad @Pavel910 please take a look

neatbyte-vnobis:
@adrians5j I have looked on that thread - it should be working with if condition that was mentioned there

SvenAlHamad:
@neatbyte-vnobis - I've tested this, and something odd is happening. The footer scripts get inserted, but, they don't get executed and I can't find the reason why. 

This is why I have in my settings:
image

And I see both header and footer content being injected. The header meta is visible and the script triggers. The footer script is visible, but it doesn't trigger. Can you check on your end please.

neatbyte-vnobis:
@SvenAlHamad I have fixed the issue and checked the preview (earlier I was looking on final page version)

SvenAlHamad:
Works fine now. Thanks for sorting it out.
@Pavel910  - this one is ready for a code review.

packages/app-page-builder/src/modules/WebsiteSettings/settingsGroups/HtmlTags.tsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+import { css } from "emotion";
+import { Bind } from "@webiny/form";
```
Pavel910, line 3: I just realized that this is the first time ever anyone used this approach, and that it actually works 🤯 
It works because internally, Bind uses a `useBind` hook, which binds to a form context.

Previously, we were prop-drilling the Bind that we got from the Form; but recently we introduce the underlying React context, so you can use the `useBind({ name: "fieldName" })` and just pass that to your UI components. Importing `Bind` directly - you just showed me another way to use my own library 🤣 🍻 

This is just a comment, no changes necessary 😅

Pavel910 (approved):
Looks great, thank you!

PR #2796, opened 2022-11-22
feat: PageBuilder list with DataTable

leopuleo:
## Changes
With this PR we introduce the new organisation in folders for the PageBuilder application.

## How Has This Been Tested?

TBD

## Documentation

TBD


package.json
```diff
@@ -44,6 +44,7 @@
     "babel-core": "^7.0.0-bridge.0",
     "babel-eslint": "^10.1.0",
     "babel-jest": "^28.1.2",
+    "babel-plugin-dynamic-import-node": "^2.0.0",
```
Pavel910, line 47: Move this dependency to `project-utils`, that's where that dependency is being used.

packages/app-folders/src/components/Tree/index.tsx
```diff
@@ -79,22 +82,29 @@ type Props = {
...
     const { showSnackbar } = useSnackbar();
 
-    useEffect(() => {
+    useDeepCompareEffect(() => {
         if (folders) {
             setTreeData(createTreeData(folders, focusedFolderId));
             setInitialOpenList(createInitialOpenList(folders, focusedFolderId));
         }
-    }, [folders]);
+    }, [Object.assign({}, folders), focusedFolderId]);
```
Pavel910, line 107: This `Object.assign` looks problematic to me, because it creates a new object on every render, so even the internal `ref` can't help, because it will always get a new object. Am I wrong?

Why not just send `folders` in there?

packages/app-page-builder/src/admin/components/Table/Row/Folder/FolderActionDelete.tsx
```diff
@@ -0,0 +1,48 @@
...
+                            ? folder.name.slice(0, 20).concat("...")
+                            : folder.name
+                    }`
+                })
+            );
+        }
+    }, [folder]);
+
+    if (!folder) {
+        return <></>;
```
Pavel910, line 44: Any reason why this is not `return null;` ?

Pavel910 (approved):

packages/app-page-builder/src/admin/components/Table/Row/Folder/FolderActionDelete.tsx
leopuleo: none in particular, it's driven by the component type `ReactElement`

packages/app-folders/src/components/Tree/index.tsx
leopuleo: Sending the folders array won't work, that's the reason why I use this workaround

Pavel910 (approved):

Pavel910 (approved):

Pavel910 (approved):

PR #2803, opened 2022-11-25
fix: advanced element settings onSave execution

### LINKED ISSUE #2776: PageBuilder - onSave callback on advanced element settings plugins is not executed
### Version

_No response_

### Operating System

_No response_

### Browser

_No response_

Define a plugin with the `onSave` callback as depicted by the following plugin type:

```tsx
export type PbEditorPageElementAdvancedSettingsPlugin = Plugin & {
    type: "pb-editor-page-element-advanced-settings";
    elementType: string;
    render(params: { Bind: BindComponent; data: any; submit: () => void }): ReactElement;
    onSave?: (data: FormData) => FormData;
};
```

The `onSave` callback is not called when the settings form is submitted.

The `onSave` callback is executed, and my `data` modifications are applied to the element.

### What do you see instead?

The callback is simply ignored, as if it's not there.

### Additional information

_No response_

### Possible solution

Invoke the callback in this `onSubmit` handler here:
https://github.com/webiny/webiny-js/blob/dev/packages/app-page-builder/src/editor/plugins/elementSettings/advanced/ElementSettings.tsx#L15-L17

Also, make the callback `async` so developers can do asynchronous stuff here (like calling an external API).

neatbyte-vnobis:
## Changes
Fix issue "PageBuilder - onSave callback on advanced element settings plugins is not executed".
Closes #2776

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elementSettings/advanced/ElementSettings.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elementSettings/advanced/ElementSettings.tsx
@@ -14,18 +14,20 @@ export const ElementSettings: React.FC = () => {
     const updateElement = useUpdateElement();
 
     const onSubmit = async (formData: FormData) => {
-        const plugin = plugins
+        const settingsPlugins = plugins
             .byType<PbEditorPageElementAdvancedSettingsPlugin>(
                 "pb-editor-page-element-advanced-settings"
             )
-            .find(pl => pl.elementType === element?.type && typeof pl?.onSave === "function");
+            .filter(pl => pl.elementType === element?.type);
 
-        let modifiedFormData = null;
-        if (typeof plugin?.onSave === "function") {
-            modifiedFormData = await plugin.onSave(formData);
+        let modifiedFormData = formData;
+        for (const plugin of settingsPlugins) {
+            if (typeof plugin?.onSave === "function") {
+                modifiedFormData = await plugin.onSave(modifiedFormData);
+            }
         }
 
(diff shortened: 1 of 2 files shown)


neatbyte-vnobis:

packages/app-page-builder/src/editor/plugins/elementSettings/advanced/ElementSettings.tsx
```diff
@@ -12,8 +13,19 @@ export const ElementSettings: React.FC = () => {
...
+                "pb-editor-page-element-advanced-settings"
+            )
+            .find(pl => pl.elementType === element?.type && typeof pl?.onSave === "function");
+
+        let modifiedFormData = null;
+        if (typeof plugin?.onSave === "function") {
+            modifiedFormData = await plugin.onSave(formData);
+        }
+
+        updateElement(merge(element, "data", modifiedFormData || formData));
```
Pavel910: This is incorrect because we can have multiple plugins for the same `elementType`. A good example is `images-list`, or `pages-list`:
CleanShot 2022-11-28 at 21 39 20


So this needs to be executed with an array of plugins in mind (use `filter` instead of `find`).

The easiest way would be:
```
let modifiedFormData = formData; // this way you don't have to think if there are many or 0 plugins to execute.

// ... iterate over plugins and call each with `modifiedFormData`

updateElement(merge(element, "data", modifiedFormData));
```

Pavel910 (changes requested):

Pavel910 (approved):

PR #2804, opened 2022-11-25
fix: page image not saved

neatbyte-vnobis:
## Changes
Fix issue "Page image not saved"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-admin/src/ui/elements/form/FileManagerElement/FileManagerElementRenderer.tsx
+++ b/packages/app-admin/src/ui/elements/form/FileManagerElement/FileManagerElementRenderer.tsx
@@ -66,7 +66,7 @@ export class FileManagerElementRenderer extends UIRenderer<
 
                         <FileManager
                             onChange={(value: FileManagerFileItem | null) =>
-                                onChange({ id: value?.id, src: value?.src })
+                                onChange(value ? { id: value.id, src: value.src } : null)
                             }
                             accept={accept}
                             images={!accept}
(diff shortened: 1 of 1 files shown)


neatbyte-vnobis:

packages/app-admin/src/ui/elements/form/FileManagerElement/FileManagerElementRenderer.tsx
```diff
@@ -75,7 +77,8 @@ export class FileManagerElementRenderer extends UIRenderer<
                                     fileManagerElement: element,
                                     showFileManager,
                                     value,
-                                    onChange
+                                    onChange: (value: FileManagerFileItem | null) =>
+                                        onChange({ id: value?.id, src: value?.src })
```
Pavel910: Ah good catch! This is something we added a few releases back (the `FileManagerFileItem` type). But, both `onChange` callbacks are actually wrong. You can call `onChange` with an object OR `null`. So calling it like you did will construct an invalid value, when `value === null`.

It should be `onChange(value ? { id: value.id, src: value.src } : null)`, because we can also unset an image, and in that case, the value will be `null`, and we want to propagate that further to the form.

Pavel910 (changes requested):

neatbyte-vnobis:
@Pavel910 please review a fix

Pavel910 (approved):
Tested this and it works as expected. Thanks!

PR #2805, opened 2022-11-27
feat: block editor changes

neatbyte-vnobis:
## Changes
Fix issues:
- Prevent user from losing their changes
- When you press save on a block in a block editor, there’s a delay before you are moved to the block list. We should display a loader if the user really needs to wait before the screen changes, or if they don’t need to wait, move them instantly to the block list screen.
- On every save, we create a new image file in our File Manager (it’s hidden, but it’s still being created). Ideally we would like to delete the previous file when uploading a new one, to keep the system clean.
- When I click edit on a block and go to the block editor. Then, when I click back in the block editor, return me back to the blocks list view, but position me inside the category I was looking previously.

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-vnobis:

packages/app-page-builder/src/blockEditor/config/eventActions/saveBlock/getPreviewImage.ts
```diff
@@ -23,10 +24,14 @@ function getDataURLImageDimensions(dataURL: string): Promise<ImageDimensionsType
     });
 }
 
-interface createdImageType {
-    data: any;
+interface CreatedImageType {
```
Pavel910: I missed this part in my reviews earlier; it would be a lot better if `getPreviewImage` returned `null` if there's no plugin, rather than an empty object. Ideally, the return type of `getPreviewImage` is `Promise<File | null>`.

packages/app-page-builder/src/blockEditor/config/eventActions/saveBlock/saveBlockAction.ts
```diff
@@ -73,7 +73,7 @@ export const saveBlockAction: BlockEventActionCallable<SaveBlockActionArgsType>
     // See `pageEditor` for an example and feel free to copy that same logic over here.
     const element = (await state.getElementTree()) as PbElement;
     // We need to grab the first block from the "document" element.
-    const createdImage = await getPreviewImage(element.elements[0], meta);
+    const createdImage = await getPreviewImage(element.elements[0], meta, state.block?.preview?.id);
```
Pavel910, line 75: Following my previous comment, `createdImage` can be `File | null`, and the `preview` below on line 99 should also be handled as such.

packages/app-page-builder/src/types.ts
```diff
@@ -22,6 +22,20 @@ export enum PageImportExportTaskStatus {
...
+    id: string;
+    name: string;
+    key: string;
+    src: string;
+    size: number;
+    type: string;
+    tags: string[];
+    meta: Record<string, any>;
+    createdOn: string;
+}
```
Pavel910, line 37: @neatbyte-vnobis I updated this type based on the actual GraphQL query being executed, as API types do not match entirely (not all fields are returned from the API). This is the actual data that you get from the API.

packages/app-page-builder/src/blockEditor/config/eventActions/saveBlock/getPreviewImage.ts
```diff
@@ -23,10 +24,14 @@ function getDataURLImageDimensions(dataURL: string): Promise<ImageDimensionsType
...
+    data: File | {};
 }
-export default async function getPreviewImage(element: any, meta: any): Promise<createdImageType> {
+export default async function getPreviewImage(
```
Pavel910: Please avoid default exports. Export named functions, components, etc. We're trying to get rid of all default exports in the system over time, as they introduce various issues; from module system compatibility, to poor "Find Usages" in IDEs.

packages/app-page-builder/src/blockEditor/config/editorBar/SaveBlockButton/SaveBlockButton.tsx
```diff
@@ -1,32 +1,55 @@
...
+    return (
+        <>
+            {loading && (
+                <LoadingWrapper>
+                    <CircularProgress label={"Saving block..."} />
+                </LoadingWrapper>
+            )}
+            <ButtonPrimary onClick={saveChanges}>Save Changes</ButtonPrimary>
+        </>
+    );
```
Pavel910: Can we make this loading indicator local to the button only? The overlay looks a little over the top. We can just add a spinner to the button, and disable the button while the saving is taking place.

The code for button with icon goes like this:
```jsx
<ButtonPrimary>
  <ButtonIcon icon={<SvgBaselineCloudDone24Px />} />
  Click to proceed
</ButtonPrimary>
```

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveDialog/graphql.ts
```diff
@@ -33,3 +33,14 @@ export const CREATE_FILE = gql`
         }
     }
 `;
+
```
Pavel910: Please add a TODO here for the core team, to export these queries from the `app-file-manager` package, to make them reusable. Thanks!

Pavel910 (changes requested):

SvenAlHamad:
@neatbyte-vnobis - regarding "Prevent user from losing their changes" - the check triggers when I press the back button in the editor, but if I press the back button in my browser, it doesn't work and I lose my changes. We should implement a check on the back button as well. (this will probably need to be a standard alert box as you can't trigger custom visual element on such action)

"When you press save on a block in a block editor, there’s a delay before you are moved to the block list. We should display a loader if the user really needs to wait before the screen changes, or if they don’t need to wait, move them instantly to the block list screen."
-> this one is looking good

Pavel910:
@neatbyte-vnobis regarding @SvenAlHamad's comment on the "back" button:
That one is tricky, because `react-router` removed support for that functionality, so we had to manually hack stuff together to make it work. Try the following:

```jsx
import { Prompt } from "@webiny/react-router";

// And mount it somewhere in that view
<Prompt when={!isPristine} message={prompt} />
```

We're using this in other places in the system, so this should do the trick.

packages/app-page-builder/src/blockEditor/config/editorBar/SaveBlockButton/SaveBlockButton.tsx
neatbyte-vnobis: @Pavel910 just in case: I am adding smth like this, right?
image

neatbyte-vnobis:
@Pavel910 I have tried to use this prompt... but no success so far.

I see that `<Prompt />` is also used in `/cms/content-models` page.
But it doesn't work (at least I do not see it locally on "back" click)
Maybe it is not fixed in this branch? or just broken...

Pavel910:
@neatbyte-vnobis it actually _does_ work in our `dev` branch (see screenshot). Let me see why it's not working in your branch 🤔 

CleanShot 2022-11-29 at 21 08 22

Pavel910:
@neatbyte-vnobis Ok, new discoveries:

I noticed this warning in your branch:
CleanShot 2022-11-29 at 21 18 33

That reminded me that `react-router` announced that they'll be removing some internals which allows you to implement this prompt. Comparing the `react-router` dependency version installed in my copy of `dev` branch, and your branch, shows that your branch pulled in `6.4.2`, while my `dev` pulled in `6.3.0` - that's why it works for me, but no longer for you.

What we'll do is lock `react-router-dom` to `6.3.0` in `@webiny/react-router` package.

Feel free to do it in this PR, and you can then properly test the Prompt component.

packages/app-page-builder/src/blockEditor/config/editorBar/SaveBlockButton/SaveBlockButton.tsx
neatbyte-vnobis: @SvenAlHamad @Pavel910 I want to confirm one thing:
We have such "saved"/"saving" marker in the Page Editor (it is showing save progress after each edit action - revision is updated when done).

On Block Editor we have same icon - but it is not used (as we save block only on "save" button click).
Should I remove it from Block Editor? or use it to show the block save progress (keep in mind that we already have a spinner inside the button)?

Other thing I want to confirm is about how to determine an unsaved changes.
1) we can track edit actions (if edits >0 - smth is unsaved)
(if there were edits, but we undo all of them - block is marked as changed, as edits > 0)
2) we can compare current block state with initial state
(if there were edits, but we undo all of them - block is marked as NOT changes, as initial and current block objects are same)


image

SvenAlHamad: Regarding the save icon - let's remove that one from the left sidebar as we don't have autosave functionality inside the block editor.

Pavel910: @neatbyte-vnobis The indicator kicks in when the actual API call is made, and it does work, so just leave it there.

Maybe you can hook into the `UpdateElementActionEvent`, and whenever it's triggered, you store your own state, which means the block is "dirty". Then you also hook into the `SaveBlockActionEvent`, and when it's executed, you can update your state as "not dirty".

Essentially you just listen for those 2 events and track some state of your own, a dedicated recoil atom, or just simple React state.

Pavel910: @neatbyte-vnobis about that indicator again... I ignored the tooltip previously; yeah it makes no sense for the Block editor. What we could do is replace that widget with a Block Editor specific one, which will show like an exclamation mark when the block is not saved 🤔

neatbyte-vnobis: @SvenAlHamad @Pavel910 
I will go on with Sven request to remove it, ok?
I will remove its state from the atom. Most likely I will replace it with "dirty" flag.
But FYI - In such case undo operation will not reset "dirty" state
Is it ok for you?

Pavel910: @neatbyte-vnobis sorry, Sven posted too quickly, and I missed his message.
So yes, remove the icon.

About the undo/redo, when those events happen, we also execute two event actions: `UpdateDocumentActionEvent` and `UpdateElementTreeActionEvent`, so you could also hook into one of those, and track the "dirty" state that way. So regardless of the action, undo or redo, just track it as "dirty".

neatbyte-vnobis:
@Pavel910 looks like all comments are fixed, please review

Pavel910 (approved):
@neatbyte-vnobis I pushed a commit which greatly simplifies the implementation, without adding new action events, removing the custom dialog, since now the <Prompt/> component works fine. Please have a look at my last commit to see how I did this without adding anything new, and essentially just using a simple `useState` to track the "dirty" state.

neatbyte-vnobis:
@Pavel910 I see you commit
Indeed your approach is much simpler in the code.
(I was not aware that we can return react in the `EventActionHandlers`)

But your approach changes one thing that may be important:
You are showing system prompt (default browser alert) for both "back to blocks list" button click and browser "back" button click.

While initial implementation was showing more styled modal on "back to blocks list" button click. And system prompt was shown as a fallback on browser "back" button click.

Maybe it is not important difference, @SvenAlHamad please let me know if we can merge this pull.

Pavel910:
@neatbyte-vnobis @SvenAlHamad I removed the styled dialog to keep the behavior consistent across all apps that have this mechanism (Page Builder, Headless CMS, Form Builder).

You actually just mount Prompt component which renders null under the hood, so it's all the same where it's located. In our case here, it's mounted within the EditorConfig, so it will be rendered within the BlockEditor only, but it will essentially just add a history listener and still render null :)

neatbyte-vnobis:
@SvenAlHamad FYI I have missed to mention that one more issue is fixed in this pull:
> When I click edit on a block and go to the block editor. Then, when I click back in the block editor, return me back to the blocks list view, but position me inside the category I was looking previously.

SvenAlHamad:
The alert on back button doesn't seem to be working. It only triggers sometimes, but not all the time. Often the back button (in browser and in the UI) doesn't work, and I get this warning in my console: 
image

neatbyte-vnobis:
@SvenAlHamad just in case - there are dependency changes in this pull, so you need to reinstall dependencies.
maybe it can be a reason of these errors?
Anyway I will take a look on it tomorrow.

Pavel910:
I did test this extensively, and it worked for me every time. I think @neatbyte-vnobis is right, you just need to run `yarn` to install new deps, then restart the admin app (`react-router-dom` was locked to a specific version to make this work).

SvenAlHamad:
I've tested this one more, and now the back button does work, BUT if you make a quick change and then press the back button, the alert doesn't trigger. But if you make a change and wait for 2-3 seconds, then it does trigger. Here's a video: 

https://user-images.githubusercontent.com/3808420/205274055-861920f3-6a97-4888-a91c-659c5c16efb4.mov

Pavel910:
@SvenAlHamad that is expected because page state update has a delay (there are reasons), so our code runs after that delay. Not saying this is "good" or "bad", but not sure if we can hook into anything earlier 🤔 I'll have to look at this and see if there are any events that can help us with this.

SvenAlHamad:
It's not a big problem if we leave it like this. Don't spend too much time on it.

PR #2816, opened 2022-12-01
feat: move button label into sidebar and scrollToElement action

neatbyte-vnobis:
## Changes
Implements issues:
- Move button label into sidebar
- Button add scroll to element option

## How Has This Been Tested?
Manual

## Documentation
None?

### DIFF
--- a/apps/theme/pageBuilder/styles/base.scss
+++ b/apps/theme/pageBuilder/styles/base.scss
@@ -1,6 +1,16 @@
 @use "./mixins/border";
 @use "./mixins/background";
 
+html {
+  scroll-behavior: smooth;
+}
+
+@media screen and (prefers-reduced-motion: reduce) {
+  html {
+    scroll-behavior: auto;
+  }
+}
+
 .webiny-pb-page {
   min-height: 100vh;
   position: relative;
--- a/packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useMemo, useState } from "react";
-import { Radio, RadioGroup } from "@webiny/ui/Radio";
 import { css } from "emotion";
 import { merge } from "dot-prop-immutable";
 import { Switch } from "@webiny/ui/Switch";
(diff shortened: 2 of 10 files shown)


SvenAlHamad:
@neatbyte-vnobis - works great, thanks!
@Pavel910  - this one is ready for code review.

apps/theme/pageBuilder/styles/base.scss
```diff
@@ -1,6 +1,16 @@
...
+html {
+  scroll-behavior: smooth;
+}
+
+@media screen and (prefers-reduced-motion: reduce) {
+  html {
+    scroll-behavior: auto;
+  }
+}
+
```
Pavel910, line 13: @neatbyte-vnobis I'm curious, how crucial is this piece of CSS for the scroll functionality?
Existing projects will not have this in their projects, and I'm wondering if this will work in some kind of a degraded manner without this CSS, or will it be entirely broken?

If this is required, we'll have to create a codemod 😭 or add instructions in the upgrade guide so people can manually add this to their projects.

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
```diff
@@ -176,7 +193,32 @@ const ActionSettingsComponent: React.FC<ActionSettingsPropsType> = ({
...
+                                    <Bind name={"scrollToElement"}>
+                                        {({ value, onChange }) => (
+                                            <Wrapper
+                                                label={"Element ID"}
+                                                containerClassName={classes.gridClass}
+                                            >
+                                                <SelectField
+                                                    value={value}
+                                                    onChange={onChange}
+                                                    placeholder={"Select id"}
```
Pavel910: I think a placeholder of `None` will look better. Also, there's a warning in the console related to this select box:

CleanShot 2022-12-01 at 21 13 44

Pavel910 (changes requested):
This part here looks off @SvenAlHamad @neatbyte-vnobis, how about we convert this to a regular dropdown? See screenshots:

CleanShot 2022-12-01 at 21 15 57

CleanShot 2022-12-01 at 21 17 54

CleanShot 2022-12-01 at 21 18 28

Also @neatbyte-vnobis let's make casing consistent, see `Scroll to element` vs. `Click Handler`; I think a regular sentence case will work here: `Click handler`

apps/theme/pageBuilder/styles/base.scss
SvenAlHamad: It will work without this, it just won't animate the scrolling, but instead it will just jump to the anchor point

SvenAlHamad:
I'm ok with making the selector a dropdown.

packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
neatbyte-vnobis: Yes, it is related to inner `SelectField` implementation - I will take a look how to fix it

apps/theme/pageBuilder/styles/base.scss
neatbyte-vnobis: @SvenAlHamad @Pavel910 
Are there any additional action items for me?
Create a codemod?
Move css to some other place?

Pavel910: Since this is not a blocking problem, I would avoid codemods for this, because they're time consuming, and this is not the only change that needs to be made to the user's project. We have a lot more stuff coming from the rest of the team, that will be "breaking change", and we won't be doing code mods for those.

I noted this change down, and we'll simply write upgrade instructions in one big upgrade guide, with all the things that will be released in 5.34.0. So from my point of view, there's nothing you need to do here.

Pavel910 (approved):

apps/theme/pageBuilder/styles/base.scss
neatbyte-vnobis: @Pavel910 Got it.
I am merging the pull.

PR #2818, opened 2022-12-01
feat: blocks list page changes

neatbyte-vnobis:
## Changes
Implements issues:
- In the blocks list view, where we have the block categories on the left side, we should also display the block icon next to the category name so it looks nicer visually. Do the same inside the Block categories list.
- In the blocks list view, if I’m inside a block category and click new block button → there is no need to ask the user to select a block category, let’s just use the category they are in currently.
- In the page editor - if I click on a block that’s linked, automatically position the right sidebar to the “Element” tab, instead of the “Style” tab.

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/admin/utils/createBlockCategoryPlugin.tsx
+++ b/packages/app-page-builder/src/admin/utils/createBlockCategoryPlugin.tsx
@@ -8,7 +8,7 @@ interface IconProps {
     category: PbBlockCategory;
 }
 
-const Icon: React.FC<IconProps> = ({ category }) => {
+export const Icon: React.FC<IconProps> = ({ category }) => {
     return (
         <FontAwesomeIcon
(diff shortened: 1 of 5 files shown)


neatbyte-vnobis:
@SvenAlHamad regarding task "In the blocks list view, where we have the block categories on the left side, we should also display the block icon next to the category name so it looks nicer visually. Do the same inside the Block categories list."

This added part "Do the same inside the Block categories list." do you mean this modal on new block creation (if none  category is selected)
image

Other place where categories are used - inside block editor (edit block category modal)
But we are using simple `<select/>` there, so we need to replace it with some other control to add the icons.

SvenAlHamad:
@neatbyte-vnobis  - good point, let's add the icon to that block list from your screenshot.
Other than that, I've tested the above points and it looks good.

neatbyte-vnobis:
@SvenAlHamad icons were added

Pavel910 (approved):

PR #2823, opened 2022-12-05
fix: folder tree ui

leopuleo:
## Changes
With this PR we improve the `FolderTree` component: 
- feat: add "loading" state
- feat: add "empty" state
- fix: keep folders open after the user clicks on another folder
- feat: improve drag and drop position feedback
- feat: improve drag and drop folder at the root level
- fix: sort folders by their name

Related to [WEB-1865](https://webiny.atlassian.net/browse/WEB-1865)

### Screenshots

Sorting
sorting

Empty List
empty

Loading List
loading

## How Has This Been Tested?

TBD

## Documentation
Documentation within the package


packages/app-folders/src/components/Tree/Placeholder/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+import { Element } from "./styled";
+import { DndItemData } from "~/types";
+
+interface Props {
+    depth: number;
+    node: NodeModel<DndItemData>;
+}
+
+export const Placeholder = ({ depth }: Props) => {
+    const left = depth * 24 + 8;
```
adrians5j, line 15: Maybe we leave a comment here, explaining why the formula?

packages/app-folders/src/components/Tree/index.tsx
```diff
@@ -93,110 +30,42 @@ export const FolderTree: React.FC<Props> = ({
...
-                        dragPreviewRender={monitorProps => (
-                            <NodePreview monitorProps={monitorProps} />
-                        )}
-                        classes={{
-                            root: "treeRoot",
-                            dropTarget: "dropTarget"
-                        }}
-                        initialOpen={initialOpenList}
+            <Title title={title} onClick={onTitleClick} isDragging={isDragging} />
+            {!folders ? (
```
adrians5j: I'd say, for better readability, we do this above the `return` line.

Having ternary and then nested ternary operators is a bit hard to read.

packages/app-folders/src/components/Tree/Node/index.tsx
```diff
@@ -19,8 +20,29 @@ type Props = {
...
+export const Node = ({
+    node,
+    depth,
+    isOpen,
+    onToggle,
+    onClick,
+    onUpdateFolder,
+    onDeleteFolder
+}: Props) => {
     const indent = depth * 24 + 8;
```
adrians5j, line 48: Mentioned the not-commented formula in a prev comment.

packages/app-folders/README.md
```diff
@@ -32,32 +63,48 @@ import { FolderTree } from "@webiny/app-folders";
...
+<FolderDialogUpdate
+    folder={folder}
+    open={dialogOpen}
+    onClose={() => setDialogOpen(false)}
+/>
 ```
 
-As you might notice, there is not `listLinks` method available from `useLinks()` hook: this is because on first mount, `listLinks` is called internally, which will either issue a network request, or load links from cache.
+### FolderDialogDelete
+`FolderDialogUpdate` component shows a dialog to allow users to delete an existing folder.
```
adrians5j:
```suggestion
`FolderDialogDelete` component shows a dialog to allow users to delete an existing folder.
```

packages/app-folders/README.md
```diff
@@ -21,9 +21,40 @@ yarn add @webiny/app-folders
 ## Usage
```
adrians5j: Tip (for future GH docs): when writing GH docs, you can also use the format in this README:
https://github.com/webiny/webiny-js/tree/fix/folder-tree/packages/wcp
https://github.com/webiny/webiny-js/tree/fix/folder-tree/packages/api-wcp
https://github.com/webiny/webiny-js/tree/fix/folder-tree/packages/app-wcp

We [discussed this](https://www.notion.so/webiny/Proposal-Packages-Docs-Reference-f6d9ed56af9a4ae9921e86228bc2d202) a long time ago, and kinda said that we want to use it, but never actually started implementing it. :)

packages/app-folders/README.md
```diff
@@ -21,9 +21,40 @@ yarn add @webiny/app-folders
...
 Make sure `FoldersProvider` has been included within the admin app.
```
adrians5j:
```suggestion
Make sure `FoldersProvider` has been included within the Admin app.
```

packages/app-folders/src/components/Tree/List/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+    const [updateDialogOpen, setUpdateDialogOpen] = useState<boolean>(false);
+    const [deleteDialogOpen, setDeleteDialogOpen] = useState<boolean>(false);
+    const [selectedFolder, setSelectedFolder] = useState<FolderItem>();
+
+    useDeepCompareEffect(() => {
+        if (folders) {
+            setTreeData(createTreeData(folders, focusedFolderId));
+            setInitialOpenList(createInitialOpenList(folders, openFolderIds, focusedFolderId));
+        }
+    }, [Object.assign({}, folders), focusedFolderId]);
```
adrians5j: Hm, can you explain `Object.assign({}, folders)` here?

Also, if really needed, `[{ ...folders }, focusedFolderId]` might be a bit easier to read.

packages/app-folders/src/components/Tree/List/index.tsx
```diff
@@ -0,0 +1,214 @@
...
+                type,
+                createdOn,
+                createdBy,
+                isFocused: focusedNodeId === id
+            }
+        };
+    });
+};
+
+const createInitialOpenList = (
```
adrians5j: Can we move these outside of this file, into sibling files or a sibling `utils.ts` file?

adrians5j (changes requested):
A couple of nitpicks, nothing spectacular.

When it comes to UI testing...

1. Noticed folders disappeared when I clicked on a folder above. Should this happen? Shouldn't the expanded folders still be visible? I'd say yes, at least that's what I'm used to seeing in other tools.

https://user-images.githubusercontent.com/5121148/205629350-fdf5a290-6d6d-4007-8bf0-c63df927e515.mov

2. Also, let's update the message here:
image

...to: 
image

3. Can we have slug auto-generated here, just so the user doesn't have to type it all the time? 
image

4. Numbers should be allowed in folder name, no? Am I missing something? 🤔 
image

5. I think there's a bug here - folders disappear when clicking the edit button:

https://user-images.githubusercontent.com/5121148/205629232-b82fedac-2c2a-475f-a8b9-6eedd9b28280.mov

That's if for now I'd say. :) 
If you feel there's no time for some of the nitpicks, please then just write them down in Jira or somewhere.

packages/app-folders/README.md
leopuleo: 👍

packages/app-folders/src/components/Tree/List/index.tsx
leopuleo: Done it, thank you!

packages/app-folders/src/components/Tree/List/index.tsx
leopuleo: Yeah, that's enough 😄

packages/app-folders/src/components/Tree/Node/index.tsx
leopuleo: 👍

packages/app-folders/src/components/Tree/index.tsx
leopuleo: Done it, thank you!

packages/app-folders/src/components/Tree/Placeholder/index.tsx
leopuleo: Done it, thank you! 🙏

leopuleo:
Hi @adrians5j,
thanks for the review. About your comment here above: 

> 1. Noticed folders disappeared when I clicked on a folder above. Should this happen? Shouldn't the expanded folders still be visible? I'd say yes, at least that's what I'm used to seeing in other tools.

It should act as you describe, but I could not reproduce the bug 😢 

> 2. Also, let's update the message here

I agree with the changes, I added to this [subtask](https://webiny.atlassian.net/browse/WEB-1878).

> 3. Can we have slug auto-generated here, just so the user doesn't have to type it all the time?

It would be a nice feature, I added a [new subtask](https://webiny.atlassian.net/browse/WEB-1940) to address this.

> 4. Numbers should be allowed in folder name, no? Am I missing something? 🤔

Absolutely, this has been already fixed by this PR -> #2812 

> 5. I think there's a bug here - folders disappear when clicking the edit button:

Nice spot! I addressed the issue with this PR -> #2826

packages/app-folders/src/components/Tree/index.tsx
```diff
@@ -93,110 +30,50 @@ export const FolderTree: React.FC<Props> = ({
...
-            await updateFolder({
-                ...item,
-                parentId: dropTargetId !== ROOT_ID ? (dropTargetId as string) : null
-            });
-        } catch (error) {
-            return showSnackbar(error.message);
+    const renderList = () => {
+        if (!folders) {
+            return <Loader />;
+        } else if (folders.length > 0) {
```
adrians5j: No need for `else if` here. We can do the following:

```ts
if (!folders) {
   return <Loader />;
} 

if (folders.length > 0) { ...
```

Makes the code a bit more readable. https://www.samanthaming.com/tidbits/23-no-else-return/

packages/app-folders/src/components/Tree/List/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+    const [updateDialogOpen, setUpdateDialogOpen] = useState<boolean>(false);
+    const [deleteDialogOpen, setDeleteDialogOpen] = useState<boolean>(false);
+    const [selectedFolder, setSelectedFolder] = useState<FolderItem>();
+
+    useDeepCompareEffect(() => {
+        if (folders) {
+            setTreeData(createTreeData(folders, focusedFolderId));
+            setInitialOpenList(createInitialOpenList(folders, openFolderIds, focusedFolderId));
+        }
+    }, [{ ...folders }, focusedFolderId]);
```
adrians5j, line 65: Looks nicer, but I still have a question: why do we need to spread here? I don't see the reason to do it, since we're doing deep compare. Maybe I'm missing smt here? 🤔

adrians5j (changes requested):
Two minor things. 

All other things looking good. 

Regarding issue 1 - could not repro that one either.

Thank you! :)

packages/app-folders/src/components/Tree/index.tsx
leopuleo: Definitely more readable 😄

packages/app-folders/src/components/Tree/List/index.tsx
leopuleo: I'm adding a comment to explain it: this need more investigation in order to remove the object spreading

adrians5j (approved):
Thanks!

PR #2827, opened 2022-12-06
feat: add rich text modal for block variable

neatbyte-vnobis:
## Changes
Implements issue "Add ability to expand the rich text editor from the sidebar"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/components/MediumEditor/index.ts
+++ b/packages/app-page-builder/src/editor/components/MediumEditor/index.ts
@@ -17,7 +17,7 @@ interface ReactMediumEditorProps {
     onSelect: (value: string) => void;
     tag: string | [string, Record<string, any>];
     options?: CoreOptions;
-    forceFocus?: boolean;
+    autoFocus?: boolean;
     [key: string]: any;
 }
 
@@ -27,7 +27,7 @@ const ReactMediumEditor: React.FC<ReactMediumEditorProps> = ({
     onChange,
     options = {},
     onSelect,
-    forceFocus
+    autoFocus
 }) => {
     const elementRef = React.useRef<HTMLElement>(null);
     const editorRef = React.useRef<MediumEditor.MediumEditor>();
@@ -109,7 +109,7 @@ const ReactMediumEditor: React.FC<ReactMediumEditorProps> = ({
     }, [handleChange, handleSelect, tagName]);
 
     useEffect(() => {
-        if (forceFocus && editorRef.current && elementRef.current) {
+        if (autoFocus && editorRef.current && elementRef.current) {
             // We need time to finish initialization of Medium Editor
             setTimeout(() => {
                 // Approach was taken from: https://github.com/yabwe/medium-editor/issues/850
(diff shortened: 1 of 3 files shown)


SvenAlHamad:
@neatbyte-vnobis - on this one, let's make two adjustments:
1. Let's still keep the ability to edit the paragraph directly in the sidebar, just add an "expand icon" to the variable name, which then opens the modal. Thing is, for shorter text editing it directly in the sidebar is handy. 
image

2. Let's limit the vertical length of the paragraph input in the sidebar so that it can't expand beyond a certain point, say 250px. I was able to achieve this by adding this css to the input field:
```
    max-height: 250px;
    overflow: scroll;
```

SvenAlHamad:
@neatbyte-vnobis - this one looks good, just one minor thing. When I open the modal view, is it possible that I'm automatically focused on the text edit field, so that I don't need to explicitly click on it with the mouse, before I can start typing. It's not a crucial feature, but if it's a quick fix, I would like to get it implemented. If it's anything more complicated, let's leave it out for now.

packages/app-page-builder/src/editor/components/MediumEditor/index.ts
```diff
@@ -17,6 +17,7 @@ interface ReactMediumEditorProps {
     onSelect: (value: string) => void;
     tag: string | [string, Record<string, any>];
     options?: CoreOptions;
+    forceFocus?: boolean;
```
Pavel910: Please rename this to `autoFocus`, as we already use that prop name in other components. Just to be consistent.

packages/app-page-builder/src/editor/plugins/elementSettings/variable/RichVariableInput.tsx
```diff
@@ -1,23 +1,64 @@
...
+    height: 400px;
+    overflow-y: auto;
+    line-height: normal;
+
+    &:has(p[data-medium-focused]) {
+        background-color: rgba(212, 212, 212, 0.7);
+    }
+
+    & .medium-editor-placeholder:after {
+        color: ${"var(--mdc-theme-text-secondary-on-background)"};
```
Pavel910: No need for `${}` around `var()`.

packages/app-page-builder/src/editor/plugins/elementSettings/variable/RichVariableInput.tsx
```diff
@@ -1,23 +1,64 @@
-import React, { useState } from "react";
+import React, { useCallback, useState } from "react";
 import styled from "@emotion/styled";
 import { CoreOptions } from "medium-editor";
+import { ReactComponent as ExpandIcon } from "~/admin/assets/expand.svg";
```
Pavel910: Please use the `@material-design-icons/svg` package for icons. It contains all icons from https://fonts.google.com/icons?selected=Material+Icons so we don't have to manually download SVGs each time, plus it saves on the bundle size. I discovered that package a few months ago and we're now switching to that across the entire admin app.

Pavel910 (changes requested):
A few minor changes and we're done here! 👍

Pavel910 (approved):
Thank you!

SvenAlHamad:
Also, all good from my side on this one.

PR #2844, opened 2022-12-08
feat: import and export blocks

neatbyte-vnobis:
## Changes
Implement feature "Add the ability to import export blocks"

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-vnobis:
@Pavel910 please review this pull

Pavel910 (changes requested):
@neatbyte-vnobis this looks good, I know it can be done better with more time (like making reusable components, etc.) but this is great for the time being. It's exactly what we need at this point in time, so no need to nitpick.

One thing is necessary to complete this PR: the changes you did to imports in the `apps/api`...
CleanShot 2022-12-20 at 15 59 14

...also need to be propagated to the `serverless-cms-aws` package:
CleanShot 2022-12-20 at 16 01 48

These handlers are pre-built when we publish packages to NPM, so in the user's projects, these are being pulled in from node_modules. Keep in mind that there are 2 versions of these handlers: `ddb` and `ddb-es`, so please use the appropriate File Manager storage operations.

A legitimate question is: `How do I test the ES version locally?` - you don't. We'll test it when everything is merged into the `next` branch. Basically everything is the same for both versions, except the storage operations plugins, so you should be quite confident by just importing the correct storage ops package. Check this as a reference: `packages/cwp-template-aws/template/ddb-es/apps/api/graphql/src/index.ts:24`

Pavel910:
@SvenAlHamad please do the UI testing when you can, as usual. Thanks a lot! 🙌

SvenAlHamad:
I'm good with this implementation for V1. For future versions we'll make some updates like the ability to select which blocks to import, as well as detect if an asset already exists, so we don't create duplicates.

Pavel910 (approved):

PR #2849, opened 2022-12-12
feat: remove unnecessary where-parser package

brunozoric:
## Changes
Remove unused and unnecessary where-parser package.


Pavel910 (approved):
So nice to see code go away 😂

PR #2861, opened 2022-12-15
fix: update and listen for page cache changes

leopuleo:
## Changes
With this PR we aim to update table list data after the user edit/publish/unpublish an entry by:

- adding Apollo `client.watchQuery()` in order to listen for cache changes;
- updating GET_PAGE graphQL cache on new revision creation.

Closes [WEB-1919](https://webiny.atlassian.net/browse/WEB-1919)


Pavel910 (approved):
Turned out pretty decent 👍

PR #2864, opened 2022-12-16
Headless Pages - UI Improvements

Pavel910:
## Changes
This PR introduces various improvements of the `ui` package, the most notable being the following:
- Accordion component now supports actions, and can be switched to non-interactive mode (this means it an only be toggled via code).
- Button, Switch, and Input components received minor types improvements.
- Individual `Tab` instances within the `Tabs` component can now be hidden/shown via the `visible` prop. This helps fix the problem with mounting/unmounting of tabs, in which case the order of tabs can change.
- Dialog action button now has a proper margin so the buttons are not stuck together.
- Tags component can now have protected tags; protected tags cannot be removed by the user.
- version of `minimatch` package was updated.

## How Has This Been Tested?
Manually.


leopuleo (approved):
👍

PR #2867, opened 2022-12-16
feat: improve blocks search functionality

neatbyte-vnobis:
## Changes
Implement task "Improve the blocks search functionality"
Includes a fix for not saved block name (regression caused by `isDirty` implementation)

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
+++ b/packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
@@ -74,10 +74,15 @@ const SORTERS: Sorter[] = [
 ];
 
 type PageBuilderBlocksByCategoriesDataListProps = {
+    filter: string;
+    setFilter: React.Dispatch<React.SetStateAction<string>>;
     canCreate: boolean;
 };
-const BlocksByCategoriesDataList = ({ canCreate }: PageBuilderBlocksByCategoriesDataListProps) => {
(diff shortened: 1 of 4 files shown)


SvenAlHamad:
I did a quick test on this one and it works fine. 
@Pavel910 - please check with regards to the API side and pagination if we need to increase any limits, when we say have 1000+ blocks, as all the search is done on client side so all the data needs to be available on the client to be searchable.

Pavel910:
@SvenAlHamad it's all good already, the API will pull all blocks from the database. I think for v1 this is good enough. Later if we see performance issues, we can transfer filtering/sorting to the API side, and introduce pagination. But in reality, if you're using DDB-only setup, it's actually cheaper to load everything once, and filter on the client, because the API still needs to load everything to do filtering/sorting.

Pavel910 (approved):
@neatbyte-vnobis Thanks! Filtering works as expected, and we'll leave this view as one of the first candidates for refactoring to reduce coupling of logic with presentation. For now, this is acceptable, as the entire system is written like that 🤷.

adrians5j:
@neatbyte-vnobis Just a quick ping that this PR has conflicts, and should be resolved before merging.

Thank you!

neatbyte-vnobis:
@adrians5j 
please take a look on new pull https://github.com/webiny/webiny-js/pull/2968 (based on up to date "next" branch and tested with latest changes)
this pull is closed (as it was based on too "old"/prerelease "next" branch)

PR #2870, opened 2022-12-16
Dynamic Zone Field

Pavel910:
## Changes
This PR introduces the "Dynamic Zone" field.

## How Has This Been Tested?
Manually.

## Documentation
User guides will be written once the feature is finalized and remains stable.


packages/api-headless-cms/src/graphqlFields/dynamicZone.ts
```diff
@@ -0,0 +1,200 @@
+import {
+    CmsEntry,
+    CmsContext,
+    CmsModelFieldToGraphQLPlugin,
+    CmsModel,
+    CmsModelField
+} from "~/types";
+import { createReadTypeName } from "~/utils/createTypeName";
+// import { createGraphQLInputField } from "./helpers";
```
leopuleo, line 9: We can remove this if we don't need it

packages/app-headless-cms/src/admin/components/Droppable.tsx
```diff
@@ -41,7 +42,11 @@ const DroppableComponent: React.FC<DroppableProps> = props => {
             isOver: monitor.isOver() && monitor.isOver({ shallow: true }),
             item: monitor.getItem()
         }),
-        drop(_, monitor) {
+        drop(item, monitor) {
+            if (typeof props.isDroppable === "function" && !props.isDroppable(item)) {
```
leopuleo, line 46: Why are we checking if `isDroppable` is a function?

leopuleo (changes requested):
Just a couple of minor fixes, thanks!
Also, I'm not sure why some actions are failing (see here below)...might be related to depending on code changes splitted across multiple PRs?

packages/api-headless-cms/src/graphqlFields/dynamicZone.ts
Pavel910: This file is WIP, the rest will be implemented in a different PR, focused only on the API side of things. The entire `read` API needs proper implementation, so I left this here.

packages/app-headless-cms/src/admin/components/Droppable.tsx
Pavel910: @leopuleo `isDroppable` is an [optional prop](https://github.com/webiny/webiny-js/pull/2870/files#diff-29d2267d0891398bbd83852cc01ef68e9572c83842969b7f9f3599ef2b4f9d33R31), so we need to make sure it's a function, before invoking it.

leopuleo: 👍

leopuleo (approved):

PR #2873, opened 2022-12-18
feat: add element id and class validation

neatbyte-vnobis:
## Changes
Implement improvement "Element ID & CSS field validator"

## How Has This Been Tested?
Manual

## Documentation
None


Pavel910 (approved):
@neatbyte-vnobis even though this works, I would recommend looking at this plugin for example: `packages/app-page-builder/src/editor/plugins/elementSettings/height/HeightSettings.tsx`

It utilizes proper form/bind/validators concepts, and it's a lot easier to write validator functions as you'll see, as simple callables that throw regular errors.

We never use `import { Validation } from "@webiny/form";` directly, consider this to be form internals, and that's what the Bind component passes to your UI components after performing validation.

I'll approve this PR, but if you would like to revisit this and implement it as we do in the other plugins, that would be great. Thanks!

Pavel910 (approved):

PR #2875, opened 2022-12-18
fix: default block category icon state fix

neatbyte-vnobis:
## Changes
Fix issue "Bug: Default icon pre-defined, but value still required"
(added initial value matches default picker image)

## How Has This Been Tested?
Manual

## Documentation
None


Pavel910 (changes requested):
@neatbyte-vnobis Let's try making this more robust. Try using the `defaultValue` prop on `Bind`, like this:

CleanShot 2022-12-20 at 16 42 48

I think this should work, and it's just always there by default.

Pavel910 (approved):

PR #2876, opened 2022-12-18
fix: prevent custom scrips from executing on prerender side

neatbyte-vnobis:
## Changes
Implement fix for "Bug: Custom scripts get executed twice"

@Pavel910 I thought that `__PS_RENDER__` check need to be added by user manually in their own script... 
I can add an `if` condition and prevent a prerender execution for a `<script>...</script>`
But it will not work for "imported" external scripts like a `<script src="some_external_script" />` (as these will not be wrapped by `if` check)

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/render/components/WebsiteScripts.tsx
+++ b/packages/app-page-builder/src/render/components/WebsiteScripts.tsx
@@ -1,6 +1,19 @@
 import React, { useEffect, useMemo } from "react";
 import { Helmet } from "react-helmet";
 
+const wrapWithRenderCondition = (innerHTML: string) => {
+    if (!innerHTML) {
+        return "";
+    }
+
+    return (
+        `// This condition prevents script execution on prerendering.\n` +
+        `if (!("__PS_RENDER__" in window)) { // This line added automatically by Webiny\n` +
+        `    ${innerHTML}\n` +
+        `} // This line added automatically by Webiny\n`
+    );
+};
+
 const parseHeaderTags = (str?: string) => {
     if (!str) {
         return [];
@@ -18,7 +31,7 @@ const parseHeaderTags = (str?: string) => {
         const newElement = React.createElement(
             nodeName,
             { ...attributes, key: index },
-            el.innerHTML
+            wrapWithRenderCondition(el.innerHTML)
         );
 
         parsedElementsArray.push(newElement);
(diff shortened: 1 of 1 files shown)


packages/app-page-builder/src/render/components/WebsiteScripts.tsx
```diff
@@ -1,6 +1,19 @@
 import React, { useEffect, useMemo } from "react";
 import { Helmet } from "react-helmet";
 
+const wrapWithRenderCondition = (innerHTML: string) => {
```
adrians5j: Curious about one thing here...

Couldn't we _not_ do this, and, instead, at the beginning of the `WebsiteScripts` component, just do the `if (!("__PS_RENDER__" in window))` check and bail if so?

So:

```tsx
const WebsiteScripts: React.FC<WebsiteScriptsProps> = ({ headerTags, footerTags }) => {
    // `WebsiteScripts` component doesn't do anything if we're
    // in the middle of a prerendering process.
    if ("__PS_RENDER__" in window) {
        return null;
    }
    
    const htmlHeadTags = useMemo(() => {
        return parseHeaderTags(headerTags);
    }, [headerTags]);
    useEffect(() => {
        const htmlFooterTags = parseFooterTags(footerTags);
        appendToBody(htmlFooterTags);
        return () => {
            removeFromBody(htmlFooterTags);
        };
    }, [footerTags]);
    return <Helmet>{htmlHeadTags?.map(element => element)}</Helmet>;
};
```

adrians5j (changes requested):
Left one comment. Just curious about one thing, might be a simpler solution.

packages/app-page-builder/src/render/components/WebsiteScripts.tsx
neatbyte-vnobis: If I understand rendering process right - no
As I see it, on prerender we are getting a final HTML which is saved and served latter as a static html file (without react).
So we need to add these scripts on prerender, but we cannot allow them to be executed (as it causes a DOM changes which are not intended).

On other hand we need them to be executable on preview page (so these can be tested before publishing)...
We may consider smth like a POST prerender html manipulation to add these scripts... but it will change a current cms flow

Pavel910: @neatbyte-vnobis the HTML is hydrated after being served to the browser, and React starts again.

adrians5j: Initially, I wasn't sure if there was some additional website scripts-related logic involved that I wasn't aware of. 

But, from what I can see now, there isn't. So, the above suggestion should work fine and the code can be simplified (`wrapWithRenderCondition` can be removed).

adrians5j (approved):

adrians5j:
Looks good @neatbyte-vnobis, merging.

Thanks!

PR #2877, opened 2022-12-19
fix(api-headless-cms): generate ref union types recursively

Pavel910:
## Changes
This PR solves an issue of generating `ref` field union types (when `ref` field accepts multiple content models) when nested within an `object` field. Essentially, the logic for union type generation was moved from the global `createSchema`, to a `createTypeField` function, because this one is being executed recursively by parent fields, so we will end up having all our union types generated for any depth.

I've also added schema error handling with the output of the invalid part of schema, so it's easier to locate syntax errors in the `headless-cms` Lambda function.

## How Has This Been Tested?
Jest tests.


packages/api-headless-cms/src/graphql/graphQLHandlerFactory.ts
```diff
@@ -118,14 +134,29 @@ const cmsRoutes = new RoutePlugin<CmsContext>(({ onPost, onOptions, context }) =
...
+                    data: null,
+                    error: {
+                        message: ex.message,
+                        code: ex.code,
+                        data: ex.data
+                    }
+                });
+            }
+
+            return reply.code(500).send();
```
Pavel910: Send the original error.

Pavel910:
This got branched via a different PR.

PR #2879, opened 2022-12-19
fix: empty folder render and loading flickering

leopuleo:
## Changes
With this PR we are fixing issues related to loading and empty state within the folders and pages list:

- fix: flickering page list showing the empty state while loading data
- fix: empty links list was not triggering the page details fetch.
- refactor: simplify `loading` by removing the bound with the current `folderId` or folder `type`
- refactor: shorten the loading enum, eg `LIST_FOLDERS` / `LIST_LINKS` -> `LIST`
- refactor: improve loading check within the list
- refactor: better handling of pages details fetch

### DIFF
--- a/packages/app-folders/src/contexts/links.tsx
+++ b/packages/app-folders/src/contexts/links.tsx
@@ -41,7 +41,7 @@ interface Props {
 }
 
 const defaultLoading = {
-    IDLE: true,
+    INIT: true,
     LIST: false,
     LIST_MORE: false,
     CREATE: false,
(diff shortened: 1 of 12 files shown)


packages/app-folders/src/contexts/links.tsx
```diff
@@ -40,10 +40,19 @@ interface Props {
...
 
+const defaultLoading = {
+    INIT: true,
+    LIST: false,
+    LIST_MORE: false,
+    CREATE: false,
+    UPDATE: false,
+    DELETE: false
+};
+
```
Pavel910: I think you should type this using the `LoadingActions` type, so that the keys are type-safe.

packages/app-folders/src/handlers.tsx
```diff
@@ -10,40 +10,33 @@ import { Loading } from "~/types";
...
-
-        return {
-            ...state,
-            [context]: {
-                ...currentContext,
-                [action]: !currentAction
-            }
-        };
-    });
+) => {
```
Pavel910, line 33: Also add use `LoadingActions` as the `action` type here. That way you have autocomplete, and type safety.

Pavel910 (changes requested):
Some minor type checks and we're good :)

Pavel910 (approved):

PR #2884, opened 2022-12-20
feat: add element toolbar changes

neatbyte-vnobis:
## Changes
Implements issues "Adjust the UI of the element list in the left sidebar" and "Click to add vs drag to add"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/components/Editor/Toolbar.tsx
+++ b/packages/app-page-builder/src/editor/components/Editor/Toolbar.tsx
@@ -51,8 +51,9 @@ const drawerStyle = css({
         marginLeft: 54,
         position: "fixed",
         top: 64,
-        width: "490px !important",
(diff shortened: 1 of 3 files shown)


Pavel910 (changes requested):
From the code side, the changes are ok.

Functionally, when `Click to add` is clicked, the toolbar should close, but it stays open for some reason.

I would also ask @SvenAlHamad to have a look at the visual implementation, as I'm not the right person to give stylistic advices 😅

SvenAlHamad:
Just tested this one....I would like to make a small change. Instead of having a width of 320px on the sidebar, let's make it 280px so we save even more space. Other than that, all good.

SvenAlHamad (approved):

Pavel910 (approved):

PR #2890, opened 2022-12-22
feat: add the ability to duplicate a page

neatbyte-vnobis:
## Changes
Resolves issue "Add the ability to duplicate a page"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-page-builder/src/graphql/graphql/pages.gql.ts
+++ b/packages/api-page-builder/src/graphql/graphql/pages.gql.ts
@@ -236,6 +236,9 @@ const createBasePageGraphQL = (): GraphQLSchemaPlugin<PbContext> => {
                     # Update page by given ID.
                     updatePage(id: ID!, data: PbUpdatePageInput!): PbPageResponse
 
+                    # Duplicate page by given ID.
+                    duplicatePage(id: ID!): PbPageResponse
+
                     # Publish page
                     publishPage(id: ID!): PbPageResponse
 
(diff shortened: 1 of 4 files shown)


SvenAlHamad:
@neatbyte-vnobis  - on this one, two minor adjustments:
1. Instead of having "Copy" as the label in the drop down menu, please say "Duplicate"
2. When a duplicated page is created, please select it so that the duplicated page selected in the left menu and its preview is show in the right preview bar

SvenAlHamad:
This one is good from my end. @Pavel910 - please check how the code implementation looks.

Pavel910 (changes requested):
@neatbyte-vnobis This solution is problematic  because the `settings` addition to the GraphQL queries is hardcoded, but in reality those settings can be extended with plugins, which means that you will lose information by implementing it this way.

It would be far better to move this whole logic to the API side, and add a `duplicatePage` mutation. It will be cleaner, and easier to handle. Plus, other developers would be able to use this new method programmatically. On the React side, you will just need to call that mutation, and when you get the data back, update the Apollo cache as you're already doing now.

neatbyte-vnobis:
@Pavel910 please review fix for your comment

packages/api-page-builder/src/graphql/graphql/pages.gql.ts
```diff
@@ -452,6 +455,31 @@ const createBasePageGraphQL = (): GraphQLSchemaPlugin<PbContext> => {
...
 
+                    duplicatePage: async (_, args: any, context) => {
+                        try {
+                            const page = await context.pageBuilder.getPage(args.id);
+                            const { id: duplicatedPageId } = await context.pageBuilder.createPage(
+                                page.category
+                            );
+
+                            const duplicatedPageData = {
+                                title: `(Copy) ${page.title}`,
```
adrians5j:
```suggestion
                                title: `${page.title} (Copy) `,
```

packages/app-page-builder/src/admin/plugins/pageDetails/header/pageOptionsMenu/PageOptionsMenu.tsx
```diff
@@ -69,6 +81,45 @@ const PageOptionsMenu: React.FC<PageOptionsMenuProps> = props => {
...
+            await client.mutate({
+                mutation: DUPLICATE_PAGE,
+                variables: { id: page.id },
+                update(cache, { data }) {
+                    if (data.pageBuilder.duplicatePage.error) {
+                        return;
+                    }
+
+                    GQLCache.addPageToListCache(cache, data.pageBuilder.duplicatePage.data);
+                    showSnackbar(`Duplicated "${page.title}".`);
```
adrians5j:
```suggestion
                    showSnackbar(`The page "${page.title}" was duplicated successfully.`);
```

adrians5j (changes requested):
A couple of tiny nit picks, but all in all we're good.
Let's just take care of these and we're good to merge.

Thank you!

adrians5j (approved):

PR #2895, opened 2022-12-24
Introducing App Aliases

adrians5j:
## Changes
With this PR, we're introducing app aliases. In other words, upon running CLI commands, instead of users having to type `apps/api` all the time, now they can just type `api`. 

```yarn webiny deploy api --env dev```

This is possible via the newly introduced `appAliases` param that's now available in the `webiny.project.ts` config file:

image

This change will not only make it easier for users to type the commands, but will also easen the communication.

All new projects will receive this configuration out of the box. Existing projects will receive these once they upgrade.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


leopuleo (approved):
👍

PR #2896, opened 2022-12-24
Enable Defining Environment Variables via `webiny.project.ts`

adrians5j:
## Changes
In some cases, users might want to define environment variables "more statically", in a config file that's pushed to git and doesn't contain secrets or other sensitive info. For example, feature flags or similar values that set up the default app behaviour.

This is now possible via the `webiny.project.ts` file, via exported config's `env` property:

image

## How Has This Been Tested?
Manually.

## Documentation
Will update https://www.webiny.com/docs/core-development-concepts/basics/environment-variables.

### DIFF
--- a/packages/cli/cli.js
+++ b/packages/cli/cli.js
@@ -1,32 +1,11 @@
 #!/usr/bin/env node
-const path = require("path");
 const yargs = require("yargs");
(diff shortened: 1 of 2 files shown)


leopuleo (approved):
Make sense! Thanks @adrians5j  👍

PR #2897, opened 2022-12-24
Introduce Feature Flags

adrians5j:
## Changes
With this PR, users are now able to use their `webiny.project.ts` file to set feature flags. For example:

image

Once defined, users can access these values in their application code easily via the new `@webiny/feature-flags` package like so:

image

The above approach works both on the backend and frontend.

## How Has This Been Tested?
Manually.

## Documentation
README and changelog.


leopuleo (approved):
Great! Thanks @adrians5j

PR #2941, opened 2023-01-12
fix: block title update fix

neatbyte-vnobis:
## Changes
Closes "Bug: When you edit an existing block the block title changes to “new block”"

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
I tested this one, and it's working as expected. Thanks for sorting it out!

Pavel910 (approved):

PR #2950, opened 2023-01-16
feat: add an indicator that the form field is required

neatbyte-vnobis:
## Changes
Implement task "Add an indicator that the field is required (both in the editor and in the UI)"

## How Has This Been Tested?
Manual

## Documentation
None


packages/cwp-template-aws/template/common/apps/theme/layouts/forms/DefaultFormLayout/fields/Textarea.tsx
```diff
@@ -36,7 +36,11 @@ export const TextareaField: React.FC<TextareaProps> = ({ field }) => {
 
     return (
         <Field>
-            <FieldLabel>{field.label}</FieldLabel>
+            <FieldLabel
```
adrians5j: Let's refactor this a bit.

Instead of the `required` flag, let's rather have `field` flag, which accepts the whole `field`. We can even get rid of the child label. Ultimately, we end up with:

```
<FieldLabel field={field}/>
```

That way we wouldn't need to pass this `required` value for every field, but the flag would be calculated internally in the `FieldLabel` component.

apps/theme/layouts/forms/DefaultFormLayout/fields/components/FieldLabel.tsx
```diff
@@ -1,11 +1,35 @@
+import * as React from "react";
 import styled from "@emotion/styled";
+import { ReactComponent as Asterisk } from "@material-symbols/svg-400/outlined/emergency.svg";
```
adrians5j: Can we use the regular `*` character here? Instead of adding a new dependency to the `theme` package? 

The less deps, the better, especially when it comes to public website-related code (`website` and `theme` packages).

The thing is - we want to keep the `theme` and `website` folders as light as possible. I get that this won't probaby

apps/theme/layouts/forms/DefaultFormLayout/fields/components/FieldLabel.tsx
```diff
@@ -1,11 +1,35 @@
...
     margin: 0 0 5px 1px;
+
     ${theme.breakpoints.mobile} {
         text-align: left !important;
     }
+
+    svg {
+        width: 10px;
+        height: 10px;
+        fill var(--mdc-theme-primary, #fa5723);
```
adrians5j: We don't want to use the SCSS vars anymore.

For all theme-related values, please rely on the `theme.ts` file. In this file, there's an example of accessing theme-defined breakpoints (`theme.breakpoints.mobile`).

packages/cwp-template-aws/template/common/apps/theme/layouts/forms/DefaultFormLayout/TermsOfServiceSection.tsx
```diff
@@ -19,7 +19,7 @@ interface Props {
     component: TermsOfServiceComponent;
 }
 
-const RteFieldLabel = styled(FieldLabel)`
+const RteFieldLabel = styled(FieldLabelStyled)`
```
adrians5j, line 22: Should we not use `FieldLabel` here instead of `FieldLabelStyled` ?

packages/cwp-template-aws/template/common/apps/theme/layouts/forms/DefaultFormLayout/fields/components/FieldLabel.tsx
```diff
@@ -1,12 +1,35 @@
...
     margin: 0 0 5px 1px;
 
     ${breakpoints["mobile-landscape"]} {
         text-align: left !important;
     }
+
+    svg {
+        width: 10px;
+        height: 10px;
+        fill var(--mdc-theme-primary, #fa5723);
```
adrians5j: Same comments apply here (deps, theme).

apps/theme/layouts/forms/DefaultFormLayout/TermsOfServiceSection.tsx
```diff
@@ -6,7 +6,7 @@ import { useBind } from "@webiny/form";
 import { Field } from "./fields/components/Field";
 import { Row } from "./Row";
 import { Cell } from "./Cell";
-import { FieldLabel } from "./fields/components/FieldLabel";
+import { FieldLabelStyled } from "./fields/components/FieldLabel";
```
adrians5j, line 9: Same question about `FieldLabel` vs. `FieldLabelStyled` applies here.

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/Field.tsx
```diff
@@ -29,26 +30,70 @@ const Actions = styled("div")({
     alignItems: "right",
     "> *": {
         flex: "1 100%"
+    },
+    ".switch-wrapper": {
+        display: "flex",
+        alignItems: "center",
+        color: "var(--mdc-theme-text-secondary-on-background)",
```
adrians5j, line 37: Just extra clarification here.

The usage of SCSS vars here is OK.

This is because these are Admin app styles, and these are still handled via SCSS.

Only Admin app styles and theme vars are handled via SCSS. The rest (Page Builder pages, Form Builder forms) is handled via the new `theme.ts` file.

If you'll have any extra questions, feel free to ping.

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/Field.tsx
```diff
@@ -29,26 +30,70 @@ const Actions = styled("div")({
...
-    const { getFieldPlugin } = useFormEditor();
-
+    const { getFieldPlugin, updateField } = useFormEditor();
     const fieldPlugin = getFieldPlugin({ name: field.name });
+
+    const isRequired = useMemo(() => {
+        return field?.validation?.some(validation => validation.name === "required");
+    }, [field]);
+
+    const handleRequiredStatusChange = useCallback(() => {
```
adrians5j:
```suggestion
    const toggleRequiredValidator = useCallback(() => {
```

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/Field.tsx
```diff
@@ -29,26 +30,70 @@ const Actions = styled("div")({
...
 const Field: React.FC<FieldProps> = props => {
     const { field, onEdit, onDelete } = props;
-    const { getFieldPlugin } = useFormEditor();
-
+    const { getFieldPlugin, updateField } = useFormEditor();
     const fieldPlugin = getFieldPlugin({ name: field.name });
+
+    const isRequired = useMemo(() => {
+        return field?.validation?.some(validation => validation.name === "required");
+    }, [field]);
```
adrians5j: I'd say `useMemo` usage here is not worth it for a couple of reasons, primarily because of the fact the the `field` reference is probably always different on each render, so the recalculation will happen. And even if the ref did not change, the `[field]` would then produce incorrect results in the UI, since that's not the value we want to use as a caching condition. 

This will suffice:

```ts
const isRequired = field?.validation?.some(validation => validation.name === "required");
```

adrians5j (changes requested):
A couple of things to check. 

Since there are duplicate files (because of the CWP template), I think it should not be to much of a hassle. 

Thank you!

SvenAlHamad:
I've tested the feature from my side, and I'm good with how it looks like.

adrians5j:
Let's resolve conflicts before moving on with another review. Thanks! @neatbyte-vnobis

adrians5j (approved):
This looks good @neatbyte-vnobis, thank you!

PR #2951, opened 2023-01-16
fix: fix unreadable values in the webhook

neatbyte-vnobis:
## Changes
Fix issue "Radio and checkbox filed values in the webhook end file export use the value field instead of the label which makes them unreadable"

## How Has This Been Tested?
Manual

## Documentation
None


adrians5j (changes requested):
Let's resolve conflicts first, so that we don't need to do another pass once they're resolved.

Thanks!

adrians5j:
Let's resolve conflicts before moving on with another review. Thanks! @neatbyte-vnobis

packages/api-form-builder/src/plugins/crud/submissions.crud.ts
```diff
@@ -202,12 +201,7 @@ export const createSubmissionsCrud = (params: CreateSubmissionsCrudParams): Subm
                 context.plugins.byType<FbFormFieldValidatorPlugin>("fb-form-field-validator");
             const { fields } = form;
 
-            const data = pick(
-                rawData,
-                fields.map(field => field.fieldId)
-            );
-
-            if (Object.keys(data).length === 0) {
+            if (Object.keys(rawData).length === 0) {
```
adrians5j, line 204: Can you explain this removal of `data` const / `lodash.pick` call? Why are we doing this change?

Not sure if we're gonna break anything with the removal (too bad no comments were left here initially). To keep it safe, I'd keep it as is, but maybe you have more insight here.

adrians5j: I just double checked my assumption, and yeah, sometimes the frontend can include other non-field pieces of data, for example: 
image

So, yeah, I'd keep the data sanitization here, would not introduce any changes.

packages/app-form-builder/src/components/Form/FormRender.tsx
```diff
@@ -159,6 +159,28 @@ const FormRender: React.FC<FbFormRenderComponentProps> = props => {
             };
         }
 
+        for (const FieldId in data) {
```
adrians5j, line 162: Let's have this code on the backend and not on the frontend. 

We can implement this extra logic in the backend's form submissions / webhooks area.

image

We could do data sanitization before sending it to triggers: 

```
packages/api-form-builder/src/plugins/crud/submissions.crud.ts:328
```

adrians5j: Also, maybe we'd want this `sanitizeFormSubmissionData` utility. We'd use it here and also with CSV imports. I saw @SvenAlHamad mentioned those as well in the initial doc. We can add the fn here: `packages/api-form-builder/src/plugins/crud/utils.ts`.

adrians5j (changes requested):
Left some comments, please check them out when you get a chance @neatbyte-vnobis.

Thank you! 🍻

neatbyte-vnobis:
@adrians5j @SvenAlHamad 
there are a lot of changes in #2957 and #2952 related to webhook logic and CSV file logic.
So we need to merge them first before updating this task (otherwise here will be conflicts, and need to retest it).

neatbyte-vnobis:
@SvenAlHamad @adrians5j
While looking on this PR and task again I see a place which worth to be confirmed:
image
Here you can see JSON (left side) with added `checkbox_label`, `select_label` and `radio_label` (as per task description).

But for CSV (right side) we have all column names that are using a "Display value" from form fields.
If I will add `_label` it will look so-so...

Maybe for CSV we need only to change row values (aka from `option1` to `Option 1`) and not add new columns?
But I am not sure if it is ok for backward compatibility.

Also on UI we have a Form Submission modal with a "copy as CSV" button:
image
here we have a UI JS to generate CSV, but for "export all form submissions" we are using an API to get the CSV file... (same lib `json2csv` is used, but code is different)
It looks not good as we have 2 places to manage CSV...

I can add API to generate CSV for one submission (by id).
But I am not sure if it worth do it on API side.
Maybe we can have some shared code to generate CSV?
But for "export all form submissions" we can have more columns (old deleted columns from old submission, all revisions are processed there), so code is much more complex there.

I plan to keep simple one submission as CSV on UI side as is now, but if you have any comments please let me know.

adrians5j:
Sorry @neatbyte-vnobis, let me get this straight.

#### 1. JSON/CSV
So, when comparing JSON/CSV in 1st sshot, you're saying that it might be better to just have a single column in the table for these special fields? 

For example, instead of having both `Select` (value: `someValue`) and `Select_label` (value: `Some value`), you're saying it might be better to just have `Select` column and human-readable value as `Some value`, right?

If so, I don't think it's a problem to include both. Also, in the future, who knows, we might add a feature where user will be able to decide which columns are to be included in the export. Let's also see what @SvenAlHamad thinks about this.

#### 2. Code
Leave it for now. It's two places at this moment, if we start reusing this more often, then we'll refactor. This way we'll also save some time. But thanks for your observation, much appreciated!

neatbyte-vnobis:
@adrians5j @SvenAlHamad 
Regarding **1. JSON/CSV**:
It is weird that in CSV we are using "display name" for columns (aka "Some value" with spaces and upper letters), but rows values are using a "value field" (aka "option1", not human readable labels).

We may modify existing columns to use "label" value (aka "Option 1") (but what about backward compatibility???).
Or we can add additional column.... But how to name it? (as we already using "display name" for a column name).
Add " (Label)" to the "display name? (aka 'Country Select (Label)").

### It is easier to understand with example (just in case so we all have same understanding):
For example we have a select with:
1. Field ID `country_select` with display label "Country Select" 
2. It has option values:
 a. "usa" with display label "United States"
 b. "eu"  with display label "European Union"

For webhook JSON currently we have next data:
```
{
  "country_select": "usa"
}
```
In scope of this issue we want to have next data:
```
{
  "country_select": "usa",
  "country_select_label": "United States",
}
```

But in the CSV export we have a mix of "label" usage for column name and option values as not readable values.
For CSV export currently we have next data:
```
column name: "Country Select"
row value: "usa"
```
But how to change it here to match webhook logic?
Expected values can be next (or you can propose some your variant):
```
// variant 1: change value to a human readable
column name: "Country Select"
row value: "United States" (was "usa")

// variant 2: leave old column as is and add new column (but how to name these columns?)
column name: "Country Select" (old column as is)
row value: "usa" (old column as is)

column name: "Country Select (Label)" (new column, name can be different if you want)
row value: "United States"

// variant 3
Some other way
```

adrians5j:
Thank you @neatbyte-vnobis for this detailed explanation. Much appreciated! 👏🏻 

So, if I understood everything correctly, you said this already exists:

```
column name: "Country Select"
row value: "usa"
```

I think that's fine. With that, I'd then just introduce your var3 into the mix:

```
column name: "Country Select (Label)" (new column, name can be different if you want)
row value: "United States"
```

That way no breaking changes are introduced, we're just adding a new column.

Hope I didn't miss smt, if so let me know please. :)

BTW before we'll be merging this, could you also attach an example CSV file, just so we're 100% on the same frequency?

neatbyte-vnobis:
@adrians5j @SvenAlHamad 
Example CSV file:
[form-submissions-8lf2d391t-form_submissions_export.csv](https://github.com/webiny/webiny-js/files/10941220/form-submissions-8lf2d391t-form_submissions_export.csv)
image

neatbyte-vnobis:
@adrians5j this PR is super old...
I have created a new one based on todays `next` branch.
See https://github.com/webiny/webiny-js/pull/3093

PR #2957, opened 2023-01-18
feat: add an email form trigger

neatbyte-vnobis:
## Changes
Implements task "Add a email trigger"

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-form-builder/src/admin/plugins/editor/triggers/emailThanks.tsx
```diff
@@ -0,0 +1,58 @@
...
+        description: "Send an email notification to the user that submitted the form.",
+        icon: <EmailIcon />,
+        renderSettings({ Bind, submit, form }) {
+            console.log(form.fields);
+            const hasEmailField = form.fields.some(field => field.name === "email");
+
+            if (!hasEmailField) {
+                return (
+                    <Alert title={t`Email field required`} type={"danger"}>
+                        {t`In order to use this trigger you need to add an email field to your form`}
```
adrians5j:
```suggestion
                        {t`In order to use this trigger you need to add an e-mail field to your form.`}
```

packages/app-form-builder/src/admin/plugins/editor/triggers/emailThanks.tsx
```diff
@@ -0,0 +1,58 @@
...
+const plugin: FbEditorTrigger = {
+    type: "form-editor-trigger",
+    name: "form-editor-trigger-email-thanks",
+    trigger: {
+        id: "email-thanks",
+        title: "Email - Thank you email",
+        description: "Send an email notification to the user that submitted the form.",
+        icon: <EmailIcon />,
+        renderSettings({ Bind, submit, form }) {
+            console.log(form.fields);
```
adrians5j:
```suggestion

```

packages/app-form-builder/src/admin/plugins/editor/triggers/emailThanks.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { FbEditorTrigger } from "~/types";
+
+const t = i18n.namespace("FormsApp.Editor.EmailThanksTriggerSettings");
+
+const plugin: FbEditorTrigger = {
+    type: "form-editor-trigger",
+    name: "form-editor-trigger-email-thanks",
+    trigger: {
+        id: "email-thanks",
+        title: "Email - Thank you email",
```
adrians5j:
```suggestion
        title: "E-mail - Thank You E-mail",
```

packages/app-form-builder/src/admin/plugins/editor/triggers/emailThanks.tsx
```diff
@@ -0,0 +1,58 @@
...
+        description: "Send an email notification to the user that submitted the form.",
```
adrians5j:
```suggestion
        description: "Send an e-mail notification to the user that submitted the form.",
```

packages/app-form-builder/src/admin/plugins/editor/triggers/emailNotification.tsx
```diff
@@ -0,0 +1,45 @@
...
+        description: "Send an email notification with form submission details.",
+        icon: <EmailIcon />,
+        renderSettings({ Bind, submit }) {
+            return (
+                <Grid>
+                    <Cell span={12}>
+                        <Bind name={"email"} validators={validation.create("email")}>
+                            <Input
+                                label={t`Email address`}
+                                description={t`Email address to which the submission details will be sent to.`}
```
adrians5j:
```suggestion
                                description={t`E-mail address to which the submission details will be sent to.`}
```

packages/app-form-builder/src/admin/plugins/editor/triggers/emailNotification.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+const t = i18n.namespace("FormsApp.Editor.EmailNotificationTriggerSettings");
+
+const plugin: FbEditorTrigger = {
+    type: "form-editor-trigger",
+    name: "form-editor-trigger-email-notification",
+    trigger: {
+        id: "email-notification",
+        title: "Email - Submission Notification",
+        description: "Send an email notification with form submission details.",
```
adrians5j:
```suggestion
        title: "E-mail - Submission Notification",
        description: "Send an e-mail notification with form submission details.",
```

adrians5j (changes requested):
Sending some visual feedback, nothing spectacular.

BTW, just for consistency, if you notice any other occurrence of the word "email" (instead of "e-mail"), please correct. 

Also, I still need to get a 2nd opinion on the API side. Might pull other team members if needed. So, in the meantime, let's resolve the visual side.

Thank you!

UPDATE: checked with the team the mailer usage. More info https://github.com/webiny/webiny-js/pull/2957/files#r1084009463.

packages/api-form-builder/src/plugins/triggers/triggerHandlers.ts
```diff
@@ -1,43 +1,135 @@
...
+        type: "form-trigger-handler",
+        name: "form-trigger-handler-email-notification",
+        trigger: "email-notification",
+        async handle({ trigger, data, form, addLog }) {
+            const email = trigger && trigger.email;
+            if (!email) {
+                return;
+            }
+
+            const mailer = createSmtpTransport(MAILER_CONFIG);
```
adrians5j: Just checked with the team.

Mailer should be used via the `context` object, and not directly via `createSmtpTransport`.

Example usage: https://www.webiny.com/docs/overview/features/mailer#usage

Thanks!

neatbyte-vnobis:
@adrians5j 
Comments are fixed. Important thing - e-mail validation was changed to support HTML-only mail body
see - packages/api-mailer/src/crud/transport/onTransportBeforeSend.ts

SvenAlHamad:
@neatbyte-vnobis - I've tried to test this. The admin side of things looks alright, but I'm not able to submit the form any more. The submit button doesn't do anything, and I don't see any errors in the console or the network panel.
This is my form: https://d1j45fd2ggkz2d.cloudfront.net/untitled-ld8vtqze

packages/api-form-builder/src/plugins/triggers/emailThanks.ts
```diff
@@ -0,0 +1,41 @@
...
+                    type: "success",
+                    message: `Successfully sent a thanks e-mail to ${data.email}`,
+                    data: {
+                        response: response.result
+                    }
+                });
+            } catch (e) {
+                addLog({
+                    type: "warning",
+                    message: `Failed to send a thanks e-mail to ${data.email}: ${e.message}`
```
adrians5j: thanks -> thank you

packages/api-form-builder/src/plugins/triggers/emailThanks.ts
```diff
@@ -0,0 +1,41 @@
...
+                const response = await context.mailer.sendMail({
+                    to: [data.email],
+                    subject: trigger.subject,
+                    text: "",
+                    html: `<pre style="font-family: 'Open Sans', sans-serif;">${trigger.content}</pre>`
+                });
+
+                addLog({
+                    type: "success",
+                    message: `Successfully sent a thanks e-mail to ${data.email}`,
```
adrians5j: thanks -> thank you

packages/api-form-builder/src/plugins/triggers/emailNotification.ts
```diff
@@ -0,0 +1,52 @@
...
+                        "Hi," +
+                        `\n\nsomeone just submitted a form: ${form.name}` +
+                        `\n\n${fieldsString}` +
+                        "\nBest," +
+                        "\n\nWebiny"
+                });
+
+                addLog({
+                    type: "success",
+                    message: `Successfully sent a notification e-mail to ${trigger.email}`,
```
adrians5j: `Successfully sent a notification e-mail to ${trigger.email}.` (added dot at the end)

packages/api-form-builder/src/plugins/triggers/emailThanks.ts
adrians5j: `Successfully sent a thanks e-mail to ${data.email}.` (added dot at the end)

adrians5j (changes requested):
A couple of nitpicks but more importantly:

1. You mentioned "e-mail validation was changed to support HTML-only mail body". Can you clarify this?

I'm looking at the code change and cannot connect your description with the actual code change:

image

2. Sven mentioned some form submission issues. I can see the code from this branch is not up-to-date with `next`. Can you pull all changes are resubmit it for another review for Sven?

Thank you!

SvenAlHamad:
I've tested this one again, and it's working fine, but I would request one change to be made. The email content field shouldn't be a plain text box, but a rich text editor instead: 
CleanShot 2023-02-01 at 12 41 12@2x

You have an example of the rich text editor if you edit the form and click the form settings, the success message is using the RTE component.

neatbyte-vnobis:
@SvenAlHamad I understand your idea, but it may be more tricky to implement
The "form submission success message" is displayed in the site UI and uses React Rich Text rendering component.
But this "email content" is used on the API side (in the backend), so I cannot use React component there to get HTML for the email.

As a solution, maybe we can save in the "settings" API 2 data fields - RTE value as a JSON (to enable editing on the setting page) and RTE value as a HTML (aka rendered on the edit page by React RTE component, it will be used for the email)

neatbyte-vnobis:
> 1. You mentioned "e-mail validation was changed to support HTML-only mail body". Can you clarify this?

@adrians5j 
This change was reverted. Validation of min 10 characters was a blocker for an "" (empty string) `text` value.
But `text` property can be dropped (not be an empty string), so this change is not required anymore.
Missed a point that `text` is nor required.

SvenAlHamad:
> @SvenAlHamad I understand your idea, but it may be more tricky to implement The "form submission success message" is displayed in the site UI and uses React Rich Text rendering component. But this "email content" is used on the API side (in the backend), so I cannot use React component there to get HTML for the email.
> 
> As a solution, maybe we can save in the "settings" API 2 data fields - RTE value as a JSON (to enable editing on the setting page) and RTE value as a HTML (aka rendered on the edit page by React RTE component, it will be used for the email)

I'm ok with that solution, but let's also get a confirmation from @adrians5j

adrians5j:
Not sure... the HTML on its own would not be enough, just double checked what's the HTML output of this component, and there are some CSS classes that get added here as well, for example (the first half of the snippet):

```<b><i><u class=\"cdx-underline\"><span class=\"cdx-text-color\" style=\"color: rgb(0, 151, 230);\">This is a test message.</span></u></i></b>
After that we have a <a href=\"https://d3pyi87q2d1m2o.cloudfront.net/form-builder/forms/63db8edd83c15500086b386b%230001\">link</a>.```

**Comment by adrians5j** [comment]
Not sure... the rendered HTML on its own would not be enough. Just double checked what's the HTML output of the RTE component, and there are some CSS classes that get added here as well, for example (the first half of the snippet):

```
<b><i><u class=\"cdx-underline\"><span class=\"cdx-text-color\" style=\"color: rgb(0, 151, 230);\">This is a test message.</span></u></i></b>
After that we have a <a href=\"https://d3pyi87q2d1m2o.cloudfront.net/form-builder/forms/63db8edd83c15500086b386b%230001\">link</a>.
```

I'm not sure if RTE is ready for rendering HTML for e-mail clients. As we all know, there are quirks / gotchas when it comes to rendering e-mail HTML. We'd need to additionally test this, which would break the scope.

If you ask me, for this first version, I'd keep it simple. I'd write it down, and maybe revisit this when we start implementing Lexical editor across the board / if users start to report.

@SvenAlHamad what do you think?

**Comment by SvenAlHamad** [comment]
Alright, in that case let's leave it as a simple text field. From my side I'm all good with merging this task.

**Comment by adrians5j** [comment]
@neatbyte-vnobis Still looking [at this](https://github.com/webiny/webiny-js/pull/2957/files#diff-3383c53f57dabdca39a0c141fbce8db784fd4b915a999d78f7cc89dba47bcff3R17):

image

Your comment: https://github.com/webiny/webiny-js/pull/2957#issuecomment-1412162731

Still not sure I completely understand this change.

@brunozoric since you're the author of this code, can you confirm this change is fine? 

It seems that now both `text` and `html` fields become optional. Would that break anything? To me it looks good, it's either one or the other.

**Comment by brunozoric** [comment]
@adrians5j
In theory, you can send an e-mail without any body. Is it good thing to do? Probably not. Can you do it - yes.
If tests are passing, everything should be fine.

**Comment by neatbyte-vnobis** [comment]
@adrians5j @brunozoric 
The tricky thing here is that we need OR `text`, OR `html` for an email body.
For this task we need an `html` body, so required `text` property was an issue (as its validation was blocking `html` usage).
So main thing here is to remove "required" from `text` validation.
(I still do not know what will happen when both fields are present in the same time, aka will `html` be used over the `text` or not)

But I agree that validation can be improved - to ensure that "one of" `text` or `html` should be present in passed data.

**Comment by brunozoric** [comment]
@neatbyte-vnobis 
zod does not support the text OR html out of the box, but it is doable: https://github.com/colinhacks/zod/issues/61
Feel free to implement the refinement method from the posts in the issue I linked.

**Comment by SvenAlHamad** [comment]
I think we all concluded that we're staying with the text field. With regards to other change requests have they been sorted? 

@neatbyte-vnobis - I have one more change from my end with regards to this. Can we allow users to use variables from the form inside the text field content. When I write {field-slug} in the the content, that it's replaced with the value from the form?

**Comment by Pavel910** [comment]
> I think we all concluded that we're staying with the text field. With regards to other change requests have they been sorted?
> 
> @neatbyte-vnobis - I have one more change from my end with regards to this. Can we allow users to use variables from the form inside the text field content. When I write {field-slug} in the the content, that it's replaced with the value from the form?

If we implement this, I would recommend `{fields:slug}` or `{fields.slug}`, so, use `:` or `.` as a delimiter. instead of `-`. And `fields` (plural), as it references all form fields, and we can later in the future implement Lexical here, and then have proper autocomplete on form fields.

**Comment by neatbyte-vnobis** [comment]
@SvenAlHamad @Pavel910  I wanted to clarify one thing about {fields.slug} variables:

What should we do?:
1) if user typed incorrect slug in the text field content (aka `some text {fields.not_exists}`)
2) if form value was not filled by user on submit (aka not required field is left empty)

By default all fields return null (except checkbox, which returns an empty array), but is a plain JS output and it is not very readable...

Some possible solutions:
1) Show `null` or `undefined` as they are
2) Replace them with empty strings (aka skip them)
3) Do not replace them and leave `{fields.slug}` as a string in the email

**Comment by SvenAlHamad** [comment]
@neatbyte-vnobis @Pavel910  - I think it's best to have an empty value in that case.

**Comment by adrians5j** [comment]
Looks good to be merged, gonna wait for @SvenAlHamad approval though, don't have e-mail set up locally. :)

**Comment by SvenAlHamad** [comment]
Let's just add one small thing to this one and merge it into next as I'm not able to test it fully anymore since the mailer package is not inside this branch anymore. 
The change we should do is just add a below the text field: "You can use {fields.slug} (example `{field.firstName}`) in the email content, which will be replaced with actual values from the form before the email is sent."

**Comment by neatbyte-vnobis** [comment]
@adrians5j 
please approve a PR so we can merge it

File: packages/app-form-builder/src/admin/plugins/editor/triggers/emailThanks.tsx [hunk 11]
```diff
@@ -37,7 +37,11 @@ const plugin: FbEditorTrigger = {
                     </Cell>
                     <Cell span={12}>
                         <Bind name={"content"}>
-                            <Input rows={5} label={t`E-mail content`} />
+                            <Input
+                                rows={5}
+                                label={t`E-mail content`}
+                                description="You can use {fields.slug} (example {field.firstName}) in the email content, which will be replaced with actual values from the form before the email is sent."
```
adrians5j: The description lacks consistency. First says `{fields.slug}` then provides `{field.firstName}` as an example (notice the `fields` / `field` difference).

Let's fix this and simplify:

> You can use {fields.FIELD_NAME} placeholders to reference form submission values. For example: "Hello {fields.firstName} {fields.lastName}!".

adrians5j (changes requested):
Let's address the description, it has an inconsistency.

adrians5j (approved):

bmccarthyco:
Curious as to the reason to select the older "e-mail" variant vs. the more common "email". This is an internet age-old dilemma and while both are valid, "email" is the more popular spelling despite a few professional journalists preferring the older "e-mail" spelling. I checked the Webiny documentation and from a cursory review, "email" is used a bit more than "e-mail" already, with the exception of `/src/docs/5.34.x/overview/features/mailer.mdx`. Furthermore, in code, `email` is used for variable names. While seemingly minor, it would make sense for Webiny to go with the more modern spelling, being a very modern platform, and would want to market itself as such.

adrians5j:
Sorry @bmccarthynn, didn't manage to look into this. If we got this feedback before we merged this, maybe we'd be able to discuss/react on time.

For now, we'll leave this as is, it's not breaking anything functionality-wise. We can rename later if we notice more people asking.

PR #2961, opened 2023-01-19
feat: add import/export capabilities to page datatable

leopuleo:
## Changes
With this PR we add import and export capabilities to page data table.

### Export pages
Users can: 
- select one or more pages from the table and trigger the "export" process by clicking the "Export" button;
- export all website pages by clicking the "Export" button.

**Note:** users can select pages to export at this stage, users can't choose folders. The folders row are not presenting the checkbox.

### Import pages
Users can trigger the import process by clicking the designated button. We add a new stage to the import process: after the page has been created (during the `process` stage) we subscribe to the `onPageAfterCreate` event topic creating the related link. 

Custom metatags we need to forward to event topics, such as `folderId`, are added to the `meta` param.

Closes [#WEB-1953](https://webiny.atlassian.net/browse/WEB-1953)

## How Has This Been Tested?
Manually


packages/api-aco/src/subscriptions/aftePageCreate.ts
```diff
@@ -0,0 +1,28 @@
...
+            } catch (error) {
+                /**
+                 * In case the link already exists for the current page, just return.
+                 */
+                if (error.code === "LINK_EXISTS") {
+                    return;
+                }
+
+                throw WebinyError.from(error, {
+                    message: "Error while creating page link.",
```
Pavel910, line 22: I would rather include the original error message, from the original `error` object, instead of masking it.

packages/api-aco/src/subscriptions/aftePageCreate.ts
```diff
@@ -0,0 +1,28 @@
...
+                    code: "AFTER_PAGE_CREATE"
```
Pavel910: Codes should include some prefix that makes them easier to find, like `ACO_AFTER_PAGE_CREATE`, then it's clear the error is coming from ACO package.

Pavel910 (changes requested):

packages/api-aco/src/subscriptions/aftePageCreate.ts
leopuleo: [Webiny.from()](https://github.com/webiny/webiny-js/blob/dev/packages/error/src/Error.ts#L28) should already use code and message from the original error, right? 

So, in case these exist already in the original error, we would use them.

packages/api-aco/src/subscriptions/aftePageCreate.ts
leopuleo: 👍

packages/api-aco/src/subscriptions/aftePageCreate.ts
Pavel910: You are right @leopuleo. Ignore this then.

Pavel910 (approved):

PR #2963, opened 2023-01-20
feat: add page templates

neatbyte-vnobis:
## Changes
Add feature "Page Templates"
Spec - https://webiny.notion.site/Page-Builder-Templates-Requirements-spec-3861f7ae1c484b599639b044d48dce38

## How Has This Been Tested?
Manual

## Documentation
None?


neatbyte-vnobis:
@SvenAlHamad @Pavel910 
this latest (and updated, and synced with "next" branch changes) pull for a "Page Templates" feature.
I have checked it with full project rebuild locally.

If you see some errors please provide me full build logs.

SvenAlHamad:
Thanks @neatbyte-vnobis 

I've tested the page builder themes, and I think it's good what we currently have, but I do have a few points that I would like you to address:

1. Add export/import options to the page templates (use the same approach as on pages, so the user can select with a check box which templates they would like to export)
2. In the list of page templates, on the right side, where we have the page template preview, can you please add the template description under then template name. 
3. Security - do we have the option to control is a user can unlock a template or now, as well as create and manage page templates or non?
4. Few minor visual adjustments we want to make:
CleanShot 2023-01-23 at 14 04 14@2x

SvenAlHamad:
@neatbyte-vnobis thanks for the recent update. I've got a few more adjustments:
1. In the list of page templates please add the description under the page template name:
CleanShot 2023-02-02 at 11 40 59@2x

2. There's a problem with the way that the security is implemented on both Page Templates and blocks. If I want to have a user that is not allowed to create/update/delete blocks and page templates, but is allowed to use them inside a page and unlink them, I can't create such a user group. The checkbox to allow users to unlink a block/page template is disabled until I give them write permissions. We should be able to allow users to unlink a block/template without needing to allow them to manage those entities. 

Other than this, I think from my end, this is good to go!

packages/api-page-builder/src/graphql/crud/pageTemplates.crud.ts
```diff
@@ -0,0 +1,358 @@
...
+    params: CreatePageTemplatesCrudParams
+): PageTemplatesCrud => {
+    const { context, storageOperations, getLocaleCode, getTenantId } = params;
+
+    const onBeforePageTemplateCreate = createTopic<OnBeforePageTemplateCreateTopicParams>();
+    const onAfterPageTemplateCreate = createTopic<OnAfterPageTemplateCreateTopicParams>();
+    const onBeforePageTemplateUpdate = createTopic<OnBeforePageTemplateUpdateTopicParams>();
+    const onAfterPageTemplateUpdate = createTopic<OnAfterPageTemplateUpdateTopicParams>();
+    const onBeforePageTemplateDelete = createTopic<OnBeforePageTemplateDeleteTopicParams>();
+    const onAfterPageTemplateDelete = createTopic<OnAfterPageTemplateDeleteTopicParams>();
```
Pavel910: Please rename these to follow the convention: `on{EntityName}AfterCreate`. So entity name goes right after `on`. That makes it easier to look for specific entity when coding. It's a convention we follow in all the apps, so in this case: `onPageTemplateBeforeCreate`, etc.

packages/api-page-builder-so-ddb-es/src/definitions/pageTemplateEntity.ts
```diff
@@ -0,0 +1,55 @@
...
+            tenant: {
+                type: "string"
+            },
+            locale: {
+                type: "string"
+            },
+            ...(attributes || {})
+        }
+    });
+};
```
Pavel910: Over time we've learned that this way of defining DDB entities is problematic, because it's super hard to extend via plugins. Let's change the entity definition to a simpler form, as we do here: https://github.com/webiny/webiny-js/blob/page-builder-blocks-v2.2-page-templates/packages/api-aco-so-ddb/src/definitions/entities.ts

It's less code, and it allows us not to worry about individual field data types, and it's easy to add new fields into the business logic, without modifying the DDB entity. Also, it's a lot easier to decouple storage-specific attributes, like PK, SK, etc. from actual data you're storing, which is now all in the `data` attribute.

packages/api-page-builder/src/graphql/crud/pageTemplates.crud.ts
```diff
@@ -0,0 +1,358 @@
...
+    title: string({ validation: validation.create("required,maxLength:100") }),
+    description: string({ validation: validation.create("maxLength:100") }),
+    content: object()
+})();
+
+const UpdateDataModel = withFields({
+    title: string({ validation: validation.create("maxLength:100") }),
+    description: string({ validation: validation.create("maxLength:100") }),
+    content: object()
+})();
```
Pavel910: Please replace `commodo` with `zod`, as we're trying to get rid of commodo completely. See example here: https://github.com/webiny/webiny-js/blob/page-builder-blocks-v2.2-page-templates/packages/api-aco/src/context/folders.ts

packages/app-page-builder-elements/src/components/Elements.tsx
```diff
@@ -31,7 +31,10 @@ export const Elements: React.FC<ElementsProps> = props => {
     const elements = props.element.elements;
 
     let parentBlockElement: ElementType;
-    if (props.element.data.blockId) {
+    if (
+        props.element.data.blockId ||
+        (props.element.type === "block" && currentRendererMeta.parentElement.data.templateId)
+    ) {
```
Pavel910: Please make this `if` statement a bit more readable:
1) extract individual conditions into local variables above the statement, like `isBlockReference` (or whatever that conditional does)
2) leave some comments on what each conditional is for, because the second one with `templateId` is unclear.

Keep in mind that this code will be maintained by someone, so inline comments would greatly help understand the intention.

packages/app-page-builder/src/pageEditor/config/ToolbarActionsPlugin.tsx
```diff
@@ -0,0 +1,135 @@
...
+                        <div>{actionsTop.map(renderPlugin)}</div>
+                    )}
+                    <div>{actionsBottom.map(renderPlugin)}</div>
+                </ToolbarActionsWrapper>
+                <Dialog open={isModalShown} onClose={onClose} className={unlinkTemplateDialog}>
+                    <DialogTitle>Unlink Template</DialogTitle>
+                    <DialogContent>
+                        <p>
+                            This page is created from a template - to change it you need to unlink
+                            it first.
```
Pavel910: Please update this sentence like this:

```
This page was created from a template. To change it, you need to unlink it first.
```

packages/app-page-builder/src/pageEditor/config/ToolbarActionsPlugin.tsx
```diff
@@ -0,0 +1,135 @@
...
+                        </p>
+                        <p>
+                            By unlinking it, any changes made to the template will no longer
+                            automatically reflect to this page.
+                        </p>
```
Pavel910: Also this:

```
By unlinking it, any changes made to the template will no longer be reflected on this page.
```

Pavel910 (changes requested):
A few change requests here, nothing complicated. And please have a look at the security comment Sven made. Thanks!

neatbyte-vnobis:
@Pavel910 please take a look on updated PR

Pavel910 (approved):

Pavel910:
@neatbyte-vnobis thanks! PR approved, and I'll merge this tomorrow (Monday)

PR #2965, opened 2023-01-20
feat: add date to form submission export

neatbyte-vnobis:
## Changes
Implements issue "Add date/time to the export"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-form-builder/src/plugins/graphql/form.ts
+++ b/packages/api-form-builder/src/plugins/graphql/form.ts
@@ -1,4 +1,5 @@
 import { parseAsync } from "json2csv";
+import { format } from "date-fns";
 import {
     ErrorResponse,
     ListResponse,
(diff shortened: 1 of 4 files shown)


packages/api-form-builder/src/plugins/graphql/form.ts
```diff
@@ -549,6 +549,12 @@ const plugin: GraphQLSchemaPlugin<FormBuilderContext> = {
                         for (let i = 0; i < submissions.length; i++) {
                             const submissionData = submissions[i].data;
                             const row: Record<string, string> = {};
+
+                            row["Date submitted"] = new Date(submissions[i].createdOn)
```
adrians5j: Let's use [`date-fns`](https://date-fns.org/) for this.

I saw it [already being used](https://github.com/webiny/webiny-js/blob/dev/packages/db-dynamodb/src/plugins/definitions/DateTimeTransformPlugin.ts) on the backend, so it should not additional size on disk.

adrians5j (changes requested):
One minor thing.

Thanks!

SvenAlHamad:
I've tested this and it's working. However let's just adjust the label so it says "Date submitted (UTC)". Later down the line we'll see how we make sure the dates are translated to the user's timezone.

adrians5j (approved):
Thanks!

PR #2968, opened 2023-01-20
feat: improve blocks search functionality

neatbyte-vnobis:
## Changes
Implement task "Improve the blocks search functionality"
Includes a fix for not saved block name (regression caused by `isDirty` implementation)

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
+++ b/packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
@@ -114,7 +114,7 @@ const BlocksByCategoriesDataList = ({
         listQuery?.data?.pageBuilder?.listBlockCategories?.data || [];
     const pageBlocksData: PbPageBlock[] = listQuery?.data?.pageBuilder?.listPageBlocks?.data || [];
 
-    const [blocksList, blockCategoriesList] = useCategoriesListData(
+    const [blocksList, filteredBlockCategoriesList] = useCategoriesListData(
         pageBlocksData,
         blockCategoriesData,
         sort,
@@ -320,7 +320,7 @@ const BlocksByCategoriesDataList = ({
             <DataList
                 title={t`Blocks`}
                 loading={Boolean(loading)}
-                data={blockCategoriesList}
+                data={filteredBlockCategoriesList}
                 actions={
                     <DataListActionsWrapper>
                         {canCreate ? (
@@ -407,7 +407,7 @@ const BlocksByCategoriesDataList = ({
                 </DialogTitle>
                 <DialogContent>
                     <React.Fragment>
-                        {isEmpty(blockCategoriesList) ? (
+                        {isEmpty(blockCategoriesData) ? (
                             <div className={noRecordsWrapper}>
                                 <Typography use="overline">
                                     There are no block categories
@@ -415,7 +415,7 @@ const BlocksByCategoriesDataList = ({
                             </div>
                         ) : (
                             <List twoLine>
-                                {blockCategoriesList.map(item => (
+                                {blockCategoriesData.map(item => (
                                     <ListItem key={item.slug} onClick={() => onSelect(item.slug)}>
                                         <ListItemGraphic>
                                             <Icon category={item} />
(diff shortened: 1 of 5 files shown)


neatbyte-vnobis:
@adrians5j 
please review this pull

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
```diff
@@ -109,13 +114,28 @@ const BlocksByCategoriesDataList = ({ canCreate }: PageBuilderBlocksByCategories
...
+    const filterBlocksData = useCallback(
+        ({ name }) => {
+            return name.toLowerCase().includes(filter);
         },
         [filter]
     );
 
+    const filteredBlocksData: PbPageBlock[] =
+        filter === "" ? pageBlocksData : pageBlocksData.filter(filterBlocksData);
+
```
adrians5j: Looks good. Can we maybe move this data filtering / sorting code into a separate function or React hook?

Something like:

```ts
// packages/app-page-builder/src/admin/views/PageBlocks/hooks/useCategoriesListData.ts

type BlockListData = Array<...>; // Please add the correct type.
type BlockCategoriesListData = Array<...>; // Please add the correct type.

function useCategoriesListData(blockCategoriesData, blockData): [BlocksList, BlockCategoriesList];
```

adrians5j (changes requested):
Looks good, tried it in the browser as well.

One minor refactor though. 

Thank you 🍻 🙂

adrians5j (changes requested):
Something's off with GH. Ignore this change request submission (but do take a look at the above one 😉).

SvenAlHamad:
can we merge and close this one?

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
```diff
@@ -435,15 +407,15 @@ const BlocksByCategoriesDataList = ({
                 </DialogTitle>
                 <DialogContent>
                     <React.Fragment>
-                        {isEmpty(categoryList) ? (
+                        {isEmpty(blockCategoriesList) ? (
```
adrians5j: `blockCategoriesList` represents filtered/sorted list of block categories.

But this dialog isn't tied to the filter/sort. In the dialog, we should always have a full list of block categories.

Did I maybe miss something here? 

Thanks!

adrians5j: BTW, maybe to be more clear, renaming `blockCategoriesList` to `filteredBlockCategoriesList` might be a better choice for the var name.

adrians5j (changes requested):
Thanks for the refactor. 

But, one thing looked odd to me so I had to ask.

neatbyte-vnobis:
@adrians5j 
please take a look on the PR changes

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
```diff
@@ -114,7 +114,7 @@ const BlocksByCategoriesDataList = ({
         listQuery?.data?.pageBuilder?.listBlockCategories?.data || [];
     const pageBlocksData: PbPageBlock[] = listQuery?.data?.pageBuilder?.listPageBlocks?.data || [];
 
-    const [blocksList, blockCategoriesList] = useCategoriesListData(
+    const [blocksList, filteredBlockCategoriesList] = useCategoriesListData(
```
adrians5j: I know it's a nitpick, but since you've applied the `blockCategoriesList` => `filteredBlockCategoriesList` (thanks!), let's be consistent and also apply the same to `blocksList` and `useCategoriesListData`. Unless I'm missing something, both of these represent filtered data.

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
```diff
@@ -379,7 +369,7 @@ const BlocksByCategoriesDataList = ({ canCreate }: PageBuilderBlocksByCategories
                 {({ data }: { data: PbBlockCategory[] }) => (
                     <ScrollList data-testid="default-data-list">
                         {data.map(item => {
-                            const numberOfBlocks = pageBlocksData.filter(
+                            const numberOfBlocks = blocksList.filter(
```
adrians5j: Do we really need to do this filter here? Can't we just use the data returned from `useCategoriesListData`?

So, `blocksList.length` ?

Sorry if I'm missing something.

adrians5j (changes requested):
One nitpick and one question.

Let's resolve these two and we merge!

adrians5j (approved):

packages/app-page-builder/src/admin/views/PageBlocks/BlocksByCategoriesDataList.tsx
neatbyte-vnobis: @adrians5j 
`useCategoriesListData` returns us full list of filtered blocks. This filter is used to get numberOfBlocks per category

PR #2971, opened 2023-01-23
feat(plugins): add support for plugin decorators

Pavel910:
## Changes
This PR adds support for registering plugin decorators. This is useful for our old POJO plugins (and we have many of those),  as very often in our code we do conditional checks if certain optional plugin properties are set, and if so, execute those functions. This adds a lot of overhead in various views where same plugin types are being used.

Using decorators we can now register decorators (classes), which decorate (wrap) the original plugin, and that way we can implement default behavior on those optional plugin functions. 

## Why not just switch the original plugins with classes? 
Because there are many plugins in the user land, which would break the app if we just assumed that everybody is now using classes. This way we can add a decorator and achieve the same result without breaking anything, and remove the redundant conditionals from our main logic.

@brunozoric @adrians5j please have a look at the tests and post your thoughts.

## How Has This Been Tested?
Jest tests.

### DIFF
--- a/packages/plugins/src/PluginsContainer.ts
+++ b/packages/plugins/src/PluginsContainer.ts
@@ -95,7 +95,7 @@ export class PluginsContainer {
                 `There is a requirement for plugin of type "${type}" to be only one registered.`
             );
         }
-        return this.decorate(list[0]) as T;
+        return list[0] as T;
     }
 
     public all<T extends Plugin>(): T[] {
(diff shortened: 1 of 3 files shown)


packages/plugins/src/PluginsContainer.ts
```diff
@@ -76,11 +95,11 @@ export class PluginsContainer {
                 `There is a requirement for plugin of type "${type}" to be only one registered.`
             );
         }
-        return list[0];
+        return this.decorate(list[0]) as T;
```
brunozoric: Unnecessary this.decorate as it is already done in `atLeastOneByType -> byType -> findByType` call.

brunozoric (changes requested):
Looks ok.
One small change. Left a comment on it.

PR #2981, opened 2023-01-25
Page Builder - Minor Fixes

adrians5j:
## Changes
This PR brings the following Page Builder-related fixes:
- ensures that the `ps-tag` rendered with Form page element contains the correct `id`
- ensures we don't end up with double `p` tags when rendering Paragraph page element
- instructs the prerendering engine to also cache `FbGetPublishedForm` query, but note that this will again change in the next PR, so this change is irrelevant

## How Has This Been Tested?
Manual / relying on existing tests.

## Documentation
Changelog.


packages/app-page-builder-elements/src/renderers/form/types.ts
```diff
@@ -45,6 +45,7 @@ export interface FormDataRevision {
 
 export interface FormData {
     id: string;
+    formId: string;
     version: number;
     parent: string;
```
adrians5j: Check `parent` property, does this still exist?

adrians5j (commented):
check `parent`

PR #2983, opened 2023-01-25
Improve internals of the Website app

Pavel910:
## Changes
This PR converts the Website app to work using the same base application framework as our current Admin app. This means we're now able to use composable components, register custom React Context providers, register HOCs (PageRenderer, Layout, MainContent, WebsiteScripts), override Layout rendering logic, etc.

The `@webiny/app` package now contains the core application framework which is now shared between the Admin and Website apps. It can also be used to create new apps by 3rd party developers (pending docs).

The following is an example of registering a HOC for the `MainContent` component, to wrap only the main page content with custom logic (be it styling, security checks, whatever):

```tsx
import React from "react";
import { Website, MainContent, createComponentPlugin } from "@webiny/app-website";

const WrapMainContentPlugin = createComponentPlugin(MainContent, Original => {
    return function MainContent(props) {
        return (
            <div style={{ padding: 20, backgroundColor: "red" }}>
                Main content wrapper
                <Original {...props} />
            </div>
        );
    };
});

export const App: React.FC = () => {
    return (
        <Website>
            <WrapMainContentPlugin />
        </Website>
    );
};
```

page-main-content-plugin

## How Has This Been Tested?
Manually (tested preview, prerendering, localhost).

## Documentation
TODO


adrians5j (approved):
Awesome stuff 🚀

PR #2996, opened 2023-02-01
Make New PB rendering work with Theme Manager

Pavel910:
## Changes
This PR adds several tweaks to make the new PB rendering mechanism work with asynchronously loaded themes. This means that themes are lazy loaded, so we can't expect the `theme` plugin to always be there. This required several TS type changes, and allow Theme Manager to trigger theme detection after plugins are registered.

## How Has This Been Tested?
Manually, by adding 2 themes, switching between them, and checking the results on the prerendered website.


adrians5j (approved):
Awesome! 🚀

PR #3000, opened 2023-02-01
fix: fix page blocks list render

neatbyte-vnobis:
## Changes
Fix issue "List of blocks bug (URGENT)"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/pageEditor/config/blockEditing/BlocksList.tsx
+++ b/packages/app-page-builder/src/pageEditor/config/blockEditing/BlocksList.tsx
@@ -66,7 +66,9 @@ const BlocksList: React.FC<BlocksListProps> = props => {
     const prevProps = useRef<BlocksListProps | null>(null);
 
     useEffect(() => {
-        rightPanelElement.current = document.querySelector(".webiny-split-view__right-panel");
+        rightPanelElement.current = document.querySelector(
+            ".webiny-split-view__right-panel-wrapper"
+        );
         setTimestamp(new Date().getTime());
     }, []);
 


neatbyte-vnobis:
@SvenAlHamad 
Here is a fix for an "List of blocks bug (URGENT)" issue.
I will keep in mind that such case can happen with this component for future tasks

adrians5j:
@SvenAlHamad we should release this in our next patch release I suppose?

If so, we'll need to change the target branch here, should be `dev` then.

neatbyte-vnobis:
@adrians5j @SvenAlHamad 
I think in this case I need to create additional PR targeting `dev`
please confirm

SvenAlHamad:
I've test this, and it looks good. @Pavel910 can we release this in the next patch release?

Pavel910:
@SvenAlHamad we'll be doing merges between `dev` and `next` tomorrow, when Adrian is back, and if necessary I'll extract this fix into a new PR against `dev` (if we jsut switch the base branch now, we'll get tons of modified files). So I'm adding a milestone for this to 5.34.3, and we'll see what to do about this during the week.

adrians5j:
We'll merge this into `next` now.

But note one thing @neatbyte-vnobis: in the future, we'll need to see what's the base branch when creating a new branch - `dev` or `next`.

For example, this fix will be released as a `5.34.3` patch version, and because of that, it needed to go into `dev`. Because this branch was created from `next`, I cannot just switch the target here from `next` to `dev`, because that would also merge a lot of changes that are already in `next`, but are not ready for `dev` yet (will be at the end of Q1).

For now, I can tell you whenever we have something that's urgent, you need to branch from `dev`. 

BTW more on our branching/release strategy can be found [here](https://www.webiny.com/docs/release-management/branching-strategy).

PR #3012, opened 2023-02-09
feat: grid and cell adjustments

neatbyte-vnobis:
## Changes
Implement task "Grid / Cell adjustments"

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
@neatbyte-vnobis - I've tested this and it works pretty good, just a few minor comments and then I'll do more more indepth round of testing making sure that the backwards compatibility is still there: 

1.

When we switch to the mobile view where by default we set the Grid wrap to “wrap”, we need to also add `flex-flow: row wrap` to the css. This is because when you have a cell with some padding applied, the cell goes outside the grid bounds and this property will fix this behavior. Note: we only apply this to when the Grid wrap is set to “wrap”

2.

Let’s rename Cell sizing drop down values like so:

“Auto” should be “Match content size”

“Full height” should be “Match grid height”

**Note**: make sure to also update the message that pops up under the vertical alignment on the cell element that references old values.

3.

By default if you haven’t set a vertical alignment on the cell, the status in the sidebar shows like it’s aligned to the middle, while actually it’s aligned to the top. The status of the alignment in the sidebar needs to be updated to the correct state, which is that by default it’s aligned to the top.

SvenAlHamad:
@neatbyte-vnobis - I think this behaviours for now are ok, but let's expand the scope a bit with this task: https://www.notion.so/webiny/Page-Builder-Bug-Fixes-Improvements-0036c7406813427694a8bda88d754e7b?p=65baea637a804fd786c61f22acc374f7&pm=s
I want to see the two changes working together as we might want to then do some additional UI tweaks before we release this.

neatbyte-vnobis:
@SvenAlHamad @adrians5j @Pavel910 
Additional tasks are done - please review the PR

SvenAlHamad:
Hey @neatbyte-vnobis - I've tested this, and I think we're close to a usable solution. I've noted a few more adjustments we need to make for this feature in this document here: https://www.notion.so/webiny/Page-Builder-Bug-Fixes-Improvements-0036c7406813427694a8bda88d754e7b?p=052b8203bc0e4b4aa1fe68d7fc905424&pm=s

neatbyte-vnobis:
@SvenAlHamad please take a look on added features (chunk 3)

SvenAlHamad:
@neatbyte-vnobis - thanks for the update. I can confirm that the new behavior on desktop works fine, and it works fine on mobile view, but only when you have a grid with a single row. On a grid with multiple rows on mobile, I'm not able to break the grid into rows anymore. Can you have a look a that please and see if you can make some adjustments to the generated css so we get the desired behavior.

SvenAlHamad:
Thanks @neatbyte-vnobis. I did a test and the mobile version is now working fine, but I did notice 2 more things we need to address:
1. When you insert a grid with 3 cells, and set say the 3/3/6 for the column width. If you then change the number of rows on the grid, the cell width resets back to the previous value. The expected behavior is that the column width doesn't change when you change the number of rows. 
2. The "wrap direction" and "Column wrap" options on mobile don't seem to be working any more, but only on Grids that have 2 or more rows.

SvenAlHamad:
Thanks @neatbyte-vnobis  - I did another round of testing and I think we're almost there. There is just one small issue I found which is that the wrap/no-wrap option doesn't work on multi row grids. Works fine on single-row grid. Otherwise I think everything else is working as expected.

packages/app-page-builder/src/contexts/PageBuilder/PageElementsProvider.tsx
```diff
@@ -49,7 +49,7 @@ export const PageElementsProvider: React.FC = ({ children }) => {
             styles: {
                 background: createBackground(),
                 border: createBorder(),
-                gridFlexWrap: createGridFlexWrap(),
+                gridSettings: createGridSettings(),
```
adrians5j: Let's name the file and property as just `grid`. No need for the `settings` word, the same we have for others.

packages/cwp-template-aws/template/common/apps/website/src/plugins/pageBuilder.ts
```diff
@@ -93,6 +94,7 @@ export default [
     background,
     border,
     gridSettings,
+    cellSettings,
```
adrians5j: The same comment we had with `gridSettings`. Let's remove the `Settings` from the name.

packages/app-page-builder-elements/src/renderers/cell.tsx
```diff
@@ -21,7 +21,7 @@ export const createCell = () => {
                     return styles;
                 }
 
-                styles.width = `${(size / 12) * 100}%`;
+                styles.width = `calc(${(size / 12) * 100}% - var(--cellWidthOffset, 0px))`;
```
adrians5j: The `var(--cellWidthOffset, 0px)` is something we'll need to remove and see how we can address it in a diff way. The new page rendering engine is completely CSS-in-JS, doesn't rely on regular CSS or CSS vars.

Maybe if you could explain why the need for this? That might give me ideas on how to do this differently.

packages/app-page-builder/src/editor/plugins/elements/cell/PeCell.tsx
```diff
@@ -88,7 +88,7 @@ const PeCell = createRenderer(
                 return styles;
             }
 
-            styles.width = `${(size / 12) * 100}%`;
+            styles.width = `calc(${(size / 12) * 100}% - var(--cellWidthOffset, 0px))`;
```
adrians5j: The same comment I made before about CSS / CSS vars usage applies here too.

packages/app-page-builder/src/editor/plugins/elementSettings/cell/index.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import { PbEditorPageElementStyleSettingsPlugin } from "../../../../types";
+import VerticalAlignSettings from "./VerticalAlignSettings";
+
+export default {
+    name: "pb-editor-page-element-style-settings-cell",
```
adrians5j:
```suggestion
    name: "pb-editor-page-element-style-settings-cell-vertical-align",
```

neatbyte-vnobis:
@SvenAlHamad fixed.
Hope all is good now.

adrians5j:
Doing code review just now, will be posting in ~30mins.

packages/app-page-builder/src/editor/plugins/elementSettings/cell/VerticalAlignSettings.tsx
```diff
@@ -0,0 +1,127 @@
...
+        <Accordion
+            title={"Vertical align"}
+            defaultValue={defaultAccordionValue}
+            icon={
+                <Tooltip content={`Changes will apply for ${config.displayMode}`}>
+                    {config.icon}
+                </Tooltip>
+            }
+        >
+            {parentAlign === "stretch" ? (
```
adrians5j, line 103: Hm... it's a bit hard for me to understand why the `parentAlign` can have the `stretch` value. Can you give me ref maybe?

From what I understood, `parentAlign` is the vertical alignment of parent grid. And values there can only be `AlignTypesEnum`. What am I missing?

adrians5j: Actually, after a bit of looking around, let's just add a comment here:

```
// When parent grid's "Column height" property is set to "Match grid height", its `data.settings.verticalAlign` property is set to "stretch".
```

adrians5j (changes requested):
Did the initial review. On top of the comments, here is some additional feedback.

### 1. Editor Glitching

The best way to test if we broke something is by opening the Welcome to Webiny page. So, I opened it, in the preview looked fine. But then I opened the editor, and noticed there is some glitching happening. Please check the attached video. Not sure if this is something that was introduced by this PR. But in `dev` this does not exist.

https://user-images.githubusercontent.com/5121148/222123590-48f4641f-7976-4cd5-aa76-c754cd0d3b53.mov

### 2. Minor Glitch When Toggling Column Height

Basically, the element height is slightly changing. Please check the attached video.

https://user-images.githubusercontent.com/5121148/222124002-eee90148-8ffc-49ec-a1b8-1620626ecf13.mov

### 3. Old Rendering Engine

This is also a question for @SvenAlHamad .

Are we introducing these features for the old rendering engine or it's just the new one? In any case, we'll need to double check if we broke something there.

Alright, that's all for now. Thanks!

packages/app-page-builder-elements/src/renderers/cell.tsx
neatbyte-vnobis: @adrians5j 
just in case, please check if you are looking on latest changes...
I see here an "outdated" tag for this file, and latest implementation is little bit different

neatbyte-vnobis: @adrians5j 
`cellWidthOffset` is used to define smth like a CSS grid gap, so cells have a spacing between them (that is configurable via UI).
The idea is that we need to reduce columns for few pixels to make a place for a gap between them.
So total width of the row is `=100%` (and not is `=100% spited between all cells + gap size * number of gaps`)

adrians5j: I understand the part related to ensuring spacing between them, and I think it makes sense.

The thing that I don't understand is the use of CSS variable `--cellWidthOffset`. No CSS variables should be used in process of calculating element styles. It's all JS / objects based.

Ultimately, this CSS var should not be here. The width offset should be pulled from element data, the same way we're already doing for, for example, `size` (`const size = element.data?.settings?.grid?.size;`).

neatbyte-vnobis:
@adrians5j
I have looked on these issues and most are fixed now (except legacy rendering - WIP with it).

Regarding flickering: after merging latest dev, I can see little bit of flickering only if dev tools are open in the browser (if not - all looks good). As I see it is more related to additional browser load to sync page with devtools DOM tree, etc.

adrians5j:
Just tried testing the flickering on `dev` and it really is happening there too! 🤯 🤯 🤯 

Alright, we'll have to check this on our end. Super weird issue.

Anyways, continuing with the rest of the review.

packages/app-page-builder/src/editor/plugins/elements/cell/index.tsx
```diff
@@ -67,7 +75,11 @@ const cellPlugin = (args: PbEditorElementPluginArgs = {}): PbEditorPageElementPl
                         ),
                         grid: {
                             size: lodashGet(options, "data.settings.grid.size", 1)
-                        }
+                        },
+                        horizontalAlignFlex: createInitialPerDeviceSettingValue(
```
adrians5j: Hm... double checking: should this be here for legacy rendering engine?

packages/app-page-builder/src/editor/plugins/elementSettings/grid/index.tsx
```diff
@@ -15,7 +16,7 @@ export default [
         name: "pb-editor-page-element-style-settings-grid-settings",
         type: "pb-editor-page-element-style-settings",
         render() {
-            return <GridSettings />;
+            return isLegacyRenderingEngine ? <></> : <GridSettings />;
```
adrians5j: Hm, I may be mistaken, but this here looks to me like we're not rendering any grid settings UI when talking about legacy rendering engine. But is this really what we want?

Shouldn't legacy systems still have at least some of the grid settings available? Like column wrap or wrap direction?

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+            value.rowGap = `${value.rowGap}px`;
+        }
+
+        const cellFullWidthConstrains =
+            rowCount === 1 ||
+            breakpointName === "mobile-landscape" ||
+            breakpointName === "mobile-portrait"; // Only applies to single-row grid or on mobile view.
+
+        // If we have flex direction set to "column" or "column-reverse",
+        // we also want to apply 100% width to direct `pb-cell` elements
```
adrians5j:
```suggestion
        // we also want to apply 100% width to direct `pb-cell` elements.
```

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+        return null;
+    }
+
+    return Object.keys(theme.breakpoints || {}).reduce((returnStyles, breakpointName) => {
+        if (!gridSettings[breakpointName]) {
+            return returnStyles;
+        }
+
+        const columns = element.data?.settings?.grid?.cellsType?.split("-")?.length || 1;
+        const columnSizes: number[] =
```
adrians5j, line 20: Let's add a comment here, saying that `cellsType` is here for backwards compat.

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+            element.data?.settings?.grid?.columnSizes ||
+            element.data?.settings?.grid?.cellsType?.split("-");
+        const rowCount = element.data.settings?.grid?.rowCount || 1;
+        const value = { ...gridSettings[breakpointName] };
```
adrians5j, line 25: For readability, let's just insert an empty line between L23 and L24.

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+    if (!gridSettings) {
+        return null;
+    }
+
+    return Object.keys(theme.breakpoints || {}).reduce((returnStyles, breakpointName) => {
+        if (!gridSettings[breakpointName]) {
+            return returnStyles;
+        }
+
+        const columns = element.data?.settings?.grid?.cellsType?.split("-")?.length || 1;
```
adrians5j: Let's rename `columns` to `columnsCount`. That's also more in line with the below `rowCount` const. Additional nitpick would be to change `rowCount` to `rowsCount`, but no need to go go that far. 🙂

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+        const columnSizes: number[] =
+            element.data?.settings?.grid?.columnSizes ||
+            element.data?.settings?.grid?.cellsType?.split("-");
```
adrians5j: You might want to convert items to integer, since splitting a string (`split("-")`) gives us an array of strings. And `columnsSizes` is an array of integers.

image

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+        const rowCount = element.data.settings?.grid?.rowCount || 1;
+        const value = { ...gridSettings[breakpointName] };
+        const cellWidthOffset = value.columnGap
```
adrians5j: `cellWidthOffset` -> `cellWidthReduction`

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+        if (value.columnGap) {
+            value.columnGap = `${value.columnGap}px`;
+        }
+        if (value.rowGap) {
+            value.rowGap = `${value.rowGap}px`;
+        }
+
+        const cellFullWidthConstrains =
+            rowCount === 1 ||
+            breakpointName === "mobile-landscape" ||
```
adrians5j, line 40: This is a bit problematic part (usage of `"mobile-landscape"` and `"mobile-portrait"`.

Breakpoints are not something that's hardcoded in the file. They are defined in the `theme.ts` file. So, technically, users could rename their breakpoints and this would stop working.

But I'm also curious if this is really needed. 

I believe the need to have cells be 100% for these smaller screens was resolved with flex-direction. I remember writing this code:

```ts
 // If we have flex direction set to "column" or "column-reverse",
        // we also want to apply 100% width to direct `pb-cell` elements.
        if (value.flexDirection !== "row") {
            value["> pb-cell"] = { width: "100%" };
        }
```

With it, the user would be able to choose where the column should stretch to 100% (on which breakpoints).

Ultimately, the question is, can we avoid hardcoding the breakpoint strings here?

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,71 @@
...
+        // If we have flex direction set to "column" or "column-reverse",
+        // we also want to apply 100% width to direct `pb-cell` elements
+        if (
+            value.flexDirection !== "row" &&
+            element.data?.settings?.verticalAlign?.[breakpointName] !== "stretch" &&
+            cellFullWidthConstrains
+        ) {
+            columnSizes.forEach(
+                (_, index) =>
+                    (value[`& > pb-cell:nth-of-type(${columns}n + ${index + 1})`] = {
```
adrians5j: Hmm, couldn't we just use `value["> pb-cell"] = { width: "100%" };` here?

packages/app-page-builder-elements/src/modifiers/styles/verticalAlign.ts
```diff
@@ -21,6 +21,48 @@ const verticalAlign: ElementStylesModifier = ({ element, theme }) => {
...
+                } else {
+                    newStyles[breakpointName].flexDirection = "row";
+                    newStyles[breakpointName].flexWrap = "nowrap";
+                }
+
+                return newStyles;
+            } else if (rowCount > 1) {
+                const columns = element.data?.settings?.grid?.cellsType?.split("-")?.length || 1;
+                const isMobileView =
+                    breakpointName === "mobile-landscape" || breakpointName === "mobile-portrait";
```
adrians5j, line 51: The same as we had in previous comment, can we go without hardcoding the breakpoint strings here?

adrians5j (changes requested):
Did another iteration. Most of the comments are trivial, one or two are something to think about.

packages/app-page-builder/src/editor/plugins/elements/cell/index.tsx
neatbyte-vnobis: I will change it, so this property is added only for non legacy editor.

packages/app-page-builder/src/editor/plugins/elementSettings/grid/index.tsx
neatbyte-vnobis: It can be done...
But column wrap or wrap direction are NOT a part of legacy engine for a VERSION 5.33.5 from the doc. 
https://github.com/webiny/webiny-js/blob/dev/docs/LEGACY_PB_DEVELOPMENT.md

Column wrap or wrap direction were initially done on old engine and migrated to a new engine latter (but I am not sure if old engine was released with them or not)

Also I am trying to simplify this task and PR (as it is already huge, and testing is not simple to check all grid cases)

CC: @SvenAlHamad maybe you have a suggestion about this?

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
neatbyte-vnobis: @adrians5j @SvenAlHamad 
I was not aware that breakpoints are configurable...
I can try your suggestion but it requires a new round of testing
And multi rows are adding more complex behavior so it is not so easy to fix.

Is there any other way to check if it is a "mobile" (small) view?

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
neatbyte-vnobis: It is needed to be compatible (CSS override rules) with separate cell widths. So we cannot simplify it as on different breakpoints these need to be overridden.

See:
https://github.com/webiny/webiny-js/pull/3012/files/a90b3ed5e8792188a09b02a42daed02b1f22e6e2#diff-13096c85e0996ae08e62aa22e75fd0a838da880d85b8d6aef414e53f7d79d18dR58

SvenAlHamad:
From my side this is now working good. I've tested the grid on a new page, and on the existing welcome to webiny page, just to see how it behaves with existing content and it's all good. There is just one minor thing, which is that you can't do no-wrap on a multi-cell grid on mobile, but I'm not even sure how to achieve this at this point. Let's leave it as it is for now, I want to get this into user's hands to start getting feedback and we also need the feature in the unstable branch so we can use it on our own website. 

@adrians5j Have a look at the code when you can and if you're good, let's merge it so we get it into the unstable release.

adrians5j:
> From my side this is now working good. I've tested the grid on a new page, and on the existing welcome to webiny page, just to see how it behaves with existing content and it's all good. There is just one minor thing, which is that you can't do no-wrap on a multi-cell grid on mobile, but I'm not even sure how to achieve this at this point. Let's leave it as it is for now, I want to get this into user's hands to start getting feedback and we also need the feature in the unstable branch so we can use it on our own website. 
> 
> @adrians5j Have a look at the code when you can and if you're good, let's merge it so we get it into the unstable release. 

Did a review today, gonna do another one tomorrow.

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
adrians5j: Did some thinking on this... and switched my answer multiple times already. 😅 

Before we proceed with any steps, a question: couldn't we do something like [this ](https://github.com/webiny/webiny-js/blob/dev/packages/app-page-builder/src/editor/plugins/elements/grid/index.tsx#L115-L124)?

So, as we can see here, upon dropping a grid element onto a page, we're assigning some default values:

```ts
gridSettings: {
                            ...createInitialPerDeviceSettingValue(
                                { flexDirection: "row" },
                                DisplayMode.DESKTOP
                            ),
                            ...createInitialPerDeviceSettingValue(
                                { flexDirection: "column" },
                                DisplayMode.MOBILE_LANDSCAPE
                            )
                        },
```

So, for desktop we're setting "row", and for landscape the "column" flex direction. 

Could we do something similar here too? So, apply a setting (or a set of settings) that will make the column stretch to 100% on mobile? That way tis modifier is just following what's set in page element's settings? And, more importantly, users have the chance to override the behaviour, which, with this solution, is essentially hardcoded.

P.S. It's funny how breakpoints (`DisplayMode` is legacy name) are hardcoded in Grid settings. That's why even if we go with your solution, we're at least in line with the rest. But yeah, let's just quickly check the above question before moving on.

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
adrians5j: Thanks.

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
adrians5j: Thanks!

packages/app-page-builder/src/editor/plugins/elements/cell/index.tsx
adrians5j: Cool.

packages/app-page-builder/src/editor/plugins/elementSettings/grid/index.tsx
adrians5j: No worries, we're aware this PR blew in scope and complexity (hence the thorough testing and code review from our side).

The column wrap and wrap direction actually WERE also introduced with the legacy rendering engine. That's how, to my surprise, I actually discovered that I needed to have this same feature implemented with the new one. 🙂 

So, I was adding it based on what I saw in the old rendering engine. 

Ultimately, because of this fact, some Grid settings should still be rendering when a user is using the legacy rendering engine. In other words, `GridSettings` component should always be rendered. It's just that some of the code within it should not be rendered with the new engine (the stuff you added in this PR).

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
adrians5j: Thanks!

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
adrians5j: Thanks!

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
adrians5j: Thanks!

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    return Object.keys(theme.breakpoints || {}).reduce((returnStyles, breakpointName) => {
+        if (!gridSettings[breakpointName]) {
+            return returnStyles;
+        }
+
+        const columnsCount = element.data?.settings?.grid?.cellsType?.split("-")?.length || 1;
+        const columnSizes: number[] =
+            element.data?.settings?.grid?.columnSizes ||
+            element.data?.settings?.grid?.cellsType?.split("-").map((x: string) => Number(x)); // cellsType is here for backward compatibility.
```
adrians5j:
```suggestion
            element.data?.settings?.grid?.cellsType?.split("-").map(Number); // cellsType is here for backward compatibility.
```

packages/app-page-builder-elements/src/modifiers/styles/verticalAlign.ts
adrians5j: Let's see what we decide [here](https://github.com/webiny/webiny-js/pull/3012#discussion_r1127680391).

adrians5j (changes requested):
We're making good progress. Thanks a lot for all the effort so far @neatbyte-vnobis .

BTW, one thing we should double check is a brand new system - do the default pages still look good there (WTF/Not found pages).

You can check that also via the `/cypress` command (which I'm sure you've already used), and checking the deployed system. If you've already been doing this testing, feel free to ignore this comment.

neatbyte-vnobis:
@adrians5j  @SvenAlHamad 

I did some more testing and it looks like we can remove this breakpoint conditions (check latest commit).
The main things that will change:

Was with breakpoint conditions:
1) We can leave Column wrap: "Wrap" (default value after increasing grid rows), while using Column height: "Match content size" to match expected behavior
2) We can't make multi-row grid wrap to one column on desktop and tablet view

Will be without breakpoint conditions:
1) We have to choose Column wrap: "No wrap" while using Column height: "Match content size" to match expected behavior
2) We can make multi-row grid wrap to one column on desktop and tablet view

image

SvenAlHamad:
@neatbyte-vnobis - I'm not sure about the latest commit change. If you add a multi-row grid, and put a content into one of the cell. Previously if you would to toggle the "Column height" property, it would behave correctly. But after the last commit, it works differently. I need to toggle the "Column wrap" to "no wrap" if I wish to set the "Column height" to "Match content size". I don't think this is what the user would expect if they just wish to change the "Column height" property.

neatbyte-vnobis:
@SvenAlHamad 
Last comment was an attempt to implement @adrians5j [suggestion](https://github.com/webiny/webiny-js/pull/3012#discussion_r1129202033) on how to drop check for breakpoint conditions (mobile size).
(according to my understanding of his comment).

It was partially successful, so I have committed it here for a demo purpose.
It can be easily reverted (as it is a separate commit)

adrians5j:
Checking this.

packages/app-page-builder-elements/src/modifiers/styles/grid.ts
adrians5j: Thanks!

adrians5j:
Sorry for delay, still reading the code and also doing some testing.

What I'm trying to do is test what defects are happening when we remove the mobile checks, and see how those can be fixed.

But, as I'm out of time, I'll just stop here and report a bug I discovered while doing testing. First I thought it was related my changes, but then discovered it wasn't.

https://user-images.githubusercontent.com/5121148/224664643-2efd1868-2e42-48af-a7e0-b16f054200d5.mov

Basically:
1. create a grid
2. set 3x3
3. put a paragraph into first row / each column
4. check different breakpoints - all works good
5. try setting a height to the grid
6. all looks good, except mobile

Note that I was not using the newest changes that you did @neatbyte-vnobis , related to removal of checks of mobile breakpoints. It was a commit before that: `91f6a849`.

image

UPDATE:
I believe this is because of the fact that when I enter 500px as grid height, it gets applied to all breakpoints, not just the current one, like desktop.

neatbyte-vnobis:
@adrians5j bug is fixed

adrians5j:
Saw the bug fixed, thanks for that. Works good here.

Regarding next steps...

Let's merge this, but with the initial version that Sven tested/confirmed (make sure to just include yday's bug fix BTW).

Once this is merged, I'll take one last look at this. Might submit an additional PR, but yeah, you don't need to concern yourself with this any longer.

Thanks again for all the work on this issue, it was a big one! 🍻

neatbyte-vnobis:
@adrians5j 
FYI I have deleted commit with breakpoints removal (so it is not longer here)
It was done before bug fix
image

adrians5j (approved):

adrians5j:
Note for myself: needs migration (added plugins in apps/admin).

PR #3017, opened 2023-02-14
feat: add form field id validation

neatbyte-vnobis:
## Changes
Implements task "Forms: Add a validator to field id"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/EditFieldDialog/GeneralTab.tsx
+++ b/packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/EditFieldDialog/GeneralTab.tsx
@@ -4,6 +4,7 @@ import { Grid, Cell } from "@webiny/ui/Grid";
 import { camelCase } from "lodash";
 import { useFormEditor } from "../../../Context";
 import { validation } from "@webiny/validation";
+import { Validator } from "@webiny/validation/types";
 import { FbFormModelField } from "~/types";
 import { FormRenderPropParams } from "@webiny/form/types";
 
@@ -32,6 +33,20 @@ const GeneralTab: React.FC<GeneralTabProps> = ({ field, form }) => {
         throw new Error("Please enter a unique Field ID");
     }, []);
 
+    const fieldIdValidator: Validator = (value: string | undefined) => {
+        if (!value) {
+            return true;
+        }
+
+        if (/^[a-zA-Z0-9_-]*$/.test(value)) {
+            return true;
+        }
+
+        throw Error(
+            'Field ID can only contain letters, numbers and chars ("-", "_") without any spaces.'
+        );
+    };
+
     const fieldPlugin = getFieldPlugin({ name: field.name });
 
     let additionalSettings: React.ReactNode = null;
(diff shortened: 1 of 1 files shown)


packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/EditFieldDialog/GeneralTab.tsx
```diff
@@ -32,6 +33,20 @@ const GeneralTab: React.FC<GeneralTabProps> = ({ field, form }) => {
         throw new Error("Please enter a unique Field ID");
     }, []);
 
+    const fieldIdValidator: Validator = (value: string | undefined) => {
```
adrians5j: 1. Let's move these two validators outside of the `GeneralTab` component. 
2. Let's assign `Validator` type to `uniqueFieldIdValidator`, the same way you did with `fieldIdValidator`. I know it's not part of your changes, still, let's sync it up.

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/EditFieldDialog/GeneralTab.tsx
```diff
@@ -32,6 +33,20 @@ const GeneralTab: React.FC<GeneralTabProps> = ({ field, form }) => {
...
+        if (!value) {
+            return true;
+        }
+
+        if (/^[a-zA-Z0-9_-]*$/.test(value)) {
+            return true;
+        }
+
+        throw Error(
+            'Field ID can only contain letters, numbers and chars ("-", "_") without any spaces.'
```
adrians5j:
```suggestion
            "Field ID may contain only letters, numbers and "-" and "_" characters."
```

adrians5j (changes requested):
Looks good, just a couple of minor changes.

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/EditFieldDialog/GeneralTab.tsx
neatbyte-vnobis: @adrians5j 
please add more details for a item 2) about `uniqueFieldIdValidator`.
It just that this validator is using not only a field value, but also uses `useFormEditor` hook to check for ID uniqueness.
So if it is outside a component with type `Validator` I have a difficulty to pass `useFormEditor` value in some way due to type restriction.

adrians5j: I hear you @neatbyte-vnobis, I missed the `useFormEditor` fact.

OK, let's then just use the `useCallback` hook when defining the `fieldIdValidator`, the same way it's done with `uniqueFieldIdValidator`. Also, see if the `Validator` can be used with the `uniqueFieldIdValidator`, again just for consistency.

neatbyte-vnobis:
@adrians5j 
Just in case, you have messaged in Slack "provided an alternative, after that we'll be good to merge".
Have I missed some code sample? Or I misunderstood and we can merged this pull as is and "alternative" is a matter of future?

neatbyte-vnobis:
@adrians5j please take a look on comment above

adrians5j:
Hey @neatbyte-vnobis , sorry for the fuss, by "alternative", I meant:

> OK, let's then just use the useCallback hook when defining the fieldIdValidator, the same way it's done with uniqueFieldIdValidator. Also, see if the Validator can be used with the uniqueFieldIdValidator, again just for consistency.

neatbyte-vnobis:
@adrians5j got it
Please take  a look on updated PR

adrians5j (approved):
Thank you @neatbyte-vnobis !

PR #3020, opened 2023-02-16
Use Apollo Client for All GraphQL API Operations

adrians5j:
## Changes
With this PR, we're no longer relying on default data loaders for **Form** and **Pages List** page elements. We're using our own, which are powered by the existing Apollo client.

This will ensure the tenant/locale headers are always included upon issuing GQL operations.

## How Has This Been Tested?
Manual.

## Documentation
Changelog.


Pavel910 (approved):
This looks perfectly ok to me, this isn't the first time we're using component factories, so I don't see any problems with this. Thanks for a quick fix, I'll be using this today!

PR #3029, opened 2023-02-17
feat: bind page-builder events to aco

leopuleo:
## Changes
With this PR we create a new package to bind Page Builder events (i.e. `onPageAfterCreate`) with ACO (creating, updating, and deleting the related search record).

### Events
The following events are intercepted by `api-page-builder-aco` package:

- `onPageAfterCreate`
- `onPageAfterCreateFrom`
- `onPageAfterDelete`
- `onPageAfterPublish`
- `onPageAfterUnpublish`
- `onPageAfterUpdate`

### Processors
Page content is processed to extract meaningful information. The following elements are processed: 

- `button`
- `image`
- `images`
- `paragraph`

### Remove `originalId` -> set directly `entryId`
We use the new feature introduced by #3019, setting the `entryId` using the external `id` passed on entry creation.
Under the hood, the system still creates both `entryId` (`abcdefg`) and revision `id` (`abcdefg#0001`).

We still keep using the id without revisions.

Closes [WEB-2062](https://webiny.atlassian.net/browse/)

## How Has This Been Tested?

Manually and Jest

## Documentation
README inside package


packages/api-aco/__tests__/graphql/record.gql.ts
```diff
@@ -32,7 +32,7 @@ export const CREATE_RECORD = /* GraphQL */ `
     mutation CreateRecord($data: SearchRecordCreateInput!) {
         search {
             createRecord(data: $data) {
-                data ${DATA_FIELD("id")}
+                data ${DATA_FIELD()}
```
Pavel910: Several things about this `DATA_FIELD()`: 
- you never use the `extra` parameter
- you can convert this into a const value (not a function), and remove the curly braces from it, then the `extra` thing can be added where necessary, without passing any arguments, like this:

```
const DATA_FIELD = /* GraphQL */ `
    id
    type
`;

export const LIST_RECORDS = /* GraphQL */ `
    query ListRecords($where: SearchRecordListWhereInput, $search: String, $limit: Int, $after: String, $sort: [AcoListSort]) {
        search {
            listRecords(where: $where, search: $search, limit: $limit, after: $after, sort: $sort) {
                data {
                    ${DATA_FIELD}
                    extraField                 # <---- extra fields
                }
                meta ${LIST_META_FIELD}
                error ${ERROR_FIELD}
            }
        }
    }
`;
```

packages/api-aco/src/record/record.so.ts
```diff
@@ -28,29 +29,41 @@ export const createSearchRecordOperations = (
...
 
     return {
         getRecordModel,
-        getRecord,
+        async getRecord({ id }) {
+            security.disableAuthorization();
+
+            const record = await getRecord(id);
+
+            security.enableAuthorization();
```
Pavel910: I have a nice utility for you to avoid this repetitive fiddling with security: https://github.com/webiny/webiny-examples/blob/master/gated-content/packages/api/src/content/onTenantAfterInstall.ts#L6-L10

packages/api-page-builder-aco/__tests__/graphql/record.gql.ts
```diff
@@ -0,0 +1,32 @@
+const DATA_FIELD = (extra = "") => /* GraphQL */ `
```
Pavel910, line 1: Same as in the related comment at the beginning of PR.

packages/api-page-builder-aco/src/page/hooks/onPageAfterCreate.hook.ts
```diff
@@ -0,0 +1,25 @@
...
+     */
+    pageBuilder.onPageAfterCreate.subscribe(async ({ page, meta }) => {
+        try {
+            const payload = await createPageRecordPayload(context, page, meta);
+            await aco.search.create<PbPageRecordData>(payload);
+        } catch (error) {
+            throw WebinyError.from(error, {
+                message: "Error while executing onPageAfterCreateHook hook",
+                code: "ACO_AFTER_PAGE_CREATE_HOOK"
+            });
```
Pavel910, line 22: My question here is: do we really want to throw an error and kill the whole page creation because a search record didn't get stored? 🤔 Not sure if this is the right approach. @adrians5j any thoughts?

packages/api-page-builder-aco/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+>;
+
+export interface PageSearchProcessor {
+    (params: PageSearchProcessorParams): string;
+}
+
+export interface PbAcoContext extends BaseContext, AcoContext, PbContext {
+    pageBuilderAco: {
+        addPageSearchProcessor(processor: PageSearchProcessor): void;
+        processPageSearchContent(content: Page): Promise<string>;
```
Pavel910: `processPageSearchContent` is a bit vague. Process how? See what I mean, what will happen after the processing takes place? It's unclear. Maybe call this `getSearchablePageContent`? 🤔

scripts/listPackagesWithTests.js
```diff
@@ -65,6 +65,12 @@ const CUSTOM_HANDLERS = {
     // Setup "api-aco" tests.
     "api-aco": () => {
         return ["packages/api-aco/* --keyword=cms:ddb --keyword=aco:base"];
+    },
+    // Setup "api-page-builder-aco" tests.
+    "api-page-builder-aco": () => {
+        return [
+            "packages/api-page-builder-aco/* --keyword=cms:ddb --keyword=aco:base --keyword=pb:ddb --keyword=pb:base --keyword=api-page-builder-aco:base"
```
Pavel910, line 72: This is nice! 👍

Pavel910 (changes requested):
A few minor changes, but overall, this looks really nice!

packages/api-page-builder-aco/src/page/hooks/onPageAfterCreate.hook.ts
leopuleo: In case of the search record creation error, the admin user won't be to list it in the page builder list. This is why I decided to kill the page creation.

Pavel910: Ah, in that case it makes sense. I'm resolving this conversation.

packages/api-page-builder-aco/src/types.ts
leopuleo: Make sense 👍

Pavel910 (approved):

PR #3030, opened 2023-02-18
feat: add infinitive scroll to the form submissions

neatbyte-vnobis:
## Changes
Implements issue "Forms: Add infinitive scroll to the form submissions"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-form-builder/src/admin/plugins/formDetails/formSubmissions/FormSubmissionsList/useSubmissions.ts
+++ b/packages/app-form-builder/src/admin/plugins/formDetails/formSubmissions/FormSubmissionsList/useSubmissions.ts
@@ -1,21 +1,19 @@
-import { useApolloClient, useMutation } from "@apollo/react-hooks";
-import { useCallback, useEffect, useReducer } from "react";
+import { useMutation, useQuery } from "@apollo/react-hooks";
+import { useCallback, useReducer } from "react";
+import debounce from "lodash/debounce";
+import get from "lodash/get";
 import { useSnackbar } from "@webiny/app-admin/hooks/useSnackbar";
-import { FetchPolicy } from "apollo-client";
-import { FbFormModel, FbFormSubmissionData } from "~/types";
+import { FbFormModel } from "~/types";
 import {
     ExportFormSubmissionsMutationResponse,
     ExportFormSubmissionsMutationVariables,
     EXPORT_FORM_SUBMISSIONS,
-    ListFormSubmissionsQueryResponse,
-    ListFormSubmissionsQueryVariables,
     LIST_FORM_SUBMISSIONS
 } from "~/admin/graphql";
 
 interface State {
     loading: boolean;
     exportInProgress: boolean;
-    submissions: FbFormSubmissionData[];
     sort: string[];
 }
 interface Reducer {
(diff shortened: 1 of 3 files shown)


SvenAlHamad:
I'm not sure if I'm missing something, but this doesn't seem to be loading new results as the user scrolls to the bottom of the list? I've added 60 submissions and it loaded all of them. How would this behave if the user has 1k submissions?

neatbyte-vnobis:
@SvenAlHamad 
Oh, I see the issue.
I need to update this PR, please skip it for now.

SvenAlHamad:
@neatbyte-vnobis - no worries. By the way, we already have an infinitive scroll implementation in the list of Pages in the Page Builder, so you can check that implementation for reference.

neatbyte-vnobis:
@SvenAlHamad it is already done (see commit from yesterday), please take a look

SvenAlHamad:
@neatbyte-vnobis - I've tested this one, and it works fine on 16:9 displays, but if you have a display that's 4:3 then you might end up with a case where you see 10 form entries and have a white space at the bottom, meaning there's no scroll, so you can't trigger a pagination to see the remaining items. I think it's best to load 20 results on the first load just to be safe.

neatbyte-vnobis:
@SvenAlHamad issue fixed, please take a look

SvenAlHamad:
@adrians5j - on this one, after the last pull I'm no longer able to run the watch. It's throwing me this error, which seems to be related to WCP:
CleanShot 2023-03-07 at 13 52 48@2x

I cleared my node_modules, re-ran yarn and yarn build, but no luck.

adrians5j:
Looks like `-p admin` is missing, can you try adding it?

This is only needed when doing development in Webiny's repo. Not needed in users' projects.

Can you give it a try maybe?

packages/app-form-builder/src/admin/plugins/formDetails/formSubmissions/FormSubmissionsList/useSubmissions.ts
```diff
@@ -96,29 +91,19 @@ export const useSubmissions = (form: Pick<FbFormModel, "id">) => {
...
-        refresh: () => loadSubmissions("network-only"),
-        submissions: state.submissions,
-        setSorter: (sort: string[]) => {
-            setState({ sort });
-        },
-        hasPreviousPage,
-        hasNextPage,
-        nextPage: () => {
-            if (!hasNextPage) {
+        loading: [listQuery].find(item => item.loading),
```
adrians5j: Let's use `some` instead of `find`, that way the flag is boolean, and no `Boolean` needs to be used in [the other file](https://github.com/webiny/webiny-js/pull/3030/files#diff-0674c038a175333987e6672afe7cac92fbd6a669e7904234c81c0df5038f4fcfR92).

packages/app-form-builder/src/admin/plugins/formDetails/formSubmissions/FormSubmissionsList/useSubmissions.ts
```diff
@@ -27,47 +23,46 @@ interface Reducer {
...
-            ListFormSubmissionsQueryResponse,
-            ListFormSubmissionsQueryVariables
-        >({
-            query: LIST_FORM_SUBMISSIONS,
-            variables: { form: form.id, sort: state.sort, after: state.cursors[state.page] },
-            fetchPolicy
-        });
+    const loadMoreOnScroll = useCallback(
+        debounce(scrollFrame => {
+            if (scrollFrame.top > 0.9) {
```
adrians5j: We should also not be doing anything if `loading: true` (despite the fact that the function is debounced).

adrians5j (changes requested):
Two minor comments, once we do these and @SvenAlHamad approves, we can merge this.

Thanks!

SvenAlHamad:
Got it ... it's working fine now. Also, the issue with the pagination is now resolved. From my end, this is good to be merged.

adrians5j (approved):
Thank you @neatbyte-vnobis !

PR #3031, opened 2023-02-18
feat: add other as an option to specific fields

neatbyte-vnobis:
## Changes
Implement issue "Forms: Add “other” as an option to specific fields"

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
This works well, but I would just suggest one improvement. When we have the "other" option active, can we hide the other input and only show it when the user clicks on the other option in the form and then we display the input and automatically focus on it.

SvenAlHamad:
I'm good with this implementation. @adrians5j or @Pavel910 this is ready for a code review.

adrians5j:
Checking.

apps/theme/layouts/forms/DefaultFormLayout/fields/Checkbox.tsx
```diff
@@ -99,6 +119,34 @@ export const CheckboxField: React.FC<CheckboxProps> = ({ field }) => {
...
+                        }}
+                    />
+                    <label htmlFor={"checkbox-" + fieldId + otherOption.value}>
+                        {otherOption.label}
+                    </label>
+                    <OtherInput
+                        name={`${fieldId}Other`}
+                        id={`${fieldId}Other`}
+                        ref={otherInputRef}
+                        disabled={!checked({ option: otherOption, value })}
```
adrians5j: Couldn't we simply render or not render the `OtherInput`, depending on the `checked` call result? A bit confused why we're using `disabled` here and then hiding the field if it's true (via above CSS).

apps/theme/layouts/forms/DefaultFormLayout/fields/Checkbox.tsx
```diff
@@ -99,6 +119,34 @@ export const CheckboxField: React.FC<CheckboxProps> = ({ field }) => {
...
+                    <CheckboxButton
+                        name={fieldId}
+                        type="checkbox"
+                        id={"checkbox-" + fieldId + otherOption.value}
+                        checked={checked({ option: otherOption, value })}
+                        onChange={e => {
+                            change({ option: otherOption, value, onChange });
+                            if (e.target.checked && otherInputRef.current) {
+                                otherInputRef.current.disabled = false;
+                                otherInputRef.current.focus();
```
adrians5j: Would the following eliminate all these lines of code?
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

Let's try that, but I believe my previous comment would also need to be applied in order for the `autofocus` to work.

packages/app-form-builder/src/admin/plugins/editor/formFields/checkboxes.tsx
```diff
@@ -21,15 +21,16 @@ const plugin: FbBuilderFieldPlugin = {
...
+                    otherOption: false
                 }
             };
         },
         renderSettings({ form }) {
             return (
                 <Grid>
                     <Cell span={12}>
-                        <OptionsList form={form} multiple />
+                        <OptionsList form={form} multiple otherOptionSwitch />
```
adrians5j:
```suggestion
                        <OptionsList form={form} multiple otherOption />
```

Let's rename to `otherOption`.

adrians5j (changes requested):
Although the feature doesn't look like it, the PR is not trivial. A lot of stuff to go through and wrap my head around.

Let's start with a couple of comments I left and we'll do additional iterations if needed.

BTW something I also noticed while giving this new feature a try. This "Allow other" text should be a bit smaller:

image

neatbyte-vnobis:
@adrians5j 
Comments above are fixed, please take a look.

neatbyte-vnobis:
@adrians5j 
please take a look on this PR
I hope we can merge it (as previous comments are fixed)

adrians5j (approved):
Nice, thanks @neatbyte-vnobis !

PR #3041, opened 2023-02-20
fix(prerendering): add support for website re-render across all tenants

Pavel910:
## Changes
This PR upgrades the PS subscriber Lambda to be able to handle a wildcard `*` tenant argument. This is necessary when we need to trigger a full website re-render across all tenants. In particular, this is needed after website redeployment. To make this work, I had to expand the storage operations with a method to fetch all tenants from the database. 

With this, we can now issue render tasks like this: `{ path: "*", tenant: "*" }`, meaning, all paths, on all tenants.

## How Has This Been Tested?
Manually, in a multi-tenant setup.


adrians5j (approved):
Nice one, thank you! :)

PR #3043, opened 2023-02-20
Improve admin app components

Pavel910:
## Changes
This PR introduces the following improvements:
- `Permissions` component can now accept an array of plugins to render. The default behavior is preserved, and works as before.
- `SearchUI` component now accepts an `onEnter` prop
- `SimpleFormFooter` component styles are fixed to reflect the global button alignment (all buttons aligned to the right)
- Admin Users installer form footer styles are adjusted to work nicely with the previous change ⬆️ 
- `@webiny/app-admin` theme now has default `h1-h6` styles defined (they got lost with the 5.34.0 release, due to theme refactors)

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/app-admin-users-cognito/src/plugins/installation.tsx
+++ b/packages/app-admin-users-cognito/src/plugins/installation.tsx
@@ -1,4 +1,5 @@
 import gql from "graphql-tag";
+import { css } from "emotion";
 import React, { useState, useCallback } from "react";
 import { useApolloClient } from "@apollo/react-hooks";
 import { Form } from "@webiny/form";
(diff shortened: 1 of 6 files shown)


adrians5j (approved):
The only thing I'm confused about is why we're returning an empty `systemPlugins` array [here](https://github.com/webiny/webiny-js/pull/3043/files#diff-579ac7531dac00bfb27a5e5d1b16cb6d8bcd9c01bbcbddc06e76f84eca7c8b12R21).

I guess it's BC?

I saw h1...h6, and of course, it's OK. Let's discuss this though on one of our calls maybe.

Otherwise, hit the merge button. :)

Pavel910:
@adrians5j yes, it's purely to respect the existing interface, because internally, Permissions component uses `systemPlugins`. This way it's less `if`s.

PR #3044, opened 2023-02-20
feat(page-builder): add slug and tags fields

Pavel910:
## Changes
This PR does the following:
- adds `slug` and `tags` fields to the Page Template entity; this was absolutely necessary for programmatic usage
- changes primary PK to use template ID: `T#${tenant}#L#${locale}#PB#TEMPLATE#${id}`
- adds GSI1 for template listing: `T#${tenant}#L#${locale}#PB#TEMPLATES`
- uses GSI1 to query for template by `slug` (note that query by slug doesn't use dataloader, as dataloader uses `batchGetItem`, which doesn't support queries by GSIs, only by primary index )
- moves `createPageFromTemplate` logic to the API; this makes it programmatically usable, and easy to maintain in one place
- makes `PbPageTemplate` GraphQL type fields `required`; this makes it easier to reason about the returned data, and write less checks for undefined values. It's easier to set [default values](https://github.com/webiny/webiny-js/compare/next...feat/pb-template-slug?expand=1#diff-50e8a965dbffbf6812806f9d59c4b6a32e0d685280f5d144a1c8c5f0ae25cdd4R81-R85), than constantly check for undefined values on the client
- improves TS types to have less `any`s and have better autocomplete on template variables and page content
- exports `useElementVariables` from the root of the package, so it is easily accessible to 3rd party developers
- removes `content` field from `ListPageTemplates` query in page templates data list; this makes API calls faster, and we don't use that data on the client side anyway.
- moves `templateId` and `templateVariables` to a dedicated `template` namespace within content element data: 
`{ template: { slug: "template-slug", variables: [] } }`, and uses template `slug` instead of exact id, as it makes programmatic usage a lot easier, and it takes less effort to copy templates between tenants, because we no longer need to worry about ID mismatches; we just get templates by slug.

## How Has This Been Tested?
Manually, via UI and programmatically.

### DIFF
--- a/packages/api-page-builder/src/graphql/graphql/pages.gql.ts
+++ b/packages/api-page-builder/src/graphql/graphql/pages.gql.ts
@@ -5,13 +5,17 @@ import {
     NotFoundResponse
 } from "@webiny/handler-graphql/responses";
 import { GraphQLSchemaPlugin } from "@webiny/handler-graphql/types";
-import { Page, PbContext, PageSecurityPermission } from "~/types";
+import { Page, PbContext, PageSecurityPermission, PageContentWithTemplate } from "~/types";
 import WebinyError from "@webiny/error";
 import resolve from "./utils/resolve";
 import { createPageSettingsGraphQL } from "./pages/pageSettings";
 import { fetchEmbed, findProvider } from "./pages/oEmbed";
 import lodashGet from "lodash/get";
 import checkBasePermissions from "~/graphql/crud/utils/checkBasePermissions";
+
+function hasTemplate(content: Page["content"]): content is PageContentWithTemplate {
+    return content?.data?.template;
+}
 
 const createBasePageGraphQL = (): GraphQLSchemaPlugin<PbContext> => {
     return {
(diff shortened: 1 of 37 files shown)


packages/app-page-builder/src/admin/views/Pages/Pages.tsx
```diff
@@ -39,8 +37,16 @@ const Pages: React.FC = () => {
     const onCreatePage = useCallback(async (template?: PbPageTemplate) => {
         setIsLoading(true);
         try {
-            const res = await create({
-                variables: { category: "static" }, // hardcoded for now
+            const MUTATION = template ? CREATE_PAGE_FROM_TEMPLATE : CREATE_PAGE;
+            const variables = {
+                // category is temporarily hardcoded
+                category: "static",
```
adrians5j, line 43: Double checking - should this stay hardcoded? Probably we need to create the UI before undoing this?

packages/api-page-builder/src/graphql/graphql/pages.gql.ts
```diff
@@ -5,14 +5,18 @@ import {
...
-import { Page, PbContext, PageSecurityPermission } from "~/types";
+import { Page, PbContext, PageSecurityPermission, PageContentWithTemplate } from "~/types";
 import WebinyError from "@webiny/error";
 import resolve from "./utils/resolve";
 import { createPageSettingsGraphQL } from "./pages/pageSettings";
 import { fetchEmbed, findProvider } from "./pages/oEmbed";
 import lodashGet from "lodash/get";
 import checkBasePermissions from "~/graphql/crud/utils/checkBasePermissions";
 
+function hasTemplate(content: Page["content"]): content is PageContentWithTemplate {
```
adrians5j, line 16: `is` operator, didn't know about this one 😮 
Will check it out.

adrians5j (approved):
Nice one, thanks!

PR #3055, opened 2023-02-24
fix(api-headless-cms-ddb-es): ddb storage plugin names

brunozoric:
## Changes
Name plugins which are registered for the StorageOperations.
This helps with debugging.

## How Has This Been Tested?
Jest and manually.


adrians5j (approved):
LGTM

PR #3056, opened 2023-02-27
fix: replace heading and quote variable inputs with rich text ones

neatbyte-vnobis:
## Changes
Fix for bug "Bug on block variables"

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
@neatbyte-vnobis - I've tested this one, and part of the problem is solved, but for whatever reason when I underline the text it the variable input in the sidebar, it show up fine, but if I do bold or italic, it only shows up in the editor, but in the sidebar that formatting is not visible. It applies both to the paragraph and the heading element.

neatbyte-vnobis:
@SvenAlHamad 
I have found few issues and I want to get your feedback on desired behavior for them.
image

Context: I am talking about Typography option (not HTML tag h1/h2 selection), so H1 === Heading1 in the dropdown, etc.

Issues with italic and bold texts are fixed for NOT bold typographers (aka h2/h3/etc).
But on bold typography H1 or Quote the Rich Text Editor component is breaking down for bold text.
It happens when user has heading H1 and is trying to apply bold to part of text...
Initially text is bold as H1 is defined as bold font, so in such case it becomes not bold (aka html `span` with CSS `weight: 400` is added).
And this span is breaking the Rich Text Editor component used for variables values on Page Editor (it causes bugs if user apply bold for part of text there).

@adrians5j @Pavel910 Maybe we can disable B (bold) option on typography H1 and Quote?
I am struggling with a workaround for this, but it looks like a bunch of if-else which is not looking good...

PS: there is a Typography bug when desktop H1 is affection tablet/mobile views (also makes them bold, even if not bold typography `Heading 2` is selected)

adrians5j:
I can double check this one, it might be a CSS-related issue.

SvenAlHamad:
If it's not an easy CSS fix, let's just remove the bold/italic option as once we finalize and merge the lexical editor this might behave differently.

neatbyte-vnobis:
@SvenAlHamad 
So I need to remove bold/italic options from Rich Text Editor used for variables, am i right?
What about underline? Or hide all 3 of them?

PS: undescore is working well, and italic can be fixed (one style need to be added), only bold is a problem

SvenAlHamad:
Let's remove all 3 options for now. This will be fixed soon once we switch to Lexical editor.

adrians5j (approved):
Thanks!

PR #3058, opened 2023-02-27
Existing VPCs Feature - Use `PulumiAppParam` For the Extended `vpc` Param

adrians5j:
## Changes
With this PR, we ensure the improved `vpc` param that's available in Pulumi apps is not just `boolean | xyzPulumiAppAdvancedVpcParams`, but `PulumiAppParam<boolean | xyzPulumiAppAdvancedVpcParams>;`.

This gives more control, as the parameter can not only be one of the two types defined in the union, but also a callback that returns it. For example:

```ts
import { createCoreApp } from "@webiny/serverless-cms-aws/enterprise";

    vpc: ({ params }) => {
        const { env } = params.run;
        if (env === "prod") {
            const subnetIds = ["subnet-1", "subnet-2", "subnet-3"];
            const securityGroupIds = ["sg-1"];
            
            return {
                useExistingVpc: {
                    elasticSearchDomainVpcConfig: { subnetIds, securityGroupIds },
                    lambdaFunctionsVpcConfig: { subnetIds, securityGroupIds }
                }
            };
        }

        // Not "prod"? Then no extra VPC config needs to be applied.
        return undefined;
    }
});

```

## How Has This Been Tested?
Manual.

## Documentation
Changelog.


Pavel910 (approved):

adrians5j:
Thanks @Pavel910 🥇

PR #3067, opened 2023-03-01
refactor: use search records inside page builder

leopuleo:
## Changes
With this PR we refactor page builder list and use the new `searchRecord` entity instead.

No major changes have been made since the first iteration of ACO implementation: the only noticeable one is the fact we are fetching records from the server each time the list is loaded and we want to fetch the most up-to-date record any time we upgrade it.

**Reasons:**
since the record creation, update and deletion are achieved by subscribing to specific topics (ie: onAfterPageCreate), we need to sync the record state "manually" in order to render up-to-date information.

Data workflow on new page creation

Closes [2141](https://webiny.atlassian.net/browse/WEB-2141)

## How Has This Been Tested?
Cypress & Jest

## Documentation
TBD


packages/app-aco/src/contexts/records.tsx
```diff
@@ -73,19 +76,44 @@ export const SearchRecordsProvider = ({ children }: Props) => {
...
-                throw new Error("`folderId` is mandatory");
+        async listRecords(
+            type?: string,
+            folderId?: string,
+            limit = 20,
+            after?: string,
+            sorting?: string[]
+        ) {
+            if (!folderId || !type) {
+                throw new Error("`folderId` and `type` are mandatory");
```
Pavel910, line 89: `folderId` OR `type`,  it's one or the other from what I understand. If they're both mandatory, then the `listRecords` method type is wrong.

packages/app-aco/src/contexts/records.tsx
```diff
@@ -73,19 +76,44 @@ export const SearchRecordsProvider = ({ children }: Props) => {
...
-            if (!folderId) {
-                throw new Error("`folderId` is mandatory");
+        async listRecords(
+            type?: string,
+            folderId?: string,
+            limit = 20,
+            after?: string,
+            sorting?: string[]
+        ) {
+            if (!folderId || !type) {
```
Pavel910, line 88: This should be `if (!folderId && !type) {`.

You have both `type` and `folderId` set to `optional`. That means that they're either both optional, or one of them is mandatory, but not both. If one of them is mandatory, you need `&&`.

packages/app-aco/src/graphql/records.gql.ts
```diff
@@ -42,9 +42,9 @@ export const CREATE_RECORD = gql`
 `;
 
 export const LIST_RECORDS = gql`
-    query ListRecords ($type: String!, $location: SearchLocationInput!, $limit: Int, $after: String) {
+    query ListRecords ($type: String!, $location: SearchLocationInput!, $limit: Int, $after: String, $sort: [AcoListSort!]) {
```
Pavel910: Regarding the `sort` array of enum values: let's convert this type from enum to a regular input type. We recently got bitten pretty badly with enums, and we want to avoid them, especially on new apps.

So the goal is to make sorters extensible via plugins, but `enum` cannot be extended in GraphQL. So we need to convert it to a regular input type:

```
enum AcoSortDirection {
    ASC
    DESC
}

input AcoSort {
    id: SortDirection
    createdOn: SortDirection
}
```

This way, we can add new fields to the `AcoSort` input via plugins. I know it's an annoying request, but it will save us lots of troubles later.

packages/app-page-builder/src/admin/plugins/routes.tsx
```diff
@@ -61,18 +61,22 @@ const plugins: RoutePlugin[] = [
             />
         )
     },
+    /**
+     * @deprecated since version 5.34
```
Pavel910: Since 5.35., hopefully :)

Pavel910 (changes requested):
Left some comments on the code, and adding these here after visual review of the app:

1) Action icons: let's keep the same icons on the menu, as we had them previously in the old view:

CleanShot 2023-03-13 at 11 52 56

2) Folders icons: let's use Material icons in the tree view, to make the UI consistent: https://fonts.google.com/icons?selected=Material+Icons&icon.query=folder&icon.style=Outlined

Use `folder` and `folder_open` (from the npm package for icons), and use the `outlined` version, to use the "lighter" version of those icons.

3) Folder dialog: 
- auto-generate slug when the "name" input is blurred, here's an example so you can just C/P: `packages/app-page-builder/src/templateEditor/config/editorBar/TemplateSettings/TemplateSettingsModal.tsx:91`

- don't align the `Cancel` button to the left, just let it sit next to the primary button. The styles have already been adjusted in one of the recent PRs so it should look nice without any custom styling. Here's a reference to what it should look like: `packages/app-page-builder/src/templateEditor/config/editorBar/TemplateSettings/TemplateSettingsModal.tsx:129`

CleanShot 2023-03-13 at 11 58 57

4) Folders and table alignment: I would align the titles of the 2 sections to look the same, and be on the same level:

CleanShot 2023-03-13 at 12 05 00

5) Apply the same button changes mentioned in 3) to the following dialogs as well:
CleanShot 2023-03-13 at 12 06 58
CleanShot 2023-03-13 at 12 09 16

packages/app-aco/src/contexts/records.tsx
leopuleo: We need to use `useRecords` to import the new method `getRecord` (useful for sinking records after page actions).

Now, we use `useRecords` without `folderId` and `type`...so we don't need to propagate these information to multiple components.

In this line, we check that `folderId` and `type` are passed to `listRecords`...both are required for this method.

Pavel910: If so, then please add a comment describing the various scenarios of calling this method. Because the method type shows that both parameters are optional.

Pavel910 (approved):

PR #3070, opened 2023-03-01
fix(app-page-builder-elements): exclude page only if Page context exists

Pavel910:
## Changes
This PR resolves a bug which is caused by the use of the `usePage()` hook in the `pages-list` renderer. `PageContext` is not always present; page template editor, and block editor, do not work with pages, but they still can use the `pages-list` element. This caused the "not-so-happy" execution path, which runs into a `usePage()` hook, which throws an error if the context is missing. 

The solution was to introduce a `useOptionalPage()` hook, which does not throw an error, and allows the developer to decide what to do, if the context is not present.

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/app-page-builder-elements/src/hooks/usePage.ts
+++ b/packages/app-page-builder-elements/src/hooks/usePage.ts
@@ -11,3 +11,7 @@ export function usePage() {
 
     return context;
 }
+
+export function useOptionalPage() {
+    return useContext(PageContext);
+}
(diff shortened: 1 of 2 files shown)


adrians5j (approved):

Pavel910:
Closing this PR as this fix is merged into `dev` branch, and we'll just merge `dev` into `next`, so this PR is not necessary.

PR #3072, opened 2023-03-02
Pulumi Stack Output - Automatically Return Custom Domains if Provided

adrians5j:
## Changes
At the moment, users follow this guide when linking custom domains with their Webiny projects: https://www.webiny.com/docs/infrastructure/additional-resources/connect-custom-domain

The problem with this is that when you start the Admin app, the app is actually still issuing API HTTP requests to the default API Cloudfront domain, and not to the custom domain that may have been assigned for the API app.

This PR addresses this and ensures that if a custom domain was provided, that it's being used in the Admin app, instead of the default Cloudfront domain.

In order to achieve this, we've updated the generation of stack output. This was only required for the API app, but I decided to apply the same change for all apps. So, from now on, for all apps, in the stack output, `appUrl` / `appDomain` and similar variables will always show custom domain, if one was assigned. Additionally, just in case, stack output will still include default Cloudfront domain, via the `cloudFrontAppUrl` / `cloudfrontAppDomain` variables.

Finally, note that users can actually provide multiple domains via the `domains.domains` param in their `webiny.application.ts` configuration file. When doing the above, we only take the first one into consideration, at least for now.

## How Has This Been Tested?
Manual.

## Documentation
Changelog. Will updated the mentioned doc article.


Pavel910 (approved):

packages/pulumi-aws/src/apps/api/createApiPulumiApp.ts
```diff
@@ -198,15 +199,34 @@ export const createApiPulumiApp = (projectAppParams: CreateApiPulumiAppParams =
...
+            let apiUrl = cloudfrontApiUrl;
+
             const domains = app.getParam(projectAppParams.domains);
             if (domains) {
                 applyCustomDomain(cloudfront, domains);
+
+                // Instead of the default Cloudfront domain, we use the first custom
+                // domain that was provided via the `domains.domains` array, and that
+                // is what will be included in the final stack output.
+                const domainsOutput = pulumi.output(domains.domains);
```
jenslauterbach: As discussed in Slack, the solution should not rely on the `domains` callback but on the underlying Pulumi `cloudfront` resource.

Since the changes from the `domains` callback are applied to that `cloudfront` resource anyway, it would be more future proof.

See: https://github.com/webiny/webiny-js/blob/de72acea7b76c87b613f65991b91894420270fba/packages/pulumi-aws/src/apps/customDomain.ts

jenslauterbach (changes requested):

packages/pulumi-aws/src/apps/api/createApiPulumiApp.ts
adrians5j: Yes! 

Will let you know once this is ready for another look @jenslauterbach!

Thank you 🙏🏻 😊

Pavel910 (approved):

PR #3074, opened 2023-03-02
Introduced `webiny build` Command

adrians5j:
## Changes
This PR introduced the long awaited `webiny build` command. With this command, users will no longer have to run `yarn webiny deploy --env xyz --no-deploy` in order to achieve the same effect.

## How Has This Been Tested?
Manually.

## Documentation
Changelog / updated list of commands on the Webiny CLI page.


Pavel910 (approved):
Nice addition 👍

PR #3075, opened 2023-03-02
Webiny CLI - Improved Error Reporting When There's No Internal Workspace Folder

adrians5j:
## Changes
Some of the Webiny CLI commands require the project application to be previously built. For example, the `webiny output` and `webiny pulumi` commands.

Prior to this PR, if the application wasn't built upon running one of the two, users would get presented with a really bad error report:

image

With this PR, we're improving the error report. In the same scenario, the user will now see the following:
image

Maybe in the future we'll find a way so that user's don't have to deal with this at all, but for now, this is the first step.

## How Has This Been Tested?
Manual.

## Documentation
Changelog.

### DIFF
--- a/packages/cli-plugin-deploy-pulumi/commands/build.js
+++ b/packages/cli-plugin-deploy-pulumi/commands/build.js
@@ -1,0 +1,30 @@
+const buildPackages = require("./deploy/buildPackages");
+const { createPulumiCommand, runHook } = require("../utils");
+
+module.exports = (params, context) => {
+    const command = createPulumiCommand({
+        name: "build",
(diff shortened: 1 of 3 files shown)


Pavel910 (approved):
Still better than the original error which tells you nothing 😂 👍

adrians5j:
Just awful I know! 🤦🏻

adrians5j:
@Pavel910 , I decided to simplify the message here 🤦🏻 

This is the new message:
image

Pavel910:
@adrians5j I like this better, good call 👍

PR #3077, opened 2023-03-03
feat: add the ability to import/export forms

neatbyte-vnobis:
## Changes
Implement task "Forms: Add the ability to import/export forms"

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-form-builder/src/admin/plugins/editor/defaultBar/ExportButton/useExportFormDialog.tsx
```diff
@@ -0,0 +1,166 @@
...
+    where?: Record<string, any>;
+    sort?: string;
+    search?: { query: string };
+}
+
+const ExportFormLoadingDialogMessage: React.FC<ExportFormsDialogProps> = props => {
+    const { exportForm } = useExportForm();
+    const {
+        exportPageData: { revisionType }
+    } = usePageBuilder();
```
neatbyte-vnobis: @adrians5j @Pavel910 
FYI: Page builder context is used here (from `app-page-builder` package).
And I am not sure if it is a good thing to use it in the forms package.

adrians5j: @Pavel910 do you know if there's anybody in the team who could provide guidance to @neatbyte-vnobis here?

Never worked with import/export, but I believe other members have.

Pavel910: @neatbyte-vnobis if this is new, and you copy/pasted this from PB export to get the overall structure, then definitely try to clear out references to PB from the form builder.

neatbyte-vnobis: @Pavel910 @adrians5j 
Current approach was done using Page Builder as a reference (to keep such pages with same code style).
I am not sure if it is a great idea to use new/existing context to just pass few input values from one modal to another on the same page...
But it was done in such way for Page export/import, so I am keeping same pattern on other pages.

I have tried to use existing Forms Context (added few values for these modals), but I have got some issues with state refresh (somehow new values were not grabbed properly from the context).
I have wasted few hours trying to fix it, but failed in the end.
So I have dropped this idea and reused Page Builder context as it is accessible there anyway (somehow??? is it ok??? just in case it is not related to my code changes :) ).

Pavel910: @neatbyte-vnobis I'll dedicate some time to this PR later today, clone your branch, and do a more close review. I'm a bit confused, so I don't want to throw random comments. Let me review this properly, and then give feedback. Thanks!

Pavel910: @neatbyte-vnobis I had a better look at this.  To be honest, this entire import/export thing needs to be redesigned to be app agnostic, and pluggable. Right now it's all hardcoded, not to mention that it's called `pageBuilder`, but now also has Form Builder stuff inside 🤦 But that's not your fault, and we don't have capacity to handle it properly. So this is all ok, keep going.

Btw. the reason why PB hook works in FB, is because the `revisionType` is just a `useState` hook (`packages/app-page-builder/src/contexts/PageBuilder/PageBuilderContext.tsx:53`), to track the selection. Very misplaced, but again, leave it for better times.

All in all, keep going with this.

One thing I noticed, is the location of these plugins, which really looks wrong, so please move them to a more logical place. The export button is NOT in the editor, and definitely not in the `defaultBar` :D 
CleanShot 2023-03-08 at 13 42 01

neatbyte-vnobis:
@Pavel910 
I have merged `next` here to resolve conflicts
Do you have any more code review comments?

Pavel910 (approved):
Thank you @neatbyte-vnobis, let's merge this!

Pavel910:
@neatbyte-vnobis I just noticed that the `cwp-template-aws` package was not updated 🤦 I will update it myself right away.

PR #3078, opened 2023-03-03
Deploy Command - Always Execute Before/After Deploy Hooks 

adrians5j:
## Changes
Prior to this PR, before/after hooks would not get triggered if the `webiny deploy` command was run with the `--preview` flag. This would cause the preview command to sometimes report changes in the deployed infra, which would not actually happen with an actual deployment.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


Pavel910 (approved):
Makes sense, it's up to developer to think about deploy/preview context, so this looks good to me.

PR #3081, opened 2023-03-07
fix: fix GraphQL error after page duplicate

neatbyte-vnobis:
## Changes
Fix issue "Use the “Duplicate page” button, then click “Edit” on the new page, GraphQL breaks. After browser reload everything is ok."

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/admin/graphql/pages.ts
+++ b/packages/app-page-builder/src/admin/graphql/pages.ts
@@ -86,17 +86,6 @@ export const DUPLICATE_PAGE = gql`
             duplicatePage(id: $id) {
                 data {
                     ${LIST_PAGES_DATA_FIELDS}
-                    settings {
-                        general {
-                            snippet
-                            tags
-                            layout
-                            image {
-                                id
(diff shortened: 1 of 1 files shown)


adrians5j (commented):
Sorry to ask, but can you briefly explain what was the issue here? Why was including these fields causing an error?  Was it because `settings` field doesn't exist?

neatbyte-vnobis:
@adrians5j 

It was causing next error:
```
Store error: the application attempted to write an object with no provided typename
but the store already contains an object with typename of PbGeneralPageSettings
for the object of id $6409a7dad103ab0008983633#0001.settings.general.
```

This issue was introduced while merging/resolving conflicts in this PR:
https://github.com/webiny/webiny-js/pull/2963/files#diff-86e943a17980e4fdaf0cba10215e44c35fa020f0b03cf19c0310b014aec350e3R76.

This code should have never been there (looks like wrong resolution of conflicts)

adrians5j (approved):

PR #3083, opened 2023-03-08
Upgrade File Manager files and settings data

Pavel910:
## Changes
This PR upgrades File Manager Storage Operations to work with better partition keys and use the `data` envelope to hold actual business-related data: `{ PK, SK, GSI1_PK, GSI1_SK, TYPE, data }` - this is the standard envelope we'll be pushing across the system over time. This will allow us to add new entity attributes easily,  and allow users to extend the system with less code, and less understanding of the underlying storage layer.

## How Has This Been Tested?
Jest tests.

## Documentation
These are internals, so no documentation is necessary at this stage.

### DIFF
--- a/packages/api-file-manager-ddb/src/operations/settings/SettingsStorageOperations.ts
+++ b/packages/api-file-manager-ddb/src/operations/settings/SettingsStorageOperations.ts
@@ -9,12 +9,13 @@ import { Entity } from "dynamodb-toolbox";
 import WebinyError from "@webiny/error";
 import defineTable from "~/definitions/table";
 import defineSettingsEntity from "~/definitions/settingsEntity";
+import { queryOne } from "@webiny/db-dynamodb/utils/query";
 
 interface SettingsStorageOperationsConstructorParams {
     context: FileManagerContext;
 }
 
-const SORT_KEY = "default";
+const SORT_KEY = "A";
 
 export class SettingsStorageOperations implements FileManagerSettingsStorageOperations {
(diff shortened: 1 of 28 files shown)


packages/api-file-manager-ddb/src/operations/settings/SettingsStorageOperations.ts
```diff
@@ -42,14 +43,18 @@ export class SettingsStorageOperations implements FileManagerSettingsStorageOper
 
     public async get(): Promise<FileManagerSettings | null> {
         try {
-            const settings = await this._entity.get({
-                PK: this.partitionKey,
-                SK: SORT_KEY
+            const settings = await queryOne<{ data: FileManagerSettings }>({
```
brunozoric: Can you check that `query` is of the same speed as the `get`? Not sure what happens in the background.

brunozoric (commented):
I have one question, check in the comments. Otherwise it looks good.

packages/api-file-manager-ddb/src/operations/settings/SettingsStorageOperations.ts
Pavel910: I converted to `queryOne` purely for TS, to have nicer return value, but now I see we do have our own `get` wrapper, like `queryOne`. Will update the code to use that!

Pavel910: @brunozoric I just updated this, check my latest commit.

PR #3085, opened 2023-03-09
feat: get lexical content text

mihajlovco:
New utility method that extracts and returns the plain text from a lexical JSON structure.

## Changes
- New method `getLexicalContentText` in the `utils` folder.
- export the method in the `index.ts`

## How Has This Been Tested?
Manually.
Screenshot 2023-03-09 at 15 50 57

PR #3087, opened 2023-03-09
feat: add Lexical Editor Content utility for api-page-builder-aco

leopuleo:
## Changes
With this PR we add a simple utility to get "text" content from the Lexical Editor payload. 

Duplicate #3085 

## How Has This Been Tested?
Jest

### DIFF
--- a/packages/api-page-builder-aco/src/utils/getLexicalContentText.ts
+++ b/packages/api-page-builder-aco/src/utils/getLexicalContentText.ts
@@ -1,17 +1,9 @@
 export const getLexicalContentText = (value: string): string => {
-    if (!isValidJSON(value)) {
-        return value;
-    }
-
-    return traverse(JSON.parse(value), "text");
-};
-
-const isValidJSON = (value: string): boolean => {
     try {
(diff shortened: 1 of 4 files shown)


packages/api-page-builder-aco/src/utils/getLexicalContentText.ts
```diff
@@ -0,0 +1,32 @@
+export const getLexicalContentText = (value: string): string => {
+    if (!isValidJSON(value)) {
+        return value;
+    }
+
+    return traverse(JSON.parse(value), "text");
```
Pavel910: I would recommend returning the parsed value from `isValidJson` to avoid double parsing. Better name the function `parseJSON`, and if it throws an error, return the original value. If it returns successfully, you have your object for traversal.

leopuleo: I just simplified it as you suggested, handling the error `catch` within `getLexicalContentText` function.

Pavel910 (approved):

PR #3089, opened 2023-03-09
Custom Domains - Wrap Domain Assignment Into a Handler

adrians5j:
## Changes
With this PR, we're further polishing the custom domains assignment. 

With it, assigning custom domains works both via the `domains` Pulumi app param, and via the `pulumi` callback. Previously, the latter would not work, because of the order of execution was incorrect, thus the domain/stack output assignment would happen too early.

### Additional Change

Prior to this PR, Website's Pulumi code would expose two Cloudfront distributions via the `app` and `delivery` properties. From now, the `app` can also be accessed via the `preview` property. More on this in the included code comment:

```
// "preview" and "app" are the same.
// We introduced "preview" just because it's the word we use when talking about
// Page Builder and "previewing" pages. In other words, the "preview" property
// contains all resources related to serving page previews, unlike "delivery",
// which is used to serve published pages to actual website visitors.
// The "app" property was still left here just or backwards compatibility.
```

## How Has This Been Tested?
Manually.

## Documentation
Changelog.

### DIFF
--- a/packages/pulumi-aws/src/apps/api/createApiPulumiApp.ts
+++ b/packages/pulumi-aws/src/apps/api/createApiPulumiApp.ts
@@ -215,16 +215,19 @@ export const createApiPulumiApp = (projectAppParams: CreateApiPulumiAppParams = 
                 dynamoDbElasticsearchTable: core.elasticsearchDynamodbTableName
             });
 
-            addDomainsUrlsOutputs({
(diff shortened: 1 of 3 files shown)


Pavel910 (approved):

adrians5j:
Thanks @Pavel910!

PR #3090, opened 2023-03-10
feat: use custom:id attribute as identity id

Pavel910:
## Changes
This PR adds a custom attribute to Cognito User Pool, called `custom:id`, which holds a Webiny-generated ID for the identity. This is necessary to support export/import of users from Cognito User Pool. The problem is that , when importing users into a Cognito User Pool, the `sub` (AWS-generated ID) changes. This means that we cannot rely on the value of the `sub` attribute. Storing our own ID into a custom attribute solves the issue, because we can now always reference this ID after importing users into a new user pool, and this ID will also stay consistent with a copy of Webiny DynamoDB table. 

This change is backwards compatible, so all existing identities will work with the `sub` attribute, but all new identities will be handled via this new `custom:id` attribute.

> Custom implementations of Cognito security can use whatever they want; this change is primarily for our out-of-the-box solution. If people do custom identities, this doesn't apply.
 
> `custom:` prefix is enforced by AWS for all non-standard attributes, so even though in Pulumi you specify simply `id`, when working with AWS SDK, you must include the `custom` prefix, and the prefix is also included in the idToken by AWS.

image

image

image

image

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/api-admin-users-cognito/src/syncWithCognito.ts
+++ b/packages/api-admin-users-cognito/src/syncWithCognito.ts
@@ -104,21 +104,15 @@ export const syncWithCognito = ({
                     {
                         Name: "email",
                         Value: username
+                    },
+                    {
+                        Name: "custom:id",
+                        Value: user.id
                     }
                 ]
             };
 
-            const response = await cognito.adminCreateUser(params).promise();
-
-            const { User } = response;
-            /**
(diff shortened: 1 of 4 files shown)


adrians5j (approved):
Looks good.

So, if available, the `custom:id` will be used, otherwise the legacy `sub`.

That was quick :)

mihajlovco (approved):

PR #3093, opened 2023-03-10
fix: fix unreadable form submission values

neatbyte-vnobis:
## Changes
Fix issue "Radio and checkbox filed values in the webhook end file export use the value field instead of the label which makes them unreadable"
See https://github.com/webiny/webiny-js/pull/2951 for previous code review comments

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-form-builder/src/plugins/crud/submissions.crud.ts
+++ b/packages/api-form-builder/src/plugins/crud/submissions.crud.ts
@@ -331,7 +331,7 @@ export const createSubmissionsCrud = (params: CreateSubmissionsCrudParams): Subm
                                 addLog: (log: Record<string, any>) => {
                                     submission.logs.push(log);
                                 },
-                                data,
+                                data: utils.sanitizeFormSubmissionData(form.fields, data),
                                 // meta,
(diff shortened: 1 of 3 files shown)


adrians5j:
Thanks for opening a new PR @neatbyte-vnobis. Continuing the discussion here.

I like the screenshot, and I'd say it's the proper way forward.

Besides @SvenAlHamad confirming this is fine, I'd also have one last question (which @SvenAlHamad my also comment on):

image

Once we confirm, we'll soon be able to merge this.

Thank you @neatbyte-vnobis !

SvenAlHamad:
Because this is a user facing information, I think it’s more readable if we don’t have the [] brackets.

adrians5j:
I agree @SvenAlHamad.

Alright @neatbyte-vnobis, let's just apply this last change and we'll most probably be good to merge.

Thanks!

P.S. Let's apply the change for both non-label and label columns (so, both **Country Radio** and **Country Radio (Label)** columns).

adrians5j (approved):
Thanks @neatbyte-vnobis !

PR #3101, opened 2023-03-14
fix: fix duplicate images when importing pages/blocks/templates

neatbyte-vnobis:
## Changes
Resolve issue "When importing pages, blocks and templates see we don’t duplicate images"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-page-builder-import-export/src/import/utils/uploadAssets.ts
+++ b/packages/api-page-builder-import-export/src/import/utils/uploadAssets.ts
@@ -29,9 +29,12 @@ export const uploadAssets = async (params: UploadAssetsParams) => {
     }
 
     // Check if files with such keys already exist
-    const [existingFiles] = await context.fileManager.files.listFiles();
+    const fileIds = files.map(file => file.id);
+    const [existingImages] = await context.fileManager.files.listFiles({
+        where: { id_in: fileIds }
+    });
     const filteredFiles = files.filter(
-        file => !existingFiles.some(existingFile => existingFile.key === file.key)
+        file => !existingImages.some(existingImage => existingImage.key === file.key)
     );
 
     // A map of temporary file keys (created during ZIP upload) to permanent file keys.
--- a/packages/api-page-builder-import-export/src/import/utils/uploadAssets.ts
+++ b/packages/api-page-builder-import-export/src/import/utils/uploadAssets.ts
@@ -28,13 +28,19 @@ export const uploadAssets = async (params: UploadAssetsParams) => {
         return oldIdToNewFileMap;
     }
 
+    // Check if files with such keys already exist
+    const [existingFiles] = await context.fileManager.files.listFiles();
+    const filteredFiles = files.filter(
+        file => !existingFiles.some(existingFile => existingFile.key === file.key)
+    );
+
     // A map of temporary file keys (created during ZIP upload) to permanent file keys.
     const uploadFileMap: UploadFileMap = new Map();
 
     // Array of file inputs, to insert into the DB.
     const createFilesInput: FileInput[] = [];
 
-    for (const oldFile of files) {
+    for (const oldFile of filteredFiles) {
         const id = mdbid();
         // We replace the old file ID with a new one.
         const newKey = `${id}/${oldFile.key.replace(`${oldFile.id}/`, "")}`;


neatbyte-vnobis:
FYI: this check is working if images are present in the DB with same `key` values.
If we import same file in NEW clean DB - we don't have old `key` values to do a check...
We still can compare all images somehow, but it is a high load operation (it is not done in scope of this pull)
Maybe we can add some hash of the files or smth like that?

packages/api-page-builder-import-export/src/import/utils/uploadAssets.ts
```diff
@@ -28,13 +28,19 @@ export const uploadAssets = async (params: UploadAssetsParams) => {
         return oldIdToNewFileMap;
     }
 
+    // Check if files with such keys already exist
+    const [existingFiles] = await context.fileManager.files.listFiles();
+    const filteredFiles = files.filter(
+        file => !existingFiles.some(existingFile => existingFile.key === file.key)
+    );
+
```
Pavel910: This will not work; this means you're potentially loading millions of files. One of our clients now is importing 3 millions of images, so you can imagine how this will not end well :D 

When you're exporting a page, let's say it has 20 images, you know the exact image IDs, so you can use the `ids` parameter on listFiles, like this:


```
const existingImages = await context.fileManager.files.listFiles({ where: { id_in: existinIds} })
```

Pavel910 (changes requested):
Let's use a more optimal query here :)

Pavel910 (approved):
Looks good, @neatbyte-vnobis. Have you tested this by doing and export and importing the page back into the same system?

SvenAlHamad:
So I tested this in a way that I exported a page from one system and imported it twice into another system. But I got all the images imported twice. I would expect that the first import would create the necessary keys, and the second import would then detect those keys and wouldn't duplicate the images. Both first and second import used the same export file.

neatbyte-vnobis:
@SvenAlHamad @Pavel910
It is expected behavior (see https://github.com/webiny/webiny-js/pull/3101#issuecomment-1468205344 )

If export is imported into the SAME system - we can compare keys and check for duplicates (which is done here).

If export is imported in OTHER system - we don't have values to compare...
Export file is not aware under which keys new images are created, so second, third, etc import will always create a new images.

What can be done:
1. As I have mentioned we can compare images themselves (or hashes if such can be fetched from somewhere).
2. Maybe we can store smth like `imported_under_key` for created images on first import.
3. Or we can import images without generation of new ID/key (but I am not sure about side effects caused by this)
4. Some other way...

Pavel910:
@SvenAlHamad @neatbyte-vnobis yes, that is a correct summary Vitalii. The problem is made more complicated with the fact that in multi-tenant system, tenant ID is NOT part of the file key. So file keys are global. Imagine you export from tenant 1, and import into tenant 2, then delete a file from one of the tenants -> that file will be deleted by file key in the S3 bucket, and all tenants are affected, if they had the page imported, and file keys were preserved.

So it's far safer to create new files than not. I don't know if there's a better way.

SvenAlHamad:
@Pavel910 @neatbyte-vnobis - I understand that importing images from one system to another, we can't check for duplicates, I'm ok with that. 

But the use-case I'm talking about is if I import a page with a few images from one system to another. I expect all the files to be created, that's ok. But if I then import another page, or even the same page again, that uses the images that I've already imported, is there a way to detect that?

Pavel910:
@SvenAlHamad no, because when you import, you get new file keys. 

Example:
- A/image.png -> this gets a new key -> X/image.png
- B/image.png -> this also gets a new key -> Y/image.png

Subsequent imports still have `A/image.png` and `B/image.png`. And there's nothing to compare with, because those files are `X/...` and `Y/....`.

SvenAlHamad:
Yeah, I see that we create new keys, but I think if we add an attribute containing the original key, like what Vitalii suggested, on subsequent imports we are able to detect that the image exists?

Pavel910:
@SvenAlHamad @neatbyte-vnobis It's worth trying, but not before I merge my file aliases PR, because it contains some changes on FM context, and I want to avoid conflicts and merges. Let's wait with this PR a bit.

neatbyte-vnobis:
@Pavel910 please let me know if I can continue with this task (were your changes done?)

Pavel910:
@neatbyte-vnobis yes, my stuff is in long ago. You're free to continue! 🚀

Pavel910:
Closing in favor of #3339

PR #3102, opened 2023-03-14
feat: move search entry into a folder

leopuleo:
## Changes
With this PR we introduce the following changes:

- Add the `EntryDialogMove` to allow users to move a search entry from the current folder to another
- Refactor `FolderDialogs` to improve the UX related to parent folder selection: instead of `Autocomplete`, we now use the `FolderTree` component.
- Add the following props to `FolderTree` component: 
  - `enableCreate`: (`boolean`) -> show or hide the "Create folder" button
  - `enableActions`: (`boolean`) -> show or hide the folder specific actions
  - `hiddenFolderId`: (`string`) -> hide a specific folder within the list

Closes [WEB-2262](https://webiny.atlassian.net/browse/WEB-2262)

## Video

https://user-images.githubusercontent.com/2866531/225258501-e3b6db34-f199-43df-a630-37877ec296bf.mov

## How Has This Been Tested?
TBD

## Documentation
Within the package.


packages/app-aco/src/components/Tree/List/utils.ts
```diff
@@ -7,33 +7,37 @@ import { ROOT_ID } from "./constants";
...
  * @param folders list of folders returned by useFolders hook.
  * @param focusedNodeId id of the current folder selected/focused.
+ * @param hiddenFolderId id of the folder you don't want to show within the list.
  * @return array of elements to render the tree component.
  */
 export const createTreeData = (
     folders: FolderItem[] = [],
-    focusedNodeId?: string
+    focusedNodeId?: string,
+    hiddenFolderId?: string
```
Pavel910: Not entirely sure what's the use case, so bare with me: would it make more sense to support an array of IDs? Is it possible that I'll want to hide multiple folders?

packages/app-aco/src/components/Tree/styled.tsx
```diff
@@ -2,7 +2,7 @@ import styled from "@emotion/styled";
 
 export const Container = styled("div")`
     position: relative;
-    margin: 0 0 32px;
+    margin: 0 0;
```
Pavel910: Nitpicking here, but `0 0` === `0`, so one zero is enough 😅

Pavel910 (commented):
One question and an irrelevant CSS nitpick. Mostly curious about the question. Other than that, this is consistent with the rest of the package, so no changes necessary. Let's resolve the question, and then I'll approve.

packages/app-aco/src/components/Tree/List/utils.ts
leopuleo: We don't need this right now, since hiding one folder does not show the whole hierarchy under it. It would be like changing a few lines of code, giving us more flexibility. I'm working on it

Pavel910 (approved):

PR #3103, opened 2023-03-14
fix: fix preview button URLs

neatbyte-vnobis:
## Changes
Fixes issue "Preview button in the revisions tab not working correctly"

## How Has This Been Tested?
Manual

## Documentation
None

PR #3112, opened 2023-03-16
feat: visual adjustments to the page preview panel

neatbyte-vnobis:
## Changes
Implements issue "Visual adjustments to the page preview panel"

SHOULD BE MERGED AFTER https://github.com/webiny/webiny-js/pull/3103

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/admin/plugins/pageDetails/header/Header.tsx
+++ b/packages/app-page-builder/src/admin/plugins/pageDetails/header/Header.tsx
@@ -1,7 +1,8 @@
-import React from "react";
+import React, { useCallback } from "react";
 import styled from "@emotion/styled";
 import { renderPlugins } from "@webiny/app/plugins";
 import { Typography } from "@webiny/ui/Typography";
+import { useConfigureWebsiteUrlDialog } from "~/admin/hooks/useConfigureWebsiteUrl";
 import { usePageBuilderSettings } from "~/admin/hooks/usePageBuilderSettings";
 import { useSiteStatus } from "~/admin/hooks/useSiteStatus";
 import { ReactComponent as LinkIcon } from "@material-design-icons/svg/round/link.svg";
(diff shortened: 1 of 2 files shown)


SvenAlHamad:
I've tested this one and noticed a small css issue, the dropdown arrow goes outside the dropdown container: 
CleanShot 2023-03-16 at 19 52 10@2x

neatbyte-vnobis:
@SvenAlHamad I have checked this, and I cannot reproduce the issue.

Maybe you need to rebuild UI on this branch? (there was similar issue in December when you saw a similar bug locally, but after rebuild it was gone). Also please let me know env OS and used browser, if issue is still there.

adrians5j:
Looks good on my end: 
image

image

packages/app-page-builder/src/admin/plugins/pageDetails/header/Header.tsx
```diff
@@ -2,46 +2,100 @@ import React from "react";
...
         <React.Fragment>
             <HeaderTitle>
-                <PageTitle>
-                    <Typography use="headline5">{page.title}</Typography>
-                </PageTitle>
+                <PageInfo>
+                    <PageTitle>
+                        <Typography use="headline6">{page.title}</Typography>
+                    </PageTitle>
+                    {isSiteRunning && (
```
adrians5j: Can we rather have the link always shown, but when clicking on it, if the site is not running, display that modal that we show in other places? Like when we click on the View page in dropdown:
image

adrians5j (changes requested):
Left a question in the code.

Also, two extra questions.

First one for @SvenAlHamad , should we have a different color for unpublished pages?

At the moment, we use green for published, everything else is yellow.

image

Second question - what happens with statuses when APW is enabled? I mean, will this work with APW enabled? Not sure how statuses work there. Can we expect statuses other than the default ones that come with basic version of Webiny?

SvenAlHamad:
Let's use a darker grey color for "unpublished" pages. 
As for the APW question @adrians5j - APW still works with those same statuses, it's not affecting them.

adrians5j (approved):

PR #3115, opened 2023-03-20
New theme typography structure and related components upgrades (WIP)

mihajlovco:
With this PR we are changing the theme typography structure so we have better flexibility to have better theme CRUD operations on the styles. In this PR are included all upgrades for old and some reimplemented components like `Quote` and `List` that need to support this structure.

In this PR we made the following changes:

### Theme implementation
`Admin/Apps/Theme`
- Changed the theme typography structure and new types (temporary at the moment but approved solution).

##  Emotion CSS classes for the lexical nodes
- Introduction of a new type `ThemeEmotionMap` keeps the typography style data by style id. Now it's faster and easier to access the styles.
- Updated WebinyThemeEditor class with `emotionMap` prop so Lexical editor nodes can access to the styles.
- Implemented utility method `toThemeEmotionMap` that goes through the `ThemeStyles` and generates `ThemeEmotionMap` styles with the Emition `css()` generation method.
- New package dependency on `@emition/core` package

## Updated `useRichTextEdtior` context
- Introduced a new `TextBlockSelection` state that represents the currently selected lexical text nodes objects and toolbar state (status of the selected node or style format like _bold_ or _italic_)
- New utility function  `getLexicalTextSelectionState` that supports the processing of the current selection used in the `FloatingToolbar` component. In this toolbar component, we listen for the _`active editor state`_ and take the current text selection (`RangeSelection` object) for further processing.

> With this solution now the toolbar action will not care about the state of the editor but only will listen for data changes specific to the action.
- New `theme` state of type `WebinyTheme`.  This object is the actual webiny theme and is provided in the lexical editor trough `LexicalEditor` component in `lexical-editor-pb-element package`. Now this component is available for `TypograpyAction`, `TypograpyDropdown`(lexical-editor-actions), and available for all editor components.
- `toolabarType` state, values:  _header_  and _paragraph_ at the moment.
- New `ThemeEmotionMap` state - `themeEmotionMap`, `setThemeEmotionMap` - keeps a map of the theme styles with keys by id.

## Implementation of the new state in the toolbar actions components
This change removes unnecessary complexity and processing

Updated components:
- BoldAction
- BulletsListAction
- CodeHeoighlightAction
- ItalicAction
- NumberedListAction
- QuoteAction
- TypographyAction
- UnderlineAction

>  FontColorAction and FontSizeAction will be addressed with  the ColorPicker updates

## New `WebinyList` and `WebinyListItem` components that support typography styles
This component is taken and adapted to our needs from the Lexical codebase.

New components 
- WebinyListNode
- WebinyListItemNode
- WebinyListPlugin and related useWebinyList  hook - in this plugin we listen for List commands and insert/update/remove the node
- New utility functions in file `utils/nodes/list-node.ts` to support manipulation of the list nodes
- Defined list commands and command payload in `commands/list-command` file .

> Known issue: When you press `tab` key instead converting the current text into a list, the focus shifts outside from the lexical editor to the right sidebar input (image in the testing section).

## New `WebinyQuote` component that support typography styles
This component is taken and adapted to our needs from the Lexical codebase.
- WebinyQuote
- WebinyQuotePlugin and related hook useWebinyQuote - listen for commands on editor event registraction
- Updated QuoteAction 
- Update TypograpyDropdown to support new quote styles listing
- New utility function `formatToQuote` that converts the current selected node into quote text block
- Defined commands  and command payload in the `commands/webiny-quote` file.

## Implemented undo/redo commands on 
- Updated package with `@lexical/history` dependency
- Introduced a new context `SharedHistoryContext` that keeps and manages the changes.
- Imported the HistoryPlugin component from lexical in the base `RitchTextEditor` component.

## Fixed Heading component on drag&drop and related display of typography styles.
 
The issue was when the user drops the heading component into the editor, the user can see only paragraph styles from the typography menu and not the heading styles.
- Updated heading data in `page-builder-elements` package
- Updated TypographyAction 
- Updated TypograpyDropdown 

## How Has This Been Tested?
Manually.

### New TextBlockSelection and toolbar action implementation
Screenshot 2023-03-28 at 17 05 51
Screenshot 2023-03-28 at 17.10.30

### New List implementation with styles

https://user-images.githubusercontent.com/82515066/228349489-737c683a-2fa3-43b8-b411-147f0666121d.mov

### undo/redo in action

https://user-images.githubusercontent.com/82515066/228349190-63b0a708-025d-431c-a152-72554e7c56b8.mov

### Heading drag&drop fix

https://user-images.githubusercontent.com/82515066/228350542-51a1a50e-cc64-4b8f-97aa-add2ba430c95.mov


packages/theme/src/types.ts
```diff
@@ -11,9 +11,27 @@ export interface StylesObject {
 
 export type ThemeBreakpoints = Record<string, string>;
 
+export type HeadingHtmlTag = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
```
mihajlovco, line 28: @Pavel910 @adrians5j New defined typography types

PR #3119, opened 2023-03-20
feat: add ability to create a page template from an existing page

neatbyte-vnobis:
## Changes
Implements issue "Ability to create a page template from an existing page".
Also includes modal dialog for new template creation (with fields title, slug, description).

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-page-builder/src/graphql/crud/pageTemplates.crud.ts
+++ b/packages/api-page-builder/src/graphql/crud/pageTemplates.crud.ts
@@ -174,11 +174,16 @@ export const createPageTemplatesCrud = (
         async createPageTemplate(this: PageBuilderContextObject, input: PageTemplateInput) {
             await checkBasePermissions(context, PERMISSION_NAME, { rwd: "w" });
 
-            const existingPageTemplate = await this.getPageTemplate({
-                where: { slug: input.slug }
-            });
+            let pageTemplateExists = true;
+            try {
+                await this.getPageTemplate({
+                    where: { slug: input.slug }
+                });
+            } catch {
+                pageTemplateExists = false;
+            }
 
-            if (existingPageTemplate) {
+            if (pageTemplateExists) {
                 throw new NotFoundError(`Page Template with slug "${input.slug}" already exists.`);
             }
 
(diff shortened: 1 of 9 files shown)


SvenAlHamad:
I tested this one - from my end it looks good.

packages/api-page-builder/src/graphql/crud/pageTemplates.crud.ts
```diff
@@ -165,6 +172,16 @@ export const createPageTemplatesCrud = (
         async createPageTemplate(this: PageBuilderContextObject, input: PageTemplateInput) {
             await checkBasePermissions(context, PERMISSION_NAME, { rwd: "w" });
 
+            const existingPageTemplate = await this.getPageTemplate(
```
adrians5j: Curious why `auth: false` here?

I don't thing this should be necessary. I'd say it's safe to assume that if users has the right to create page templates, that they also has the right to read them.

If we agree on this, then we can also remove the `options` param added to the `getPageTemplate` fn.

packages/api-page-builder/src/graphql/crud/pageTemplates.crud.ts
```diff
@@ -378,6 +395,51 @@ export const createPageTemplatesCrud = (
...
+        async createTemplateFromPage(pageId, data) {
+            const page = await context.pageBuilder.getPage(pageId);
+            if (!page) {
+                throw new NotFoundError(`Page "${pageId}" was not found!`);
+            }
+
+            const templateVariables: Array<{ blockId: string; variables?: PageBlockVariable[] }> =
+                [];
+
+            const templateElements = page.content?.elements?.map((block: PbPageElement) => {
```
adrians5j, line 416: I don't quite understand the purpose of this map call. Can you clarify?

For example, why are we iterating only over first set of elements and not all elements recursively (every element can have child elements)?

packages/app-page-builder/src/admin/views/PageTemplates/CreatePageTemplateDialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+    & .mdc-dialog__surface {
+        width: 600px;
+        min-width: 600px;
+    }
+`;
+
+const slugValidator: Validator = (value: string) => {
+    if (!value.match(/^[a-z]+(\-[a-z]+)*$/)) {
+        throw new Error(
+            "Page Template slug must consist of only 'a-z' and '-' characters (for example: 'page-template-slug')."
```
adrians5j, line 23: Page Template -> Page template

adrians5j (approved):

PR #3121, opened 2023-03-20
feat: improve folders and search records listing 

leopuleo:
## Changes

With this PR we expose `useAcoList()` hooks that exposes:
- `folders` and `records` state, coming from the respective state;
- `listItems` method, to fetch more or sorted `records`; 
- `meta` (coming from `records` state);
- `listTitle`, the current folder name;
- `loading` states.

Using `useAcoList()` allows us to sort both `records` and `folders`:
- `records` sorting is done server-side;
- `folders` sorting is done client-side, since the folder list tree has been already fetched from the server.

Closes [https://webiny.atlassian.net/browse/WEB-2279](WEB-2279)

## How Has This Been Tested?
Jest + Cypress

## Documentation
Inline

### DIFF
--- a/packages/app-aco/src/contexts/records.tsx
+++ b/packages/app-aco/src/contexts/records.tsx
@@ -1,6 +1,5 @@
 import React, { ReactNode, useState } from "react";
 import { useApolloClient } from "@apollo/react-hooks";
-import orderBy from "lodash/orderBy";
 import unionBy from "lodash/unionBy";
 
 import { apolloFetchingHandler, loadingHandler } from "~/handlers";
@@ -29,8 +28,9 @@ import {
     Meta,
     UpdateSearchRecordResponse,
     UpdateSearchRecordVariables,
-    ListSort
+    ListDbSort
 } from "~/types";
+import { handleDbSorting, handleTableSorting } from "~/sorting";
 
 interface SearchRecordsContext {
     records: SearchRecordItem[];
@@ -41,7 +41,7 @@ interface SearchRecordsContext {
         folderId?: string;
         limit?: number;
         after?: string;
-        sort?: ListSort;
+        sort?: ListDbSort;
     }) => Promise<SearchRecordItem[]>;
     getRecord: (id: string) => Promise<SearchRecordItem>;
     createRecord: (record: Omit<SearchRecordItem, "id">) => Promise<SearchRecordItem>;
@@ -78,7 +78,7 @@ export const SearchRecordsProvider = ({ children }: Props) => {
         loading,
         meta,
         async listRecords(params) {
-            const { type, folderId, after, limit = 5, sort: sorting } = params;
+            const { type, folderId, after, limit, sort: initialSorting } = params;
 
             /**
              * Both folderId and type are optional to init `useRecords` but required to list records:
@@ -102,15 +102,12 @@ export const SearchRecordsProvider = ({ children }: Props) => {
             }
 
             // Remove records in case of sorting change and not a paginated request.
-            if (sorting && !after) {
+            if (initialSorting && !after) {
                 setRecords([]);
             }
 
             const action = after ? "LIST_MORE" : "LIST";
-            const sort =
-                sorting && Object.keys(sorting).length > 0
-                    ? sorting
-                    : ({ savedOn: "DESC" } as unknown as ListSort);
+            const sort = handleDbSorting(initialSorting);
 
             const { data: response } = await apolloFetchingHandler(
                 loadingHandler(action, setLoading),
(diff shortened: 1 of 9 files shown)


packages/app-aco/src/contexts/records.tsx
```diff
@@ -78,7 +78,7 @@ export const SearchRecordsProvider = ({ children }: Props) => {
         loading,
         meta,
         async listRecords(params) {
-            const { type, folderId, after, limit, sort: sorting } = params;
+            const { type, folderId, after, limit, sort: initialSorting } = params;
```
Pavel910: The word `initial` implies that there can be subsequent sort. But we also know that when you change sorting parameters, you start from page 1 (well, zero). So the word `initialSorting` is misleading. Keep it as `sort`.

packages/app-aco/src/contexts/records.tsx
```diff
@@ -102,15 +102,12 @@ export const SearchRecordsProvider = ({ children }: Props) => {
             }
 
             // Remove records in case of sorting change and not a paginated request.
-            if (sorting && !after) {
+            if (initialSorting && !after) {
```
Pavel910: To continue my previous comment:  here the comment says: `in case of sorting change`. So if it changed, then it's not `initialSorting`.

packages/app-aco/src/contexts/records.tsx
```diff
@@ -102,15 +102,12 @@ export const SearchRecordsProvider = ({ children }: Props) => {
...
                 setRecords([]);
             }
 
             const action = after ? "LIST_MORE" : "LIST";
-            const sort =
-                sorting && Object.keys(sorting).length > 0
-                    ? sorting
-                    : ({ savedOn: "DESC" } as unknown as ListSort);
+            const sort = handleDbSorting(initialSorting);
```
Pavel910: When reading this code, I'm left wondering what does `handleDbSorting` really do? Let's try to give meaningful names to functions. For example, here you do something with sort, and return a new value. Name the function so it reflects what the new value is.

packages/app-aco/src/contexts/records.tsx
```diff
@@ -129,13 +126,7 @@ export const SearchRecordsProvider = ({ children }: Props) => {
...
 
             // Adjusting sorting while merging records with data received from the server.
-            const fields = [] as string[];
-            const orders = [] as Array<"asc" | "desc">;
-            for (const [field, order] of Object.entries(sort)) {
-                fields.push(field);
-                orders.push(order.toLowerCase() as "asc" | "desc");
-            }
-            setRecords(records => orderBy(unionBy(data, records, "id"), fields, orders));
+            setRecords(records => handleTableSorting(unionBy(data, records, "id"), sort));
```
Pavel910: Similarly, here we have `handleTableSorting`. What does it do? If it sorts records, call it `sortTableItems` (for example). If it returns a value, `getSortedTableItems`. You get the idea. The reader must understand what you're doing without going into the function and reading the implementation.

packages/app-aco/src/sorting.tsx
```diff
@@ -0,0 +1,24 @@
+import orderBy from "lodash/orderBy";
+
+import { ListDbSort, ListTableSort, ListTableSortDirection } from "~/types";
+
+export const handleDbSorting = (initial?: ListDbSort): ListDbSort => {
```
Pavel910: 5 files into the PR, I finally got to the function body. This function validates the sorter, and returns a default sort if parameter is invalid. Name it that: `getValidSort`, or `validateOrGetDefaultSort`, just some ideas.

packages/app-aco/src/sorting.tsx
```diff
@@ -0,0 +1,24 @@
...
+    if (!initial || Object.keys(initial).length === 0) {
+        return { savedOn: "DESC" } as unknown as ListDbSort;
+    }
+
+    return initial;
+};
+
+export const handleTableSorting = (records: any[], sort?: ListDbSort): any[] => {
```
Pavel910: This one is `sortRecords`.

Pavel910 (changes requested):

packages/app-aco/src/contexts/records.tsx
leopuleo: Thanks for the suggestion, I left `sorting` so I can use `sort` as query param later in the code

packages/app-aco/src/contexts/records.tsx
leopuleo: I changed in `sortTableItems`...makes sense

packages/app-aco/src/sorting.tsx
leopuleo: Thank you, I changed it in `sortTableItems`

packages/app-aco/src/sorting.tsx
leopuleo: 👍 I changed it into `validateOrGetDefaultDbSort`, it explains better what the method does.

Pavel910 (approved):

PR #3124, opened 2023-03-21
feat: enrich webhook with Google Ads data

neatbyte-vnobis:
## Changes
Implements issue "When sending a webhook enrich it with url data fragments and google ads details"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-form-builder/src/plugins/crud/forms.models.ts
+++ b/packages/api-form-builder/src/plugins/crud/forms.models.ts
@@ -103,7 +103,13 @@ export const FormSubmissionCreateDataModel = withFields({
             submittedOn: string({
                 value: new Date().toISOString()
             }),
-            url: object()
+            url: fields({
+                value: {},
+                instanceOf: withFields({
+                    location: string({}),
+                    query: object()
+                })()
+            })
         })()
     }),
     form: fields({
--- a/packages/api-form-builder/src/plugins/crud/forms.models.ts
+++ b/packages/api-form-builder/src/plugins/crud/forms.models.ts
@@ -102,7 +102,8 @@ export const FormSubmissionCreateDataModel = withFields({
             ip: string({}),
             submittedOn: string({
                 value: new Date().toISOString()
-            })
+            }),
+            url: object()
         })()
     }),
     form: fields({
--- a/packages/api-form-builder/src/plugins/crud/forms.models.ts
+++ b/packages/api-form-builder/src/plugins/crud/forms.models.ts
@@ -106,7 +106,7 @@ export const FormSubmissionCreateDataModel = withFields({
             url: fields({
                 value: {},
                 instanceOf: withFields({
-                    location: string({}),
+                    location: string(),
                     query: object()
                 })()
             })
--- a/packages/api-form-builder/src/plugins/crud/utils.ts
+++ b/packages/api-form-builder/src/plugins/crud/utils.ts
@@ -125,3 +125,19 @@ export const sanitizeFormSubmissionData = (fields: FbFormField[], data: { [key: 
 
     return data;
 };
+
+export const flattenObj = (
+    obj: Record<string, any>,
+    parent: string,
+    res: Record<string, string> = {}
+) => {
+    for (const key in obj) {
+        const propName = parent ? parent + "_" + key : key;
+        if (typeof obj[key] == "object") {
+            flattenObj(obj[key], propName, res);
+        } else {
+            res[propName] = obj[key];
+        }
+    }
+    return res;
+};
--- a/packages/api-form-builder/src/plugins/crud/utils.ts
+++ b/packages/api-form-builder/src/plugins/crud/utils.ts
@@ -126,7 +126,10 @@ export const sanitizeFormSubmissionData = (fields: FbFormField[], data: { [key: 
     return data;
 };
 
-export const flattenObj = (
+/**
+ * Converts deep submission meta object into flat object suitable for CSV.
+ */
+export const flattenSubmissionMeta = (
     obj: Record<string, any>,
     parent: string,
     res: Record<string, string> = {}
@@ -134,7 +137,7 @@ export const flattenObj = (
     for (const key in obj) {
         const propName = parent ? parent + "_" + key : key;
         if (typeof obj[key] == "object") {
-            flattenObj(obj[key], propName, res);
+            flattenSubmissionMeta(obj[key], propName, res);
         } else {
             res[propName] = obj[key];
         }
--- a/packages/api-form-builder/src/plugins/graphql/form.ts
+++ b/packages/api-form-builder/src/plugins/graphql/form.ts
@@ -7,7 +7,7 @@ import {
     Response
 } from "@webiny/handler-graphql/responses";
 import { GraphQLSchemaPlugin } from "@webiny/handler-graphql/types";
-import { sanitizeFormSubmissionData } from "~/plugins/crud/utils";
+import { sanitizeFormSubmissionData, flattenObj } from "~/plugins/crud/utils";
 import { FormBuilderContext, FbFormField } from "~/types";
 
 const plugin: GraphQLSchemaPlugin<FormBuilderContext> = {
@@ -214,6 +214,7 @@ const plugin: GraphQLSchemaPlugin<FormBuilderContext> = {
             type FbSubmissionMeta {
                 ip: String
                 submittedOn: DateTime
+                url: JSON
             }
 
             type FbListSubmissionsMeta {
@@ -554,6 +555,22 @@ const plugin: GraphQLSchemaPlugin<FormBuilderContext> = {
                         }
 
                         /**
+                         * Add meta fields.
+                         */
+                        for (let i = 0; i < submissions.length; i++) {
+                            const flattenExportMeta = flattenObj(
+                                submissions[i].meta.url || {},
+                                "meta_url"
+                            );
+
+                            for (const metaKey in flattenExportMeta) {
+                                if (!fields[metaKey]) {
+                                    fields[metaKey] = metaKey;
+                                }
+                            }
+                        }
+
+                        /**
                          * Build rows.
                          */
                         for (let i = 0; i < submissions.length; i++) {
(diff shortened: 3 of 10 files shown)


neatbyte-vnobis:
@Pavel910 @adrians5j 
I have a question: what to do if utm params are not present in the url?
Now webhook will send `null`s in such case (for missing values, or all of them if there are none utm parameters).
But it is possible to drop `utm` object completely too if all are missing.
Or drop only `null` values

packages/app-page-builder-elements/src/renderers/form/FormRender/functions/createFormSubmission.ts
```diff
@@ -61,11 +61,26 @@ export default async ({
...
+    const query = new URLSearchParams(location.search);
+
     const variables = {
         revision: formData!.id,
         reCaptchaResponseToken,
         data: formSubmissionData,
-        meta: { ip: await getClientIp() }
+        meta: {
+            ip: await getClientIp(),
+            dateSubmitted: new Date().toISOString(),
```
adrians5j: For consistency sake, let's just rename `dateSubmitted` to `submittedOn` (we have other date/times in other parts of the system using the same naming convention, for example `createdOn`).

packages/api-form-builder/src/plugins/triggers/webhook.ts
```diff
@@ -20,7 +21,16 @@ const plugin: FbFormTriggerHandlerPlugin = {
                 const response = await got(url, {
                     method: "post",
                     json: true,
-                    body: data
+                    body: {
+                        ...data,
+                        _meta: {
+                            dateSubmitted: format(
```
adrians5j: Same `submittedOn` comment I left earlier.

packages/api-form-builder/src/plugins/triggers/webhook.ts
```diff
@@ -20,7 +21,16 @@ const plugin: FbFormTriggerHandlerPlugin = {
                 const response = await got(url, {
                     method: "post",
                     json: true,
-                    body: data
+                    body: {
+                        ...data,
+                        _meta: {
```
adrians5j: Let's have this as `meta` instead of `_meta`.

adrians5j (changes requested):
Looks good @neatbyte-vnobis .

Left a couple of comments, nothing special. 

I know the initial doc said to use `dateSubmitted` and `_meta` keys.

For the `dateSubmitted`, I think it's better we use `submittedOn`, so that we're more consistent with the rest of the system.

For the `_meta`, I think there's no need for the `_` prefix. Again, we have the `meta` property in other parts of the system, so having it the same here will again keep things more consistent.

SvenAlHamad:
On this one I have one feedback, and then a few additional items that I got from our marketing person, which we would like to implement. 

The first feedback items is that it seems that the UTM tags are hard coded. I tried adding another tag `utm_shoe_size` to the end of a sample url that I share in the original task definition, and the new tag wasn't picked up. 

In addition to that, our marketing person asked that we store the tags and include them in both json and csv exports to that they can do analysis on the submission results and render some analytics data. 

So with that I suggest we do the following changes:

1. Don't just read a fixed list of UTM tags, but instead let's read all query parameters. So say you have a path like this:

`https://example.com/path/?utm_tag1=value1&some_tag=value`

Under the `_meta.url` let's have the `location` as we do currently, and then instead of `utm` key, let's have `query` key that just extract the query params as they are in the path and inserts them as key-value pairs under that parameter, so we have something like this:

```
{
	... form payload data
	_meta: {
		dateSubmitted: '2023-01-16 11:38:22'
		url: {
			location: 'https://example.com/path/?utm_tag1=value1&some_tag=value'
			query: {
				utm_tag1: 'value1'
				some_tag: 'value'
			}
		}
	}
}
``` 
2. The second change is, let's store this information under the submission in the database so we can have it in our export.
3. Let's modify the copy JSON and copy CSV options to include this data, as well the multi submission CSV export. 
On the CSV side let's just flatten the structure so we have a propery like `_meta_url_query_utm_tag1`.

I know it's a lot of changes, but it's mostly my mistake I haven't thought this through properly in the start and now we have to circle back.

adrians5j:
Just a small note @SvenAlHamad , in my [last comment](https://github.com/webiny/webiny-js/pull/3124#pullrequestreview-1352191666) I suggested to @neatbyte-vnobis that we change `_meta` property to just `meta` and also `dateSubmitted` to `submittedOn`.

Main reason is: consistency with the rest of the system.

packages/api-form-builder/src/types.ts
```diff
@@ -54,6 +54,7 @@ export interface FbFormTriggerHandlerParams {
     addLog: (log: Record<string, any>) => void;
     trigger: FbFormTriggerData;
     data: FbSubmissionData;
+    meta: Record<string, any>;
```
adrians5j: Let's have this as `FbSubmissionMeta`.

packages/api-form-builder/src/plugins/crud/forms.models.ts
```diff
@@ -102,7 +102,8 @@ export const FormSubmissionCreateDataModel = withFields({
             ip: string({}),
             submittedOn: string({
                 value: new Date().toISOString()
-            })
+            }),
+            url: object()
```
adrians5j: Can we be more specific here?

An example I pulled from another piece of code:


```ts
    reCaptcha: fields({
        value: {},
        instanceOf: withFields({
            enabled: boolean(),
            /**
             * Note: We've replaced "i18nString()" with "string()"
             */
            errorMessage: string({
                value: "Please verify that you are not a robot."
            })
        })()
    })
```

packages/api-form-builder/src/plugins/graphql/form.ts
```diff
@@ -214,6 +214,7 @@ const plugin: GraphQLSchemaPlugin<FormBuilderContext> = {
             type FbSubmissionMeta {
                 ip: String
                 submittedOn: DateTime
+                url: JSON
```
adrians5j: Can we be more specific here? Why do we want to use JSON here?

packages/api-form-builder/src/plugins/graphql/form.ts
```diff
@@ -553,6 +554,22 @@ const plugin: GraphQLSchemaPlugin<FormBuilderContext> = {
                             }
                         }
 
+                        /**
+                         * Add meta fields.
+                         */
+                        for (let i = 0; i < submissions.length; i++) {
+                            const flattenExportMeta = flattenObj(
```
adrians5j: Since this is really related to form submission meta key, let's maybe rename this to `flattenSubmissonMeta`. Having it more explicit gives the info that its logic is specific to form submission meta, which it is.

Let's do the same on the frontend side.

packages/app-form-builder/src/admin/plugins/formDetails/formSubmissions/FormSubmissionsList/FormSubmissionDialog.tsx
```diff
@@ -67,8 +67,24 @@ const renderFieldValueLabel = (field: FbFormModelField, value: string): string =
     return getFieldValueLabel(field, value);
 };
 
+const flattenObj = (obj: Record<string, any>, parent: string, res: Record<string, string> = {}) => {
```
adrians5j: Let's apply the same rename here. Let's add a comment why we need this as well.

packages/api-form-builder/src/plugins/crud/utils.ts
```diff
@@ -125,3 +125,19 @@ export const sanitizeFormSubmissionData = (fields: FbFormField[], data: { [key:
 
     return data;
 };
+
```
adrians5j, line 128: Let's add a comment why we need this as well.

adrians5j (changes requested):
A couple of minor tweaks. Should be ready to merge after we're done.

packages/api-form-builder/src/plugins/triggers/webhook.ts
```diff
@@ -20,7 +21,13 @@ const plugin: FbFormTriggerHandlerPlugin = {
                 const response = await got(url, {
                     method: "post",
                     json: true,
-                    body: data
+                    body: {
+                        ...data,
+                        meta: {
+                            submittedOn: format(new Date(meta.submittedOn), "yyyy-MM-dd HH:mm:ss"),
+                            url: meta.url
```
adrians5j, line 30: Curious why are we directly setting `url: meta.url` and not simply spreading `meta` object here. So:

```ts
 meta: {
   ...meta,
   submittedOn: format(new Date(meta.submittedOn), "yyyy-MM-dd HH:mm:ss")
}
```

packages/api-form-builder/src/plugins/graphql/form.ts
```diff
@@ -553,6 +559,22 @@ const plugin: GraphQLSchemaPlugin<FormBuilderContext> = {
                             }
                         }
 
+                        /**
+                         * Add meta fields.
+                         */
+                        for (let i = 0; i < submissions.length; i++) {
+                            const flattenExportMeta = flattenSubmissionMeta(
```
adrians5j:
```suggestion
                            const flattenedSubmissionMeta = flattenSubmissionMeta(
```

packages/api-form-builder/src/plugins/crud/forms.models.ts
```diff
@@ -102,6 +102,13 @@ export const FormSubmissionCreateDataModel = withFields({
             ip: string({}),
             submittedOn: string({
                 value: new Date().toISOString()
+            }),
+            url: fields({
+                value: {},
+                instanceOf: withFields({
+                    location: string({}),
```
adrians5j:
```suggestion
                    location: string(),
```

adrians5j (changes requested):
Super close, we should be done in like 5mins and merge!

adrians5j (changes requested):
Super close, we should be done in like 5mins and merge!

packages/api-form-builder/src/plugins/triggers/webhook.ts
neatbyte-vnobis: @SvenAlHamad @adrians5j just in case:
if we will add whole meta object - IP will be added too
as meta contains user IP address
but it is sensitive data and may not be allowed to gather in some countries

adrians5j: Good point, thanks for pointing this out @neatbyte-vnobis .

OK, let's just add a comment:

```
// We don't spread the full `meta` object in order to ensure sensitive data 
// doesn't end up being included (at the moment, that's IP address).
```

neatbyte-vnobis:
@SvenAlHamad @adrians5j 
Any updates on this PR?

adrians5j (approved):

PR #3131, opened 2023-03-23
feat: introduce block variable support to more elements

neatbyte-vnobis:
## Changes
Implements issue "Introduce block variable support to all other elements".

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
Hey @neatbyte-vnobis - on this one, I think I might have missed one item in my specification, and that is the variable support for the icon element.

neatbyte-vnobis:
@SvenAlHamad 
This PR was updated and icon variable support was added (this includes some changes to editor icon component to support it).

Important thing: I have checked the page preview screen and fond out that embed elements are not displaying correctly, now it is fixed.
Also I have added required validation (placeholder, description, and url fields validation), it was missing and it could cause issues without it.

neatbyte-vnobis:
@SvenAlHamad @adrians5j 
Any updates on this PR?

adrians5j (changes requested):
Did a quick manual test. Some notes below ⬇️
 

1. Scratched 1st note. Sorry.

2. When I tried to test Twitter, I received the following err message after I dropped the block onto a new page ⬇️ 
image

BTW can confirm the same happened when I tried YT.

packages/api-page-builder/src/graphql/elementProcessors/embed.ts
```diff
@@ -15,7 +15,7 @@ export default new ContextPlugin<PbContext>(context => {
         const variables = useElementVariables(block, element);
         const value = variables?.length > 0 ? variables[0].value : null;
 
-        if (value !== null) {
+        if (value !== undefined && value !== null) {
```
adrians5j, line 18: Why not just do `typeof value === 'string'` here? 🤔

adrians5j (changes requested):
One question here.

PR #3133, opened 2023-03-23
feat: add support for resumable migrations

Pavel910:
## Changes
This PR adds support for resumable migrations. In reality, migrations are limited by Lambda function timeout. With this PR, we track the timeout, and store checkpoints to the DB, so we can continue where we left of last time. Repeated invoking of Lambda function is automated, using the `resume: true` flag in the response. If it's set - we invoke the Lambda, and do so for as long as migration process keeps returning that flag.

2 new utilities were added to the `data-migration` package: `runMigration` and `printReport`, both used in after-deploy hooks as well as manual CLI command execution.

## How Has This Been Tested?
Jest tests.


packages/data-migration/src/MigrationRunner.ts
```diff
@@ -196,10 +238,16 @@ export class MigrationRunner {
             ids.add(id);
         }
     }
+
+    private async createCheckpoint(migration: DataMigration, checkpoint: unknown) {
```
brunozoric, line 243: Maybe `createCheckpoint` can throw some `MigrationCheckpointFailed` error if it fails.
It could contain the JSON.stringified data, which user can use to continue migration manually?

brunozoric (approved):

PR #3139, opened 2023-03-27
fix: fix building graphql types for nested dynamic zones

ghost:
Changed the way the type is created within the dynamic zone field to include nested dynamic zones.

### DIFF
--- a/packages/app-headless-cms/src/admin/plugins/fields/dynamicZone.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fields/dynamicZone.tsx
@@ -58,13 +58,13 @@ export const dynamicZoneField: CmsEditorFieldTypePlugin = {
         },
         graphql: {
             queryField({ model, field }) {
-                const prefix = `${createTypeName(model.modelId)}_${createTypeName(field.fieldId)}`;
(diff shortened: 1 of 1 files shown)


brunozoric:
hi @alexvanloon-dh 
Unfortunately it is not only the GraphQL types on the UI side but some stuff in the API.
We need to sort that first and then we will evaluate your PR.

thanks :)

brunozoric:
Hi @alexvanloon-dh 

At the moment we cannot support Dynamic Zone in the Object field. We have some architectural changes to implement before we can make this possible.

There is a [PR](https://github.com/webiny/webiny-js/pull/3160) which will disable adding Dynamic Zone in the Object field.

ghost:
Hi @brunozoric 

Thanks for replying, do you have any idea if/when this will be possible? That will help us to decide to wait or find a workaround.

brunozoric:
Hi @alexvanloon-dh 

best case scenario: end of next querter (start of July)
worst case: not sure exactly as we need to do a lot architectural changes regarding CMS Models to make this work properly

ghost:
Ok thanks

brunozoric:
Dynamic Zone will work in Object field in 5.38.0. Closing...

PR #3146, opened 2023-03-28
fix: fix cursor positioning for button label

neatbyte-vnobis:
## Changes
Fix issue "Button label - cursor positioning issue"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elementSettings/components/InputField.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elementSettings/components/InputField.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState, useEffect } from "react";
 import { css } from "emotion";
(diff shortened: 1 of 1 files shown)


neatbyte-vnobis:
@SvenAlHamad @adrians5j @Pavel910 
Any updates on this PR?

adrians5j (changes requested):
A bit on the fence whether we should merge this or not.

On one side, it fixes the problem. On other hand, maybe we should've found the reason why this is happening? If I had to guess, cursor ending at the end of the input on each keystroke is a sign of complete component re-render. It's just a guess.

Can we double check why this is happening? Like spend 10-15minutes on it, and if we see it requires more time, then we leave this and leave a comment why this was left in this state.

Thanks @neatbyte-vnobis

PR #3147, opened 2023-03-28
Theme typography migration for Lexical Editor and other packages

mihajlovco:
In this PR included migration to the new typography types, a new upgrade of the emotion package, and lexical upgrades to the List and Quote components.

### Theme implementation affected:
- `apps/theme/theme.ts`- Changed the theme typography structure.
- `apps/theme` static pages updated
- upgrade the lexical editor to emotion 11
- cwp templates changes
- In `app-page-builder` and `app-page-builder-elements` implementation of the `DecoratedTheme` for interfaces, types and expression
- implemented `cssById` method as an extension to typography styles so users can access to single style 

##  Emotion CSS classes for the lexical nodes
- Introduction of a new type `ThemeEmotionMap` keeps the typography style data by style id. Now it's faster and easier to access the styles.
- Updated WebinyThemeEditor class with `emotionMap` prop so Lexical editor nodes can access to the styles.
- Implemented utility method `toThemeEmotionMap` that goes through the `ThemeStyles` and generates `ThemeEmotionMap` styles with the Emition `css()` generation method.
- New package dependency on `@emition/core` package

## Updated `useRichTextEdtior` context
- Introduced a new `TextBlockSelection` state that represents the currently selected lexical text nodes objects and toolbar state (status of the selected node or style format like _bold_ or _italic_)
- New utility function  `getLexicalTextSelectionState` that supports the processing of the current selection used in the `FloatingToolbar` component. In this toolbar component, we listen for the _`active editor state`_ and take the current text selection (`RangeSelection` object) for further processing.

> With this solution now the toolbar action will not care about the state of the editor but only will listen for data changes specific to the action.
- New `theme` state of type `WebinyTheme`.  This object is the actual webiny theme and is provided in the lexical editor trough `LexicalEditor` component in `lexical-editor-pb-element package`. Now this component is available for `TypograpyAction`, `TypograpyDropdown`(lexical-editor-actions), and available for all editor components.
- `toolabarType` state, values:  _header_  and _paragraph_ at the moment.
- New `ThemeEmotionMap` state - `themeEmotionMap`, `setThemeEmotionMap` - keeps a map of the theme styles with keys by id.

## Implementation of the new state in the toolbar actions components
This change removes unnecessary complexity and processing

Updated components:
- BoldAction
- BulletsListAction
- CodeHeoighlightAction
- ItalicAction
- NumberedListAction
- QuoteAction
- TypographyAction
- UnderlineAction

>  FontColorAction and FontSizeAction will be addressed with  the ColorPicker updates

## New `WebinyList` and `WebinyListItem` components that support typography styles
This component is taken and adapted to our needs from the Lexical codebase.

New components 
- WebinyListNode
- WebinyListItemNode
- WebinyListPlugin and related useWebinyList  hook - in this plugin we listen for List commands and insert/update/remove the node
- New utility functions in file `utils/nodes/list-node.ts` to support manipulation of the list nodes
- Defined list commands and command payload in `commands/list-command` file .

> Known issue: When you press `tab` key instead converting the current text into a list, the focus shifts outside from the lexical editor to the right sidebar input (image in the testing section).

## New `WebinyQuote` component that support typography styles
This component is taken and adapted to our needs from the Lexical codebase.
- WebinyQuote
- WebinyQuotePlugin and related hook useWebinyQuote - listen for commands on editor event registraction
- Updated QuoteAction 
- Update TypograpyDropdown to support new quote styles listing
- New utility function `formatToQuote` that converts the current selected node into quote text block
- Defined commands  and command payload in the `commands/webiny-quote` file.

## Implemented undo/redo commands on 
- Updated package with `@lexical/history` dependency
- Introduced a new context `SharedHistoryContext` that keeps and manages the changes.
- Imported the HistoryPlugin component from lexical in the base `RitchTextEditor` component.

## Fixed Heading component on drag&drop and related display of typography styles.
 
The issue was when the user drops the heading component into the editor, the user can see only paragraph styles from the typography menu and not the heading styles.
- Updated heading data in `page-builder-elements` package
- Updated TypographyAction 
- Updated TypograpyDropdown 

## How Has This Been Tested?
Manually.

### New TextBlockSelection and toolbar action implementation
Screenshot 2023-03-28 at 17 05 51
Screenshot 2023-03-28 at 17.10.30

### New List implementation with styles

https://user-images.githubusercontent.com/82515066/228349489-737c683a-2fa3-43b8-b411-147f0666121d.mov

### undo/redo in action

https://user-images.githubusercontent.com/82515066/228349190-63b0a708-025d-431c-a152-72554e7c56b8.mov

### Heading drag&drop fix

https://user-images.githubusercontent.com/82515066/228350542-51a1a50e-cc64-4b8f-97aa-add2ba430c95.mov


packages/app-page-builder-elements/src/createRenderer.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { usePageElements } from "~/hooks/usePageElements";
 import { Renderer, Element } from "~/types";
-import { Theme, StylesObject } from "@webiny/theme/types";
+import { DecoratedTheme, StylesObject } from "@webiny/theme/types";
 import { RendererProvider } from "~/contexts/Renderer";
 import { CSSObject, ClassNames } from "@emotion/react";
 
 interface GetStylesParams {
-    theme: Theme;
+    theme: DecoratedTheme;
```
adrians5j: Something to discuss maybe, should the name of the theme object be just `Theme` and not `DecoratedTheme`. Should be easier for users to understand, word `Decorated` might confuse them.

packages/app-page-builder-elements/src/modifiers/styles/text.ts
```diff
@@ -17,7 +18,10 @@ const text: ElementStylesModifier = ({ element, theme }) => {
 
         let breakpointStyles: CSSObject = {};
 
-        const typographyStyles = theme?.styles?.typography?.[values.typography];
+        const typographyStyles = getTypographyStyleById(
```
adrians5j, line 21: I assume we need this utility here because this will search typography within all categories, by ID. Right?

packages/app-page-builder-elements/src/utils/index.ts
```diff
@@ -165,3 +165,24 @@ export const elementDataPropsAreEqual = (prevProps: RendererProps, nextProps: Re
...
+ * */
+export const getTypographyStyleById = (
+    typographyId: string,
+    typography?: Typography
+): Record<string, any> | undefined => {
+    if (!typography) {
+        return undefined;
+    }
+    for (const key in typography) {
+        // @ts-ignore
```
adrians5j: Can we get rid of the ts-ignore here? If not, we usually leave a comment that justifies the usage.

packages/app-page-builder-elements/src/utils/index.ts
```diff
@@ -165,3 +165,24 @@ export const elementDataPropsAreEqual = (prevProps: RendererProps, nextProps: Re
...
     return prevRendererMetaHash === nextRendererMetaHash;
 };
+
+/*
+ * Desc: CSSObject style
+ * */
+export const getTypographyStyleById = (
+    typographyId: string,
+    typography?: Typography
+): Record<string, any> | undefined => {
```
adrians5j: Can `Record<string,any>` be improved here?

packages/app-page-builder/src/editor/plugins/elements/heading/elementText.ts
```diff
@@ -20,8 +20,9 @@ const lexicalText = JSON.stringify({
                 direction: "ltr",
                 format: "",
                 indent: 0,
-                type: "paragraph",
-                version: 1
+                type: "heading",
```
adrians5j, line 23: Why is the default here `heading`, with `h1` tag?

Let's maybe leave a comment.

packages/app-theme-manager/src/modules/themes/index.tsx
```diff
@@ -79,7 +79,11 @@ const WebsiteSettingsSelection = gql`
 const WebsiteSettings: React.FC = () => {
     return (
         <Fragment>
-            <Group name={"theme"} label={"Theme"} querySelection={WebsiteSettingsSelection}>
+            <Group
+                name={"theme"}
+                label={"DecoratedTheme"}
```
adrians5j: Should `DecoratedTheme` be here?

packages/cwp-template-aws/template/common/apps/theme/theme.ts
```diff
@@ -95,6 +121,7 @@ const theme = createTheme({
     breakpoints,
     styles: {
         colors,
+        // @ts-ignore
```
adrians5j: Should not be here.

packages/lexical-editor-actions/src/components/TypographyDropDown.tsx
```diff
@@ -0,0 +1,79 @@
...
+} from "@webiny/lexical-editor";
+import { usePageElements } from "@webiny/app-page-builder-elements";
+import { TypographyStyle } from "@webiny/theme/types";
+import { TypographyValue } from "@webiny/lexical-editor/types";
+
+export const TypographyDropDown = () => {
+    const { value, applyTypography } = useTypographyAction();
+    const { theme } = usePageElements();
+    const [styles, setStyles] = useState<TypographyStyle[]>([]);
+    const typographyStyles = theme.styles?.typography;
```
adrians5j: Let's not try to overdo the usage of `?.`.

For example, here, `theme.styles?.typography;` can be `theme.styles.typography;`, because `styles` and `typography` are not optional. It will always be in the theme object.

The same goes for other cases, for example below where we're accessing different types of typographies. Instead of `theme.styles?.typography?.headings`, `theme.styles.typography?.headings` is also fine.

packages/lexical-editor-actions/src/components/TypographyDropDown.tsx
```diff
@@ -0,0 +1,79 @@
...
+    const { textBlockSelection } = useRichTextEditor();
+    const textType = textBlockSelection?.state?.textType;
+
+    const hasTypographyStyles = (): boolean => {
```
adrians5j: Why the need for a function here? I guess just having `const hasTypographyStyles = !!typographyStyles` will work too.

packages/lexical-editor-actions/src/components/TypographyDropDown.tsx
```diff
@@ -0,0 +1,79 @@
...
+            }
+        }
+    }, [textType]);
+
+    return (
+        <>
+            {theme && hasTypographyStyles() ? (
+                <DropDown
+                    buttonClassName="toolbar-item typography-dropdown"
+                    buttonAriaLabel={"Formatting options for font size"}
```
adrians5j: Is this label valid?

packages/lexical-editor/src/commands/webiny-list.ts
adrians5j, line 1: Refer to @Pavel910 's comments on file naming conventions.

typings/emotion/index.d.ts
```diff
@@ -32,12 +32,42 @@ export type ThemeBreakpoints = {
...
+ * Decorated typography types
+ */
+export type DecoratedThemeTypographyStyles = ThemeTypographyStyleItems & {
+    stylesById: (id: string) => CSSObject | undefined;
+};
+
+export type DecoratedTypography = Record<TypographyType, DecoratedThemeTypographyStyles>;
+
+interface DecoratedThemeStyles extends Omit<ThemeStyles, "typography"> {
+    colors: Record<string, any>;
```
adrians5j: Do we need to again mention all of the fields when we're extending from `ThemsStyles` ?

packages/theme/src/types.ts
```diff
@@ -25,16 +25,52 @@ export type ThemeBreakpoints = {
...
+ */
+export type TypographyType = "headings" | "paragraphs" | "quotes" | "lists" | string;
+export type TypographyStyle = {
+    id: string;
+    name: string;
+    tag: string;
+    styles: CSSObject;
+};
+
+export type Typography = Record<TypographyType, Readonly<TypographyStyle[]>>;
```
adrians5j, line 39: I see the `Readonly`, but I'm also thinking about the `const` discussion we had before, and if it's possible to remove it. 

Maybe we can.

If we'll have time, let's double check this too.

packages/theme/src/index.ts
```diff
@@ -1,3 +1,27 @@
-import { Theme } from "~/types";
+import { DecoratedTheme, Theme } from "~/types";
+import { DecoratedTypography } from "../../../typings/emotion";
```
adrians5j: Should not be imported from here. Must be imported from the package itself.

packages/lexical-editor/src/utils/styleObjectToString.ts
```diff
@@ -0,0 +1,14 @@
+// @ts-ignore
```
adrians5j: Is this really neccessary?

packages/lexical-editor/src/utils/styleObjectToString.ts
```diff
@@ -0,0 +1,14 @@
...
+import reactToCSS from "react-style-object-to-css";
+
+/*
+ * Converts CSS style objects to string
+ * Example:
+ *  { fontSize: '10px' } => "font-size: 10px"
+ * */
+export const styleObjectToString = (styleObject: Record<string, any>): string => {
```
adrians5j: Can we change `Record<string, any>` to `CSSObject`? 

Also, check if we can get rid of the explicit `: string` return type. Let's have TS infer as much as possible.

packages/lexical-editor/src/types.ts
```diff
@@ -1,3 +1,97 @@
...
+    link: NodeState;
+    typography: NodeState;
+    fontColor: NodeState;
+    list: NodeState;
+    quote: NodeState;
+    textType: LexicalTextType;
+};
+
+/*
+ * @description Represent set of data from the current selection of the text and nodes selected by the user.
```
adrians5j: For consistency sake, no need to use these `@description` tags when writing comments.

In other code, you won't find any usages of it.

It's a nitpick, but yeah, keeping the code as consistent as possible does help in maintaining the project.

packages/lexical-editor/src/types.ts
```diff
@@ -1,3 +1,97 @@
...
+export type HeadingHtmlTag = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+export type ParagraphHtmlTag = "p";
+export type QuoteBlockHtmlTag = "quoteblock";
+
+// Typography
+export type TypographyHTMLTag = HeadingHtmlTag | ParagraphHtmlTag | ListHtmlTag | QuoteBlockHtmlTag;
+
+export type TypographyValue = {
+    // CSSObject type
+    css: Record<string, any>;
```
adrians5j: I guess we can change then this: 
```suggestion
    css: CSSObject
```

packages/lexical-editor/src/types.ts
```diff
@@ -1,3 +1,97 @@
...
+/* Nodes */
+export interface WebinyThemeNode {
+    /*
+     * @desc Get theme style id
+     * */
+    getStyleId: () => string;
+}
+
+/*
+ * Contains ids of the styles and emotion generated classes
```
adrians5j: In general, when writing comments, we try to be grammatically correct.

So, this would be more consistent with the rest of the codebase:

```suggestion
 * Contains IDs of the styles and Emotion generated classes.
```

packages/lexical-editor/src/types.ts
```diff
@@ -1,3 +1,97 @@
...
+    id: string;
+    tag: TypographyHTMLTag;
+    // Display name
+    name: string;
+};
+
+/* Nodes */
+export interface WebinyThemeNode {
+    /*
+     * @desc Get theme style id
```
adrians5j: The same applies here when it comes to `@desc` (check my above's `@description` comment, in this same file).

packages/lexical-editor/src/plugins/WebinyListPLugin/WebinyListPlugin.ts
```diff
@@ -0,0 +1,21 @@
+import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
+import { useEffect } from "react";
+import { WebinyListNode } from "~/nodes/list-node/WebinyListNode";
+import { WebinyListItemNode } from "~/nodes/list-node/WebinyListItemNode";
+import { useWebinyList } from "~/hooks/useWebinyList";
```
adrians5j, line 5: Please refer to @Pavel910's comments on naming things.

packages/lexical-editor/src/utils/toThemeEmotionMap.ts
```diff
@@ -0,0 +1,34 @@
+import { ThemeEmotionMap, TypographyHTMLTag } from "~/types";
+import { WebinyTheme } from "~/themes/webinyLexicalTheme";
+
+export const toThemeEmotionMap = (
```
adrians5j: Let's describe what this function does.

packages/lexical-editor/src/utils/toThemeEmotionMap.ts
```diff
@@ -0,0 +1,34 @@
...
+    for (const key in typographyStyles) {
+        const typographyTypeData = typographyStyles[key] as {
+            id: string;
+            tag: TypographyHTMLTag;
+            name: string;
+            styles: Record<string, any>;
+        }[];
+        if (typographyTypeData) {
+            typographyTypeData.forEach(styleItem => {
+                // 'le' is for shorter url generation
```
adrians5j: This is unclear. A better comment would be nice to have.

packages/lexical-editor/src/components/ToolbarActions/ItalicAction.tsx
```diff
@@ -1,18 +1,26 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { FORMAT_TEXT_COMMAND } from "lexical";
+import { useRichTextEditor } from "~/hooks/useRichTextEditor";
 
 /**
  * Toolbar action. On toolbar, you can see the button that is italic.
```
adrians5j: Not sure if this comment provides any value.

packages/lexical-editor/src/components/ToolbarActions/CodeHighlightAction.tsx
```diff
@@ -1,6 +1,7 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { FORMAT_TEXT_COMMAND } from "lexical";
+import { useRichTextEditor } from "~/hooks/useRichTextEditor";
 
 /**
  * Toolbar action. User can highlight the selected text.
```
adrians5j: Same as my prev comment - not sure if this comment provides any value.

adrians5j (changes requested):
Initial review done.

packages/lexical-editor/src/types.ts
mihajlovco: yup, we can update to `CSSObject` now, previously we didn't want to depend on emotion for this type. 

Thanks

packages/lexical-editor/src/types.ts
mihajlovco: That is ok, we need to maintain the comments as well,  I will know in the future what to do.

packages/lexical-editor/src/utils/toThemeEmotionMap.ts
mihajlovco: Good catch actually, I renamed this function to `toTypographyEmotionMap` and described what is mapped.

packages/lexical-editor/src/utils/styleObjectToString.ts
mihajlovco: yes, have some strange TS errors that I can't manage to get rid of:
` TS7016: Could not find a declaration file for module 'react-style-object-to-css'. '/Users/sashomihajlov/Documents/Webiny/webiny-js/node_modules/react-style-object-to-css/index.js' implicitly has an 'any' type.`
if I write the import declaration to  `const reactToCSS = require("react-style-object-to-css");` typescript will be ok with this but `reactToCSS` will be undefined.

mihajlovco: there are no types for react-style-object-to-css

packages/lexical-editor/src/utils/styleObjectToString.ts
mihajlovco: I will leave `string` for now, let me know if this must change. We can discuss it.

packages/lexical-editor/src/plugins/WebinyListPLugin/WebinyListPlugin.ts
mihajlovco: At the moment I will leave this for the polishing phase in favor of time because Lexical is sensitive to these kinds of changes. I need to test everything again.

packages/app-page-builder-elements/src/modifiers/styles/text.ts
mihajlovco: yes, we need to find `CSSObject` styles  by ID from specific typography types like `headings`.

packages/theme/src/types.ts
mihajlovco: yes, let's check, I guess we need to test the real webiny deployed project

typings/emotion/index.d.ts
mihajlovco: I had a problem, TS didn't recognize the other properties at some point. I need to check again but I did have that case.

packages/lexical-editor-actions/src/components/TypographyDropDown.tsx
mihajlovco: Thanks, I have changed to 'Typography formatting options'

packages/lexical-editor-actions/src/components/TypographyDropDown.tsx
mihajlovco: thanks, reduced the optional changing in the statement expressions

packages/cwp-template-aws/template/common/apps/theme/theme.ts
mihajlovco: Corrected

packages/app-page-builder-elements/src/utils/index.ts
mihajlovco: Corrected

mihajlovco: removed

packages/app-page-builder/src/editor/plugins/elements/heading/elementText.ts
mihajlovco: Set comment, hope is understandable. Let me know if I need to change.

packages/app-theme-manager/src/modules/themes/index.tsx
mihajlovco: corrected

packages/app-page-builder-elements/src/createRenderer.tsx
mihajlovco: yes, we can discuss that one

adrians5j: A relatively easy solution for this would be to rename the `Theme` that we have in the types.ts to, say, `BaseTheme` or `RawTheme`, or whatever. Then, we leave the `DecoratedTheme` as is, but we create a new `Theme = `DecoratedTheme`. That way users would be able to still import `Theme`. 

It's essentially all about DX.

If we were able to go the straightforward way, and simply do renames without of the above manouvers, I can see this PR's number of files also being reduced significantly.

packages/app-page-builder-elements/src/modifiers/styles/text.ts
adrians5j: It might not be a bad idea to simply id `stylesById` on the `typography` level in the theme object, but that's for another day. Gonna mark this one as resolvled.

packages/app-page-builder-elements/src/utils/index.ts
adrians5j: Ty

packages/app-page-builder-elements/src/utils/index.ts
adrians5j: Something to check.

packages/app-page-builder/src/editor/plugins/elements/heading/elementText.ts
adrians5j: You know what, I just figured this is a file located inside of the `heading` folder. Sorry about that, we can remove the comment all together. 🙂

packages/lexical-editor/src/commands/webiny-list.ts
adrians5j: Smt to check.

typings/emotion/index.d.ts
adrians5j: Smt to check.

packages/theme/src/types.ts
adrians5j: Let's see if we'll have the time.

packages/lexical-editor/src/utils/styleObjectToString.ts
adrians5j: Cool, let's just add a comment why this ts-ignore is here.

packages/lexical-editor/src/plugins/WebinyListPLugin/WebinyListPlugin.ts
adrians5j: Cool, let's leave this for the final polishing phase. Will leave this as unresolved.

packages/lexical-editor/src/utils/toThemeEmotionMap.ts
adrians5j: Cool.

And extra question here, is this part needed?

https://github.com/webiny/webiny-js/pull/3147/files#diff-8d69c7352b11a72d3c3d186151d4238357e6e341259afb01e96d0939519c51eaR19-R24

(the `as { ...` part)

packages/lexical-editor/src/components/ToolbarActions/UnderlineAction.tsx
```diff
@@ -1,13 +1,21 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
 import { FORMAT_TEXT_COMMAND } from "lexical";
+import { useRichTextEditor } from "~/hooks/useRichTextEditor";
 
 /**
  * Toolbar action. On toolbar, you can see the button that is underline.
```
adrians5j: Double check all actions for this redundant comment.

packages/app-page-builder-elements/src/utils/index.ts
mihajlovco: corrected to `CSSObject`

packages/lexical-editor/src/utils/toThemeEmotionMap.ts
mihajlovco: I am using `as {...}` part to describe the style object because, in the Lexical package, we do not import and use `Theme` type but `WebinyTheme` type where `styles` are described with `Record<string, any>`. 
It's for documentation purposes.

[You can check this file: https://github.com/webiny/webiny-js/pull/3147/files/e8913d34ca453a98ed5d2649e9d66508bb8096b3#diff-b290cd45c1b0eca83c2a1be02bf66ada24c0ab25b88446eb7daf46910b0a974apackages/lexical-editor/src/themes/webinyLexicalTheme.ts
](https://github.com/webiny/webiny-js/blob/e8913d34ca453a98ed5d2649e9d66508bb8096b3/packages/lexical-editor/src/themes/webinyLexicalTheme.ts)

packages/app-page-builder-elements/src/createRenderer.tsx
mihajlovco: I think BaseTheme is better in this context, I have implemented the solution, `Theme` now is an alias for `DecoratedTheme`, and yes the good news is that PR files are reduced now :), let me know if I need to change something today.

packages/app-page-builder/src/editor/plugins/elements/heading/elementText.ts
mihajlovco: removed

typings/emotion/index.d.ts
mihajlovco: I removed the other fields, seems to work fine, maybe at that moment I have other errors not related to this. Now it's fine.

packages/theme/src/types.ts
```diff
@@ -25,16 +25,54 @@ export type ThemeBreakpoints = {
...
     [key: string]: any;
 }
 
-export interface Theme {
+export interface BaseTheme {
     breakpoints: ThemeBreakpoints;
     styles: ThemeStyles;
 }
+
+export interface DecoratedTheme {
```
adrians5j: Let's not export `DecoratedTheme`, just so it doesn't confuse the users or the users accidentally don't use this. Let's make sure they only use `Theme`. Also, with this approach, we are sure nobody used `DecoratedTheme` and we can safely later remove it.

typings/emotion/index.d.ts
```diff
@@ -32,24 +32,55 @@ export type ThemeBreakpoints = {
...
+    typography: DecoratedTypography;
+}
+
+export interface BaseTheme {
     breakpoints: ThemeBreakpoints;
     styles: ThemeStyles;
 }
 
-type WTheme = Theme;
+export interface DecoratedTheme {
```
adrians5j: Same comment I had in theme/types.ts file.

adrians5j (changes requested):
One minor detail to resolve, and we might be able to merge this to start user testing.

packages/lexical-editor/src/components/ToolbarActions/UnderlineAction.tsx
mihajlovco: Checked and removed all.

typings/emotion/index.d.ts
mihajlovco: done

packages/theme/src/types.ts
mihajlovco: done

adrians5j (approved):

PR #3158, opened 2023-03-30
fix(api-headless-cms): dynamic zone in object

brunozoric:
## Changes

PR fixes dynamic zone in object field.

## How Has This Been Tested?
Jest and manually.


brunozoric:
Postponed for Q2 due to architectural decisions.

PR #3167, opened 2023-03-31
feat(api-page-builder): replace commodo with zod

brunozoric:
## Changes
This PR replaces the Commodo with the Zod library in API Page Builder package.

## How Has This Been Tested?
Jest, manually and Cypress.


packages/api-page-builder/src/graphql/crud/settings/validation.ts
```diff
@@ -0,0 +1,98 @@
...
+                        if (!value) {
+                            return value;
+                        }
+                        const { id } = parseIdentifier(value);
+                        return id;
+                    })
+            })
+            .passthrough()
+            .optional(),
+        theme: zod.string().optional()
```
Pavel910: Pease remove the `theme` attribute. This was hardcoded for Theme Manager, but is no longer necessary, now that we can properly extend GraphQL Schema. I'll test this after merge, with Theme Manager.

Pavel910 (approved):

PR #3173, opened 2023-04-04
fix: decrease api-migration memorySize

leopuleo:
## Changes
With this PR we reduce the `memorySize` to `3008` in order to fix error while deploying in case of newly created AWS accounts.

## How Has This Been Tested?
Manually


Pavel910 (approved):
What a dumb limitation, and even worse error reporting on Pulumi side 🤦 
Glad you ran into a solution! 🚀

PR #3175, opened 2023-04-04
feat: add accordion element

neatbyte-vnobis:
## Changes
Implements issue "Accordion element".

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-vnobis:
@adrians5j regarding our previous discussion about the styling:
> For the styles, it depends whether we want them to be editable by user or not.
> A) For example, button element styles are exposed in the theme.ts file, so users can edit their visual side. By using the same approach here, this enables the dev to just tweak certain visual aspects of the accordion. The downside I have in mind - we'll pollute the theme object with a bunch of styles related to accordion.
> B) Another solution is to give the dev the ability to provide the accordion renderer. Something like we already have with Pages List, we provide the default renderer, but, via plugins, users can provide new renderers / render list of pages in a different way. That way, all visual customizations are done via the renderer / emotion / by dev - from scratch.
> C) Finally, we can do a combo, where we don't expose any styles via the theme.ts by default (as there will probably be a lot of styles), but let the user do overrides via the theme.styles.elements.accordion property. ICYMI, the new engine will always, for each element, look into additional styles that might've ended up in `theme.styles.elements.[ELEMENT_NAME]`
> If you ask me, I'd do C)

I am not sure current implementation is exactly what you want, please take a look on it.

SvenAlHamad:
@neatbyte-vnobis I looked at both the tabs and the accordion implementation. I found the tab implementation having a much better user experience, for one single reason, that I'm able to interact with the component. Would it be possible to adjust the accordion implementation in a similar way, so that the user can click on the expand/contract button in the editor to open/close the accordion?

neatbyte-vnobis:
@SvenAlHamad 
Sure it can be done
The only thing is that this open/close state will not be saved (on preview and final page all items will be collapsed).
FYI we can add an item setting for default state (open/close)

SvenAlHamad:
That's fine, I need the expand/collapse option in the editor just for the editor experience. In some cases there might be a larger accordion and it's not always needed that all items are expanded and similar. The states don't need to be saved. The open/closed state for accordion we already have in the settings and that's fine for this implementation.

SvenAlHamad:
Just to confirm, from my end, this is looking good now.

packages/app-page-builder-elements/src/renderers/accordionItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+import { useRenderer } from "~/hooks/useRenderer";
+import { usePageElements } from "~/hooks/usePageElements";
+
+export type AccordionItemRenderer = ReturnType<typeof createAccordionItem>;
+
+export const createAccordionItem = () => {
+    return createRenderer(() => {
+        const { getElement } = useRenderer();
+        const element = getElement();
+        const { theme } = usePageElements();
```
adrians5j: `theme` can be imported from `useRenderer`, no need for this.

packages/app-page-builder-elements/src/renderers/accordionItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+        const [isOpen, setOpen] = useState(
+            element?.data?.settings?.accordionItem?.expanded || false
+        );
+
+        theme.styles.elements["accordion-item"] = {
```
adrians5j: Same comment as in [here](https://github.com/webiny/webiny-js/pull/3184#discussion_r1185856761).

packages/app-page-builder-elements/src/renderers/accordionItem.tsx
```diff
@@ -0,0 +1,90 @@
...
+
+                ".accordion-item-title": {
+                    ...theme.styles.typography.paragraph1,
+                    fontSize: "20px",
+                    fontWeight: 700,
+                    paddingRight: "24px"
+                },
+
+                ".accordion-item-arrow-icon": {
+                    "-webkit-mask-image": `url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNOC4xMiA5LjI5IDEyIDEzLjE3bDMuODgtMy44OGEuOTk2Ljk5NiAwIDEgMSAxLjQxIDEuNDFsLTQuNTkgNC41OWEuOTk2Ljk5NiAwIDAgMS0xLjQxIDBMNi43IDEwLjdhLjk5Ni45OTYgMCAwIDEgMC0xLjQxYy4zOS0uMzggMS4wMy0uMzkgMS40MiAweiIvPjwvc3ZnPg==")`,
```
adrians5j: Let's have these stored in a const, at the top of the file or in a separate file. Just not inline, a bit verbose.

packages/app-page-builder/src/editor/plugins/elements/accordionItem/PeAccordionItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    .accordion-item-title {
+        font-size: 20px;
+        font-weight: 700;
+        padding-right: 24px;
+        color: ${props => (props.isOpen ? "var(--mdc-theme-primary)" : "")};
+    }
+
+    .accordion-item-arrow-icon {
+        -webkit-mask-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNOC4xMiA5LjI5IDEyIDEzLjE3bDMuODgtMy44OGEuOTk2Ljk5NiAwIDEgMSAxLjQxIDEuNDFsLTQuNTkgNC41OWEuOTk2Ljk5NiAwIDAgMS0xLjQxIDBMNi43IDEwLjdhLjk5Ni45OTYgMCAwIDEgMC0xLjQxYy4zOS0uMzggMS4wMy0uMzkgMS40MiAweiIvPjwvc3ZnPg==");
```
adrians5j: The same comment re inline strings I posted above. Thanks!

packages/app-page-builder/src/editor/plugins/elements/accordionItem/PeAccordionItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+        mask-image: url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNOC4xMiA5LjI5IDEyIDEzLjE3bDMuODgtMy44OGEuOTk2Ljk5NiAwIDEgMSAxLjQxIDEuNDFsLTQuNTkgNC41OWEuOTk2Ljk5NiAwIDAgMS0xLjQxIDBMNi43IDEwLjdhLjk5Ni45OTYgMCAwIDEgMC0xLjQxYy4zOS0uMzggMS4wMy0uMzkgMS40MiAweiIvPjwvc3ZnPg==");
+        background-color: ${props =>
+            props.isOpen
+                ? "var(--mdc-theme-primary)"
```
adrians5j: If this is an element that will be rendered on a page / public website, than vars like these should definitely not be used. All of the theme-related variables are to be used from the theme object.

packages/app-page-builder/src/editor/plugins/elements/accordionItem/PeAccordionItem.tsx
```diff
@@ -0,0 +1,96 @@
...
+            </>
+        );
+    },
+    {
+        baseStyles: {
+            boxSizing: "border-box",
+            display: "flex",
+            flexDirection: "column",
+            width: "100%",
+            borderBottom: "2px solid var(--mdc-theme-on-background)"
```
adrians5j: Let's double check these usages of CSS vars. As mentioned in prev comment, these should not be used.

CSS env vars are only to be used when theming Admin app, because these are the vars used by the underlying material UI lib that we use. When it comes to PB elements, it's all CSS-in-JS / Emotion. BTW we have new docs soon to be publsihed here : https://docs-webiny-com-git-theming-docs-webiny.vercel.app/docs/page-builder/theming/introduction

adrians5j (changes requested):
Let's check a couple of things.

packages/app-page-builder/package.json
```diff
@@ -40,6 +40,7 @@
     "@webiny/react-composition": "0.0.0",
     "@webiny/react-router": "0.0.0",
     "@webiny/telemetry": "0.0.0",
+    "@webiny/theme": "0.0.0",
```
adrians5j: No need for this, the app already depends on `@webiny/app-theme`, which is the proper package to use when importing theme-related things.

packages/app-page-builder-elements/src/renderers/accordionItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+import styled from "@emotion/styled";
+import { Theme } from "@webiny/theme/types";
+import { Elements } from "~/components/Elements";
+import { createRenderer } from "~/createRenderer";
+import { useRenderer } from "~/hooks/useRenderer";
+
+const arrowIcon =
+    'url("data:image/svg+xml;base64,PHN2ZyB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciIHdpZHRoPSIyNCIgaGVpZ2h0PSIyNCIgdmlld0JveD0iMCAwIDI0IDI0Ij48cGF0aCBkPSJNOC4xMiA5LjI5IDEyIDEzLjE3bDMuODgtMy44OGEuOTk2Ljk5NiAwIDEgMSAxLjQxIDEuNDFsLTQuNTkgNC41OWEuOTk2Ljk5NiAwIDAgMS0xLjQxIDBMNi43IDEwLjdhLjk5Ni45OTYgMCAwIDEgMC0xLjQxYy4zOS0uMzggMS4wMy0uMzkgMS40MiAweiIvPjwvc3ZnPg==")';
+
+const AccordionItem = styled.div<{ theme: Theme }>`
```
adrians5j: did you really need to type the `theme` property here? Wasn't this automatically already the case?

adrians5j: I checked and saw it's not working. This is b/c of the outdated `typings/emotion/index.d.ts` file. Once you update the branch, should be working automatically. No need to do `theme: Theme`.

packages/app-page-builder-elements/src/renderers/accordionItem.tsx
```diff
@@ -0,0 +1,83 @@
...
+    border-bottom: 2px solid ${props => props.theme.styles.colors.color5};
+`;
+
+const AccordionItemHeader = styled.div<{ theme: Theme; open?: boolean }>`
+    position: relative;
+    cursor: pointer;
+    padding: 12px 10px 32px 10px;
+
+    .accordion-item-title {
+        ${props => props.theme.styles.typography.paragraph1};
```
adrians5j: Unfortunately, this is not how it works anymore. :(

```suggestion
        ${props => props.theme.styles.typography["paragraphs"].styledById("paragraph1")};
```

More info at https://www.webiny.com/docs/page-builder/theming/theme-object#typography

adrians5j: BTW the rest of the props (e.g. colors) should be fine, no changes were introduced there.

packages/app-page-builder/src/editor/plugins/elementSettings/accordion/AccordionItemsList.tsx
```diff
@@ -0,0 +1,211 @@
...
+import Accordion from "~/editor/plugins/elementSettings/components/Accordion";
+import { moveInPlace, useSortableList } from "~/hooks/useSortableList";
+import { Collapsable } from "~/editor/plugins/toolbar/navigator/StyledComponents";
+import { useUpdateElement } from "~/editor/hooks/useUpdateElement";
+import { ReactComponent as DragIndicatorIcon } from "~/editor/plugins/toolbar/navigator/assets/drag_indicator_24px.svg";
+import { ReactComponent as DeleteIcon } from "~/editor/assets/icons/delete.svg";
+import { createElement } from "~/editor/helpers";
+import { PbEditorElement } from "~/types";
+
+const accordionStyles = css`
```
adrians5j, line 18: If possible, let's not use `emotion` package. We'll be removing this in the near future.

We only want to rely on `@emotion/styled` and `@emotion/react` for styling.

adrians5j: On a second thought, just so I don't dwell on this too much, let's ignore this one. But for future PRs, please have this in mind. Thanks!

adrians5j (changes requested):
We're almost there. I'd say the most important thing now is to get this branch up-to-date with `next`. Left some comments below.

Also, while testing, I noticed there's a visual diff between editor and public website:
image

Thanks!

neatbyte-vnobis:
> We're almost there. I'd say the most important thing now is to get this branch up-to-date with `next`. Left some comments below.
> 
> Also, while testing, I noticed there's a visual diff between editor and public website: image
> 
> Thanks!

As we've decided with @SvenAlHamad expand/collapse option in the editor is just for the editor experience.
(in the editor all items are expanded by default, and on rendering they are collapsed by default)
To control public website expand state there is option inside accordion item settings "Expand by default" which is off by default.

That is why accordion items are expanded inside editor (expanded by default inside the editor to show the content) and collapsed on public website (collapsed by default or  affected by setting "Expand by default")

adrians5j:
Hey @neatbyte-vnobis, I should've added a description in the screenshot. It's actually not expanded/collapsed that I meant here, it's the colors. On the left side it's orange, on the right it's black.

neatbyte-vnobis:
> Hey @neatbyte-vnobis, I should've added a description in the screenshot. It's actually not expanded/collapsed that I meant here, it's the colors. On the left side it's orange, on the right it's black.

Title is orange when accordion item is expanded and black when collapsed. Same logic is present both in the editor and on public website

image

PR #3184, opened 2023-04-06
feat: add tabs element

neatbyte-vnobis:
## Changes
Implements issue "Tabs element".

image

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-page-builder-elements/src/renderers/tabs.tsx
```diff
@@ -0,0 +1,67 @@
...
+export type TabsRenderer = ReturnType<typeof createTabs>;
+
+export const createTabs = () => {
+    return createRenderer(() => {
+        const { getElement } = useRenderer();
+        const element = getElement();
+        const { theme } = usePageElements();
+        const [selectedTabElement, setSelectedTabElement] = useState(element.elements[0]);
+
+        theme.styles.elements["tabs"] = {
```
adrians5j: Mutating the `theme` object is not advised.

Rather than doing this, let's just create styled components above the `createTabs` and use them below.

adrians5j (changes requested):
Checked 15/22 files.

Re code, I left only one comment for now.

I'm also adding a couple of extra ntoes.

#### 1. Deleting Grid/Cell

If I delete Grid/Cell in the Tab, I can't drag and drop anything to it. I can only click / select element. Check video.

https://user-images.githubusercontent.com/5121148/236421521-aa1c6dcc-13f4-48a2-a3fc-6a18bd25dc38.mov

#### 2. Smaller Screens

I see when I enter mobile view, the tabs are just shrunk. Should this be done in a diff way? Maybe question for @SvenAlHamad ?

image


#### 3. Full Controll When Clicking on Tabs

At the moment, users edit the tab label by clicking on the tab, and then via the STYLE tab in the right sidebar (orange arrows). But if I want to delete a tab or move it around, I need to click on the Tabs, and then go to ELEMENT tab and do my adjustments there.

image

Seems to me it'd be easier to have everything just on the Tabs level. Less clicking and less confusion.

@SvenAlHamad thoughts?

#### 4. Migration

Note that, as it was the case in last quaterly release, this and some of the PRs that we're doing here will require a migration script. I'll handle that, I'll just tag PRs with `needs-migration` label, so it's easier for me to find it later.

SvenAlHamad:
Regarding:

2. Smaller Screens
-> I agree, having the list of tabs in one row that the user can scroll would be the preference. This solution can also work on larger screens if there are too many tabs that don't fit in one row.

3. Full Controll When Clicking on Tabs
-> This one I was thinking the same, but then when you come to carousel/slider component where you might have additional options on the item level, you will have a different UX for tabs, vs the UX on carousel/slider. I would prefer to have it the same, so I think keeping it as it is might be the better option.

packages/app-page-builder/src/editor/plugins/elements/tabs/PeTabs.tsx
```diff
@@ -0,0 +1,85 @@
...
+    background-color: ${props =>
+        props.active ? props.theme.styles.colors.color5 : props.theme.styles.colors.color6};
+    cursor: pointer;
+    ${props => props.theme.styles.typography.headings.stylesById("heading6")};
+    font-size: 14px;
+    transition: all 0.2s;
+    z-index: 40;
+
+    &:hover {
+        background-color: var(--mdc-theme-background);
```
adrians5j: CSS var leftover here.

adrians5j (changes requested):
One minor comment and resolve conflicts. Other than that, we're ready to merge.

adrians5j (approved):

PR #3195, opened 2023-04-11
Upgrade Theme Code (Use Theme via Newly Introduced Emotion Context)

adrians5j:
## Changes
This PR upgrades existing `apps/theme` code so that it uses the new approach of accessing the central theme object, and that is via Emotion's theme context. In other words, theme object is never imported directly via an import statement, but it's accessed via the `theme` prop that the Emotion passes to styled components.

## How Has This Been Tested?
Manual.

## Documentation
Potentially update changelog. But will revisit the upgrade process.


mihajlovco:
Looks good

mihajlovco (approved):

PR #3197, opened 2023-04-11
VPC Endpoints - Use Current AWS Region Instead Of Hardcoded `eu-central-1`

adrians5j:
## Changes
Prior to this PR, when the `useVpcEndpoints` feature would be used, the `eu-central-1` region would be used instead of the active one. This has now been addressed.

## How Has This Been Tested?
Manually.

## Documentation
Won't create any docs as the feature is not being used at the moment.


brunozoric:
@adrians5j will this break existing deployments?

adrians5j:
No, we did not integrate this with any of our EE clients (this feature is EE-only).

brunozoric (approved):

PR #3199, opened 2023-04-11
feat: add api-file-manager - aco integration

leopuleo:
## Changes
- Add `@webiny/api-file-manager-aco` package
- Add `listTags` method in ACO `search` content model.

## How Has This Been Tested?
Jest

## Documentation
Inline and packages docs


packages/api-aco/src/record/record.gql.ts
Pavel910, line 1: I would recommend adding a `tags` field resolver, which returns an empty array in case no tags were assigned originally. That way, you don't need to handle the empty value on the client side. Maybe even make it so that if search record is being created without tags, set an empty array, that way even programmatically, you'll always have as default value.

Then you can also set `tags: [String!]!` in the schema.

packages/api-aco/src/record/record.so.ts
```diff
@@ -57,6 +60,47 @@ export const createSearchRecordOperations = (
...
+
+                const meta = {
+                    hasMoreItems: false,
+                    totalCount: tags.length,
+                    cursor: null
+                };
+
+                return [tags, meta];
+            });
+        },
```
Pavel910: This implementation will not cut it. We already have users with millions of records in File Manager, now imagine loading 5 millions records to read unique tags. Won't go. I asked @brunozoric about it, and we'll need to add some generic functionality to our storage ops to allow this type of queries utilizing more optimal ways, so we can ask storage ops something like: `give me a list of unique values for given field on given model`. That way we'll be able to write a solid Elasticsearch query.

packages/cwp-template-aws/template/ddb-es/apps/api/graphql/src/index.ts
```diff
@@ -112,6 +113,7 @@ export const handler = createHandler({
         }),
         createAco(),
         createAcoPageBuilderContext(),
+        createAcoFileManagerContext(),
```
Pavel910, line 116: User project upgrade will be necessary for this, so open a task in Jira so we don't forget this.

Pavel910 (changes requested):
This looks great for the most part, but there _is_ an issue that we'll need to address, and we'll need @brunozoric for this. Bruno is informed on the issue, and I'll let you guys organize the work on this with @m0nikaza. We'll need to add some functionality to CMS storage ops to make this available to everyone.

packages/cwp-template-aws/template/ddb-es/apps/api/graphql/src/index.ts
brunozoric: Why not just make the the ACO part of the PB and FM? Its always loaded, right? We can just push it when creating the PB or FM context.

Pavel910: @brunozoric because it's not a requirement for PB or FM API to work. ACO is like an extra app on top of everything. That's why I want to keep it as separate packages, which can be replaced if necessary without digging through PB/FM/CMS respectively.

packages/api-aco/src/record/record.so.ts
brunozoric: this was resolved in the #3222

packages/api-aco/src/record/record.gql.ts
leopuleo: Thanks @Pavel910 👍 
I addressed this in #3228, which already merged into the `next` branch.

brunozoric (approved):
Everything looks fine. I'm just wondering if we should have a batchCreate method in the CMS - for example when uploading multiple files current code goes through a for loop and calls create methods one after another.
Not something we can introduce right now, but something to think about.

PR #3200, opened 2023-04-11
fix(data-migration): check for the unstable version

brunozoric:
## Changes
Add a check for the unstable version (0.0.0) into methods for migrations (ddb and ddb+es).

## How Has This Been Tested?
Manually.


adrians5j (approved):
LGTM

PR #3209, opened 2023-04-14
feat: add a date and time fields as field options

neatbyte-vnobis:
## Changes
Implements issue "Forms: Add a date and time fields as field options".

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-form-builder/src/admin/plugins/editor/formFields/components/Date.tsx
```diff
@@ -0,0 +1,224 @@
...
+        value: "+13:00",
+        label: "UTC+13:00"
+    },
+    {
+        value: "+14:00",
+        label: "UTC+14:00"
+    }
+];
+
+export const Date = ({ fieldFormat, Bind } : { fieldFormat: string, Bind: BindComponent }) => {
```
neatbyte-vnobis: @Pavel910 @adrians5j 
Maybe we have a better place where we can save this re-usable component?
If yes, then where should we create it?

packages/app-form-builder/src/types.ts
```diff
@@ -41,6 +41,10 @@ export type FbBuilderFormFieldValidatorPlugin = Plugin & {
             setValue: (name: string, value: any) => void;
             setMessage: (message: string) => void;
             data: FbBuilderFieldValidator;
+            // We need to return this optional "field" property in the case where we want to render different fields based on it's type or format
+            formFieldData?: {
```
adrians5j, line 45: I saw we're passing this here ```packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/EditFieldDialog/ValidatorsTab.tsx:78```.

Let's rather have the whole `form` param here.
image

packages/app-form-builder/src/render/plugins/validators/lteDate.ts
```diff
@@ -0,0 +1,28 @@
+import { validation } from "@webiny/validation";
+import { FbFormFieldValidatorPlugin } from "~/types";
+
+const plugin: FbFormFieldValidatorPlugin = {
+    type: "fb-form-field-validator",
+    name: "form-field-validator-lte-date",
+    validator: {
+        name: "lteDate",
```
adrians5j: Here we're using `lteDate`, and below `dateLte` (so, not the same order of the word `date` and `lte`). Can we fix this? The file name is also `lteDate.ts`.

packages/api-form-builder/src/plugins/validators/gteDate.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const plugin: FbFormFieldValidatorPlugin = {
+    type: "fb-form-field-validator",
+    name: "form-field-validator-gte-date",
+    validator: {
+        name: "gteDate",
+        validate: async (value, validator) => {
+            const regex = /(^\d*:\d*)/gm;
+
+            const validValue = `${validator.settings.value}${
```
adrians5j: This code is confusing. Can you divide it into multiple lines, maybe use specific var names, or at least add a comment.

packages/api-form-builder/src/plugins/validators/gteDate.ts
```diff
@@ -0,0 +1,28 @@
+import { validation } from "@webiny/validation";
+import { FbFormFieldValidatorPlugin } from "~/types";
+
+const plugin: FbFormFieldValidatorPlugin = {
+    type: "fb-form-field-validator",
+    name: "form-field-validator-gte-date",
+    validator: {
+        name: "gteDate",
+        validate: async (value, validator) => {
+            const regex = /(^\d*:\d*)/gm;
```
adrians5j: The same goes with this regex. Just naming this `regex` doesn't provide much info on what `/(^\d*:\d*)/gm;` is actually trying to do.

Please do the same for the `lteDate` validator.

packages/app-form-builder/src/admin/plugins/editor/formFieldValidators/lteDate.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { Date } from "~/admin/plugins/editor/formFields/components/Date";
+
+const plugin: FbBuilderFormFieldValidatorPlugin = {
+    type: "form-editor-field-validator",
+    name: "form-editor-field-validator-date-lte",
+    validator: {
+        name: "lteDate",
+        label: "Earlier or equal",
+        description: "Entered date/time must be equal or earlier compared to the provided date.",
+        defaultMessage: "Date/time is later than the provided one.",
```
adrians5j: This err message is a bit confusing. What is the "provided one" ?

I think, ideally, we'd want to have the exact date being included in the error message. @SvenAlHamad what do you think?

Like "Date must be later than X.". The same would go for "Date must be earlier than X.".

apps/theme/layouts/forms/DefaultFormLayout/fields/Date.tsx
```diff
@@ -0,0 +1,280 @@
...
+import { InputField, StyledInput } from "./Input";
+import { StyledSelect } from "./Select";
+import { Field as FieldWrapper } from "./components/Field";
+import { useBind, BindComponentRenderProp } from "@webiny/form";
+import { FormRenderFbFormModelField } from "@webiny/app-form-builder/types";
+import { FieldLabel } from "./components/FieldLabel";
+import { FieldErrorMessage } from "./components/FieldErrorMessage";
+import styled from "@emotion/styled";
+
+const UTC_TIMEZONES = [
```
adrians5j: Noticed this twice, plus also used in `packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/utils.tsx:16`.

Let's move this array into `@webiny/utils` package (`utcTimezones.ts`).

adrians5j (changes requested):
Left some comments.

Also, let's use Date/Time as the name of the field, the same we have in HCMS.

image

Finally, let's also back merge `next` and resolve conflicts.

adrians5j (changes requested):
Left some comments.

Also, let's use Date/Time as the name of the field, the same we have in HCMS.

image

image

packages/app-form-builder/src/admin/plugins/editor/formFieldValidators/lteDate.tsx
SvenAlHamad: Ideally yes as it will help the user to better understand how they can correct this error.

packages/api-form-builder/src/plugins/validators/dateGte.ts
adrians5j, line 1: Having just `date` in the name might cause confusion. From what I can see, this does both date and time validation. Let's make the filename reflect that.

packages/api-form-builder/src/plugins/validators/dateGte.ts
```diff
@@ -0,0 +1,31 @@
...
+const plugin: FbFormFieldValidatorPlugin = {
+    type: "fb-form-field-validator",
+    name: "form-field-validator-date-gte",
+    validator: {
+        name: "dateGte",
+        validate: async (value, validator) => {
+            // With this regex we check whether our value is time (11:59) and not a date
+            // We need it in order to apply correct validation rule (example: #line 23)
+            // Because for date and time validation rules are different (check @webiny/validation)
+            const timeValidator = /(^\d*:\d*)/gm;
```
adrians5j: This variable can go below in the code. Not used immediately.

packages/api-form-builder/src/plugins/validators/dateGte.ts
```diff
@@ -0,0 +1,31 @@
...
+
+            const validValue = `${validator.settings.value}${
```
adrians5j, line 13: This is also confusing. Let's add comments why we're doing this.

packages/api-form-builder/src/plugins/validators/dateGte.ts
```diff
@@ -0,0 +1,31 @@
+import { validation } from "@webiny/validation";
+import { FbFormFieldValidatorPlugin } from "~/types";
+
+const plugin: FbFormFieldValidatorPlugin = {
+    type: "fb-form-field-validator",
+    name: "form-field-validator-date-gte",
+    validator: {
+        name: "dateGte",
+        validate: async (value, validator) => {
+            // With this regex we check whether our value is time (11:59) and not a date
```
adrians5j: Let's be grammatically correct when writing comments. For example, here we're missing dots at the end of each sentence. Sorry to nitpick, but tiny details like these matter to us.

packages/api-form-builder/src/plugins/validators/dateGte.ts
adrians5j: Make sure to apply all of the things mentioned in this file in the `dateLte.ts` file and frontend counterpart files too. Thanks!

packages/app-form-builder/src/admin/plugins/editor/formFields/components/Date.tsx
adrians5j, line 1: Let's rename to `DateTime.tsx`. Thanks!

packages/app-form-builder/src/admin/plugins/editor/formFields/dateAndTime.tsx
adrians5j, line 1: `dateAndTime` rename to `dateTme`.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dateTime/DateTimeWithTimezone.tsx
```diff
@@ -13,6 +12,7 @@ import {
 } from "./utils";
 import { CmsModelField } from "~/types";
 import { BindComponentRenderProp } from "@webiny/form";
+import { UTC_TIMEZONES } from "@webiny/utils";
```
adrians5j, line 15: Awesome 🚀

apps/theme/layouts/forms/DefaultFormLayout/fields/Date.tsx
adrians5j: Awesome 🚀

apps/theme/layouts/forms/DefaultFormLayout/Field.tsx
```diff
@@ -1,12 +1,12 @@
 import React from "react";
 import { FormRenderFbFormModelField } from "@webiny/app-form-builder/types";
-
 import { InputField } from "./fields/Input";
 import { SelectField } from "./fields/Select";
 import { RadioField } from "./fields/Radio";
 import { CheckboxField } from "./fields/Checkbox";
 import { TextareaField } from "./fields/Textarea";
 import { HiddenField } from "./fields/Hidden";
+import { DateField } from "./fields/Date";
```
adrians5j, line 9: As mentioned in prev comment, should be `DateTimeField`. Thanks!

adrians5j (changes requested):
A couple of comments to check. I'll do one final usability test once we resolve these.
 
Additional notes ⬇️ 

1. I don't see any changes related to [this](https://github.com/webiny/webiny-js/pull/3209#discussion_r1206702680) comment from @SvenAlHamad. Should we maybe to these in a separate PR? What would be more ideal for you? We can do it in a separate PR too or even write it for later.

2. CI fails, there's a static check-related issue in there I saw.

packages/app-form-builder/src/admin/plugins/editor/formFieldValidators/lteDate.tsx
neatbyte-vnobis: @adrians5j @SvenAlHamad 

Currently in Form Builder validations user CAN specify his own "custom error message" (plain fixed string) for each added validation.
If this "custom error message" is left empty - UI shows `defaultMessage` (also a plain fixed string, just hard coded in the code).

For this `defaultMessage` we can modify plugin TS interface to accept a string of function (this function will return a string with more dynamic error message, for example with max value, etc).

Few things to confirm:
1) maybe it worth to improve `defaultMessage` for all types (integers, strings, etc, not only dates)?

2) how to show a date value for a case with timezone? (plain +1, +2 ??? GMT+2 ??? or also with a timezone name "EEST, Eastern European Summer Time") the easiest way is to use an ISO date format, but it is not "human readable" for common user

3) should we add this more detailed error message to an API validation too??? it has separate implementation inside API package

neatbyte-vnobis: @adrians5j @SvenAlHamad ?

adrians5j (approved):

adrians5j:
All good, let's just resolve the CI error and also let's quickly comment on [this](https://github.com/webiny/webiny-js/pull/3209#discussion_r1206702680). Should we write this down for future? @SvenAlHamad @neatbyte-vnobis

SvenAlHamad:
> All good, let's just resolve the CI error and also let's quickly comment on [this](https://github.com/webiny/webiny-js/pull/3209#discussion_r1206702680). Should we write this down for future? @SvenAlHamad @neatbyte-vnobis

Yeah, let's park that improvement for later.

PR #3228, opened 2023-04-21
feat: add tags to search records

leopuleo:
## Changes
With this PR: 

- we add `tags` to ACO search records - via `api-aco` package
- add page builder `tags` to ACO search records - via `api-page-builder-aco` package
- fix migration script for pages -> aco - via `migrations` package

## How Has This Been Tested?
Jest


brunozoric (approved):
Everything looks fine. Good that you set the tags to be empty array by default. One less thing to worry about.

PR #3233, opened 2023-04-24
fix(api-aco): prefix search record

brunozoric:
## Changes
Because the ACO uses the CMS as underlying storage an issue poped out, due to single table design, where ACO search record and the CMS Entry, being indexed for search, have same PK in both DDB and DDB+ES storage operations.

## How Has This Been Tested?
Jest and manually.

### DIFF
--- a/packages/api-aco/src/record/record.gql.ts
+++ b/packages/api-aco/src/record/record.gql.ts
@@ -89,9 +89,6 @@ export const searchRecordSchema = new GraphQLSchemaPlugin<AcoContext>({
             id: async parent => {
                 const { id } = parseIdentifier(parent.id);
                 return removeAcoRecordPrefix(id);
-            },
-            entryId: async parent => {
-                return removeAcoRecordPrefix(parent.entryId);
             }
         },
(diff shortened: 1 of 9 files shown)


packages/api-aco/src/record/record.gql.ts
```diff
@@ -83,6 +85,15 @@ export const searchRecordSchema = new GraphQLSchemaPlugin<AcoContext>({
         }
     `,
     resolvers: {
+        SearchRecord: {
+            id: async parent => {
+                const { id } = parseIdentifier(parent.id);
+                return removeAcoRecordPrefix(id);
+            },
+            entryId: async parent => {
```
leopuleo: I would remove it, since it has the same value of `id`. 
If we decide to keep it, we should add it to `SearchRecord` type

leopuleo (changes requested):
Thanks Bruno! It looks good to me, I left a comment about the GraphQL resolvers. The rest is 👍

leopuleo (approved):

PR #3253, opened 2023-04-26
feat(app-headless-cms): simple checkbox ref renderer

brunozoric:
## Changes
This PR adds a simple checkbox/radio button renderer for the reference field.
It loads ALL the entries in the models selected for the given reference field. It is not meant to be used with models which have a lot of entries as there is no pagination.

## How Has This Been Tested?
Manually.

Screenshot 2023-05-11 at 15 08 02


Pavel910 (approved):
Awesome! This was exactly what I needed, especially the checkboxes for multi-select.

PR #3259, opened 2023-04-26
feat: app-file-manager - aco integration

leopuleo:
## Changes
With this PR we add ACO to File Manager. The legacy viewer is still available, we can make it available via `FEATURE_FLAG`.

Screenshot 2023-04-26 at 20 42 50

### Features:
- New table list view available.
- Both Grid and Table list is generated using ACO search records.
- Users can organise files in folders.
- Tags rendered are coming from ACO search records.
- Search is integrated with ACO.
- `File` record is fetched only when the user opens the "File Details" drawer.
- The file mime-type filtering is handled via tags.

## How Has This Been Tested?
TBD - Check existing Cypress integration

## Documentation
TBD


packages/app-aco/src/components/FolderGrid/ActionDelete.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { MenuItem } from "@webiny/ui/Menu";
+
+import { ListItemGraphic } from "./styled";
+
+const t = i18n.ns("app-aco/folder-grid/action-delete");
+
+interface Props {
+    onClick: () => void;
+}
+export const ActionDelete = ({ onClick }: Props): ReactElement => {
```
brunozoric: Why not use the `React.VFC<Props>` type? Also, when you use the ReactElement you cannot return null or undefined as a component result.

There are multiple components with code like this, but I won't tag them.

packages/app-aco/src/components/FolderGrid/FolderGrid.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { Grid } from "~/components/FolderGrid/styled";
+
+import { FolderItem } from "~/types";
+
+interface FolderGridProps {
+    folders: FolderItem[];
+    onFolderClick: (folder: FolderItem) => void;
+}
+
+export const FolderGrid: ({ folders, onFolderClick }: FolderGridProps) => null | JSX.Element = ({
```
brunozoric: Unnecessary type, just use the `React.VFC<FolderGridProps>`.

packages/app-aco/src/components/TagList/Loader.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+
+import { Skeleton } from "@webiny/ui/Skeleton";
+
+import { LoaderContainer, SkeletonWrapper } from "./styled";
+
+export const Loader = () => {
```
brunozoric: When no props, use the `React.VFC`.

packages/app-aco/src/graphql/records.gql.ts
```diff
@@ -85,3 +86,15 @@ export const DELETE_RECORD = gql`
         }
     }
 `;
+
+export const LIST_TAGS = gql`
+    query ListTags($where: SearchRecordTagListWhereInput) {
+        search {
+            listTags(where: $where) {
+                data
```
brunozoric: Can you return an object here?
```
data {
        tag
    }
```
We will upgrade the query to return the amount of times a tag was used in the records. So it will look like:
```
data {
        tag
        count
    }
```

brunozoric (commented):
Small things I left comments about. Otherwise looks good.

packages/app-file-manager/src/components/Table/RecordActionDelete.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+import { FileItem } from "@webiny/app/types";
+import { SearchRecordItem } from "@webiny/app-aco/types";
+
+import { ListItemGraphic } from "./styled";
+
+const t = i18n.ns("app-admin/file-manager/components/table/record-action-delete");
+
+interface RecordActionDeleteProps {
+    record: SearchRecordItem<FileItem>["data"];
```
brunozoric, line 20: Isnt the `data` key actually `FileItem`?

brunozoric (approved):
I left you a single comment. No need to address it in this PR.

PR #3273, opened 2023-05-02
fix(headless-cms): data loader caching

brunozoric:
## Changes
This PR is a collection of fixes/refactoring.
* Benchmark default output must always be last in the execution line
* Force delete CMS Entry which does not exist in the DynamoDB (DDB+ES systems)
* Fix data loader caching for the CMS DDB systems
* Fix data loader caching for the CMS DDB+ES systems
* Export everything from the root of the db-dynamodb package

## How Has This Been Tested?
Jest and manually.

## Documentation
Changelog


packages/api-headless-cms/__tests__/testHelpers/useGraphQLHandler.ts
```diff
@@ -54,7 +54,12 @@ export const useGraphQLHandler = (params: GraphQLHandlerParams = {}) => {
 
     const core = createHandlerCore(params);
 
-    const plugins = new PluginsContainer(core.plugins.concat([acceptIncomingChanges()]));
+    const plugins = new PluginsContainer(
+        core.plugins.concat([
+            // createOutputBenchmarkLogs(),
```
adrians5j: This line of code is commented. This is fine?

packages/api-headless-cms/src/types.ts
```diff
@@ -2272,7 +2272,7 @@ export interface CmsEntryContext {
     /**
      * Delete entry with all its revisions.
      */
-    deleteEntry: (model: CmsModel, id: string) => Promise<void>;
+    deleteEntry: (model: CmsModel, id: string, force?: boolean) => Promise<void>;
```
adrians5j: Wondering if it'd be better to have `options` object as the 3rd param here. In case we need to add more options in the future. 

Also, can we add a short comment explaining what `force` does?

I'd also add a comment in the GQL schema for this param.

packages/api-headless-cms/__tests__/testHelpers/outputBenchmarkLogs.ts
```diff
@@ -0,0 +1,27 @@
+import pathPackage from "path";
+import writeJson from "write-json-file";
+import { ContextPlugin } from "@webiny/handler";
+
+/**
+ * Do not use in the CI!
```
adrians5j: Why do we don't want to use this in CI? A bit more explanation might be useful.

packages/api-headless-cms/__tests__/contentAPI/references.test.ts
```diff
@@ -1060,4 +1060,61 @@ describe("entry references", () => {
...
+                publish: true
+            });
+        }
+
+        const [result] = await articleRead.listArticles({
+            limit: 1000
+        });
+
+        expect(result.data.listArticles.data).toHaveLength(totalCount);
+        expect(result).toMatchObject({
```
adrians5j, line 1111: How is this assuring there aren't multiple DB calls? 🤔

packages/api-headless-cms-ddb/src/operations/entry/index.ts
```diff
@@ -371,12 +372,13 @@ export const createEntriesStorageOperations = (
 
     const deleteEntry: CmsEntryStorageOperations["delete"] = async (initialModel, params) => {
         const { entry } = params;
+        const id = entry.id || entry.entryId;
```
adrians5j, line 376: So, from what I understand here, the passed `id` can now be a revision ID or the entry ID, right?

packages/api-headless-cms-ddb/src/operations/entry/dataLoader/createBatchScheduleFn.ts
```diff
@@ -0,0 +1,10 @@
+import { CMS_ENTRY_BATCH_SCHEDULE_WAIT } from "./constants";
+
+export const createBatchScheduleFn = () => {
```
adrians5j, line 10: Would be probably useful to have a comment here, explaining what this function is.

packages/api-headless-cms-ddb/src/operations/entry/dataLoader/createBatchScheduleFn.ts
```diff
@@ -0,0 +1,10 @@
...
+    return (callback: () => void) => {
+        if (CMS_ENTRY_BATCH_SCHEDULE_WAIT <= 0) {
+            callback();
+        }
+        setTimeout(callback, CMS_ENTRY_BATCH_SCHEDULE_WAIT);
```
adrians5j, line 15: This `setTimeout` call is interesting. Is it possible for the Lambda function to return the result and this `setTimeout` call still not be resolved?

Or is this just for testing purposes? 🤔

packages/api-headless-cms-ddb/src/operations/entry/dataLoader/constants.ts
```diff
@@ -0,0 +1,2 @@
+const batchScheduleWaitEnv = Number(process.env.WEBINY_CMS_ENTRY_BATCH_SCHEDULE_WAIT || "0");
```
adrians5j: BTW it might be useful for us to start following the newest naming convention for env vars, the one we [released with 5.35](https://www.webiny.com/docs/release-notes/5.35.0/changelog#introduced-application-name-based-environment-variable-naming-3095).

packages/api-headless-cms/src/crud/contentEntry.crud.ts
```diff
@@ -1089,16 +1096,25 @@ export const createContentEntryCrud = (params: CreateContentEntryCrudParams): Cm
...
-        });
+        const storageEntry = (await storageOperations.entries.getLatestRevisionByEntryId(model, {
+            id
+        })) as CmsEntry;
 
-        if (!storageEntry) {
-            throw new NotFoundError(`Entry "${entryId}" was not found!`);
+        if (!storageEntry && !force) {
+            throw new NotFoundError(`Entry "${id}" was not found!`);
+        } else if (force) {
```
adrians5j: #### 1. Can we add comments maybe here? For example, I'm not clear what's L1111 doing / what's the intention here.

#### 2. Also, if I understood correctly, here we do:

Step1: If there's no `storageEntry` and `force` is false, just throw.
Step2: If `force` is true, then we do additional steps with `await deleteEntryHelper ...`

What's weird to me here this `else if (force)`. It's only checking the value of `force`. Shouldn't the complete condition here be `!storageEntry && force` ? If so, maybe a small refactor could be done here as well.

adrians5j (changes requested):
Checked the PR, with the comments below, two additional notes I'd add:

1. I know it's hard sometimes, but the PR could've been split into 2 or 3 standalone PRs.

2. I'm not super familiar with all of the HCMS internals (don't remember when's the last time I was doing anything here), so, although I believe everything is fine, especially when you wrote tests, it is possible I missed something. But, for me, I think the biggest value here was to actually see what changes are happening on the HCMS. In other words, just to kinda stay in the loop. 

Buy yeah, overall, great job as usual 🚀 🙂

packages/api-headless-cms/__tests__/testHelpers/useGraphQLHandler.ts
brunozoric: yes, this is to output logs in the files, for my local testing. ill change that to use some env variable

packages/api-headless-cms/__tests__/contentAPI/references.test.ts
brunozoric: debugger inspection, unfortunately. cannot validate via tests atm as we do not have access to the data loaders when doing the API calls. i can comment out test for now as i validated it is ok.

packages/api-headless-cms-ddb/src/operations/entry/dataLoader/createBatchScheduleFn.ts
brunozoric: It is not possible as this is the callback which loads the data from the DynamoDB. I added some comments above the function.

packages/api-headless-cms/src/types.ts
brunozoric: I added the options and explained what the force does.

packages/api-headless-cms/__tests__/testHelpers/outputBenchmarkLogs.ts
brunozoric: Explanation added.

packages/api-headless-cms-ddb/src/operations/entry/dataLoader/createBatchScheduleFn.ts
brunozoric: Added some explanation.

packages/api-headless-cms/src/crud/contentEntry.crud.ts
brunozoric: Changed according to the comment. I also added comments above this part of the code.

packages/api-headless-cms-ddb/src/operations/entry/dataLoader/constants.ts
brunozoric: Changed.

adrians5j (approved):

PR #3274, opened 2023-05-02
Lexical editor - fixed error message when dropdown menu is open

mihajlovco:
With this PR we resolved the error message in the console after opening the dropdown menu from the lexical toolbar.

Submitted screenshot:
image

## Changes
- Moved the dropdown component out of the button component
- Now the dropdown position is calculated
- dropdown will be opened after the calculation of the dropdown position is finished (there is no visible delay)

## How Has This Been Tested?
Manually.

**Note!** Warning is shown from the Grammarly plugin.

https://user-images.githubusercontent.com/82515066/235667074-10a4e392-861e-4dc7-a370-0da6204e10fb.mov

### DIFF
--- a/packages/lexical-editor/src/ui/DropDown.tsx
+++ b/packages/lexical-editor/src/ui/DropDown.tsx
@@ -200,9 +200,8 @@ export function DropDown({
     }, [dropDownRef, buttonRef, showDropDown, stopCloseOnClickSelf]);
 
     const displayContainer = (): CSSProperties => {
-        // shows the container only when dropdown position is calculated to prevent
-        // blinking and visibility of the element visibility first on left and
-        // after on the right position
+        // shows the container only when dropdown position is calculated to prevent blinking.
+        // without this, window will be visible first on left (0px) and after a millisecond on the right position
         return positionIsCalculated ? { display: "block" } : { display: "none" };
     };
 
--- a/packages/lexical-editor/src/ui/DropDown.tsx
+++ b/packages/lexical-editor/src/ui/DropDown.tsx
@@ -6,7 +6,7 @@
  *
  */
 
-import { ReactNode, useCallback, useEffect, useMemo, useRef, useState } from "react";
+import { CSSProperties, ReactNode, useCallback, useEffect, useMemo, useRef, useState } from "react";
 import * as React from "react";
 
 type DropDownContextType = {
@@ -43,7 +43,7 @@ export function DropDownItem({
     }, [ref, registerItem]);
 
     return (
-        <button className={className} onClick={onClick} ref={ref} title={title}>
+        <button className={className} onClick={onClick} ref={ref} title={title} type="button">
             {children}
         </button>
     );
@@ -122,7 +122,7 @@ function DropDownItems({
         <DropDownContext.Provider value={contextValue}>
             <div
                 className={`lexical-dropdown ${showScroll ? "" : "no-scroll"}`}
-                ref={dropDownRef ?? null}
+                ref={dropDownRef}
                 onKeyDown={handleKeyDown}
             >
                 {children}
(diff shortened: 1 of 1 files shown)


packages/lexical-editor/src/ui/DropDown.tsx
```diff
@@ -162,42 +166,70 @@ export function DropDown({
...
+        }
+    }, [dropDownRef, buttonRef, showDropDown, stopCloseOnClickSelf]);
 
-        return () => {
-            document.removeEventListener("click", handle);
-        };
-    }, [buttonRef, showDropDown, stopCloseOnClickSelf]);
+    const displayContainer = (): CSSProperties => {
+        // shows the container only when dropdown position is calculated to prevent blinking.
+        // without this, window will be visible first on left (0px) and after a millisecond on the right position
```
adrians5j:
```suggestion
        // To prevent blinking, we show the container only when the dropdown position is calculated.
        // Without this, window would be visible first on left (0px), and after a millisecond on the right side.
```

adrians5j: Thanks for the comment BTW.

adrians5j: BTW we could probably use a `useMemo` here, and have `positionIsCalculated` set as its cache key.

packages/lexical-editor/src/ui/DropDown.tsx
```diff
@@ -162,42 +166,70 @@ export function DropDown({
...
+        const dropDown = dropDownRef.current;
+
+        if (showDropDown && button !== null && dropDown !== null) {
+            dropDown.style.top = "44px";
+            dropDown.style.left = `${button.offsetLeft}px`;
+            setPositionIsCalculated(true);
         }
+    }, [dropDownRef, buttonRef, showDropDown]);
+
+    // @ts-ignore - keeping the original logic from lexical that works
```
adrians5j: Can we go without a ts-ignore here? In general, we try to avoid them, unless there's a good reason to use it.

Can you explain in more detail why we're using ts-ignore here?

packages/lexical-editor/src/ui/DropDown.tsx
```diff
@@ -162,42 +166,70 @@ export function DropDown({
...
         const button = buttonRef.current;
-        if (button === null || !showDropDown) {
-            return;
+        const dropDown = dropDownRef.current;
+
+        if (showDropDown && button !== null && dropDown !== null) {
+            dropDown.style.top = "44px";
+            dropDown.style.left = `${button.offsetLeft}px`;
+            setPositionIsCalculated(true);
         }
```
adrians5j:
```suggestion
     if (showDropDown && dropDown && button) {
            dropDown.style.top = "44px";
            dropDown.style.left = `${button.offsetLeft}px`;
            setPositionIsCalculated(true);
        }
```

A bit more elegant maybe? ⬆️

packages/lexical-editor/src/ui/DropDown.tsx
```diff
@@ -150,10 +150,14 @@ export function DropDown({
     stopCloseOnClickSelf?: boolean;
     showScroll?: boolean;
 }): JSX.Element {
+    const dropDownRef = useRef<HTMLDivElement>(null);
     const buttonRef = useRef<HTMLButtonElement>(null);
+    // Prevent blinking pre- / post-calculation of the position of the dropdown
```
adrians5j:
```suggestion
    // Used to prevent flickering of the dropdown while calculating the dropdown position.
```

adrians5j (changes requested):
A couple of minor comments. But I tried the editor, and the errors is no longer there. Thanks!

packages/lexical-editor/src/ui/DropDown.tsx
mihajlovco: Here it's complaining that all paths do not return value. Previously I have some changes to eliminate that but there is no reason, works as expected. I will just remove the comment here.

mihajlovco: suggestion: `// @ts-ignore - it's ok that all paths do not return value`

PR #3275, opened 2023-05-02
fix(api-headless-cms): delete multiple entries

brunozoric:
## Changes
This PR adds delete multiple entries (by given IDs). Max entries allowed in multiple delete method is 50 due to 30 seconds API Gateway limit.

Also, removing a few of unused / unnecessary tests in the PR.

## How Has This Been Tested?
Jest.


adrians5j (approved):
Looks good!

PR #3276, opened 2023-05-03
feat: add carousel element

neatbyte-vnobis:
## Changes
Implements issue "Carousel element".

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-page-builder-elements/src/renderers/carousel.tsx
```diff
@@ -0,0 +1,99 @@
...
+import { useRenderer } from "~/hooks/useRenderer";
+import { usePageElements } from "~/hooks/usePageElements";
+import { registerCarousel } from "~/modifiers/carousel/index";
+
+export type CarouselRenderer = ReturnType<typeof createCarousel>;
+
+const CarouselWrapper = styled.div`
+    display: flex;
+    justify-content: center;
+    --swiper-navigation-color: var(--mdc-theme-primary);
```
adrians5j: Let's remove the usage of CSS vars.

packages/app-page-builder-elements/src/modifiers/carousel/index.ts
```diff
@@ -0,0 +1,8 @@
+import { initializeSwiper } from "./initializeSwiper";
+
+export const registerCarousel = () => {
+    new Promise<void>(async resolve => {
```
adrians5j: This is OK, but note that this code should not be placed within modifiers.

With AOS, it was registered as a modifier because animation is part of the element styles (selectable via right sidebar in the editor).

With this, we can have Swiper initialized in the actual page element renderer component.

packages/app-page-builder/src/editor/plugins/elements/carousel/PeCarousel.tsx
```diff
@@ -0,0 +1,166 @@
...
+    useEffect(() => {
+        const swiperContainer = swiperRef?.current;
+        const params = {
+            rebuildOnUpdate: true,
+            injectStyles: [
+                `
+                    .swiper-button-next,
+                    .swiper-button-prev {
+                        width: 40px;
+                        color: var(--mdc-theme-primary);
```
adrians5j: CSS vars usage here.

adrians5j (changes requested):
First iteration done.

Note that updating the branch with the latest could will also be required.

packages/app-page-builder/src/editor/contexts/EditorPageElementsProvider/ElementControls.tsx
```diff
@@ -11,7 +11,7 @@ import { uiAtom } from "~/editor/recoil/modules";
 // Lists elements that, when empty, can receive other elements as children using
 // drag and drop. For now, the element types that are hardcoded. Down the road,
 // we might want to expose this, enabling users to create more complex elements.
-const EMPTY_DROPPABLE_ELEMENTS = ["block", "cell"];
+const EMPTY_DROPPABLE_ELEMENTS = ["block", "cell", "carousel-element"];
```
adrians5j: We finally fixed this, and the fix is in `dev` branch. Will move the changes into `next` now.

Essentially, you won't need this. You'll just add the `canReceiveChildren: true` ([example](https://github.com/webiny/webiny-examples/blob/master/page-builder/element-with-child-elements/apps/theme/pageElements/contentGalleryDropzone/admin/index.tsx#L12)) in Carousel's PB Editor plugin and that's it. Also, the `onReceived` is no longer needed. 

I added the needed changes on previously implemented Tab and Accordion components. But I think this PR doesn't need any further changes, at least from what I saw it the editor plugin.

packages/app-page-builder-elements/src/renderers/carousel.tsx
```diff
@@ -0,0 +1,120 @@
...
+    namespace JSX {
+        interface IntrinsicElements {
+            "swiper-container": { [key: string]: any };
+            "swiper-slide": { [key: string]: any };
+        }
+    }
+}
+
+export const createCarousel = () => {
+    registerCarousel();
```
adrians5j: This will still be practically immediately executed, because the `createCarousel` is executed immediately, when the plugins are registered. That's eager load.


So, we need it to get executed when the component is actually being rendered.

I'd say just initializing Swiper's stuff on the first component render should be fine. No need even to do those setInterval calls (with AOS it was a bit different).

packages/app-page-builder/src/editor/plugins/elements/carousel/PeCarousel.tsx
```diff
@@ -0,0 +1,100 @@
...
+        }
+    }, [childrenElements.length]);
+
+    const navProps = {
+        navigation: arrowsToggle,
+        pagination: bulletsToggle
+    };
+
+    return (
+        <PeCarouselWrapper>
```
adrians5j, line 86: Can't we just import the rendered from `app-page-builder-elements` ?

E.g. like we do it here: packages/app-page-builder/src/editor/plugins/elements/button/PeButton.tsx:8

adrians5j (commented):
Left a couple of comments.

Additional Notes ⬇️ 

1. Should this field be so small? @SvenAlHamad Should we make it wider?

image

2. We have a cut off here, not sure if we can do anything about it? 
image

Thanks!

P.S. Just merged `dev` -> `next`.

SvenAlHamad:
Regarding the drop zone, ideally we would make it full width by default, so it looks the same as what we have in an empty grid block.

packages/app-page-builder/src/editor/contexts/EditorPageElementsProvider/ElementControls.tsx
neatbyte-vnobis: @adrians5j 
Got a message from Ilya:
I've done everything that you said, removed `onReceived` function, added `canReceiveChildren` to the `PbCarousel`, but  drag-n-drop functionality is not working.

Also drag-n-drop not working for a Tabs component (it should be working, yes?).
I've checked it on the latest next branch too - same, drag-n-drop is not working for tabs.
Can I ask you to confirm this? I want to be sure it is not some local issue before trying to fix it.
(I have rebuilded `next` successfully today, so my testing was done on latest codebase)

Maybe I misunderstand use case (aka steps)?
My steps are:
1. Add Tabs element
2. Remove default inner Grid from this Tabs element
3. All some other element instead (drag-n-drop is not working on this step)
And click on "Add element" is NOT working well.

packages/app-page-builder/src/editor/plugins/elements/carousel/PeCarousel.tsx
neatbyte-ivelychko: Maybe you are right!

But I have to clarify why we use separate react component `PeCarousel` instead of using a prerendering component `Carousel`.

1 - We need to use `useEffect` inside of `PeCarousel` to update Carousel whenever we add or remove a slide;
see here - https://github.com/webiny/webiny-js/pull/3276/files#diff-816934b2321ac03eb3e6bbdb09eced3c835698943e814901b642d999ec8965d2R68

2 - Also we use a ref to get an access to Swiper API methods;

3 - (Not actual anymore as [bug reported by Vitalii](https://github.com/nolimits4web/swiper/issues/6672) was already fixed in Swiper lib) Initially there were fixes to properly show a Carousel when arrows and bullets control settings were changed. (when we can configure in Carousel setting whether we want to show or hide arrows or bullets);

4 - All these things related to a Carousel settings changes or slides changes are possible only inside editor (published page will never see such case, as there is no data change after Swipper initialization);

We can make it as you have mentioned, but in such case I have to add `useEffect` and `ref` to the prerendering [Carousel](https://github.com/webiny/webiny-js/pull/3276/files#diff-009e5dab9be1a7d2a95e5ed3fe58f7e45a423b8858fd754f6f099591260e5a2eR79) component, even though published page does not need them.

packages/app-page-builder/src/editor/contexts/EditorPageElementsProvider/ElementControls.tsx
adrians5j: ### Tabs Isssue

Yes. Although definitely not something user would expect, that's actually to be expected from the code point of view. Because every element basically has their `target` property set to `cell, block`. While the grid>cell is in tab, it works. The moment you remove the grid/cell, you no longer have an element that can go into the leftover `tab` page element.

There are two ways to approach this. We either prevent grid>cell from being removed in the tabs > tab element, or we go through each page element, and add the `tab` element to their `target` property. So, we'd end up with `cell, block, tab`. 

Both ideas kinda are hacks, but you know what, let's resolve this in a separate PR. For this PR, let's pretend the user would not remove the grid > cell in the tabs > tab element.

packages/app-page-builder-elements/src/renderers/carousel.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    register.register();
+};
+
+const registerCarousel = () => {
+    new Promise<void>(async resolve => {
+        await initializeSwiper();
+        resolve();
+    });
+};
```
adrians5j: Let's simplify this and simply put 

```ts
 if (typeof window === "undefined") {
        return;
    }

    const register = await import("swiper/element/bundle");

    register.register();
```

into `useEffect` that only executes on the initial component render.

packages/app-page-builder/src/editor/plugins/elements/carousel/PeCarousel.tsx
```diff
@@ -0,0 +1,102 @@
+import React, { useRef, useEffect } from "react";
+import styled from "@emotion/styled";
+import { createRenderer, useRenderer, Elements } from "@webiny/app-page-builder-elements";
+import { Element } from "@webiny/app-page-builder-elements/types";
+import { useRecoilValue } from "recoil";
+import { elementWithChildrenByIdSelector } from "~/editor/recoil/modules";
+import { register } from "swiper/element/bundle";
+
+register();
```
adrians5j: Let's also have this in a `useEffect` call. No need to have this called immediately.

packages/app-page-builder/src/editor/plugins/elements/carousel/PeCarousel.tsx
```diff
@@ -0,0 +1,102 @@
...
+    const { getElement, meta } = useRenderer();
+    const element = getElement();
+    const elementWithChildren = useRecoilValue(
+        elementWithChildrenByIdSelector(element.id)
+    ) as Element;
+    const childrenElements = elementWithChildren?.elements;
+
+    const { arrowsToggle = true, bulletsToggle = true } = element?.data?.settings?.carousel;
+
+    const zIndex = meta.depth * 10;
```
adrians5j, line 57: Let's put an explanation here. Why are we doing this?

packages/app-page-builder/src/editor/plugins/elements/carousel/PeCarousel.tsx
adrians5j: All good, I can see now essentially the components differ so much that it makes sense to have separate components. Resolving this convo.

adrians5j (changes requested):
A couple of minor comments.

Should be mergeable soon. 

I left some comments regarding the tabs not workign when you remove grid > cell. Check them out. All in all, I say we write this down for future.

P.S Please check CI errors. We'll need to address those too before merge.

adrians5j:
As agreed with @SvenAlHamad , let's update the `target` property of `Grid` page element so that it can also be dropped into these new elements that accept children, like tab and carousel. 

With this approach, even if the user removes the grid > cell from these elements, they'll be able to drop a new grid back.

@neatbyte-vnobis @neatbyte-ivelychko

adrians5j (approved):

adrians5j (approved):

PR #3277, opened 2023-05-03
fix(api-aco): models for multiple tenants and locales

brunozoric:
## Changes
This PR removes the default tenant and locales added to the ACO models as they are not working when switching the tenant and locale in the code.

## How Has This Been Tested?
Jest.


leopuleo (approved):
👍 thanks Bruno

PR #3279, opened 2023-05-04
(PoC) Lexical nodes refactor - WIP 

mihajlovco:
## Goals
- Create heading and paragraph custom nodes. The goal is to remove dependence on typography nodes or lexical so we have more control over the actions.
- Create a theme text node that can handle all styles from the theme and custom/inline styling.
- Create styling patch for the theme text node


mihajlovco:
PoC is done.

PR #3282, opened 2023-05-04
Lexical List and Quote components - fix copy pasting from another source

mihajlovco:
## Changes
- Added `PastPlugin` to preserve the `new lines` in the text

### Numbered and bullet list
- WebinyListNode - set theme `styleId` by default on dom creation and update 

### Quote
- WebinyListNode - set theme `styleId` by default on dom creation and update 

### How styles are found?
Both components will get theme styles by searching for the `first occurrence `of the `HTML tag` in the `ThemeEmotionMap` in the lexical editor. 

### Fallback: 
Yes, we have a fallback in case the user doesn't have any style lexical will provide some default occurrence of the Quote and Lists as default styles, further the user can apply the desired style from the typography menu.

### How Has This Been Tested?
Manually.

Tested on Google Docs, various web sites (W3School, space.com), the Notion app, and Apple Notes app.
- `Bold` and `italic` work in all of the mentioned apps and the `underline` is not working for Notion and Apple notes because they have some custom implementation for the underline (that behavior is approved by @SvenAlHamad ). 

### List testing

https://user-images.githubusercontent.com/82515066/236257493-bc01436b-4797-4d3a-80cc-e92450490c64.mov

#### Block variables for List testing

https://user-images.githubusercontent.com/82515066/236412154-35cde87f-a2e8-46d2-a800-c88c990ad027.mov

### Quote testing

https://user-images.githubusercontent.com/82515066/236257627-8e8c9cb9-c8c4-4a2d-93a1-0a0b1cdbfe8c.mov

#### Block variables for Quote testing

https://user-images.githubusercontent.com/82515066/236412638-96dcbc44-401f-4221-9d92-db2955b1a9da.mov


packages/lexical-editor/src/plugins/PastePlugin/PastePlugin.tsx
```diff
@@ -0,0 +1,34 @@
...
+ *
+ */
+
+import { useLexicalComposerContext } from "@lexical/react/LexicalComposerContext";
+import { COMMAND_PRIORITY_NORMAL, PASTE_COMMAND } from "lexical";
+import { useEffect } from "react";
+
+export function PastePlugin(): JSX.Element | null {
+    const [editor] = useLexicalComposerContext();
+    // @ts-ignore - it's ok that all paths do not return value
```
adrians5j: Instead of ts ignoring this, let's rather just have a `return` statement where needed. From what I see, it's below L25.

packages/lexical-editor/src/utils/findTypographyStyleByHtmlTag.ts
```diff
@@ -0,0 +1,14 @@
+import { ThemeEmotionMap, ThemeEmotionStyle } from "~/types";
+
+export const findTypographyStyleByHtmlTag = (
+    htmlTag: string,
+    themeEmotionMap: ThemeEmotionMap
+): ThemeEmotionStyle | undefined => {
```
adrians5j: If TS can infer the return type, no need to explicitly set it here.

packages/lexical-editor/src/utils/findTypographyStyleByHtmlTag.ts
```diff
@@ -0,0 +1,14 @@
...
+    for (const key in themeEmotionMap) {
```
adrians5j: I see we're using `styledId` in the `ThemeEmotionMap` interface. Let's replace `key` with `styleId` here, just to make things more consistent / more readable.

packages/lexical-editor/src/types.ts
```diff
@@ -83,16 +83,18 @@ export interface WebinyThemeNode {
...
+    name: string;
+    styles: Record<string, any>;
+    // emotion generated class
+    className: string;
+};
+
 /*
  * Contains IDs of the styles and Emotion generated classes.
  */
 export type ThemeEmotionMap = {
```
adrians5j, line 98: When possible, we normally prefer interfaces over types. A nitpick, but it does help making the whole codebase more consistent.

packages/lexical-editor/src/nodes/ListNode/WebinyListNode.ts
```diff
@@ -137,11 +143,35 @@ export class WebinyListNode extends ElementNode {
...
+        }
+
+        const style = findTypographyStyleByHtmlTag(tag, themeEmotionMap);
+        if (style) {
+            this.__themeStyleId = style.id;
+        }
+    }
+
+    hasThemeStyle(): boolean {
+        return this.__themeStyleId !== undefined && this.__themeStyleId.length > 0;
```
adrians5j:
```suggestion
        return !!this.__themeStyleId;
```

packages/lexical-editor/src/nodes/ListNode/WebinyListNode.ts
```diff
@@ -61,6 +62,11 @@ export class WebinyListNode extends ElementNode {
             dom.setAttribute("start", String(this.__start));
         }
 
+        // update styles for different tag styles
```
adrians5j: This comment is needed? Could be improved if so. Atm doesn't bring any value.

packages/lexical-editor/src/nodes/ListNode/WebinyListNode.ts
```diff
@@ -137,11 +143,35 @@ export class WebinyListNode extends ElementNode {
         };
     }
 
+    setDefaultThemeListStyleByTag(tag: string, theme: WebinyTheme) {
```
adrians5j, line 145: In general, I don't understand why is this needed or what's this fixing. 

I mean, lists are being styled on my end fine, without this logic in place.

https://user-images.githubusercontent.com/5121148/236429980-953367fa-fd6b-422f-9745-e043428210ac.mov




Can you clarify a bit?

apps/theme/theme.ts
```diff
@@ -138,7 +138,10 @@ export const typography = {
             }
         }
     ],
-    lists: [{ id: "list", name: "List 1", tag: "ul", styles: { ...paragraphs, fontSize: 17 } }]
+    lists: [
+        { id: "list", name: "List 1", tag: "ul", styles: { ...paragraphs, fontSize: 17 } },
+        { id: "list1", name: "Numbered List 1", tag: "ol", styles: { ...paragraphs, fontSize: 17 } }
```
adrians5j: This looks odd. If this will ultimately be needed, we'll want to probably adjust the id and label to be in more sync.

Note that if we can avoid changing theme.ts (in general user project files), we should do it. B/c otherwise we'll need to introduce a migration script, which is certainly not ideal. Basically, migration scripts are reserved for minor releases, not patch ones.

adrians5j (changes requested):
Left some code-level comments.

A couple of extra notes.

### 1. `dev` branch

These fixes will not wait for the quarterly release. They should be relased in one of the following patch releases, like 5.35.1 or 5.35.2. So, targeting the `dev` branch is required, instead of `next`.

### 2. Additional Potential Bugs 

Not sure if this should be tacked within this PR or not. Not sure how much out-of-scope are these. If not, let's create extra tickets in Jira. @m0nikaza 

#### 1. Switching List Type - Issue In Typography Selector

https://user-images.githubusercontent.com/5121148/236433235-a2f8c1b2-ce23-4516-8e22-706acb373616.mov

In the video, I'm selecting unordered list, and I can see List 1 in typography selector. Once I switch to numbered list, shouldn't the typography selector automatically select Numbered List 1? Instead, it stays on List 1, and I have to manually select Numbered List 1.

#### 2. Clicking on the List Type Breaks Styles

In the video, I'm selecting ordered list. Then I blur / close the lexical toolbar. Once I reopen the toolbar, clicking again on the ordered list icon (in order to just switch back to regular text), removes the default paragraph styles. Ideally, maybe Paragraph 1 styles would get automatically selected, because we are back in the paragraph mode. Prolly low prioirty, but still noted it.

https://user-images.githubusercontent.com/5121148/236433654-1e9b40db-ba15-4e64-bdbe-83b88adb4041.mov

#### 3. Pasting outside content into quote breaks styles

In the video, I have a quote, and I'm pasting some text from notion in it. The quote gets destroted, and I'm left with the default text. Not sure if that's the intended behaviour or not.

https://user-images.githubusercontent.com/5121148/236434583-c457e9cc-3836-4e66-8591-336123c6e329.mov

### 3. PR Description

Can we add a link to the original issue? When possible, let's do this for PRs. It's gonna be more clear what's the actual issue the PR is trying to tackle. 

Having `WebinyListNode - set theme styleId by default on dom creation and update` is fine, but it doesn't tell me which actual issue that we've experiences this code change is solving.

---

In any case, for all of these extra notes, we can decide if we want to tackle them at all. If so, then decide whether we want to do it within this PR or separate.

packages/lexical-editor/src/nodes/ListNode/WebinyListNode.ts
mihajlovco: When you try to copy and paste some lists from other web pages or google docs will not be styled properly.

apps/theme/theme.ts
mihajlovco: It's a specific style for numbered lists but users can add that anyway. In this case if `ol` is not found I can apply the same styles (from the bullet list) to the numbered list. Or I will try to find at least one list style to apply as a default.

mihajlovco:
This PR is not valid anymore. For List and Quote have new PR's that will be merged into dev branch:
New list PR: [link](https://github.com/webiny/webiny-js/pull/3300) 
New quote PR: [link](https://github.com/webiny/webiny-js/pull/3303)

PR #3290, opened 2023-05-09
Fix(lexical): Error message in the console after opening the dropdown menu

mihajlovco:
With this PR we resolved the error message in the console after opening the dropdown menu from the lexical toolbar.

Submitted screenshot:
image

## Changes
- Moved the dropdown component out of the button component
- Now the dropdown position is calculated
- dropdown will be opened after the calculation of the dropdown position is finished (there is no visible delay)
- Updated lexical to latest version 0.10.0

## How Has This Been Tested?
Manually.

**Note!** Warning is shown from the Grammarly plugin.

https://user-images.githubusercontent.com/82515066/235667074-10a4e392-861e-4dc7-a370-0da6204e10fb.mov


packages/lexical-editor/src/ui/DropDown.tsx
```diff
@@ -162,42 +166,70 @@ export function DropDown({
...
+    }, [dropDownRef, buttonRef, showDropDown]);
 
-        const handle = (event: MouseEvent) => {
-            const target = event.target;
-            if (!button.contains(target as Node)) {
-                setShowDropDown(false);
-            }
-        };
-        document.addEventListener("click", handle);
+    // @ts-ignore - it's ok that all paths do not return value
```
adrians5j: We still have TS ignore here?

TS is complaining here and it's a for a good reason. Let's fix the error by adding the return statement where needed.

I have to repeat - using @ts-ignore is just for cases where it's really tricky to fix, or fixing would take unreasonable amount of time. This is a trivial TS issue, let's address it properly.

adrians5j (changes requested):
Took another look, let's resolve the TS issue and I believe we should be fine. Will give this another test in the browser and confirm.


Extra two notes. The upgrade of Lexical packages is fine, but it should've been done in a separate PR probably. 

Also, the title of this PR is `Toolbar error when dropdown is open on action click`. Not only I have to decypher this message in order to figure its actual meaning, but also, this is not the correct title. Title says one thing, the first sentence in the body says something else:

image

Let's get these things in order for future PRs. Slight improvements like this make PR reviews more enjoyable.

adrians5j (approved):
Alright, let's merge this once we get 5.35.1 through the door, and have this released with 5.35.2 release.

Thanks!

PR #3292, opened 2023-05-09
Remove and Update `merge` Dependency Where Appropriate

adrians5j:
## Changes
This PR remove and update `merge` dependency where appropriate.

We're removing this because NPM reports 1.2.x versions as a security vulnerability.

The package can be removed in three of our packages (not actually used). For user's projects, we've bumped it to v2.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


ma-schmidt-de:
Awesome. Thank you, Adrian.

mihajlovco (approved):
Looks good.

PR #3295, opened 2023-05-09
fix(lexical): fixed copy-paste content from other sources for paragraph node

mihajlovco:
With this PR we fixed the copy/paste behavior for the styles of the `paragraph` node. When the user will copy and paste content from other sources paragraph need to have the theme paragraph style.

## Changes
- Introduced new `BaseParagraphNode` that inherits Lexical's `ParagraphNode` node.
- Implements theme styling
- By default on paragraph creation, the first style for the HTML tag `p` will be applied to the paragraph node
> BaseParagraphNode will replace the Paragraph node and the default behavior of creation will be taken by the custom node

- Change the code in `formatToParagraph` node utility method, `$wrapNodes` _**deprecated**_ lexical method is changed with the new `$setBlocksType` method that will format any HTML tag like `blockquote` and `ul` for example back to `<p>` paragraph HTML tag. 

## How Has This Been Tested?
Manually

Test video:

https://github.com/webiny/webiny-js/assets/82515066/17e4d6e9-b319-4990-b10c-fb577e74377f


packages/app-page-builder/src/editor/plugins/elements/paragraph/elementText.ts
```diff
@@ -20,14 +20,12 @@ const lexicalText = JSON.stringify({
...
-                styleId: "paragraph1",
-                name: "Paragraph 1",
                 direction: "ltr",
-                typographyStyles: {},
+                styles: [],
                 format: "",
                 indent: 0,
                 tag: "p",
-                type: "typography-el-node",
+                type: "base-paragraph-node",
```
adrians5j: We are changing the `styles` and `type` here. Will this affect existing paragraphs added to a page before 5.35.1? Will this break anything?

packages/lexical-editor/src/nodes/webinyNodes.ts
```diff
@@ -32,5 +33,12 @@ export const WebinyNodes: ReadonlyArray<
     OverflowNode,
     MarkNode,
     FontColorNode,
-    TypographyElementNode
+    TypographyElementNode,
+    BaseParagraphNode,
+    {
```
adrians5j, line 45: Why do we need this replacement? Can we add a comment here?

packages/lexical-editor/src/utils/nodes/formatToParagraph.ts
```diff
@@ -1,18 +1,17 @@
...
     $isRangeSelection,
     DEPRECATED_$isGridSelection,
     LexicalEditor
 } from "lexical";
-import { $wrapNodes } from "@lexical/selection";
+import { $createBaseParagraphNode } from "~/nodes/BaseParagraphNode";
+import { $setBlocksType } from "@lexical/selection";
 
-export const formatToParagraph = (editor: LexicalEditor) => {
+export const formatToParagraph = (editor: LexicalEditor, typographyStyleId?: string) => {
```
adrians5j, line 10: Should this be named "formatToBaseParagraph"?

adrians5j: Having a comment that explains why we need this might be useful too.

packages/lexical-editor/src/plugins/TypographyPlugin/TypographyPlugin.tsx
```diff
@@ -16,8 +12,8 @@ export const TypographyPlugin: React.FC = () => {
             ADD_TYPOGRAPHY_ELEMENT_COMMAND,
             payload => {
                 const selection = $getSelection();
-                if ($isRangeSelection(selection)) {
-                    $wrapNodes(selection, () => $createTypographyNode(payload.value));
+                if ($isRangeSelection(selection) && payload.value.id && payload.value.tag === "p") {
```
adrians5j, line 15: The thing I don't understand here... this is a TypographyPlugin, and we're checking if `payload.value.tag === "p"`. Why is this only working with `p` tags and not like h1?

Some comments are in order here I'd say. 🙂

packages/lexical-editor/src/nodes/BaseParagraphNode.ts
```diff
@@ -0,0 +1,190 @@
...
+        node.setFormat(serializedNode.format);
+        node.setIndent(serializedNode.indent);
+        node.setDirection(serializedNode.direction);
+        node.setThemeStyles(serializedNode.styles);
+        return node;
+    }
+
+    override exportJSON(): SerializeBaseParagraphNode {
+        return {
+            ...super.exportJSON(),
```
adrians5j: Not a single comment in this file. 😕

Let's add some comments.

Most importantly, I don't understand why we need BaseParagraph that extends the one from Lexical. I read your Slack messages, but didn't still understand what's going on.

I'm not saying we need to comment on every single method and line, but for starters, let's have a summary at the top of the file, explaining what are the key methods and what they do.

Thanks!

adrians5j (changes requested):
First iteration done!

packages/app-page-builder/src/editor/plugins/elements/paragraph/elementText.ts
mihajlovco: This will not affect existing nodes. This is the case when you drop a new paragraph node, so we will continue with the new node.

packages/lexical-editor/src/nodes/webinyNodes.ts
mihajlovco: yes, I will comment, and respond here, for the custom implementation of inserting the theme styles. For rendering and editing updates.

packages/lexical-editor/src/plugins/TypographyPlugin/TypographyPlugin.tsx
mihajlovco: This typography plugin implements the event when the user will select one of the typography in the dropdown to apply to the currently selected node in this case paragraph (heading will be included as well).

**Additional explanation**:
The typography node itself is not flexible and modular as the separate heading and paragraph nodes in controlling the insert/updates. 
So we created `BaseHeadingNode` and `BaseParagraphNode` where all theme styles will be applied.

packages/lexical-editor/src/nodes/webinyNodes.ts
```diff
@@ -11,6 +11,8 @@ import { TypographyElementNode } from "~/nodes/TypographyElementNode";
 import { WebinyListNode } from "~/nodes/list-node/WebinyListNode";
 import { WebinyListItemNode } from "~/nodes/list-node/WebinyListItemNode";
 import { WebinyQuoteNode } from "~/nodes/WebinyQuoteNode";
+import { BaseParagraphNode } from "~/nodes/BaseParagraphNode";
+import { ParagraphNode } from "lexical";
 
 export const WebinyNodes: ReadonlyArray<
```
adrians5j, line 20: Let's also add a comment here, saying this is a list of all of the nodes that Webiny's Lexical implementation supports OOTB.

packages/lexical-editor/src/nodes/webinyNodes.ts
```diff
@@ -32,5 +33,17 @@ export const WebinyNodes: ReadonlyArray<
     OverflowNode,
     MarkNode,
     FontColorNode,
-    TypographyElementNode
+    TypographyElementNode,
+    BaseParagraphNode,
+    /*
+     * We inherit and replace the native ParagraphNode because in this way we can take advantage from the native lexical processing
```
adrians5j: Let's simplify this comment, it's a bit confusing.

Let's just say:

> In order to provide additional Webiny-related functionality, we override Lexical's `ParagraphNode` node. More info on overriding can be found here: https://lexical.dev/docs/concepts/node-replacement.

packages/lexical-editor/src/nodes/webinyNodes.ts
```diff
@@ -11,6 +11,8 @@ import { TypographyElementNode } from "~/nodes/TypographyElementNode";
 import { WebinyListNode } from "~/nodes/list-node/WebinyListNode";
 import { WebinyListItemNode } from "~/nodes/list-node/WebinyListItemNode";
 import { WebinyQuoteNode } from "~/nodes/WebinyQuoteNode";
+import { BaseParagraphNode } from "~/nodes/BaseParagraphNode";
+import { ParagraphNode } from "lexical";
```
adrians5j: We've been talking about this on our call.. But, you know what, let's do this like the following:

```ts
import { ParagraphNode } from "~/nodes/ParagraphNode";
import { ParagraphNode as LexicalParagraphNode } from "lexical";
```

Let's also apply the same approach in `~/nodes/ParagraphNode`.

Hope I did not miss anything.

packages/lexical-editor/src/nodes/types.ts
```diff
@@ -0,0 +1,18 @@
...
+    setTypography: (typographyStyleId: string) => void;
+    getTypographyStyleId: () => string | undefined;
+    clearTypographyStyle: () => void;
+    hasTypographyStyle: () => boolean;
+}
+
+export interface TextNodeThemeStyles {
+    getThemeStyles: () => ThemeStyleValue[];
+    setThemeStyles: (styles: ThemeStyleValue[]) => void;
+}
```
adrians5j:
```suggestion
export interface TextNodeThemeStyles {
    getThemeStyles: () => ThemeStyleValue[];
    setThemeStyles: (styles: ThemeStyleValue[]) => void;
}

export interface TypographyStylesNode {
    setTypography: (typographyStyleId: string) => void;
    getTypographyStyleId: () => string | undefined;
    clearTypographyStyle: () => void;
    hasTypographyStyle: () => boolean;
}
```

Let's do a minor reorder.

packages/lexical-editor/src/nodes/BaseParagraphNode.ts
```diff
@@ -0,0 +1,205 @@
...
+
+export type SerializeBaseParagraphNode = Spread<
+    {
+        styles: ThemeStyleValue[];
+        type: "base-paragraph-node";
+    },
+    SerializedParagraphNode
+>;
+
+export class BaseParagraphNode
```
adrians5j: Naming-related comment was left in a previous comment.

adrians5j (changes requested):
We're close to merging this. Just a couple of tweaks and we should be good.

adrians5j (approved):
Good to be merged for 5.35.2.

PR #3298, opened 2023-05-10
fix(lexical): copy paste headings from other source

mihajlovco:
With this PR we fixed the copy/paste behavior for the styles of the `heading` node. When the user will copy and paste titles from other sources heading needs to have the theme style specific to the HTML tag (h1-h6).

## Changes
- Introduced new `BaseHeadingNode` that inherits Lexical's `HeadingNode` node.
- Implements theme styling
- By default on heading creation, the first style for the HTML tag `h1` will be applied to the heading node except is specified in the constructor.

> BaseHeadingNode will replace the Heading node and the default behavior of creation will be taken by the custom node
- Change the code in the `formatToBaseHeading` node utility method.

## How Has This Been Tested?
Manually

Heading testing video:

https://github.com/webiny/webiny-js/assets/82515066/0daa97de-de5b-4fca-849a-24d3c82a307b


packages/app-page-builder/src/editor/plugins/elements/heading/elementText.ts
```diff
@@ -17,13 +17,11 @@ const lexicalText = JSON.stringify({
...
                 ],
-                styleId: "heading1",
-                name: "Heading 1",
                 direction: "ltr",
-                typographyStyles: {},
+                styles: [],
                 format: "",
                 indent: 0,
-                type: "typography-el-node",
+                type: "base-heading-node",
```
adrians5j: On the back of my comments re BaseParagraph naming... can we use `heading-node` here instead of using the word `base` ? I'd like to avoid `base` since that could imply it's a base for something, which it isn't. It's just a node.

packages/lexical-editor/src/nodes/webinyNodes.ts
```diff
@@ -40,5 +40,16 @@ export const WebinyNodes: ReadonlyArray<
         with: () => {
             return new BaseParagraphNode();
         }
+    },
+    /*
+     * We inherit and replace the native HeadingNode, so we can take advantage from the native lexical processing
```
adrians5j: Same comment I had in prev PR.

adrians5j (changes requested):
I started reviewing the PR, but then I realized it might be better do deal with the "base" thingie first.

Can we also get rid of the word "base" here, for the Heading node
?

adrians5j (approved):

PR #3300, opened 2023-05-10
fix(lexical): missing style on the list when user copy and paste from other sources

mihajlovco:
With this PR we fixed the problem of missing style on the list when user copy and paste from other sources.

## Changes
- Changes `WebinyListNode` to intercept and add theme styles to the list by default.
- Fallback to default style in case the user has removed the previously used style.
> For example if I have now style with an id list, and remove that one, the node on creation and update will try to find the first occurrence of the style list for the specified `ol` or `ul` tag. 

## How Has This Been Tested?
Manually.


packages/lexical-editor/src/nodes/list-node/BaseListNode.ts
adrians5j, line 1: Let's move this file up, into `nodes` folder, and have it import additional code from the `list-node` folder. But let's rename the `list-node` folder to `ListNode`, and, in the same way we did it in prev PRs, let's rename `BaseListNode` to `ListNode`.

adrians5j (changes requested):
For starters, let's make things a bit more consistent.

packages/lexical-editor/src/utils/nodes/listNode.ts
adrians5j, line 1: If this is only related to `ListNode`, I don't see the reason why this should be located outside of the `nodes` folder. Is this only present in `ListNode` ?

packages/lexical-editor/src/components/ToolbarActions/QuoteAction.tsx
```diff
@@ -17,6 +17,7 @@ export const QuoteAction = () => {
             formatToQuote(editor, hasQuoteStyles ? DEFAULT_QUOTE_ID : undefined);
             return;
         }
+        debugger;
```
adrians5j:
```suggestion

```

adrians5j (changes requested):
Did a quick review, still need to actually test it in browser.

packages/lexical-editor/src/components/ToolbarActions/QuoteAction.tsx
mihajlovco: removed thanks

packages/lexical-editor/src/nodes/ListNode.ts
```diff
@@ -40,27 +41,40 @@ export class WebinyListNode extends ElementNode {
...
+        this.__start = start || 1;
     }
 
     static override getType() {
         return "webiny-list";
     }
 
+    /*
+     *
+     * */
```
adrians5j:
```suggestion

```

packages/lexical-editor/src/nodes/ListNode.ts
```diff
@@ -40,27 +41,40 @@ export class WebinyListNode extends ElementNode {
...
     override createDOM(config: EditorConfig): HTMLElement {
         const tag = this.__tag;
         const dom = document.createElement(tag);
+        const wTheme = config.theme as WebinyTheme;
 
         if (this.__start !== 1) {
             dom.setAttribute("start", String(this.__start));
         }
 
+        // if styleId is not set or user removed from theme, set default style.
```
adrians5j:
```suggestion
        // If styleId is not set or user removed from theme, set default style.
```

packages/lexical-editor/src/nodes/ListNode.ts
```diff
@@ -137,25 +146,56 @@ export class WebinyListNode extends ElementNode {
...
+        if (!tag) {
+            return;
+        }
+
+        const themeEmotionMap = theme?.emotionMap;
+        if (!themeEmotionMap) {
+            return;
+        }
+
+        const style = findTypographyStyleByHtmlTag(tag, themeEmotionMap);
```
adrians5j, line 159: I noticed this doesn't work for ordered list. Reported it here https://webiny.slack.com/archives/C01BL618QLA/p1685357931418519

adrians5j (changes requested):

packages/lexical-editor/src/nodes/ListNode.ts
mihajlovco: Functionally works in the correct way, if there is no theme style the `typography action will be removed` from the toolbar. 
We currently don't have a theme style for the ordered list. That would require migration and we will postpone for another version.

packages/lexical-editor/src/nodes/ListNode.ts
mihajlovco: done

packages/lexical-editor/src/nodes/ListNode.ts
mihajlovco: done

adrians5j (approved):

PR #3303, opened 2023-05-10
fix(lexical): copy paste quote block from different sources

mihajlovco:
With this PR we fixed the copy/paste behavior for the styles of the `BaseQuoteNode` node. When the user will copy and paste quotes with `quoteblock` tag from other sources quote needs to have the theme style.

## Changes
- Introduced new `BaseQuoteNode` that inherits Lexical's `QuoteNode` node.
- Implements theme styling
- By default on quote creation, the first style for the HTML tag `quoteblock` will be applied to the quote node except as specified in the constructor.

> BaseQuoteNode will replace the Quote node and the default behavior of creation will be taken by the custom node
- Change the code in the `formatToBaseQuote` node utility method.

## How Has This Been Tested?
Manually

Quote testing video:

https://github.com/webiny/webiny-js/assets/82515066/3c9fefab-93ed-4c84-970c-540ca81aa651


packages/lexical-editor/src/components/ToolbarActions/QuoteAction.tsx
```diff
@@ -14,13 +14,11 @@ export const QuoteAction = () => {
             // Try to set default quote style, when the action button is clicked for first time
             const DEFAULT_QUOTE_ID = "quote";
             const hasQuoteStyles = themeEmotionMap && themeEmotionMap[DEFAULT_QUOTE_ID];
-
             formatToQuote(editor, hasQuoteStyles ? DEFAULT_QUOTE_ID : undefined);
-            setIsActive(true);
             return;
         }
+        debugger;
```
adrians5j:
```suggestion

```

packages/lexical-editor/src/plugins/WebinyQuoteNodePlugin/WebinyQuoteNodePlugin.ts
```diff
@@ -1,14 +1,14 @@
...
 import { useWebinyQuote } from "~/hooks/useWebinyQuote";
 
 export function WebinyQuotePlugin(): null {
     const [editor] = useLexicalComposerContext();
 
     useEffect(() => {
-        if (!editor.hasNodes([WebinyQuoteNode])) {
-            throw new Error("WebinyQuoteNodePlugin: WebinyQuoteNode is e not registered on editor");
+        if (!editor.hasNodes([QuoteNode])) {
+            throw new Error("WebinyQuoteNodePlugin: BaseQuoteNode is e not registered on editor");
```
adrians5j: Double check this message.

adrians5j (changes requested):
A couple of comments, but I'd also like to report something I noticed ⬇️ 

The video shows me selecting a block that was created previously, with an older version of Lexical. Clicking on it throws an `parseEditorState: type "base-heading-node" + not found` error.

https://github.com/webiny/webiny-js/assets/5121148/a073894a-825a-47cd-9524-5ffd06d63883

Not sure if we want to tackle this in this PR or not. Prolly not as it's related to header. But it is something we should double check.

packages/lexical-editor/src/components/ToolbarActions/QuoteAction.tsx
mihajlovco: > A couple of comments, but I'd also like to report something I noticed ⬇️
> 
> The video shows me selecting a block that was created previously, with an older version of Lexical. Clicking on it throws an `parseEditorState: type "base-heading-node" + not found` error.
> 
>  old-block.mov 
> Not sure if we want to tackle this in this PR or not. Prolly not as it's related to header. But it is something we should double check.

You are using the old version of the heading node before we did the refactoring, just create a new block with the new heading element and everything will be ok, in the future we will not change this node type name

Will not be problem for users because they don't have this heading node.

mihajlovco: debugger is removed, thanks

packages/lexical-editor/src/plugins/WebinyQuoteNodePlugin/WebinyQuoteNodePlugin.ts
mihajlovco: thanks for this catch, I also renamed the hook and the plugin removing the `Webiny` prefix from the name.

adrians5j (approved):

PR #3304, opened 2023-05-11
feat(lexical): toolbar text alignment actions

mihajlovco:
With this PR we are introducing the new toolbar text alignment actions.

## Changes
- The toolbar for the heading and paragraph component has a new text alignment dropdown menu.
- Alignment actions: Alignment left, Alignment right, Align center, and Justify center
- Indent actions: Indent text, outdent text
- Right side text align menu will no longer exist for the lexical components
## How Has This Been Tested?
Manually.

## Testing the alignment and indent actions

https://github.com/webiny/webiny-js/assets/82515066/502111ba-8961-4871-95f6-f4e20ced360c

## Testing the right sidebar menu - text settings options and alignment no longer exist for the lexical components
@SvenAlHamad if you have some pages with the legacy components please test if everything it's ok.

https://github.com/webiny/webiny-js/assets/82515066/2d145c61-4e4b-4616-92f8-2d5ccbf24a6b


adrians5j (changes requested):
Before reviewing the code, I tried to use the new feature. Here are some of the issues I experienced (if you think some of these should we addressed in a separate PR, we can do that too).

1. When I select a word (by double clicking on it), I see an error in the console.
image

2. When I select a word, I don't see the toolbar. Only when I select the whole block, then I get to see it. I guess this is connected to the number 1.
image

3. When I make something a link, I can't get the toolbar anymore. It's the link special toolbar that is always shown - even if I've defined the URL of the link.
image

mihajlovco:
@adrians5j for the link toolbar that is expected behavior. Together with @SvenAlHamad we decided when the link is selected standard floating toolbar (with actions like bold...) to not be visible because it causes two toolbars at once to be visible (one on top of the other) and that is not good UX. 
- About the other issues I will check.

mihajlovco:
@adrians5j for the link toolbar that is expected behavior. Together with @SvenAlHamad we decided when the link is selected standard floating toolbar (with actions like bold...) to not be visible because it causes two toolbars at once to be visible (one on top of the other) and that is not good UX.

- About the other issues I will check.
#### Update: solved  the issue, now the toolbar is visible on the word selection

https://github.com/webiny/webiny-js/assets/82515066/bf302655-40f6-47ee-9675-66c9724a77b8

packages/lexical-editor/src/components/ToolbarActions/TextAlignmentAction.tsx
```diff
@@ -0,0 +1,71 @@
...
+    INDENT_CONTENT_COMMAND,
+    OUTDENT_CONTENT_COMMAND
+} from "lexical";
+import { useRichTextEditor } from "~/hooks/useRichTextEditor";
+import { Compose, makeComposable } from "@webiny/react-composition";
+import { TextAlignmentActionContext } from "~/context/TextAlignmentActionContextProps";
+
+/*
+ * Base composable action component that is mounted on toolbar action as a placeholder for the custom toolbar action.
+ * Note: Toa add custom component access trough @see LexicalEditorConfig API
```
adrians5j: Check grammar.

packages/lexical-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -30,15 +27,49 @@ interface FloatingToolbarProps {
...
-    const [activeEditor, setActiveEditor] = useState(editor);
+    const { toolbarType, setToolbarType } = useRichTextEditor();
 
     useEffect(() => {
         if (toolbarType !== type) {
             setToolbarType(type);
         }
     }, [type]);
 
+    function mouseMoveListener(e: MouseEvent) {
```
adrians5j: Can we `useCallback` here? If possible, let's avoid recreating the same functions on each render.

packages/app-page-builder/src/editor/plugins/elementSettings/text/TextSettings.tsx
```diff
@@ -196,17 +196,19 @@ const TextSettings: React.FC<TextSettingsProps> = ({ defaultAccordionValue, opti
...
-            icon={
-                <Tooltip content={`Changes will apply for ${activeDisplayModeConfig.displayMode}`}>
-                    {activeDisplayModeConfig.icon}
-                </Tooltip>
-            }
-        >
-            <>
-                {!usingLexicalEditor && (
+        <>
+            {!usingLexicalEditor ? (
```
adrians5j: Let's rather do `if (usingLexicalEditor) { return null }`.

adrians5j (changes requested):
A couple of minor changes.

packages/lexical-editor/src/components/ToolbarActions/TextAlignmentAction.tsx
mihajlovco: I changed the comment a bit, let me know what you think.

```
Base text alignment dropdown composable component.
Note: To add a custom dropdown component use @see LexicalEditorConfig API.
```

packages/lexical-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -30,15 +27,50 @@ interface FloatingToolbarProps {
...
+                const y = e.clientY;
+                const elementUnderMouse = document.elementFromPoint(x, y);
+
+                if (!popupCharStylesEditorRef.current.contains(elementUnderMouse)) {
+                    // Mouse is not over the target element => not a normal click, but probably a drag
+                    popupCharStylesEditorRef.current.style.pointerEvents = "none";
+                }
+            }
+        }
+    }, []);
```
adrians5j: Do we want to add `popupCharStylesEditorRef` as a dep here? I'd check other callbacks as well.

packages/lexical-editor/src/components/Toolbar/Toolbar.tsx
```diff
@@ -30,15 +27,50 @@ interface FloatingToolbarProps {
...
+    const { toolbarType, setToolbarType } = useRichTextEditor();
 
     useEffect(() => {
         if (toolbarType !== type) {
             setToolbarType(type);
         }
     }, [type]);
 
+    const mouseMoveListener = useCallback((e: MouseEvent) => {
+        if (popupCharStylesEditorRef?.current && (e.buttons === 1 || e.buttons === 3)) {
```
adrians5j: Let's add some comments here, explaining what these do. The `e.buttons === 1 || e.buttons === 3` ... what is that?

adrians5j (changes requested):

packages/lexical-editor/src/components/Toolbar/Toolbar.tsx
mihajlovco: yes, I have added the dependency, seems that everything it's working fine. I checked the console as well.

packages/lexical-editor/src/components/Toolbar/Toolbar.tsx
mihajlovco: I found what these magic numbers are :D, added the comments as well. Because is the official standard I left them. 
[Mozzila docs link](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons)

Event `buttons` numbers:
1 = The left button
2 = The right button
4 = The middle button
8 = The fourth button (Browser Back)
16 = The fifth button (Browser Forward)

Why in the code do we test for 3? This is how it works:
If you click for example the first and second buttons on the mouse at the same time in that case you will get 3 from the `e.buttons` because 1(left) + 2(right) = 3 
I didn't know that this exist. :)

adrians5j (approved):

PR #3308, opened 2023-05-12
Enable Configuration of Shared ElasticSearch Domain via `webiny.application.ts`

adrians5j:
## Changes
Instead of setting up a shared ElasticSearch domain via environment variables, this PR enables users to set this up simply via their `webiny.application.ts` files (Core and API ones).

Example:

```ts
import { createCoreApp } from "@webiny/serverless-cms-aws";

// Here we're listing all environments that will use the shared ElasticSearch domain.
const ENVIRONMENTS_USING_SHARED_ES_DOMAIN = ["staging", "prod"];

    elasticSearch: ({ params }) => {
        const { env } = params.run;
        if (ENVIRONMENTS_USING_SHARED_ES_DOMAIN.includes(env)) {
            return {
                domainName: "my-elastic-search-shared-domain",
                indexPrefix: `${env}_`
            };
        }

        return undefined;
    }
});
```

## How Has This Been Tested?
Manually.

## Documentation
Changelog + addition in Modify Cloud Infra doc.

### DIFF
--- a/packages/pulumi-aws/src/apps/api/createApiPulumiApp.ts
+++ b/packages/pulumi-aws/src/apps/api/createApiPulumiApp.ts
@@ -23,9 +23,9 @@ export interface CreateApiPulumiAppParams {
     elasticSearch?: PulumiAppParam<
         | boolean
         | {
-        domainName: string;
-        indexPrefix: string;
-    }
+              domainName: string;
+              indexPrefix: string;
+          }
     >;
 
     /**
@@ -255,7 +255,6 @@ export const createApiPulumiApp = (projectAppParams: CreateApiPulumiAppParams = 
             }
         }
     }
-
 
     return withCommonLambdaEnvVariables(app);
 };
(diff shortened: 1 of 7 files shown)


packages/pulumi-aws/src/apps/api/createApiPulumiApp.ts
```diff
@@ -16,6 +16,18 @@ import { tagResources, withCommonLambdaEnvVariables, addDomainsUrlsOutputs } fro
...
 export interface CreateApiPulumiAppParams {
+    /**
+     * Enables ElasticSearch infrastructure.
+     * Note that it requires also changes in application code.
+     */
+    elasticSearch?: PulumiAppParam<
+        | boolean
+        | {
+              domainName: string;
+              indexPrefix: string;
```
Pavel910, line 27: both should be optional, so this is missing `?`

packages/pulumi-aws/src/apps/core/createCorePulumiApp.ts
```diff
@@ -20,7 +20,13 @@ export interface CreateCorePulumiAppParams {
      * Enables ElasticSearch infrastructure.
      * Note that it requires also changes in application code.
      */
-    elasticSearch?: PulumiAppParam<boolean>;
+    elasticSearch?: PulumiAppParam<
+        | boolean
+        | {
+              domainName: string;
+              indexPrefix: string;
+          }
```
Pavel910: Optional

Pavel910 (changes requested):

Pavel910 (approved):

PR #3309, opened 2023-05-15
fix: fix grid issues and after save redirect

neatbyte-vnobis:
## Changes
Fix issue "Grid width: auto is not getting applied"
Fix issue "State issue when moving a grid"
Fix issue "After save redirect"
Fix issue "The save indicator icon is still to close to the button label"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder-elements/src/utils/index.ts
+++ b/packages/app-page-builder-elements/src/utils/index.ts
@@ -49,6 +49,58 @@ export const isPerBreakpointStylesObject = ({
     return false;
 };
 
+// Detect if certain style is a transformed per-breakpoint CSS style object, or a transformed regular CSS properties.
+export const isBreakpointCssStyle = ({
+    breakpoints,
+    styleName
+}: {
+    breakpoints: ThemeBreakpoints;
+    styleName: string;
+}): boolean => {
+    for (const breakpointName in breakpoints) {
+        if (breakpoints[breakpointName] === styleName) {
+            return true;
+        }
+    }
+    return false;
+};
+
+// separate style parts (modifiers) are assigned to different screen sizes one by one
+// this means that "responsive" style properties are assigned
+// in different order compared to theme.breakpoints order
+// it causes not guaranteed CSS order (that causes UI issues)
+// so we need to recreate styles object in proper order
+// (here "order" refers to a for...in cycle order while iterating over the breakpoints)
+// (so we need to recreate styles object adding properties in valid order)
+// see "traversal order" section of for-in loop - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#description
+const recreateStyleObject = ({
+    breakpoints,
+    cssObject
+}: {
+    breakpoints: ThemeBreakpoints;
+    cssObject: CSSObject;
+}) => {
+    const recreatedStyles: Record<string, any> = {};
+
+    // copy plain styles as is
+    for (const styleName in cssObject) {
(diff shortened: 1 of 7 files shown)


SvenAlHamad:
I did a quick test on this one and it seems the issue is resolved now.

adrians5j:
Quick question - should this be merged in our quarterly release or in the next patch? 

Asking b/c if it's the latter, we'll need to target `dev` branch.

packages/app-page-builder/src/types.ts
```diff
@@ -517,7 +517,7 @@ export type PbEditorPageElementPlugin = Plugin & {
     onReceived?: (params: {
         state: EventActionHandlerCallableState;
         meta: EventActionHandlerMeta;
-        source: PbEditorElement | DragObjectWithTypeWithTarget;
+        source: PbElement | DragObjectWithTypeWithTarget;
```
adrians5j: Can you explain why we did this change here?

packages/app-page-builder-elements/src/utils/index.ts
```diff
@@ -49,6 +49,58 @@ export const isPerBreakpointStylesObject = ({
...
+
+// separate style parts (modifiers) are assigned to different screen sizes one by one
+// this means that "responsive" style properties are assigned
+// in different order compared to theme.breakpoints order
+// it causes not guaranteed CSS order (that causes UI issues)
+// so we need to recreate styles object in proper order
+// (here "order" refers to a for...in cycle order while iterating over the breakpoints)
+// (so we need to recreate styles object adding properties in valid order)
+// see "traversal order" section of for-in loop - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#description
+const recreateStyleObject = ({
```
adrians5j: I want to understand the actual issue better.

From what I understood, you're saying that the `Object.keys` sometimes returns the property names in a different order?

image

adrians5j (changes requested):
First view.

packages/app-page-builder-elements/src/utils/index.ts
neatbyte-vnobis: No, the issue is not with `Object.keys()`
The issue is with JS `for-in` loop inside @emotion/css lib.
The order of grabbed "property keys" is defined by CHRONOLOGICAL order of added "properties"
(it mean that first added property is grabbed on first iteration, and last added - on last)
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#description

(of cause I ignore properties that uses a NUMBER as a key, as such are not appearing in our code)

When we build element style we loop over element 'modificators', but not all of the have definition for all screensizes in theme. So screen sizes are added to CSS object in order different from theme object.

adrians5j: Sorry for taking so long to respond to this....

Can you answer me one thing - which exact issue this changes solves?

I was looking at the Notion doc and thought maybe it was [this one](https://www.notion.so/webiny/Grid-issues-and-a-few-minor-bugs-56bdf708bd574530b872a6ceec441692?pvs=4#519607cb5f0041e08a98c05d86f3977d), but not sure.

neatbyte-vnobis: Exactly this one.
First subtask inside it:
image

The problem with styles was not only with a grid but with other elements too.

We can jump into the call, so I can explain issue more clearly.

PR #3312, opened 2023-05-16
feat: File Manager - ACO integration

leopuleo:
## Changes
With this PR we add ACO to File Manager. 
Screenshot 2023-04-26 at 20 42 50

### Features
- New table list view available.
- Both Grid and Table list is generated using ACO search records.
- Users can organise files in folders.
- Tags rendered are coming from ACO search records.
- Search is integrated with ACO.
- File record is fetched only when the user opens the "File Details" drawer.
- The file mime-type filtering is handled via tags.

## How Has This Been Tested?
Jest + Cypress

## Documentation
TBD


Pavel910 (approved):
Let's squash merge this!

PR #3328, opened 2023-05-24
fix: folder creation in root

leopuleo:
## Changes
With this PR we fix a bug introduced by the 5.36.0 version: creating a folder in the root using the "Create folder" button in the top right corner. After completing the folder creation process, this was not saved with the right `parentId`.

## How Has This Been Tested?
Manually

### DIFF
--- a/packages/app-aco/src/components/FolderTree/List/utils.ts
+++ b/packages/app-aco/src/components/FolderTree/List/utils.ts
@@ -21,7 +21,7 @@ export const createTreeData = (
 
             return {
                 id,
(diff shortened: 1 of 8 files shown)


brunozoric:
@leopuleo 
This will fix the future creation of the folders. For the existing ones to be listed you must add the root check in this method:
https://github.com/webiny/webiny-js/blob/fix/folder-in-root-creation/packages/app-aco/src/hooks/useAcoList.ts#L49

brunozoric (approved):

PR #3329, opened 2023-05-24
fix: fix element thumbnail visual issues

neatbyte-vnobis:
## Changes
Fix issue "Block Thumbnail visual issues"

## How Has This Been Tested?
Manual

## Documentation
None


SvenAlHamad:
Hey @neatbyte-vnobis - I don't think we want to add the font definitions into the `global.scss` file...several reasons for that. 

Can we explore another potential approach. Instead of generating thumbnails, can we render the block content directly inside the preview container. Similarly how the page preview is rendered in the drawer, would that be possbile? 
The benefit of this approach is that if a user updates the theme file, say they change the primary color, currently the image previews don't get updated with that new information. Rendering the preview directly would reflect this change immediately.

adrians5j:
Did a quick POC for @neatbyte-vnobis [here](https://webiny-community.slack.com/archives/C03FG2P74N9/p1685361477393139?thread_ts=1685350159.443329&cid=C03FG2P74N9), let's see how far we get.

packages/app-page-builder/src/pageEditor/config/blockEditing/BlockPreview.tsx
```diff
@@ -46,6 +45,7 @@ const BlockPreview: React.FC<BlockPreviewProps> = props => {
                 {onDelete && (
                     <Styled.DeleteBlock>
                         <ConfirmationDialog
+                            style={{ zIndex: 99 }}
```
adrians5j: Why this change?

packages/app-page-builder/src/pageEditor/config/blockEditing/EditBlockDialog.tsx
```diff
@@ -60,7 +60,7 @@ const EditBlockDialog: React.FC<EditBlockDialogProps> = props => {
     }));
 
     return (
-        <Dialog open={open} onClose={onClose} className={narrowDialog}>
+        <Dialog open={open} onClose={onClose} className={narrowDialog} style={{ zIndex: 99 }}>
```
adrians5j: Why this change? Also, it could also be embedded into the `narrowDialog`?

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveDialog/domToImage.ts
adrians5j, line 1: Do we even need any of these changes? Since we're not snapshotting blocks anymore?

adrians5j (changes requested):
Starting the review with a couple of quick comments.

Additional questions below.

### 1. Image issue
Adding an image works:


But when saving, I get this error:

image

Wondering if simply pulling in all changes from `next` would resolve this issue? 🤔 

### 2. Still using image?
Also, shouldn't this modal also be rendering the block, instead of displaying its sshot? Did I miss smt here? 🤔 
image

packages/app-page-builder/src/pageEditor/config/blockEditing/BlockPreview.tsx
neatbyte-ivelychko: We need this change because `ConfirmationDialog` and `EditBlockDialog` were rendering below `All Blocks` component, and because of that the user couldn't see those modals and couldn't interact with them.
So to fix that issue I've simply added bigger `z-index` so it can render properly.

Знімок екрана 2023-07-11 о 10 36 17

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveDialog/domToImage.ts
neatbyte-ivelychko: I think we don't need them anymore. 
I haven't deleted this file, because I needed confirmation that it is safe to delete it

neatbyte-ivelychko:
> Additional questions below.
> ### 1. Image issue
About this one
I've double checked it locally and on the admin app, and seems like it works as expected (saving image and not throwning an error)

But I want to say that I was having the same issue as this one when I was working on Dynamic Images (Dynamic Pages).

neatbyte-ivelychko:
> ### 2. Still using image?

Regarding this question, I've double checked and I want to say that we do use Component Preview instead of Image Preview. Maybe you need to try to re-deploy admin, because for me it works as expected

Знімок екрана 2023-07-12 о 16 25 16

packages/app-page-builder/src/pageEditor/config/blockEditing/BlockPreview.tsx
adrians5j: Why not just do 100? Why 99? 🤔 

Can we add a comment into the code, just so one day when we get back, know why this is here?

packages/app-page-builder/src/pageEditor/config/blockEditing/EditBlockDialog.tsx
```diff
@@ -19,15 +18,16 @@ import { Form, FormOnSubmit } from "@webiny/form";
 import styled from "@emotion/styled";
 import { PbEditorBlockCategoryPlugin, PbEditorBlockPlugin } from "~/types";
 
-const narrowDialog = css({
-    ".mdc-dialog__surface": {
-        width: 600,
-        minWidth: 600
+const StyledDialog = styled(Dialog)`
+    z-index: 99;
```
adrians5j: Provided a comment [here](https://github.com/webiny/webiny-js/pull/3329#discussion_r1259558781).

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveDialog/domToImage.ts
adrians5j: Let's delete the file and see if we break anything. I'd say it's safe to do it.

packages/app-page-builder/src/blockEditor/config/eventActions/saveBlock/getPreviewImage.ts
adrians5j, line 1: Let's delete this file.

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveAction.tsx
```diff
@@ -82,46 +63,11 @@ const SaveAction: React.FC = ({ children }) => {
...
-        const previewImage = await fileUploaderPlugin.upload(blob, { apolloClient: client });
-
-        const createFile: FileInput = {
-            ...previewImage,
-            tags: [],
-            meta: {
-                ...meta,
-                private: true
-            }
+        formData.preview = {
```
adrians5j: Do we need this on the backend? 🤔 

I think not, maybe I'm wrong?

packages/app-page-builder/src/admin/components/EditorPluginsLoader.tsx
```diff
@@ -100,6 +113,13 @@ export const EditorPluginsLoader: React.FC<EditorPluginsLoaderProps> = ({ childr
     if (location.pathname.startsWith("/page-builder/pages") && loaded.render) {
         return children as unknown as React.ReactElement;
     }
+    /**
+     * This condition is for the list of page blocks.
+     * Page template can be selected at this point.
```
adrians5j:
```suggestion
     * Blocks can be selected at this point.
```

adrians5j (changes requested):

packages/app-page-builder/src/pageEditor/config/blockEditing/BlockPreview.tsx
```diff
@@ -46,6 +45,9 @@ const BlockPreview: React.FC<BlockPreviewProps> = props => {
                 {onDelete && (
                     <Styled.DeleteBlock>
                         <ConfirmationDialog
+                            // We need to have this z-index because without it Delete Block Dialog will be rendered below.
```
adrians5j: Please recheck this grammar. I think the dot after the `rendered below` is a mistake?

Sorry for the nitpick, but yeah, we're OSS, and companies tend to check our code sometimes before making their decision. I know this is just a drop in the sea, but yeah, gotta start somewhere. :)

packages/app-page-builder/src/pageEditor/config/blockEditing/EditBlockDialog.tsx
```diff
@@ -19,15 +18,18 @@ import { Form, FormOnSubmit } from "@webiny/form";
 import styled from "@emotion/styled";
 import { PbEditorBlockCategoryPlugin, PbEditorBlockPlugin } from "~/types";
 
-const narrowDialog = css({
-    ".mdc-dialog__surface": {
-        width: 600,
-        minWidth: 600
+const StyledDialog = styled(Dialog)`
+    // We need to have this z-index because without it Edit Block Dialog will be rendered below.
```
adrians5j: Same comment re grammar.

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveAction.tsx
adrians5j: This question still remains.

adrians5j (changes requested):
Close!

Left minor comment and there's a question that's still unanswered.

Tested it manually, looks good. Glad we've managed to deleted some of those dom-to-image files.

One final thing to check on the UI side:

image

This list is rendered a bit weirdly. Can we double check that before merging?

Thanks!

neatbyte-ivelychko:
> One final thing to check on the UI side:
> This list is rendered a bit weirdly. Can we double check that before merging?

Sure, will check it

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveAction.tsx
neatbyte-ivelychko: Maybe we don't. But I need to spend a little bit more time on this question

adrians5j (changes requested):
Looks good. We just need to revisit [this question](https://github.com/webiny/webiny-js/pull/3329#discussion_r1266367663).

neatbyte-ivelychko:
@adrians5j l've checked whether we need to save `preview` on the back-end, and it looks like we don't, because we are rendering element itself instead of image. I will remove `preview` from the api.
https://github.com/webiny/webiny-js/blob/f5370b57d50768573b9455ce8a85dfaf441ffdb6/packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveAction.tsx#L66

neatbyte-ivelychko:
@adrians5j I've removed `preview` from the api

packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveAction.tsx
neatbyte-ivelychko: @adrians5j I have removed `preview` form the API, as we no longer need it

packages/app-page-builder/src/types.ts
```diff
@@ -632,8 +632,7 @@ export type PbEditorBlockPlugin = Plugin & {
     blockCategory: string;
     tags: string[];
     create(): PbEditorElement;
-    image?: Partial<File>;
-    preview?(): ReactElement;
+    preview(): ReactElement;
```
adrians5j, line 635: Do we need this still here? 🤔

adrians5j (approved):

PR #3337, opened 2023-05-30
Fix/5.36.x aco file migration empty index

leopuleo:
## Changes
With this PR we fix 5.36.0/001 migration bug, preventing migration failure in the case of an empty file-manager index.

## How Has This Been Tested?
Jest

### DIFF
--- a/packages/migrations/__tests__/migrations/5.36.0/001/ddb-es/001.test.ts
+++ b/packages/migrations/__tests__/migrations/5.36.0/001/ddb-es/001.test.ts
@@ -135,6 +135,27 @@ describe("5.36.0-001", () => {
         }
     };
 
+    const insertEmptyFileIndexes = async () => {
+        const tenants = createTenantsData().map(tenant => tenant.data.id);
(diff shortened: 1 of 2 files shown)


Pavel910 (approved):

Pavel910 (changes requested):
The code itself is fine, but looking at the target branch, `dev` is a wrong target. This should be fixed in a branch created from 5.35.3 tag, and released from that branch. Then we can also merge this into `dev` to keep the fixed code.

Also the PR says it's for File Manager, but the code fixes Page Builder.

leopuleo:
> The code itself is fine, but looking at the target branch, `dev` is a wrong target. This should be fixed in a branch created from 5.35.3 tag, and released from that branch. Then we can also merge this into `dev` to keep the fixed code.
> 
> Also the PR says it's for File Manager, but the code fixes Page Builder.

Thanks Pavel, should be ok now.

Pavel910 (approved):

PR #3339, opened 2023-05-31
fix: fix duplicate images on import

neatbyte-vnobis:
## Changes
Resolve issue "When importing pages, blocks and templates see we don’t duplicate images"

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-file-manager/src/graphql/index.ts
+++ b/packages/api-file-manager/src/graphql/index.ts
@@ -143,6 +143,8 @@ export const createGraphQLSchemaPlugin = () => {
                 tag_not_startsWith: String
                 id_in: [ID!]
                 id: ID
+                importedUnderKey_in: [String!]
+                importedUnderKey: String
                 createdBy: ID
             }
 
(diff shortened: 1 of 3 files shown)


adrians5j:
Will leave this one to @Pavel910.

Pavel910:
@neatbyte-vnobis went through your changes. I understand the idea you came up with; but there are 2 things:
1) I don't like that new key in the GraphQL schema, in the root of FileInput. Do we even need to have it exposed in GraphQL? I think not, it's a very internal functionality, which we'll use programmatically.

2) I'm currently working on File Manager overhaul, using CMS as storage layer, and we're also doing some work on GQL schema, so I won't be able to merge this PR directly. But, the overall approach looks ok to me, and I'll apply the same approach in my branch, to the new API, and also put this new key into `file.meta.originalKey`. 

So I would say this PR is done, you don't need to do anything else here. I'm taking this over, and will apply the logic to my branch.

adrians5j:
Can we close this PR? 🤔 
https://github.com/webiny/webiny-js/pull/3101

neatbyte-vnobis:
> Can we close this PR? thinking #3101

Yes, we can

Pavel910:
Closing in favor of #3401.

PR #3349, opened 2023-06-06
Page Builder - Multiple Fixes

adrians5j:
## Changes
This PR introduces the following Page Builder fixes.

### 1. Enabled `target` Property On the `PbEditorPageElementPlugin` Plugin

The `target` property on the `PbEditorPageElementPlugin` plugin enables user to define into which existing elements the element that's being dragged can be dropped. Prior to this PR, this didn't work, because the property essentially wasn't taken into consideration in the new `ElementControlOverlay` and related React components.

With this PR, users can now use the `target` property on the `PbEditorPageElementPlugin` plugin to explicitly list elements into which the dragged element can be dropped.

### 2. Introduced the `canReceiveChildren` Property On the `PbEditorPageElementPlugin` Plugin

Prior to this PR, if a user wanted to mark an element as an element that can receive additional children elements e.g. via drag and drop, they would need to define the `onReceived` callback via the `PbEditorPageElementPlugin` plugin. While preparing this PR, I detected that the code that the user needs to write is a bit verbose, and could be abstracted away.

So, with this PR, if a user wants to mark an element as an element that can receive additional children elements, they simply define the `canReceiveChildren: true` on the `PbEditorPageElementPlugin` plugin, and that's it. Once that's in place, the only thing that remains is to define new elements that have their `target` property set correctly.

### 3. `target` Is Now Also Considered When Adding Elements By Clicking In the Left Sidebar

While doing the above, I also detected that when adding an element onto a page via the "Click to Add" feature (vs. just dragging and dropping), the `target` prop defined on the `PbEditorPageElementPlugin` plugin would not be taken into consideration. 

This has now been taken care of. No matter if the user drags/drops the element onto an existing element, or adds it via the mentioned "Click to Add" feature, the `target` prop will be considered. In other words, the element won't be added into the existing element, no matter which approach was used.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


Pavel910 (approved):
Lots of good stuff here. The whole thing is still a mess, but until we get to a better overall architecture, this will have to do.

ma-schmidt-de:
Amazing! Powerful improvements yet simple to configure for element developers. Love it.

PR #3361, opened 2023-06-15
fix: fix back button inside block editor

neatbyte-vnobis:
## Changes
Resolve issue "Back button on block editor doesn’t always work"

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-page-builder/src/blockEditor/config/editorBar/BackButton/BackButton.tsx
```diff
@@ -1,22 +1,29 @@
...
 export const BackButtonPlugin = createComponentPlugin(EditorBar.BackButton, () => {
     return function BackButton() {
+        const location = useLocation();
         const navigate = useNavigate();
+        const [block] = useBlock();
 
         const onClick = useCallback(() => {
-            navigate(-1);
-        }, [navigate]);
+            if (location.key === "default") {
```
adrians5j: Can we explain this if statement via comments? Why are we checking against key === `default` ? 🤔

packages/app-page-builder/src/blockEditor/config/editorBar/SaveBlockButton/SaveBlockButton.tsx
```diff
@@ -33,7 +34,11 @@ const DefaultSaveBlockButton: React.FC = () => {
                     debounce: false,
                     onFinish: () => {
                         setLoading(false);
-                        history.push(`/page-builder/page-blocks`);
+                        if (location.key === "default") {
```
adrians5j: Let's put the same comment here as well.

adrians5j (changes requested):

adrians5j (approved):

PR #3366, opened 2023-06-19
fix: fix import/export process get stuck

neatbyte-vnobis:
## Changes
Resolve issue "Page import/export is unreliable".

## How Has This Been Tested?
Manual

## Documentation
None


packages/api-page-builder-import-export/src/client.ts
```diff
@@ -13,35 +13,9 @@ export async function invokeHandlerClient<TParams>({
...
-         */
-        cookies: (request as any).cookies
-    };
-    // Invoke handler
     await context.handlerClient.invoke<TParams & any>({
         name: name,
-        payload: {
-            ...payload,
-            ...invocationArgs
-        },
```
Pavel910: @neatbyte-vnobis can you please explain how does code removal solve the issue?
I assume it has something to do with our migration to `fastify` some time ago, but maybe you can give us more details? Thanks!

neatbyte-vnobis: @Pavel910 
This payload data is not used anymore.
It was containing a LOT of nested data as a payload (headers and HUGE body)

The problem with failing exports was:
if too many entities are exported - this payload was increasing in size (>256KB) and was causing a failing AWS lambda (due to payload size).

PS: not sure regarding `fastify` part...

Pavel910: Ok, thanks. I mean, if this works, then great. The original code was an ugly hack to being with, so, I'm glad if we can remove that.

Pavel910: I'll need to test this on a multi-tenant system, because this `x-tenant` header is important for the import/export Lambda to work on a correct tenant. If we just remove it from headers, it might be an issue.

Pavel910: @neatbyte-vnobis I just tested this, and to make this work on a multi-tenant system, we need to have `headers` and `httpMethod` in the payload. Like this:

```
        payload: {
            ...payload,
            headers: {
                ["x-i18n-locale"]: request.headers["x-i18n-locale"],
                ["x-tenant"]: request.headers["x-tenant"] || tenantId
            },
            httpMethod: request.method
        }
```

Pavel910 (changes requested):
Please add the code snippet from the comment, and then we can merge this.

neatbyte-vnobis:
@Pavel910 I have just tested this fix after resolving conflicts and merging with the `next` branch, and it appears that the fix is still relevant.

Pavel910:
> @Pavel910 I have just tested this fix after resolving conflicts and merging with the `next` branch, and it appears that the fix is still relevant.

Awesome, thank you! I'll merge this then 👍

Pavel910 (approved):

PR #3370, opened 2023-06-21
refactor: change default root folder to lowercase

brunozoric:
## Changes
This PR changes the default folder from `ROOT` to `root`.
It also includes the migrations to change the existing folder entry parentId property from ROOT / root to null.

## How Has This Been Tested?
Jest and manually.


packages/app-headless-cms/src/admin/components/ContentEntryForm/useContentEntryForm.ts
```diff
@@ -20,8 +20,8 @@ import { CmsContentEntry, CmsEditorFieldRendererPlugin, CmsModelField } from "~/
 import { useContentEntry } from "~/admin/views/contentEntries/hooks/useContentEntry";
 import { plugins } from "@webiny/plugins";
 import { getFetchPolicy } from "~/utils/getFetchPolicy";
-import { ROOT_ID } from "@webiny/app-aco/components/FolderTree/List/constants";
 import { useRecords } from "@webiny/app-aco";
+import { ROOT_FOLDER } from "@webiny/app-aco/constants";
```
leopuleo: Since `root` is lowercase for both entries and folders, we can import `ROOT_FOLDER` from `packages/app-headless-cms/src/admin/constants.ts`

packages/app-headless-cms/src/admin/constants.ts
```diff
@@ -1,5 +1,5 @@
 export const CMS_ACO_APP = "CmsEntry";
-export const FOLDER_ID_DEFAULT = "ROOT";
+export const ROOT_FOLDER = "root";
```
leopuleo: You can handle it the same way you did for `@webiny/app-file-manager` and `@webiny/app-page-builder`

packages/migrations/src/migrations/5.37.0/002/ddb-es/AcoFolderMigration.ts
```diff
@@ -0,0 +1,333 @@
...
+                    index: foldersIndexName,
+                    body: {
+                        ...this.createElasticsearchFolderBody(tenantId, localeCode),
+                        size: 500,
+                        search_after: status
+                    },
+                    callback: async (folders, cursor) => {
+                        batch++;
+                        logger.info(
+                            `Processing batch #${batch} in group ${groupId} (${folders.length} pages).`
```
leopuleo: `pages` -> `folders`

packages/migrations/__tests__/migrations/5.37.0/002/ddb/002.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+        assertNotError(error);
+        const grouped = groupMigrations(data.migrations);
+
+        expect(grouped.executed.length).toBe(0);
+        expect(grouped.skipped.length).toBe(1);
+        expect(grouped.notApplicable.length).toBe(0);
+    });
+
+    it("should not run if no pages found", async () => {
```
leopuleo, line 52: `pages` -> `folders` (but it doesn't really matter 😃)

I would also cover the case of an empty folder index, as we did [here](https://github.com/webiny/webiny-js/blob/next/packages/migrations/__tests__/migrations/5.36.0/001/ddb-es/001.test.ts#L219).

leopuleo (commented):
Thanks Bruno, I left a couple of comments here there, but it looks good to me 👍

adrians5j (approved):

PR #3372, opened 2023-06-21
feat: add search bar into page builder and headless cms applications

leopuleo:
## Changes
With this PR we add search bar into the page builder and headless cms applications.

## How Has This Been Tested?
Jest + Manually


packages/app-aco/src/hooks/useAcoList.ts
```diff
@@ -117,66 +168,113 @@ export const useAcoList = <T = GenericSearchData>(params: UseAcoListParams) => {
...
         });
 
         const currentFolder = originalFolders?.find(folder => folder.id === folderId);
         setListTitle(currentFolder?.title);
-    }, [originalFolders, folderId]);
+    }, [originalFolders, folderId, isSearch]);
 
     /**
      * Any time we receive a `records` list or `folderId` update:
+     * - we return all `records` in case of search query;
```
Pavel910: You're not returning anything; you're setting state :)

packages/app-aco/src/hooks/useAcoList.ts
```diff
@@ -117,66 +168,113 @@ export const useAcoList = <T = GenericSearchData>(params: UseAcoListParams) => {
...
             search: undefined
         });
+
+        // Reset search flag any time we receive a new `folderId`
+        setIsSearch(false);
     }, [folderId]);
 
     /**
      * Any time we receive a `folders` list update:
+     * - we return an empty array in case of search;
```
Pavel910: You're not returning anything; you're setting state :)

packages/app-aco/src/hooks/useFolders.ts
Pavel910, line 1: Why does this file exist? I mean, the functions you create in `useMemo` are just calling those same functions on the context. So why not just return context from this hook, and that's it? That way you only need to manage the context functions.

This file right now doesn't have its own state, it doesn't change/wrap any context function,.... it's a pass-through. Just return `context` from this hook.

packages/app-headless-cms/src/admin/views/contentEntries/Table/Main.tsx
```diff
@@ -1,66 +1,39 @@
...
-    } = useAcoList({
-        folderId: initialFolderId || FOLDER_ID_DEFAULT
-    });
+        isSearch,
+        search,
+        setSearch,
+        listMoreRecords,
+        sorting,
+        setSorting
+    } = useContentEntriesList({ folderId });
```
Pavel910: I would recommend _not_ to destructure this, but assign it to a variable, like this:

```
const list = useContentEntriesList({ folderId });
```

I did this in File Manager, and it makes reading the component a lot easier, you know exactly if it's a local state/callback, or something coming from a hook. Moreover, you won't need to modify this destructuring when you add new properties, they'll simply be part of the `list` object.

packages/app-page-builder/src/admin/components/Table/Header/Search/Search.tsx
```diff
@@ -0,0 +1,17 @@
...
+    onChange: (value: string) => void;
+}
+
+export const Search: React.VFC<SearchProps> = ({ value, onChange }) => {
+    return (
+        <SearchWrapper>
+            <SearchUI value={value} onChange={onChange} />
+        </SearchWrapper>
+    );
+};
```
Pavel910: This component is already copied twice in this PR. Can we make this component part of the `app-aco` package, and just use it from there?

packages/app-page-builder/src/admin/views/Pages/Table/Main.tsx
```diff
@@ -28,34 +26,37 @@ interface Props {
...
+        isListLoadingMore,
+        isSearch,
+        search,
+        setSearch,
+        selected,
+        onSelectRow,
+        sorting,
+        setSorting,
+        listMoreRecords
+    } = usePagesList({ folderId });
```
Pavel910: Same recommendation here; don't destructure.

Pavel910 (changes requested):

Pavel910 (approved):

Pavel910 (approved):

Pavel910 (approved):

PR #3375, opened 2023-06-23
fix(app-form-builder): added timeout limit to ipify xhr request

SvenAlHamad:
## Changes
Added a timeout of 1.5s to the ipify API request. We use the ipify API to get the client's IP address when they submit the form. The problem is that ipify API sometimes takes even 20s to respond or it times out. Long waiting times could lead to abandoned form submissions. To get around that we've added the timeout limit, so if ipify doesn't respond within 1.5s, we submit the form and log `0.0.0.0` as. the client's ip.

## How Has This Been Tested?
Manually


brunozoric (approved):

Pavel910 (approved):
Nice trick! 👍

PR #3379, opened 2023-06-26
fix: change revisions order to latest on top

neatbyte-vnobis:
## Changes
Resolve issue "List revisions - latest on top".

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/api-page-builder/__tests__/graphql/pages.deletion.test.ts
+++ b/packages/api-page-builder/__tests__/graphql/pages.deletion.test.ts
@@ -176,10 +176,10 @@ describe("deleting pages", () => {
                             publishedOn: expect.stringMatching(/20/),
                             revisions: [
                                 {
-                                    id: p1v1.id,
+                                    id: p1v3.id,
                                     locked: false,
                                     status: "draft",
-                                    version: 1
+                                    version: 3
                                 },
                                 {
                                     id: p1v2.id,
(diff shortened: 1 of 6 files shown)


neatbyte-vnobis:
@Pavel910 @adrians5j @SvenAlHamad 
While working on this feature I've seen that Revision Selector behaves unexpectedly, when we select one of the revisions (v7 for example) and reopen Revision Selector menu, instead of highlighting selected revision (v7) it will highlight the first revision in the list.

So I want to know what should we do about it, should we leave it as it is right now (highlight first revision in the list) or should we change it to highlight revision that has been selected?

image

adrians5j:
Let's rather do this sorting on the backend.

Just because the backend already has this functionality built, it's just a matter of exposing it via GQL field params.

This is the CRUD method:
https://github.com/webiny/webiny-js/blob/next/packages/api-page-builder/src/graphql/crud/pages.crud.ts#L1252

Storage operations, both DDB and DDB+ES, support the `sort` param, e.g.:

packages/api-page-builder-so-ddb/src/operations/pages/index.ts:771

adrians5j (changes requested):
Let's rather do this sorting on the backend.

Just because the backend already has this functionality built, it's just a matter of exposing it via GQL field params.

This is the CRUD method:
https://github.com/webiny/webiny-js/blob/next/packages/api-page-builder/src/graphql/crud/pages.crud.ts#L1252

Storage operations, both DDB and DDB+ES, support the `sort` param, e.g.:

packages/api-page-builder-so-ddb/src/operations/pages/index.ts:771

adrians5j (approved):

adrians5j (changes requested):
Let's check failing tests before merge.

PR #3402, opened 2023-07-05
Page Builder - Fixed Drag and Drop-related Issues

adrians5j:
## Changes
This PR addresses two PB editor / drag-and-drop-related issues.

1. When creating a new page and immediately dropping the default block and any page element into its grid > cell page elements, the dropped page element would not be selectable. Only after a full page refresh, the interactivity would work as expected. With this PR, this is no longer the case.

2. Sometimes, dragging and dropping an element above or below an existing element would simply make the element disappear. This was also resolved with this PR.

## How Has This Been Tested?
Manual.

## Documentation
N/A

### DIFF
--- a/packages/app-page-builder/src/editor/helpers.ts
+++ b/packages/app-page-builder/src/editor/helpers.ts
@@ -13,6 +13,7 @@ import {
 } from "~/types";
 import {
     CreateElementActionEvent,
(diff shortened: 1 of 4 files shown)


Pavel910 (approved):
That was one sneaky bug! 🐞

PR #3405, opened 2023-07-06
Enable Responsive Styles For Lexical Typography

adrians5j:
## Changes
This PR enables having responsive styles within the `typography` section in the theme object, which wasn't possible before.

Note that this is basically quick and dirty solution to this problem. We're basically just passing down a separate "theme transformer" callback, which gets executed upon assigning styles to nodes within the editor. If the callback exists, the editor will run it, otherwise not.

If there will be time, we'll probably want to refactor this in the future.

It would be much easier if the editor just received theme classes that were calculated outside of the editor, like:

```ts
// ...

export const LexicalEditor: React.FC<LexicalEditorProps> = ({ tag, value, onChange, ...rest }) => {
    const { theme } = usePageElements();
    return (
        <ClassNames>
            {({ css }) => {
                const typography = getTypographyVariantsFromTheme({
                    theme,
                    css
                });

                return (
                    <>
                        {isHeadingTag(tag) && (
                            <HeadingEditor
                                theme={{
                                    typography,
                                    colors: theme.colors,
                                    fonts: theme.fonts
                                }}
                                typography={typography}
                                colors={theme.colors}
                                value={value}
                                onChange={onChange}
                                {...rest}
                            />
                        )}
                        {isParagraphTag(tag) ? (
                            <ParagraphEditor
                                typography={typography}
                                colors={theme.colors}
                                value={value}
                                onChange={onChange}
                                {...rest}
                            />
                        ) : null}
                    </>
                );
            }}
        </ClassNames>
    );
};

```

This way we would not have to maintain "emotionMap" and have to do any transformations within editor. This is not its domain, ideally the editor would just receive a list of typography options with classes already generated upon passing the props. 

I've also seen some renaming would be useful when it comes to typography-related code. We want to have the following convention (the same is used in docs):

```ts
interface Typography {
   [typographyCategory: string]: Array<TypographyVariant>
}
```

Basically, things like heading, paragraph, list in the theme object are a category of typography, and every entry in it is a typography variant (like h1, h2, h3, ...).

## How Has This Been Tested?
Manual.

## Documentation
Changelog.


packages/app-page-builder-elements/src/renderers/heading.tsx
```diff
@@ -16,7 +17,18 @@ export const createHeading = () => {
         const __html = element.data.text.data.text;
 
         if (isValidLexicalData(__html)) {
-            return <LexicalHtmlRenderer theme={theme} value={__html} />;
+            return (
+                <LexicalHtmlRenderer
+                    theme={theme}
+                    themeStylesTransformer={styles => {
```
mihajlovco, line 23: When I try to build the project shows `Parameter 'styles' implicitly has an 'any' type.`.

mihajlovco (approved):
Looks good for now, next, we will refactor the code and implement your suggestion @adrians5j .

PR #3419, opened 2023-07-13
feat: add cypress tests for block categories

neatbyte-vnobis:
## Changes
Implements issue "Add cypress tests for block categories".

## How Has This Been Tested?
Cypress and manually.

## Documentation
None


adrians5j:
Didn't dive deeper into this PR, but I gotta say at first look, looks good. 

Will try to do a review tomorrow.

neatbyte-vnobis:
@SvenAlHamad @adrians5j 
In this PR we have Cypress tests for Block Categories. These tests have some improvements implemented according to the doc I have shared with Sven (but without Page Objects or more complex testing approaches).

Also there are 2 bug fixes: refresh button for a Block Categories list and slug validation fix (now it is synced with API validation).


FYI: same issue with refresh button is present on `Page Categories` page too (no onclick handler at all - so button do nothing)
Please let me know if we need to fix it too.
image

packages/app-page-builder/src/admin/views/BlockCategories/validators.ts
```diff
@@ -1,7 +1,7 @@
 export const blockCategorySlugValidator = (value: string): boolean => {
-    if (!value.match(/^[a-z]+(\-[a-z]+)*$/)) {
+    if (!value.match(/^[a-z0-9]+(-[a-z0-9]+)*$/)) {
         throw new Error(
-            "Block Category slug must consist of only 'a-z' and '-' characters (for example: 'block-category-slug')"
+            "Block Category slug must consist of only 'a-z', '0-9' and '-' (for example: 'some-slug' or 'some-slug-2')"
```
adrians5j: Thanks for addressing this, super annoying. Funny enough, I just reported this a couple of seconds ago in our internal Slack. 🤣 

BTW:

```suggestion
            "Block category slug must consist of only 'a-z', '0-9' and '-' (for example: 'some-slug' or 'some-slug-2')"
```

cypress/integration/admin/pageBuilder/blockCategories/searchAndSortBlockCategories.spec.js
```diff
@@ -0,0 +1,143 @@
...
+                    icon: `block-category-icon-${id}`,
+                    description: `${i} Cool Block Category Description ${id}`
+                }
+            }).then(menu => {
+                entries.push(menu);
+            });
+        }
+    });
+
+    after(() => {
```
adrians5j, line 33: I have a question related this `after` callback.

What happens if a test fail? `after` won't get called, right?

I'm asking because we have retry mechanism set up, so if this tests start running again, and there are leftover items from the previous run (b/c it failed and `after` was not called), then they will prolly fail.

What we've seen before that works is having "delete all test records" in the `before` block, and then the 'create three block categories' that you have here.

That way we ensure test data is deleted before this is run. There is less chance (if any) for leftover test records to exist.

Wdyt?

adrians5j (changes requested):
Left some initial comments. Overall it looks good.

And yes for the the additional missing-handler fix you mentioned.

cypress/integration/admin/pageBuilder/blockCategories/searchAndSortBlockCategories.spec.js
neatbyte-vnobis: @adrians5j 
Yes, you are right.
I have checked the docs & best practices and looks like you are right.

That is a pain as we do not have ONE common command `resetDB` (and with dynamoDB+elastik it is not a trivial task).

I propose to move "delete" part to the `beforeEach` (before data creation).
It still can leave a leftover data if test fails (so it will affect other tests), but at least retry case will be covered.

In general - it look like we need to find a smart way to `resetDB` (but it is more like an Automation Tests design improvement, not an improvement for one test).

adrians5j (approved):

adrians5j:
Before merging, need to see if Cypress tests are 🟢 .

adrians5j:
This PR is good to be merged, but, since there might be another PR that will tackle this, I won't be merging it now.

Will resolve this soon enough.

adrians5j:
Closing this as, ultimately, these tests are will be included with another PR.

PR #3422, opened 2023-07-14
feat(api-headless-cms-ddb-es): entry values modifier

brunozoric:
## Changes
This PR adds the CMS Entry Values Modifier plugin via which users can modify what exactly goes into the Elasticsearch index.

This is basically meant to be used in apps using the CMS storage but having their own GraphQL schema.

## How Has This Been Tested?
Jest and manually.

## Usage
```typescript
import { createCmsEntryElasticsearchValuesModifier } from "@webiny/api-headless-cms-ddb-es";
const plugin = createCmsEntryElasticsearchValuesModifier({
    models: ["targetedModelId"],
    modifier: ({ setValues, values }) => {
        const title = `Audit log: ${values.title}`;
        // we set the values using the `setValues` function
        // in the prev object, there are values set by other plugins, if any.
        setValues(prev => {
            // Note that a field you are setting MUST exist in the model.
            return {
                // This will only make title be written into the Elasticsearch index.
                title
            };
        });
    }
});
// register the plugin as you would every other
```


packages/api-elasticsearch/src/compression.ts
```diff
@@ -13,7 +13,7 @@ const getCompressionPlugins = (plugins: PluginsContainer): CompressionPlugin[] =
 export const compress = async (
     pluginsContainer: PluginsContainer,
     data: Record<string, any>
-): Promise<Record<string, any>> => {
+): Promise<Record<string, any> | string> => {
```
Pavel910: `string` return type seems wrong.

packages/api-elasticsearch/src/compression.ts
```diff
@@ -34,7 +34,7 @@ export const compress = async (
 export const decompress = async (
     pluginsContainer: PluginsContainer,
     data: Record<string, any>
-): Promise<Record<string, any>> => {
+): Promise<Record<string, any> | string> => {
```
Pavel910: `string` return type seems wrong.

Pavel910 (changes requested):
Have a look at those return types. The rest seems fine.

Pavel910 (approved):

Pavel910 (approved):

PR #3426, opened 2023-07-17
fix: own permission in case of full access

leopuleo:
## Changes
In case of full access, `canAccessOnlyOwn` should return `false`. This PR fixes this behaviour. 

## How Has This Been Tested?
Manually


adrians5j (approved):

adrians5j:
Nice fix, tnx!

PR #3429, opened 2023-07-18
feat(migrations): add default folderId to all cms entries

brunozoric:
## Changes
This PR adds default location.folderId to all the CMS entries which do not have it.

I also moved all 5.37.0 >001 migrations to next higher number:
* 002 to 003
* 003 to 004
* 004 to 005

## How Has This Been Tested?
Jest tests run through 7,000 records but locally it was tested on 700,000.


packages/migrations/src/migrations/5.37.0/002/utils/getDecompressedData.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const convertToBuffer = (value: string | Buffer) => {
+    if (typeof value === "string") {
+        return Buffer.from(value, TO_STORAGE_ENCODING);
+    }
+    return value;
+};
+
+export const getDecompressedData = async <T>(data: any): Promise<T | null> => {
+    if (data?.compression !== GZIP) {
```
adrians5j, line 15: No need to do any decompressions if data.compression !== GZIP? I assume other compression algos are not used anymore, it's only GZIP these days?

adrians5j (approved):
Let's merge this.

PR #3436, opened 2023-07-21
feat: implemented multi step form

neatbyte-vnobis:
## Changes
Implements issue "Add the ability to build multi-step forms".

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-ivelychko:
Functionality is ready. 
Fixed API Tests.

neatbyte-ivelychko:
@SvenAlHamad @adrians5j 

I have several questions regarding Form behaviour:
  1 - How many fields we can store in one row, for now there are no limitations, we can add as many fields as we want in           the row, but the design looks awful. If we have 3 fields in one row it looks okay (Image 1), but if we have more than 3 fields then the fields are bleeding out (Image 2). What should we do, should we introduce limitation (like max 3 fields in the row)?
  
(IMAGE 1) Знімок екрана 2023-07-21 о 14 00 21

(IMAGE 2) Знімок екрана 2023-07-21 о 14 00 38

  2 - Let's say that we have created 4 form step, then we swithced to the preview tab, navigated to the last step and we decided to delete last step and the question is, what should we do with the preview, should we reset all fields values and reset preview step to the first step in the list, or should we just reset preview step to the first step? As an example below now we reset our preview to the first step, whenever we remove step.
  

https://github.com/webiny/webiny-js/assets/129154672/dd1af1bd-a943-451a-9b15-3d0c51ef5f1b

neatbyte-ivelychko:
@adrians5j what should we do about migration? Should I implement it? Or someone else gonna do that?

@SvenAlHamad & @adrians5j also I need your opinion on my comment above https://github.com/webiny/webiny-js/pull/3436#issuecomment-1645403649

adrians5j:
Hey @neatbyte-ivelychko, thanks for the questions.

Regarding how many fields in a row...I can see the current limit is 4 fields per row, plus on a wider screen, it looks fine. 

image

But, if it's not too much trouble, we can add this so we have a horizontal scrollbar on smaller screens:

image

For now I guess it's fine.

Regarding the form reset, having the form completely reset or resetting to step 1 is both fine at this point I guess.

adrians5j:
Leave the migration to us BTW.

neatbyte-ivelychko:
@adrians5j okay, got it. 
I think that I will add scrollbar.

neatbyte-vnobis:
> Leave the migration to us BTW.

@adrians5j I have an idea - for forms import we can add support for OLD exports (without new `step` property).
It will allow users to import some older exports (or exports from older Webiny deploys).

It will be almost same code as an migration - just not for db records, but for import archive records

neatbyte-ivelychko:
@adrians5j also what should we do with cypress tests, should we add more of them in this PR (I think that we need to add tests for steps and also fix already existing if they fail) or we will do that in another PR?

SvenAlHamad:
I did a quick test on this one and have some feedback:
- When we add a new step, a step should have some default title, for example it can be just "Step"
- when you drag and drop the step fields, it works in an odd way, try creating two steps, and then move the first step below the second one, you'll see that it has a strange way of working

neatbyte-ivelychko:
Also I did a quick research on how to handle drag-n-drop in Cypress
Here is the link -> https://stackoverflow.com/questions/55361499/how-to-implement-drag-and-drop-in-cypress-test

packages/app-form-builder/src/admin/components/FormEditor/Context/functions/getFieldPosition.ts
```diff
@@ -1,11 +1,11 @@
-import { FbFormModel, FbFormModelField, FieldIdType, FieldLayoutPositionType } from "~/types";
+import { FbFormModelField, FieldIdType, FieldLayoutPositionType } from "~/types";
 
 interface GetFieldPositionResult extends Omit<FieldLayoutPositionType, "index"> {
     index: number;
 }
 interface GetFieldPositionParams {
     field: FbFormModelField | FieldIdType;
-    data: FbFormModel;
+    data: any;
```
neatbyte-ivelychko: Will fix this `any` type

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/FormStep/FormStep.tsx
```diff
@@ -0,0 +1,297 @@
...
+            }
+
+            // If source pos is set, we are moving an existing field.
+            if (pos) {
+                if (pos.index === null) {
+                    console.log("Tried to move Form Field but its position index is null.");
+                    console.log(source);
+                    return;
+                }
+                const fieldId = (pos.formStep.id !== formStep.id ? pos.formStep : formStep).layout[
```
neatbyte-ivelychko: Will add a comment that would explain what are we doing here

adrians5j:
Sorry to ask, this PR can be reviewed?

Re migrations, basically, since the form data structure will become different once this PR is merged, we'll need to write a migration script that upgrades all existing form entries in the DB, so that existing data conforms with the new code. But we'll handle that. That's why I added `needs-migration` label to this PR, so that when we reach the release phase, I can easily see what needs migration to be written from our side.

You did mention import/exports though, and that's a good point. If it's not like a week of work, it would certainly be good to be able to import forms created with older version of Webiny into the latest version. So yeah, feel free to check that one one and ping for any further questions.

Finally, just a note on Cy tests. We've decided we'll be taking over here, so, feel free to finish any WIP Cy tests, but no need to create new ones from now on.

neatbyte-ivelychko:
@adrians5j Yeah, it is ready to be reviewed

packages/api-form-builder-so-ddb-es/src/definitions/form.ts
```diff
@@ -73,6 +73,9 @@ export const createFormEntity = (params: Params): Entity<any> => {
             layout: {
```
adrians5j: Do we still need this?

packages/api-form-builder-so-ddb/src/definitions/form.ts
```diff
@@ -79,6 +79,9 @@ export const createFormEntity = (params: Params): Entity<any> => {
             layout: {
```
adrians5j: Do we need this here also?

apps/theme/package.json
```diff
@@ -15,6 +15,7 @@
     "@webiny/form": "0.0.0",
     "@webiny/react-rich-text-renderer": "0.0.0",
     "@webiny/react-router": "0.0.0",
+    "@webiny/ui": "0.0.0",
```
adrians5j: Do we need this here? 🤔 

UPDATE: Actually no, check my next comment.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -39,8 +58,55 @@ const DefaultFormLayout: FormLayoutComponent = ({
     // Is the form successfully submitted?
     const [formSuccess, setFormSuccess] = useState(false);
 
+    // State to show fields corresponding to the step.
+    const [currentStep, setCurrentStep] = useState<number>(0);
+
     // All form fields - an array of rows where each row is an array that contain fields.
-    const fields = getFields();
+    // @ts-ignore
```
adrians5j: We'll want to check this ts-ignore here.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -1,12 +1,12 @@
-import React, { useState } from "react";
-import { Form } from "@webiny/form";
+import React, { useState, useEffect } from "react";
+import { Form, FormAPI } from "@webiny/form";
+import { ButtonDefault, ButtonPrimary } from "@webiny/ui/Button";
```
adrians5j: Basically, the form layout was imagined as an as lightweight as possible code. Meaning, we had `SubmitButton` in there, totally coded from scratch and not using `@webiny/ui` package.

We need to continue the same approach here.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -73,17 +146,39 @@ const DefaultFormLayout: FormLayoutComponent = ({
...
-                    >
-                        {formData.settings.submitButtonLabel || "Submit"}
-                    </SubmitButton>
+                    {/*
+                        If the form has more than one step then the form will be recognized as a Multi Step Form,
+                        so it means that we need to render form step handlers to switch between steps.
+                    */}
+                    {isMultiStepForm && (
+                        <ButtonsWrapper>
+                            <ButtonDefault onClick={handlePrevStep} disabled={currentStep === 0}>
```
adrians5j: As mentioned in prev comment, we should be using `SubmitButton` here that was already here.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -39,8 +58,55 @@ const DefaultFormLayout: FormLayoutComponent = ({
...
+    const fields = getFields(currentStep);
+
+    // Check if the form is a multi step.
+    const isMultiStepForm = formData.steps.length > 1;
+
+    const handleNextStep = () => {
+        setCurrentStep(prevStep => (prevStep += 1));
+    };
+
+    const handlePrevStep = () => {
```
adrians5j: We need to handle step back / step fwd controls and state internally, within the app-form-builder package's component. The same thing we're doing for example with `submit`, pulled from `FormLayoutComponent`. So, let's add those controls above.

The idea is to have these FB layouts as straightforward as possible, and be able to provide as much of OOTB functionality as possible. With this approach, if the user wanted to create another layout, they'd need to c/p again the steps state management code.

packages/app-page-builder-elements/src/renderers/form/types.ts
```diff
@@ -43,12 +43,18 @@ export interface FormDataRevision {
     createdBy: FormDataCreatedBy;
 }
 
+export interface FbFormStep {
```
adrians5j: For consistency, let's rename this to `FormDataStep`.

packages/app-form-builder/src/types.ts
```diff
@@ -394,7 +416,7 @@ export interface FbReCaptchaInput {
 }
 
 export interface FbFormSettingsInput {
-    layout: FbFormSettingsLayoutInput;
+    steps: FbFormStep[];
```
adrians5j, line 426: The naming here is good. 👍🏻

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/FormStep/FormStep.tsx
adrians5j: Good point.

packages/app-form-builder/src/admin/components/FormEditor/Context/functions/getFieldPosition.ts
adrians5j: Good point.

adrians5j (changes requested):
Did the first review, managed to go over like ~60% of files.

Overall looking good, but yeah, there's always stuff to polish / improve.

Please check my comments and let me know if there will be any questions.

Once we deal with this first batch of feedback, I'll continue with the rest.

neatbyte-ivelychko:
@adrians5j I've implemented first part of suggested changes.

packages/app-page-builder-elements/src/renderers/form/FormRender.tsx
```diff
@@ -200,6 +209,9 @@ const FormRender: React.FC<FormRenderProps> = props => {
         getDefaultValues,
         getFields,
         submit,
+        handleNextStep,
```
adrians5j: For consistency, let's name this as simply `goToNextStep` and `goToPreviousStep`.

But overall, nice job, this is the way to go 🚀

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -40,7 +63,35 @@ const DefaultFormLayout: FormLayoutComponent = ({
     const [formSuccess, setFormSuccess] = useState(false);
 
     // All form fields - an array of rows where each row is an array that contain fields.
-    const fields = getFields();
+    const fields = getFields(currentStep);
+
+    // Check if the form is a multi step.
+    const isMultiStepForm = formData.steps.length > 1;
```
adrians5j: Let's also have this passed down to us, with the next step / previous step callbacks. Thanks!

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -40,7 +63,35 @@ const DefaultFormLayout: FormLayoutComponent = ({
...
+
+    // Validate fields for current step with "form.validateInput" function,
+    // if current step is invalid then we should block posibility to move to the next step,
+    // but if step is valid then user can switch to the next step.
+    const validateCurrentStepFields = (form: FormAPI) => {
```
adrians5j: Let's also have this as an internal function users can immediately use.

But one thing though... When you call `submit` (the one passed by the `Form` component, L122), the rendered fields will be automatically validated. And then the `Form` component only calls your `onSubmit` callback if the validation is successful.

In other words, the validation of fields should probably be:

1) handled by the @webiny/app-form-builder's Form component internally.
2) should be internally handled by the `submit` callback / the @webiny/form package which triggers it automatically for us.

Hope I didn't miss anything here.

packages/app-form-builder/src/components/Form/FormRender.tsx
```diff
@@ -55,15 +56,30 @@ const FormRender: React.FC<FbFormRenderComponentProps> = props => {
         });
     }, [data.id]);
 
+    // We need this useEffect in case when user has deleted a step and he was on that step on the preview tab,
+    // so it won't trigger an error when we trying to view the step that we have deleted,
+    // we will simpy change currentStep to the first step.
+    useEffect(() => {
+        setCurrentStep(0);
+    }, [data.steps]);
```
adrians5j: To be more precise, we could also do `data.steps.length` here, right?

packages/app-form-builder/src/components/Form/FormRender.tsx
```diff
@@ -73,8 +89,12 @@ const FormRender: React.FC<FbFormRenderComponentProps> = props => {
         return fields.find(field => field.fieldId === id) || null;
     };
 
-    const getFields = (): FormRenderFbFormModelField[][] => {
-        const fieldLayout = cloneDeep(layout);
+    // We need to have "stepIndex" prop in order to get corresponding fields for the current step.
+    const getFields = (stepIndex: number): FormRenderFbFormModelField[][] => {
```
adrians5j: Let's have `stepIndex = 0` here, just so if a user knows he's working with single-step forms, he can just call `getFields()` and that's it.

packages/app-form-builder/src/components/Form/FormRender.tsx
```diff
@@ -73,8 +89,12 @@ const FormRender: React.FC<FbFormRenderComponentProps> = props => {
...
+        // We need this check in case we deleted last step and at the same time we were previewing it.
+        const stepFields =
+            steps[stepIndex] === undefined ? steps[steps.length - 1] : steps[stepIndex];
```
adrians5j, line 112: Wouldn't the current step just be reset when the user deletes the step? So, in the preview tab, we'd just start from the beginning.

If the above assumption is correct, do we need this code here?

packages/api-form-builder/__tests__/forms.test.ts
```diff
@@ -290,7 +368,10 @@ describe('Form Builder "Form" Test', () => {
         const { id } = create.data.formBuilder.createForm.data;
 
         // Add fields definitions
-        await updateRevision({ revision: id, data: { fields } });
+        await updateRevision({
+            revision: id,
+            data: { fields, steps: [{ title: "", layout: [] }] }
```
adrians5j: Seeing this test here, not only it might be better to have a title here just saying "Test step" at lest, but also maybe have validation that ensures title is not an empty string. Can we try to see if this is easily doable and if so, can we do it?

packages/api-form-builder/__tests__/formSubmissionSecurity.test.ts
```diff
@@ -170,6 +187,12 @@ describe("Forms Submission Security Test", () => {
                             ...formA,
                             savedOn: expect.stringMatching(/^20/),
                             fields: expect.any(Array),
+                            steps: [
+                                {
+                                    title: null,
```
adrians5j: Confused a bit why we sometimes have `""`, and sometimes `null` here. Am I missing something here?

But in any case, if we add the "required" validator I mentioned in my prev comments, this should no longer be an issue.

adrians5j (changes requested):
Another iteration done. Thanks for the work you did here so far.

With code comments, here are a couple of comments related to UI / manual testing I did.

#### 1.Styling of "Steps" In Editor

Should we adjust the styling here a bit @SvenAlHamad ? I'm thinking the same thing we have in our CMS? Check sshot. 

image

I also like how this looks:
image

In our case, this could be the step name.

#### 2. Step Name Modal Improvements

Please check sshot.

image

#### 3. Add Step Button

Not sure how to improve here.. IDK, but it looks a bit weird. Maybe @SvenAlHamad has an idea how to improve this?

image

neatbyte-ivelychko:
@SvenAlHamad could you please take a look at these comments https://github.com/webiny/webiny-js/pull/3436#pullrequestreview-1574207881

adrians5j:
Tip: a couple of UI-related questions for you @SvenAlHamad above.

neatbyte-vnobis:
@SvenAlHamad please take a look on the questions above

SvenAlHamad:
With regards to the UI questions above on #1 and #2 I agree with Adrian's suggestions. As for #3, I would just switch places between the icon and the text, so the text is on the right and icon on the left. This is similar to the design we have for dynamic zone in HCMS, so I'm ok with keeping this style instead of adding a button.

packages/app-form-builder/src/components/Form/FormRender.tsx
neatbyte-ivelychko: We need this block of code because of the way useState works. It changes `currentStep` value with a delay and because of that we get `undefined`.

neatbyte-ivelychko:
@adrians5j I've implemented second part of suggested changes

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -27,6 +47,10 @@ const DefaultFormLayout: FormLayoutComponent = ({
     getFields,
     getDefaultValues,
     submit,
+    validateCurrentStepFields,
+    goToPreviousStep,
+    isMultiStepForm,
+    currentStep,
```
adrians5j, line 54: Let's rename this to `currentStepIndex`. And can we have `currentStep` be an actual step object?

It will simplify the `stepTitle` code below. We can also have the "in case we deleted last step and at the same time we were previewing it" internally, and return the appropriate step, without user having to deal with this.

Sorry for nitpicking, but this is the code users have in their project. Trying to make it as simple as possible as have a great DX is important.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -73,17 +104,44 @@ const DefaultFormLayout: FormLayoutComponent = ({
...
+                                Previous Step
+                            </DefaultButton>
+                            {currentStep === formData.steps.length - 1 ? (
+                                <SubmitButton onClick={submit} loading={loading} fullWidth={false}>
+                                    {formData.settings.submitButtonLabel || "Submit"}
+                                </SubmitButton>
+                            ) : (
+                                <SubmitButton
+                                    onClick={() => {
+                                        validateCurrentStepFields(form);
```
adrians5j: Maybe we misunderstood each other, but the goal here was for the user to simply call `submit`. Note that, as mentioned, the `submit` callback (coming from L95) validates the fields that were rendered in the form. Meaning, no need to have a new `validateCurrentStepFields` callback that users need to think about. 

Once the form processes validation, it calls the `onSubmit` callback provided in L94, which is again coming as a prop. Meaning we know what to do internally - we basically just go to the next step if there are any steps after. 

Sorry, but if you'll need an extra call, LMK.

adrians5j (changes requested):
Some comments, didn't have a lot of time. I'll try to do more as soon as possible, today or tomorrow.

apps/theme/layouts/forms/DefaultFormLayout.tsx
neatbyte-ivelychko: @adrians5j 
Actually I do that extra validation because without it the method `goToNextStep` would always work and the form would be submited (even though validation failed). Maybe we can pass `goToNextStep` method inside of the `Form` component and call it there?

adrians5j: Can't we just call `submit` here, and, in the `onSubmit` callback (which will be called after the `<Form>` component has validated the rendered fields) do the next step? Since it's defined internally, the `onSubmit` callback can internally do all the logic of "go next step" or "finish form submission - show success". 

I hope I'm making sense. :)

neatbyte-ivelychko: Yeah, you're right!
I just did not realize that we could do that in the `onSubmit`.

packages/app-page-builder-elements/src/renderers/form/types.ts
```diff
@@ -77,8 +84,12 @@ export interface ErrorResponse {
 export type FormLayoutComponentProps<T = any> = {
     getFieldById: Function;
     getFieldByFieldId: Function;
-    getFields: () => FormRenderComponentDataField[][];
+    getFields: (stepIndex: number) => FormRenderComponentDataField[][];
```
adrians5j: I think we mentioned in one of the previous reviews that we could have this an optional argument. Just in case when the form is not multi-step, so the user doesn't have to provide step.

packages/app-page-builder-elements/src/renderers/form/types.ts
```diff
@@ -77,8 +84,12 @@ export interface ErrorResponse {
...
     getDefaultValues: () => { [key: string]: any };
+    goToPreviousStep: () => void;
```
adrians5j, line 89: I don't see `goToNextStep` here... hmmm 🤔 

It's basically because we don't need it? Because the switching to next step will be done by the `onSubmit` call?

packages/app-page-builder-elements/src/renderers/form/types.ts
```diff
@@ -77,8 +84,12 @@ export interface ErrorResponse {
...
+    validateCurrentStepFields: (form: FormAPI) => void;
```
adrians5j: There's a possibility we'll be able to get rid of this method. We'll see.

adrians5j (changes requested):
A bit more comments. UI looks better.

BTW, when I adda a step 2, I get this error in the console:
image

packages/app-page-builder-elements/src/renderers/form/types.ts
neatbyte-ivelychko: @adrians5j yes, because of that.

neatbyte-ivelychko: I have reverted this change, and now I am passing this prop down to `DefaultFormLayout` so we can use it inside `submitForm` function.

packages/app-page-builder-elements/src/renderers/form/types.ts
neatbyte-ivelychko: @adrians5j Hi, could you please check pr

neatbyte-ivelychko: @adrians5j sorry for bothering, but could you please check this pr again 🙏🏻

packages/app-page-builder-elements/src/renderers/form/FormRender.tsx
```diff
@@ -85,8 +102,8 @@ const FormRender: React.FC<FormRenderProps> = props => {
         return fields.find(field => field.fieldId === id) || null;
     };
 
-    const getFields = (): FormRenderComponentDataField[][] => {
-        const fieldLayout = structuredClone(layout) as FormDataFieldsLayout;
+    const getFields = (stepIndex = 0): FormRenderComponentDataField[][] => {
+        const fieldLayout = structuredClone(steps[stepIndex].layout) as FormDataFieldsLayout;
```
adrians5j, line 109: Any particular reason for the structured clone here?

apps/theme/layouts/forms/DefaultFormLayout/DefaultButton.tsx
adrians5j, line 1: Let's refactor these buttons. 

Let's create `theme/layouts/forms/DefaultFormLayout/buttons` folder, and in there have maybe base `DefaultButton`, and also have `SubmitButton` use `DefaultButton`. Let's avoid c/p-ing code.

Maybe it'd be easier to use the `SubmitButton` code as the new `DefaultButton` code, just because it also handles full width of the wrapper.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -40,17 +65,20 @@ const DefaultFormLayout: FormLayoutComponent = ({
...
     /**
      * Once the data is successfully submitted, we show a success message.
      */
     const submitForm = async (data: Record<string, any>): Promise<void> => {
-        setLoading(true);
-        const result = await submit(data);
-        setLoading(false);
-        if (result.error === null) {
-            setFormSuccess(true);
+        if (isMultiStepForm && currentStepIndex !== formData.steps.length - 1) {
```
adrians5j: Won't dwell on this further, but just one minor final detail here... can we have `currentStepIndex !== formData.steps.length - 1` replaced with `!isLastStep` ?

`isLastStep` can also come from props. We can also add `isFirstStep`.

After that, we're done with this file.

packages/app-form-builder/src/admin/components/FormEditor/Draggable.tsx
```diff
@@ -24,7 +24,7 @@ type EndDrag = (item: DragObjectWithType, monitor: DragSourceMonitor) => void;
 
 export interface DraggableProps extends BeginDragProps {
     children: DraggableChildrenFunction;
-    beginDrag?: BeginDrag | BeginDragProps;
+    beginDrag?: BeginDrag | BeginDragProps | any;
```
adrians5j: Can we address the `any` here?

packages/app-form-builder/src/admin/components/FormEditor/Droppable.tsx
```diff
@@ -24,13 +24,13 @@ export interface IsVisibleCallableParams {
     type: string;
     isDragging: boolean;
     ui: string;
-    pos?: Partial<FieldLayoutPositionType>;
+    pos?: Partial<FieldLayoutPositionType> | Record<any, any>;
```
adrians5j: Can we improve `Record<any, any>;` here?

packages/app-form-builder/src/admin/components/FormEditor/Droppable.tsx
```diff
@@ -24,13 +24,13 @@ export interface IsVisibleCallableParams {
...
 }
 export interface IsVisibleCallable {
     (params: IsVisibleCallableParams): boolean;
 }
 export interface OnDropCallable {
-    (item: DragObjectWithType): DroppableDropResult | undefined;
+    (item: DragObjectWithType & { ui: string }): DroppableDropResult | undefined;
```
adrians5j: Hm, why `{ ui: string }`? Can you provide some context?

If this is something that has to be here, can we add a comment as well here?

adrians5j (changes requested):
Left some comments, please check. 

Also some extra comments ⬇️ 

#### 1. Success Screen Issue in Preview Tab
When I submit a form in preview, I see success screen which is good. 

image

But I can't test the form again without refreshing the screen. So, maybe at least what we could do here is if revisiting the preview screen, we reset the form to step 1 if previously a success msg was shown. We could also maybe just show the success screen for 2-3 seconds, and reset back to step 1. Not sure what would be better.

#### 2. Visual Improvements
Second, I can see this form got visual improvements. I like where we're going. 

Let's see if @SvenAlHamad has anything to mention here. One thing that crossed my mind was the shadows that are created around each step (I know, it's how accordion does it). 

image

Removing the shadows made everything look a bit more like HCMS, but yeah, don't know if that's what we want.

image

I know it's not super important, but I think it might improve consistency. Also, might positively affect the potential change described in item 3 below.

#### 2.1 Delete Icon
Also, the delete and edit icons are kinda weirdly positioned IMO. 

image

Can we maybe position the pen near the step title? Or, can we have the delete step button in a dropdown, the same way it's done on the HCMS side? That way only pen and the three dots icons would be present for each step. I kinda like this solution better, but would like @SvenAlHamad to maybe share his thoughts.

#### 3. Split Steps
image

Can we try splitting the steps like we have it with fields in HCMS? @SvenAlHamad wdyt?


All in all, maybe @SvenAlHamad will have some feedback. Maybe we can even let him do final polishes once this is merged.

Either way works.

packages/app-page-builder-elements/src/renderers/form/FormRender.tsx
neatbyte-ivelychko: No, I just forgot to remove extra cloned function.

neatbyte-ivelychko: hm, I don't know because it was there before my changes

packages/app-form-builder/src/admin/components/FormEditor/Droppable.tsx
neatbyte-ivelychko: We need that `{ ui: string }` because `DragObjectWithType` is the type from the library and it does not have fields that we set through `beginDrag`. Like in our case we need `ui` to know what type of field we are dragging, if it is a `step` then we will block posiblity to drop steps into steps.
https://github.com/webiny/webiny-js/pull/3436/files#diff-33b595e920d5882dbdc98d6382274a874914bbfc92ff1b4c27df2ff8a711e48cR206

P.S. I will probably move `DragObjectWithType & { ui: string }` into separate type.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -73,17 +108,54 @@ const DefaultFormLayout: FormLayoutComponent = ({
...
-                        {formData.settings.submitButtonLabel || "Submit"}
-                    </SubmitButton>
+                    {/*
+                        If the form has more than one step then the form will be recognized as a Multi Step Form,
+                        so it means that we need to render form step handlers to switch between steps.
+                    */}
+                    {isMultiStepForm && (
+                        <ButtonsWrapper>
+                            <Button
+                                type="default"
```
adrians5j: Let's have the `type` have `default` as its default value. So, in this case, no need to pass `type="default"`.

apps/theme/layouts/forms/DefaultFormLayout.tsx
```diff
@@ -40,17 +66,25 @@ const DefaultFormLayout: FormLayoutComponent = ({
...
+            setLoading(false);
+            if (result.error === null) {
+                setFormSuccess(true);
+                if (resetFormAfterSubmit) {
+                    // This function will reset Success Message screen in 3 seconds after showing it,
+                    // it will also reset view to the first step.
+                    resetFormAfterSubmit(() => setFormSuccess(false));
+                }
+            }
         }
```
adrians5j:
```TS
if (isLastStep) {
    setLoading(true);
    const result = await submit(data);
    setLoading(false);
    if (result.error === null) {
        setFormSuccess(true);
        if (resetFormAfterSubmit) {
            // This function will reset Success Message screen in 3 seconds after showing it,
            // it will also reset view to the first step.
            resetFormAfterSubmit(() => setFormSuccess(false));
        }
    }
} else {
    goToNextStep();
}
```

Just switched the order of true/false code.

packages/app-form-builder/src/admin/components/FormEditor/Droppable.tsx
adrians5j: This one is still pending. Or do we want to wait here until we implement the `onDrop` refactor?

apps/theme/layouts/forms/DefaultFormLayout.tsx
adrians5j: Ignore this comment, I refactored this by removing `resetFormAfterSubmit` and implementing the `layoutRenderKey` value in `packages/app-form-builder/src/components/Form/FormRender.tsx:48`.

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/FormStep/EditFormStepDialog.tsx
```diff
@@ -0,0 +1,106 @@
...
+                        isOpened: false,
+                        id: null
+                    })
+                }
+            >
+                <DialogHeader>
+                    <span>Change Step Title</span>
+                </DialogHeader>
+                <DialogBody>
+                    <Input
```
adrians5j: We usually do not use the `Input` component with a `Form` component above it. Check `packages/app-aco/src/components/Dialogs/DialogCreate.tsx:94` for an example (I'm sure you've seen this before too).

adrians5j (changes requested):
A couple of quick things to check on the code level and I think we should be good.

I'll just ping @SvenAlHamad to give us his answers re UI.

packages/app-form-builder/src/admin/components/FormEditor/Droppable.tsx
adrians5j: Thanks, maybe a short comment about this. Would be nice to have when somebody needs to revisit this in the future.

neatbyte-ivelychko:
Probably I just forgor to remove it.
Also I think that we should do onDrop refactor in the separate PR. Because it would be easier to review those changes.
https://github.com/webiny/webiny-js/pull/3436#discussion_r1328705263

neatbyte-ivelychko:
@adrians5j I have added a form for Edit Step Modal, also added comments for types.

SvenAlHamad:
In general I agree with what Adrian said, and have two more things to add from my end:

1. When a user is creating a new form and we automatically insert the first step, name the step as "Step 1", instead of just "Step".

2. The background of the accordion should be white, like what we have on Dynamic Zones in Headless CMS.
Example - current design in the form builder:
CleanShot 2023-09-19 at 10 56 06@2x

Example - current design in the HCMS Dynamic Zone  field
CleanShot 2023-09-19 at 10 56 57@2x


As a general rule, I think we should try to make the Form Builder looks as close to the Headless CMS, particularly to the Dynamic Zone field as we invested time to polish that one out and it looks good. So when in a dilemma, just look at that one first.

neatbyte-ivelychko:
@SvenAlHamad @adrians5j
This is how it looks now, could you please confirm this is what you want.
Btw. this is only screenshot and the code has not been pushed yet.
Знімок екрана 2023-09-19 о 14 09 56

Знімок екрана 2023-09-19 о 14 13 29

adrians5j:
Thanks for the sshots @neatbyte-ivelychko !

Gonna share some of my thoughts, but let's also hear what @SvenAlHamad will have to say before doing anything. Just don't want us doing one thing, and then changing stuff again.

1] I like the separation of steps. 👍🏻 
image

2] I can see we still have these icons like before:

image

I was thinking maybe doing the same like the following (red arrows in the sshot, this was mentioned above in "2.1 Delete Icon" section):

image

But I guess it's fine for the time being. Maybe @SvenAlHamad can confirm we're good with how things look here.

3] White Accordion Bg

The only reason I was against having white bg of the accordion is now we have 2 white colors here:

image

This is what I had in mind and how it looks with the accordion bg not being changed (old sshot so there is no separation between steps yet):

image

So in my mind, I'd do it like that, but yeah, let's see what @SvenAlHamad thinks.

SvenAlHamad:
I only have a comment regarding point 3, I still think we should keep it as white. Here's a screenshot from HCMS dynamic zone: 
CleanShot 2023-09-21 at 09 52 22@2x

I think we should have it look the same in the Form Builder.

adrians5j:
Thanks @SvenAlHamad 🙏🏻 

@neatbyte-ivelychko , it looks like the screenshot you submitted is approved and we can continue with the implementation. Thanks!

neatbyte-ivelychko:
@adrians5j do I need to do (2.1 Delete Icon" section) or we can do that later?

adrians5j:
No need. The screenshot good to go.

neatbyte-ivelychko:
@adrians5j what should we do next? Should we merge this PR and then start working on `onDrop` function refactoring for d-n-d?

neatbyte-ivelychko:
@adrians5j I have pushed the latest requested changes

adrians5j (changes requested):
I was about to merge this, but while manually testing, I noticed we can no longer move form fields side by side, for example:

image

With the new code, I just get above/below drop zones, not the left/right:
image

We should check this out before merging.

neatbyte-ivelychko:
@adrians5j Maybe you were moving row instead of field? Because we cannot drop rows in rows.
I just checked and I can move fields in fields.

Here is the demo:
https://github.com/webiny/webiny-js/assets/129154672/3fe35890-713f-49d1-9287-1595313dad32

neatbyte-ivelychko:
@adrians5j so after merging latest next, api test for form builder are failing because of prerendering, should we fix them right now? Because before merging they were working

adrians5j:
It's an issue that was introduced by the FB prerendering PR yes. We both missed failing API tests before merging.

I'll see now if I can take care of it quickly.

adrians5j (approved):

PR #3438, opened 2023-07-24
Page Builder - Fixed Flushing of Pages

adrians5j:
## Changes
The `flush` method of the prerendering service's handler was reading storage name from the prerendering service settings (`prerendering.storage.name`). But, since these settings don't exist for a long time, the flush logic below the if condition would never be executed.

Not sure why this was still here, probably it was left by an accident? @Pavel910 

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


Pavel910 (approved):
@adrians5j nice find. Yes, this was left by mistake, and since we used `lodash.get` here, TS was unable to catch the error on a string path :(

PR #3449, opened 2023-07-27
fix(api-headless-cms): permissions checks

brunozoric:
## Changes

The main part of this PR is the export of access checks for models and groups. Adrian removed those when he introduced new security stuff.

PR also includes:
* attach permissions to the CmsContext so they are available to the public
* remove unnecessary locale parameter from model and group permission checks as it already exists on both group and model
* group permission check params now require only a small part of the group object to be sent into it
* model permission check params now require only a small part of the model object to be sent into it

## How Has This Been Tested?
Jest and manually.


adrians5j (dismissed):
Thanks!

leopuleo (approved):

PR #3453, opened 2023-08-04
fix(migrations): elasticsearch 5.37.0 002 validation

brunozoric:
## Changes
This PR adds a complex data structure to be upgraded within the 002 migration.
It also checks for the Elasticsearch data integrity.

## Additional
All content entry CRUD methods now return regular entry instead of the storage one.
DynamicZone storage converter can be ran multiple times over same entry object and the data will be correct.

## How Has This Been Tested?
Jest.

## Note
In the code there are 100 articles being generated over 2 tenant and 2 locales.
Locally i tested on 1000 articles over 7 tenants and 10 locales - 70k entries, which produce:
* 350k regular DynamoDB records (3 revisions + L + P)
* 140k Elasticsearch DynamoDB Records (L + P)
* 140k Elasticsearch records in 70 different indexes (L + P)


packages/api-headless-cms/src/types.ts
```diff
@@ -2136,6 +2136,7 @@ export interface OnEntryPublishErrorTopicParams {
 export interface OnEntryBeforeRepublishTopicParams {
     entry: CmsEntry;
     model: CmsModel;
+    storageEntry: CmsEntry;
```
Pavel910: I would prefer not to expose storage layer objects to 3rd party developers. If this is not necessary, let's not add this. Also, if you know other places that expose this, let's remove it from those places.

Pavel910 (changes requested):

Pavel910 (approved):

PR #3462, opened 2023-08-09
fix(data-migration): logger with trace level

brunozoric:
## Changes
PR adds a `@webiny/logger` package via which a user can configure the pino logger before it is created. Or just get the logger and use our default settings.

## How Has This Been Tested?
Jest and manually.

## Documented
Not yet. Going to keep this private for the moment.


Pavel910 (approved):
Thanks, we'll be using this package in many places!

PR #3463, opened 2023-08-09
fix: page re-render on form publish [OLD]

neatbyte-vnobis:
## Changes
Implements issue "When a form is published, re-render all the pages containing it".

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-ivelychko:
Also I've implemented few more changes:
   1 - Page rerender in case that page is using the form that has been deleted. Before this change the page was trying to render form that was no longer existing, so it was crashing page. Now it shows `Form not found`;
   2 - Page rerender in case that page is using form revision that has been deleted. Same issue as above, it was crashing page. Showing `Form not found` message now;
   3 - Also updated `Delete` feature for `Form Preview` tab, now it acctually deletes entire form instead of selected revision.

neatbyte-ivelychko:
@adrians5j this pr is ready to be reviewed.

neatbyte-ivelychko:
@adrians5j could please check this pr

apps/api/graphql/src/index.ts
```diff
@@ -77,6 +77,11 @@ export const handler = createHandler({
         prerenderingServicePlugins({
             eventBus: String(process.env.EVENT_BUS)
         }),
+        createFormBuilder({
```
adrians5j, line 80: Why this change? 🤔

packages/api-page-builder/src/prerendering/hooks/afterFormDeletetion.ts
```diff
@@ -0,0 +1,19 @@
+import { ContextPlugin } from "@webiny/api";
+import { PbContext } from "~/graphql/types";
+import { FormBuilderContext } from "@webiny/api-form-builder/types";
+
+export default () => {
+    return new ContextPlugin<PbContext & FormBuilderContext>(
+        async ({ pageBuilder, formBuilder }) => {
+            /**
+             * If page contains a form that has been deleted,
```
adrians5j, line 9: Looking at this code... can we have things done the other way? 

Have these event handlers in `api-form-builder`? 

This is because we wan't to keep form-related logic in its `api-form-builder` package, and not move that logic into PB package.

That way, we'll be adding PB as the dependency of FB, not FB as the dep of PB.

packages/app-form-builder/src/page-builder/admin/plugins/components/PeFormElement.tsx
```diff
@@ -34,7 +34,8 @@ const PeForm: FormRenderer = props => {
                     } catch {
                         return apolloClient
                             .query({ query: gql(GET_PUBLISHED_FORM), variables })
-                            .then(({ data }) => data.formBuilder.getPublishedForm.data);
+                            .then(({ data }) => data.formBuilder.getPublishedForm.data)
+                            .catch(() => null);
```
adrians5j, line 38: Why this change? Can we have a comment here?

packages/app-form-builder/src/admin/plugins/formDetails/previewContent/HeaderComponents/DeleteForm.tsx
```diff
@@ -0,0 +1,72 @@
...
+import * as queries from "../../../../graphql";
+import { removeFormFromListCache } from "~/admin/views/cache";
+import { usePermission } from "~/hooks/usePermission";
+import { FbRevisionModel } from "~/types";
+
+interface DeleteRevisionProps {
+    form: FbRevisionModel;
+    revision: FbRevisionModel;
+}
+const DeleteRevision: React.FC<DeleteRevisionProps> = ({ form, revision }) => {
```
adrians5j, line 18: I saw your comment:

>  Also updated Delete feature for Form Preview tab, now it acctually deletes entire form instead of selected revision.

What do you mean by this?

I mean, users should be able to delete just a specific revision of the form. Not sure I understand what's going on here. 🤔

packages/api-page-builder/src/prerendering/hooks/afterFormDeletetion.ts
adrians5j, line 1: Let's by consistent with naming. Instead of "Deletion", let's use "Delete", like we did with `afterFormRevisionDelete`.

packages/api-page-builder/src/prerendering/hooks/afterFormUpdate.ts
```diff
@@ -0,0 +1,19 @@
...
+import { FormBuilderContext } from "@webiny/api-form-builder/types";
+
+export default () => {
+    return new ContextPlugin<PbContext & FormBuilderContext>(
+        async ({ pageBuilder, formBuilder }) => {
+            /**
+             * If page contains a form and the version of that form is "latest published revision",
+             * then it should trigger re-render for all pages that contain that form.
+             */
+            formBuilder.onFormAfterPublish.subscribe(async ({ form }) => {
```
adrians5j, line 12: What about if a revision has been updated and we have a page that's showing an exact revision?

packages/api-page-builder/src/prerendering/hooks/afterFormUpdate.ts
adrians5j, line 1: BTW the name of this file should use `Publish` verb, not `Update`.

adrians5j (changes requested):
First iteration done. 

I can see one or two additional reviews happening here, but overall we're heading in the right direction.

packages/api-page-builder/src/prerendering/hooks/afterFormUpdate.ts
neatbyte-ivelychko: @adrians5j hm, I think that we cannot update specific form version if it has been published. Also if page shows a specific form revision and we decide to publish new revision version, the page that uses specific revision not gonna be re-rendered

packages/api-page-builder/src/prerendering/hooks/afterFormDeletetion.ts
neatbyte-ivelychko: @adrians5j so you want me to move all form-builder related handlers to the api-form-builder and just import them in `api-page-builder`?

packages/app-form-builder/src/page-builder/admin/plugins/components/PeFormElement.tsx
neatbyte-ivelychko: If we add a form on the page and then we delete that form we get error `Form "64ee005792862d000941dc7a#0003" was not found!`. on the `Page Builder Editor` page. This is because the query in `catch` block is trying to query a FORM that does not exist. And the `catch` helps us to avoid that error.

neatbyte-ivelychko: @adrians5j also It was a quick fix, I will imrove it

apps/api/graphql/src/index.ts
adrians5j: This question still open.

packages/api-page-builder/src/prerendering/hooks/afterFormDeletetion.ts
adrians5j: Let's have all of the handlers defined in the `api-form-builder` package.

Probably just moving the file into the package will be good enough (ofc you'll need to import it in relevant files).

But yeah, as mentioned, this will require us to list PB as a dep of FB, because we're accessing the `prerendering` property on the `pageBuilder` context, both of which are not defined I believe in FB's Context object.

```packages/api-form-builder/src/types.ts:438```

adrians5j (changes requested):
Left some comments. Understood the UI change part (deletion of form instead of form revision). Thanks for that.

Will do another review where I analyse all of the prerendering handlers.

neatbyte-vnobis:
Please refer to the new PR for this feature #3541

PR #3472, opened 2023-08-14
Fix button with named element reference

mihajlovco:
With this PR we fixed the problem of creating a reference link to the same page using Lexical editor. Now users can link elements to the same page.

## Changes
- New custom Lexical `LinkNode` class that overrides default link node.
   - In [createDOM](packages/lexical-editor/src/nodes/link-node.ts) method, we sanitize the URL with our custom method.
- New `isUrlLinkReference` method  - detect link reference - checks if URL is starting with (#) hashtag
- Update to existing `sanitize` method to recognize the URLs that are link references to other elements on the page.
- In the Lexical link floating popup, when the user enters the link reference, the checkbox "Open in new page" will be disabled (check the test video). 

## Technical problem with the current lexical solution
- Default link [sanitizer method](https://github.com/facebook/lexical/blob/cf0b053a9694af3a6beac42f5fd581c01e5c7a6a/packages/lexical-link/src/index.ts#L165) in Lexical `LinkNode` class, doesn't allow reference links like `#link-to-my-title` to be set in the `href` attribute.
- `LexicalHtmlRenderer` uses the same default `LinkNode` node so the component can't render the link references.

## How Has This Been Tested?
Manually.

## Tests

https://github.com/webiny/webiny-js/assets/82515066/1cfb5595-e486-47b3-8866-36a4d1530710

### DIFF
--- a/packages/lexical-editor/src/nodes/link-node.ts
+++ b/packages/lexical-editor/src/nodes/link-node.ts
@@ -1,0 +1,97 @@
+import {
+    LinkAttributes,
+    LinkNode as BaseLinkNode,
+    SerializedAutoLinkNode,
+    SerializedLinkNode as BaseSerializedLinkNode
+} from "@lexical/link";
+import { DOMConversionMap, DOMConversionOutput, EditorConfig, NodeKey, Spread } from "lexical";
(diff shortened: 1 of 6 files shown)


mihajlovco:
PR with a better title to the problem. https://github.com/webiny/webiny-js/pull/3474

PR #3474, opened 2023-08-15
Lexical link to support reference to element in the same page

mihajlovco:
With this PR we fixed the problem of creating a reference link to the same page using Lexical editor. Now users can link elements to the same page.

## Changes
- New custom Lexical `LinkNode` class that overrides default link node.
   - In [createDOM](https://github.com/webiny/webiny-js/pull/3474/files#diff-40531d17943b6f9bc2f010a2769055a6050e37c456d0ee41e0ed98bc1631c20f) method, we sanitize the URL with our custom method.
- New `isUrlLinkReference` method  - detect link reference - checks if URL is starting with (#) hashtag
- Update to existing `sanitize` method to recognize the URLs that are link references to other elements on the page.
- In the Lexical link floating popup, when the user enters the link reference, the checkbox "Open in new page" will be disabled (check the test video). 

## Technical problem with the current lexical solution
- Default link [sanitizer method](https://github.com/facebook/lexical/blob/cf0b053a9694af3a6beac42f5fd581c01e5c7a6a/packages/lexical-link/src/index.ts#L165) in Lexical `LinkNode` class, doesn't allow reference links like `#link-to-my-title` to be set in the `href` attribute.
- `LexicalHtmlRenderer` uses the same default `LinkNode` node so the component can't render the link references.

## How Has This Been Tested?
Manually.

## Tests

https://github.com/webiny/webiny-js/assets/82515066/1cfb5595-e486-47b3-8866-36a4d1530710

### DIFF
--- a/packages/lexical-editor/src/nodes/webinyNodes.ts
+++ b/packages/lexical-editor/src/nodes/webinyNodes.ts
@@ -43,7 +43,7 @@ export const WebinyNodes: ReadonlyArray<
     FontColorNode,
     TypographyElementNode,
 
-    // The following code replaces the built-in Lexical nodes with our custom nodes. The built-in nodes are
+    // The following code replaces the built-in Lexical nodes with our custom ones. The built-in nodes are
     // imported from the Lexical package, and then replaced with our custom nodes. This is the only way to
     // replace the built-in nodes.
     // https://lexical.dev/docs/concepts/node-replacement
--- a/packages/lexical-editor/src/nodes/webinyNodes.ts
+++ b/packages/lexical-editor/src/nodes/webinyNodes.ts
@@ -16,9 +16,7 @@ import { QuoteNode } from "~/nodes/QuoteNode";
 import { ImageNode } from "~/nodes/ImageNode";
 import { LinkNode } from "~/nodes/link-node";
 
-/*
- * This is a list of all the nodes that Webiny's Lexical implementation supports OOTB.
- * */
+// This is a list of all the nodes that our Lexical implementation supports OOTB.
 export const WebinyNodes: ReadonlyArray<
     | Klass<LexicalNode>
     | {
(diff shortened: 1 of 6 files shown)


packages/lexical-editor/src/nodes/webinyNodes.ts
```diff
@@ -62,5 +62,20 @@ export const WebinyNodes: ReadonlyArray<
         with: () => {
             return new QuoteNode();
         }
+    },
+    LinkNode,
```
adrians5j, line 69: Not related that much to this PR but... 

I'm still not 100% sure on why in this list we're first registering e.g. `LinkNode`, and then registering the replacement of all `BaseLinkNode` with `LinkNode`. 

Can you remind me... The base nodes are already registered somewhere in Lexical lib's internals? That's why we don't see the registration of base nodes like `BaseLink` here, right?

packages/lexical-editor/src/plugins/FloatingLinkEditorPlugin/FloatingLinkEditorPlugin.tsx
```diff
@@ -162,6 +163,7 @@ function FloatingLinkEditor({
                         <input
                             type={"checkbox"}
                             checked={linkUrl.target === "_blank"}
+                            disabled={isUrlLinkReference(linkUrl.url)}
```
adrians5j, line 166: Why is this here? Can we add a comment?

adrians5j (changes requested):
A couple of minor comments.

packages/lexical-editor/src/nodes/webinyNodes.ts
mihajlovco: We need to register our node as the default node for the link. We must provide only one node for the link, and our custom node is now in charge of the links. 

When lexical is traversing and find the built-in `link` node, checks if we have registered a custom default node, then on the `with: (node: BaseLinkNode)` line, will try to transfer the constructor data and create an instance of our custom `LinkNode` with type` link-node`. 

In this way, lexical has a mechanism to support old and new links. The old type `link` will be converted to the custom type `link-node`.

packages/lexical-editor/src/plugins/FloatingLinkEditorPlugin/FloatingLinkEditorPlugin.tsx
adrians5j: You know what.. I just figured it's disabling this checkbox here: 
image

Resolving this, no need to do anything here.

packages/lexical-editor/src/nodes/webinyNodes.ts
adrians5j: All good. Just wasn't sure about the "The base nodes are already registered somewhere in Lexical lib's internals?" part.

Let's add a comment about this just for future, and we can merge this (once we release 5.37.1).

packages/lexical-editor/src/plugins/FloatingLinkEditorPlugin/FloatingLinkEditorPlugin.tsx
mihajlovco: When we link the elements on the same page, we want the page to scroll to the linked element on the same page.

To prevent the user to open reference links like `<a href="#scroll-to-my-title">link</a>` in another tab, I disabled the checkbox option user to select that possibility.

image

PR #3487, opened 2023-08-21
feat: Icon picker component (WIP)

neatbyte-vnobis:
## Changes
Implements issue "Icon picker component".

### Implementation steps
Most of the implementation is described [here](https://webiny.notion.site/Icon-picker-component-Improvements-fc5e6e2387694457ac2eee9f629d7048).

For custom icon support IconPickerConfig was added under `packages/app-admin/src/components/IconPicker/config/index.tsx`.
For the most part it is similar to what we've done with audit logs [here](https://github.com/webiny/webiny-js/blob/4df5051b097523beeb765d4d7efca3b136fdd284/packages/app-audit-logs/src/config/list/AuditLogsListConfig.tsx).
It allows to add custom icons/emojis inside [DefaultIcons](https://github.com/webiny/webiny-js/pull/3487/files#diff-6e6691e6103e3bdd98b77e58da07410d2b990c9b7b18ca11fadd78961e282549R32). Also there is an [example](https://github.com/webiny/webiny-js/pull/3487/files#diff-95f5dd4cde2fbc8ad1716c4421b3c0c2bacd78c8912db6e1b57807f29c4d306eR21) of how it can be used.

### UI
Color Picker UI is currently only available on `/page-builder/block-categories` page for test purposes.
image

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-vnobis:
@Pavel910 @SvenAlHamad  I have a couple questions about Icon Picker:
1) Emojis need to be displayed by category?
There is a `People` category on design, but nothing about it in doc. I'am asking, because it may not be that easy to implement with `react-virtualized` `Grid` component.

2) How will we store our custom icons?
I see two possible solutions:
 a) Create a separate api for custom icons.
 b) Use the existing `File Manager` api. We can add flag (e.g. `isIcon`) under file meta property, when a user adds an icon to `Icon Picker`. And remove it when the user deletes it from `Icon Picker`.

neatbyte-vnobis:
@Pavel910 @SvenAlHamad
Please take a look at the questions above.

packages/app-admin/src/components/IconPicker/IconPicker.tsx
```diff
@@ -0,0 +1,330 @@
...
+                        });
+                    }}
+                    color={color}
+                    isActive={item.name === value.name}
+                >
+                    <IconRenderer icon={item} size={32} />
+                </Cell>
+            );
+        };
+    }, [filteredIcons, color]);
```
Pavel910: I think this is overcomplicated; Why return a new function? Further down in JSX, you just call `renderCell()` when passing it to props, so you can essentially just do this:

```tsx
const renderCell = useCallback(({ columnIndex, key, rowIndex }) => { .... }, [filteredIcons, color]);
```
And then you pass `renderCell` as a `cellRenderer` prop without invoking it.

packages/app-admin/src/components/IconPicker/IconPicker.tsx
```diff
@@ -0,0 +1,330 @@
...
+
+    const tabsRef = useRef<TabsImperativeApi>();
+    const { icons } = useIconPickerConfig();
+
+    // const emojisByCategory = groupBy(
+    //     icons.filter(icon => icon.type === "emoji"),
+    //     "category"
+    // );
+    const emojis = icons.filter(icon => icon.type === "emoji");
+    const defaultIcons = icons.filter(icon => icon.type === "icon");
```
Pavel910: This is something you want to memoize with `useMemo`.

packages/app-admin/src/components/IconPicker/config/IconPackProvider.tsx
```diff
@@ -0,0 +1,20 @@
...
+import { Property } from "@webiny/react-properties";
+
+import { Icon, IconProps } from "./Icon";
+
+export type IconPackProviderProps = {
+    provider: () => IconProps[];
+};
+
+export const IconPackProvider = ({ provider }: IconPackProviderProps) => {
+    const icons = provider();
```
Pavel910: Can we make this work with async providers? Think code splitting... Would be nice to allow icons to be lazy loaded, or even loaded from an API. That way they'll only load when IconPicker is actually being used in the UI.

Pavel910 (changes requested):

SvenAlHamad:
I was testing the icon component implementation and have a few UX related points that require a bit of polish:
- improve the search: if I search for "glass" or "magnifying glass" I don't get the icon of a magnifying glass; but if I search for "magnifying-glass" then I get it. We should improve the search so that in both these cases I get the expected result. 
- empty state: when we have an empty icon field, can add the magnify glass as the default placeholder, just lower the opacity on the icon in that case so that we hint to the user this is an empty state for the icon field
- when I select and icon, automatically close the window; currently I need to click manually outside the window to close it
- when I have a search query in one tab (say I wrote "boat" in the input field in the icons tab) and if I switch to the emoji tab, keep the same value in the input

Pavel910:
Attaching this git patch file here. This contains all changes from the iconpicker PR in one place, so it's easier to integrate with the latest branch. Merging of this is not possible, so we need to carefully pick changes. Part of these changes will be merged in one of the upcoming PRs, but I'm still keeping this here, not to lose it accidentally.

[iconpicker.patch](https://github.com/user-attachments/files/17543129/iconpicker.patch)

PR #3491, opened 2023-08-23
feat: Add conditional fields and conditional form steps

neatbyte-vnobis:
## Changes
Implements issue "Add conditional fields and conditional form steps".

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/apps/admin/src/plugins/formBuilder.ts
+++ b/apps/admin/src/plugins/formBuilder.ts
@@ -13,6 +13,7 @@ import editorFieldNumber from "@webiny/app-form-builder/admin/plugins/editor/for
 import editorFieldRadioButtons from "@webiny/app-form-builder/admin/plugins/editor/formFields/radioButtons";
 import editorFieldCheckboxes from "@webiny/app-form-builder/admin/plugins/editor/formFields/checkboxes";
 import editorFieldDateTime from "@webiny/app-form-builder/admin/plugins/editor/formFields/dateTime";
+import editorFieldConditionGroup from "@webiny/app-form-builder/admin/plugins/editor/formFields/conditionGroup";
 import editorFieldFirstName from "@webiny/app-form-builder/admin/plugins/editor/formFields/contact/firstName";
 import editorFieldLastName from "@webiny/app-form-builder/admin/plugins/editor/formFields/contact/lastName";
 import editorFieldEmail from "@webiny/app-form-builder/admin/plugins/editor/formFields/contact/email";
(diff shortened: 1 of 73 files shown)


neatbyte-ivelychko:
[WIP]

neatbyte-ivelychko:
@adrians5j @SvenAlHamad 

I have some questions regarding Conditon Group:
  1) Should `Condition Group` be hidden by default or it should be visible by default?
If it should be hidden by default then maybe we should have only one action option and it is `Show` in the `Rules` for the `Condition Group`.
Because if conditions are not met then `Condition Group` still `NOT` be visible, and if conditions are met then `Condition Group` will be visible.

And otherwise -  if `Condition Group` should be visible by default then we should do everything vise versa to the sentence above.
What do you think?

  2) `Condition Group` as a field for `Rules` conditions
Can we select `Condition Group` as a field for `Rules` conditions (as we can now select any other field, like `First Name`) for `Step`.
Or can we select fields for `Step Rules` conditions that are inside of `Condition Group`?


3) Should Condition Group have `Rules attached` or  `Rules are broken` tag?
As of now if `Step` has rules then it has the `Rules attached` tag, and if rules of the `Step` are broken then we add `Rules are broken` tag. Should we also add same tags for the `Condition Group`?
Знімок екрана 2023-08-23 о 12 18 44

adrians5j:
Will leave this questions for @SvenAlHamad to answer.

SvenAlHamad:
Hey @neatbyte-ivelychko regarding your questions:

1. Can we allow the user to choose what the default behaviour would be - we can do that by having a "Default behavior" accordion under the rules. Something like this: 

CleanShot 2023-08-29 at 13 37 54@2x

2. For step rules I would say the user can choose any of the fields inside the condition group, but the condition group it self is not something the user can select. 

3. We can have the "rules attached" tag to keep the experience consistent. I'm not sure what "rules are broken" tag means, can you explain please?

neatbyte-ivelychko:
@SvenAlHamad 
"Rules are broken" tag means that something wrong with rules for the step. For instance in `Step 2` in `Rules` we have created a condition with a field `First Name` from `Step 1`. And then we decided to move `Step 1` below `Step 2`, if we do that, the condition for `Step 2` that contains field `First Name` from `Step 1` would not work. Because `Step 2` has no longer access to the fields from `Step 1`.

SvenAlHamad:
I see, it makes sense. In that case can we use our Alert component and display a message in the rules tab. The message can say: 

"At the moment one or more of your rules are broken. To correct the state please check your rules and ensure they are referencing fields that still exists and are place inside the current or one of the previous steps."

PR #3495, opened 2023-08-25
fix: fix rerenders on element select

neatbyte-vnobis:
## Changes
Resolve issue "Continuous rerenders when selecting element inside page/block/template editors".

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveDialog.tsx
+++ b/packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveDialog.tsx
@@ -66,7 +66,7 @@ const SaveDialog = (props: Props) => {
         setTimeout(async () => {
             setPbElement((await getElementTree({ element })) as PbElement);
         });
-    });
+    }, []);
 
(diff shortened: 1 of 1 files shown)


packages/app-page-builder/src/editor/plugins/elementSettings/save/SaveDialog.tsx
```diff
@@ -66,7 +66,7 @@ const SaveDialog = (props: Props) => {
         setTimeout(async () => {
             setPbElement((await getElementTree({ element })) as PbElement);
         });
-    });
+    }, []);
```
adrians5j: Should we rather use `element.id` as the cache key? Could be more precise.

Also, this whole `useEffect` block was added probably with the improved block previews PR, but I guess I missed this. Can we add a short comment explaining why this is needed?

adrians5j (changes requested):
Quick review.

adrians5j (approved):

PR #3497, opened 2023-08-27
fix(rte-lexical-renderer) - information about react mismatching versions

mihajlovco:
With this PR we want to update `README.md` file and inform users how to resolve the problem of mismatching versions between our component which is on version `17.0.2` and the 3rd party custom React app on version `18.x.x`.

## Changes
- Provide a new topic for users to solve this issue in `README.md` file.
- Update documentation for `Verdaccio` commands. (in addition to this PR)

## How Has This Been Tested?
Manually. 
Tested in custom React application.

Preview of the error 
image


packages/react-rich-text-lexical-renderer/README.md
```diff
@@ -96,12 +101,42 @@ const myTheme = {
...
+  },
+  ...
+  "resolutions": {
+    "react": "18.2.0"
+  },
+  ...
+}
+```
+
+Then run `yarn install`.
```
Pavel910: You should make it clear that this solution only works for `yarn` projects, not `npm`. In general this whole section is very stiff, and robot-like. There's no natural flow to it. 

> it's possible to encounter errors.

No, it's not a possibility, it's a fact. An error WILL slap you.

> You will see this error message on the screen

When will I see this error? Who will throw the error? Also, that error is just an excerpt from the full error. Show the full error  so developers can quickly compare their error and the one you're documenting.

You should also add a link to the Yarn docs, so people can learn more about resolutions: https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/

Give it a few more tries, try to make it more human, like you're explaining it to a 4-year old.

Pavel910 (changes requested):

packages/react-rich-text-lexical-renderer/README.md
```diff
@@ -96,12 +101,73 @@ const myTheme = {
...
+When you try to use `RichTextLexicalRenderer` component in React `v18` application you will see this error on the
+screen:
+
+![React application error for mismatch of the React versions](./images/react-renderer-versisons-conflict-error.png)
+
+This is because our `@webiny/react-rich-text-lexical-renderer` package and the React application have
+different versions of React. Our rich text renderer component is using `v17.0.2`, and the React application is
+using `v18.x.x`.
+
+> You can check the React version dependencies by running following command:
```
Pavel910:
```
You can check which React versions are requested by various dependencies by running the following command
```

packages/react-rich-text-lexical-renderer/README.md
```diff
@@ -96,12 +101,73 @@ const myTheme = {
...
+> - `yarn why react` for `yarn` users.
+> - `npm ls react` for `npm` users.
+
+To resolve this problem, we need to define the versions of the `@webiny/react-rich-text-lexical-renderer`
+package, and it's dependencies to `v18.x.x`.
```
Pavel910:
```
To resolve this problem, we need to force all dependencies to use the same version of React.
```

packages/react-rich-text-lexical-renderer/README.md
```diff
@@ -96,12 +101,73 @@ const myTheme = {
...
+
+### Instructions for `yarn` users
+
+To define the `v18.x.x` of React to all package dependencies that use the old version of React, we need to
```
Pavel910: To force `yarn` to resolve dependencies across the project to the exact versions we're looking for, use the `resolutions` field in the root `package.json` file.

packages/react-rich-text-lexical-renderer/README.md
```diff
@@ -96,12 +101,73 @@ const myTheme = {
...
+{
+  ...
+  "resolutions": {
+    "react": "18.x.x"
+  },
+  ...
+}
 ```
+
+After, run `yarn install` in your terminal, and `@webiny/react-rich-text-lexical-renderer` package and inside
```
Pavel910: Make this sentence shorter:

```
Once the `resolutions` field is defined, run `yarn` to apply the new config.
```

packages/react-rich-text-lexical-renderer/README.md
```diff
@@ -96,12 +101,73 @@ const myTheme = {
...
+{
+  ...
+  "overrides": {
+    "react": "^18.x.x"
+  },
+  ...
+}
+```
+
+After, run `npm install` in your terminal, and `@webiny/react-rich-text-lexical-renderer` package and inside
```
Pavel910: Same here:

```
Once the `overrides` field is defined, run `npm install` to apply the new config.
```

Pavel910 (changes requested):
@mihajlovco much better, has better flow and clear guidance. Just a few comments on grammar/clarity of sentences, and we're done here.

packages/react-rich-text-lexical-renderer/README.md
mihajlovco: Changed.

Pavel910 (approved):

PR #3501, opened 2023-08-28
feat: add bulk actions to HCMS

leopuleo:
## Changes
With this PR, we allow bulk actions to headless CMS entries. The default bulk actions are:

- Publish
- Unpublish
- Move
- Delete

Third-party developers can add bulk actions: a new article describing the process will be added to the developer docs.

## How Has This Been Tested?
Jest + manually

## Documentation
Inline + Developer Doc article (TBD)


packages/app-admin/src/components/Buttons/Buttons.tsx
```diff
@@ -34,7 +36,7 @@ export const Buttons: React.VFC<ButtonsProps> = props => {
     return (
         <>
             {props.actions.map(action => (
-                <ButtonContainer key={action.name}>
+                <ButtonContainer key={action.name} className={"button-container"}>
```
Pavel910: Adding CSS classes randomly is a bad idea. It's better to either add props to the component,  or simply create a component variant, like `<NoMarginButtonContainer>`,  and you can even keep it local to this view. 

```tsx
const NoMarginButtonContainer = styled(ButtonContainer)`
    margin: 0;
`;
```

packages/app-admin/src/components/BulkActions/Worker.ts
```diff
@@ -0,0 +1,138 @@
...
+     * @param callback - The callback function to process each item.
+     * @param {T[]} chunk - The chunk of items to be processed.
+     * @returns {Promise<void>}
+     */
+    private async processChunk(
+        callback: ({ item, allItems, report }: CallbackParams<T>) => void | Promise<void>,
+        chunk: T[]
+    ): Promise<void> {
+        for (const item of chunk) {
+            await callback({ item, allItems: this._items, report: this._report });
```
Pavel910, line 135: Do you think it's work adding `pRetry` here, or do you want the developer to handle failed executions/api calls in their callback?

packages/app-headless-cms/src/admin/components/ContentEntries/BulkActions/styles.tsx
```diff
@@ -0,0 +1,29 @@
...
+    padding: 0 16px;
+`;
+
+export const ButtonsContainer = styled.div`
+    display: flex;
+    align-items: center;
+
+    .button-container {
+        margin: 0;
+    }
```
Pavel910: This is connected to my other comment on randomly adding classes. Nobody knows this exists, it's not easily discoverable, it's not in TS autocomplete, so let's avoid these hidden magic classes.

Pavel910 (changes requested):
Just a few comments here, and we are good. In general, it's great that you developed a system first (the configuration), then followed the same approach for every action.

packages/app-admin/src/components/Buttons/Buttons.tsx
leopuleo: Here is the problem: We are using the same component to render [Entry Editor Actions](https://www.webiny.com/docs/headless-cms/extending/customize-entry-editor-actions) and now the Entry List Bulk actions. 

I don't like to use the CSS class but I need to target the single action from different parts of the UI.

I changed it: now I'm exporting the styled component from the `@webiny/app-admin` package so I can target it inside the specific component.

packages/app-admin/src/components/BulkActions/Worker.ts
leopuleo: I love having a retry strategy but would not bake it inside the Worker, keeping it simple and straightforward. 

I noticed that we are already exposing a helper function that wraps [pRetry](https://github.com/webiny/webiny-js/blob/next/packages/utils/src/executeWithRetry.ts)...developer could use it outside the Worker logic.

Pavel910 (approved):

PR #3508, opened 2023-08-30
fix(data-migration): add log stream link and print logs in CI output

Pavel910:
## Changes
With this PR, we've added 3 things to improve the DX related to data migrations:  
- we print a CloudWatch log stream link at the beginning and at the end of each migration run, so it's easy to locate the exact log stream that is connected to the particular migration run
- we pull logs from the log stream, and print them to the CI console, so they're available to developers who might not have access to the AWS account that is doing the deployment of the project (large teams where developers and dev-ops engineers are separate company units)
- you can now force-execute a specific migration by using an ENV variable, for example:  `WEBINY_MIGRATION_FORCE_EXECUTE_5_36_0_001`

CleanShot 2023-08-30 at 21 02 51

image

## How Has This Been Tested?
Manually, and Jest tests.


packages/data-migration/src/cli/InteractiveCliStatusReporter.ts
```diff
@@ -0,0 +1,59 @@
...
+    report(migrationStatus: MigrationStatus): void {
+        const { status, migrations, context } = migrationStatus;
+        this.clearLine();
+
+        const currentLogStreamName = context?.logStreamName;
+        if (currentLogStreamName) {
+            this.logReporter.initializeStream(currentLogStreamName);
+        }
+
+        if (status === "running") {
```
brunozoric, line 29: These ifs are good candicates for for a switch statement, with a default behavior when status is not one of the running, init or done.

packages/data-migration/src/cli/LogStream.ts
```diff
@@ -0,0 +1,76 @@
...
+    private nextPage: string | undefined;
+
+    private constructor(logGroupName: string, logStreamName: string) {
+        this.logGroupName = logGroupName;
+        this.logStreamName = logStreamName;
+        this.cloudWatchLogs = new CloudWatchLogs();
+    }
+
+    getLogStreamLink() {
+        const replacements = [
```
brunozoric, line 18: Where did you find what needs to be replaced?
Wouldnt the encodeURIComponent help?

brunozoric (approved):
I left a few comments. But looks ok. Will test as soon as the system I have in place fills with tons of data.

packages/data-migration/src/cli/InteractiveCliStatusReporter.ts
Pavel910: This is just a reporter, meaning, it prints stuff to CLI. Here I'm only interested in these statuses, others I just ignore. As for `switch`, I have a better idea there, but for this release I won't spend any more time on making this any more advanced than what it is.

packages/data-migration/src/cli/LogStream.ts
Pavel910: 😂 I had to look at CloudWatch URLs in the browser. And they're not using encodeURIComponent, but some custom encoding here, see the difference:

encodeURIComponent:
```
%2Faws%2Flambda%2Fwby-data-migration-8a6968a%2Flog-events%2F2023%2F08%2F24%2F%5B%24LATEST%5De41d9ccf18fe487c9599ad4274ceef3f
```

AWS encoding:
```
$252Faws$252Flambda$252Fwby-data-migration-8a6968a/log-events/2023$252F08$252F24$252F$255B$2524LATEST$255De41d9ccf18fe487c9599ad4274ceef3f
```

They encode some parameters, but leave some parts like `/log-events/` in its raw form...

brunozoric (approved):

PR #3516, opened 2023-09-12
feat: add createdOn column to ACO tables

leopuleo:
## Changes
With this PR, we add a new column `createdOn` to all ACO tables: the column will also be available for sorting.

**Bonus:** we add a new `initialSorting` prop to the `DataTable` component, which defines the initial sorting state and handles the `tableSorting` state within the component itself.

## How Has This Been Tested?
Manually + Cypress


brunozoric (approved):
Everything looks fine.

PR #3531, opened 2023-09-14
Add support for PB page force-update

Pavel910:
## Changes
This PR adds an ability to force-update pages, and turn off publishing of lifecycle events. Force-update means that updates will be executed even if the page is `locked`. This is only exposed via programmatic usage; GraphQL API does not expose this feature.

image

## How Has This Been Tested?
Manually.


Pavel910:
After in-depth analysis of the problem, we decided to drop this feature. It doesn't make any sense in the context of  prerendering, even when using external delivery systems.

Issue #3510, opened 2023-08-31
inconsistent pb element save-behavior in editor


woeps:
### Version

5.37.2

### Operating System

Linux

### Browser

Firefox 116.0.2 (64-bit)

### What are the steps to reproduce this bug?

Drop one of the following pb elements onto a page in the editor (without any further modification of the element's settings):

- basic
  - icon
  - image gallery
- layout
  - grid
  - carousel
  - tabs
  - accordion
- form
  - form
- embeds
  - soundcloud
  - vimeo
  - youtube
  - iframe
- social
  - pinterest
  - twitter
- code
  - codesandbox

Preview the page (no matter if published or not).

### What is the expected behavior?

To see the newly dropped element or at least a placeholder noting "this element is not yet fully configured" (or similar).

### What do you see instead?

The newly dropped element is not visible at all.

### Additional information

The elements are are not saved (`updatePage` mutation won't get called) upon dropping the listed elements.

The following elements trigger the `updatePage` mutation immediately upon dropping it onto a page:

- basic
  - heading
  - paragraph
  - button
  - list of pages

This inconsistent behavior lead to quite some confusion in my team currently evaluation webiny. I believe the behavior should be consistent.

This concern was first raised on [Slack](https://webiny-community.slack.com/archives/C014Y0HGJ0Y/p1692890250009589).

### Possible solution

Trigger the `updatePage` for any element being dropped onto the page (in the pb editor) and have elements, which need some settings to correctly render, a "fallback" to render a warning about not being fully configured.

PR #3540, opened 2023-09-20
fix: fix inconsistent pb element save-behavior in editor

neatbyte-vnobis:
## Changes
Resolve issue: [Inconsistent pb element save-behavior in editor](https://github.com/webiny/webiny-js/issues/3510).

1) Page revision now updates after adding an element to the page.
2) Added fallback message to empty elements (`image`, `embeds`) on website.
  image

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx
```diff
@@ -39,9 +39,14 @@ const ActionSettingsComponent: React.FC<ActionSettingsPropsType> = ({
 }) => {
     const updateElement = useUpdateElement();
 
-    const { actionType } = element.data?.action || { actionType: "link" };
+    const { actionType } = element.data?.action || {};
 
     const updateSettings: FormOnSubmit = data => {
+        // Skip update if nothing is changed.
+        if (data.actionType === actionType) {
```
adrians5j: Let's double check this. It looks like now I can't save button action settings at all. 
image

packages/app-page-builder/src/editor/recoil/actions/createElement/action.ts
```diff
@@ -1,8 +1,9 @@
+import { UpdateDocumentActionEvent } from "~/editor/recoil/actions/updateDocument";
 import { CreateElementEventActionArgsType } from "./types";
 import { EventActionCallable } from "~/types";
 
 export const createElementAction: EventActionCallable<CreateElementEventActionArgsType> = () => {
     return {
-        actions: []
+        actions: [new UpdateDocumentActionEvent()]
```
adrians5j, line 7: Let's add a comment why this step please.

packages/app-page-builder/src/types.ts
```diff
@@ -180,12 +180,12 @@ export type PbElementDataType = {
     blockId?: string;
     variables?: PbBlockVariable[];
     action?: {
-        href: string;
-        newTab: boolean;
-        clickHandler: string;
+        href?: string;
+        newTab?: boolean;
```
adrians5j: Should the same changes be applied in `packages/app-page-builder-elements/src/renderers/button.tsx:80` ?

packages/app-page-builder/src/editor/plugins/elements/button/index.tsx
```diff
@@ -70,7 +70,8 @@ const buttonElementPluginsFactory = (args: PbEditorElementPluginArgs = {}) => {
                                 "center",
                                 DisplayMode.DESKTOP
                             ) as unknown as "flex-start" | "center" | "flex-end"
-                        }
+                        },
+                        action: { actionType: "link", href: "" }
```
adrians5j: Why this change? Can you just add a comment?

Sorry for nitpkicking, just wanted to be 100% sure I understand what's going on here.

adrians5j (changes requested):
Left a couple of comments.

packages/app-page-builder/src/editor/recoil/actions/createElement/action.ts
neatbyte-vnobis: Did some more investigation on this.

Currently, Page Builder does not send update document requests to API when an element is added to a page. However, based on my testing, some elements (such as `Heading`, `Paragraph`, `Image`, `Button`, `Pages List` and `Form`) do trigger a document update.

For instance, adding a button element triggers this [updateSettings](https://github.com/webiny/webiny-js/blob/next/packages/app-page-builder/src/editor/plugins/elementSettings/action/ActionSettings.tsx#L44) function, when added to the page. Elements utilizing the Lexical Editor (like `Heading` and `Paragraph`) are triggering this [onChange](https://github.com/webiny/webiny-js/blob/next/packages/lexical-editor-pb-element/src/components/LexicalText.tsx#L44) function.

By adding new `UpdateDocumentActionEvent()` to the `createElementAction` we ensure the document is updated after adding a new element to the page. But since the elements mentioned already trigger a document update on their own, this addition now results in two separate API calls when they are added.

I've tried to fix this for the `Button`, but the fix didn't work as expected, so I reverted these `Button` changes.

neatbyte-vnobis:
@adrians5j Is this PR ready to be merged?

PR #3541, opened 2023-09-20
feat: page re-render on form publish

neatbyte-vnobis:
## Changes
Implements issue "When a form is published, re-render all the pages containing it".

## How Has This Been Tested?
Manual

## Documentation
None


neatbyte-ivelychko:
## What has been implemented
- afterFormPublish: triggers page rerender in case page that contains a form with a selected `Latest Published Revision`;
- afterFormRevisionDelete: triggres page rerender in case page that contains a form with a specific `Published Revision`;
- afterFormDelete: triggers page rerender in case page that contains a specific form.
##
@adrians5j regarding your suggestion:
Знімок екрана 2023-09-12 о 13 30 41

So I have tried to implement it and I figured out that the latestPublished.id and form.id are gonna be always the same, so the IF statement will be OMITTED.

packages/api-form-builder/src/prerendering/index.ts
```diff
@@ -0,0 +1,6 @@
+import { prerenderingHandlers } from "@webiny/api-page-builder/prerendering/prerenderingHandlers";
```
adrians5j: Hm, why do we need this? I don't think we need it. 🤔

apps/api/graphql/src/index.ts
```diff
@@ -10,6 +10,7 @@ import {
 } from "@webiny/api-page-builder/graphql";
 import { createStorageOperations as createPageBuilderStorageOperations } from "@webiny/api-page-builder-so-ddb";
 import pageBuilderPrerenderingPlugins from "@webiny/api-page-builder/prerendering";
+import formBuilderPrerenderingPlugins from "@webiny/api-form-builder/prerendering";
```
adrians5j: Let's have this embedded in the `createFormBuilder` factory. We don't want to have this visible in this file.

packages/api-form-builder/src/prerendering/hooks/afterFormDelete.ts
```diff
@@ -0,0 +1,22 @@
...
+    return new ContextPlugin<FormBuilderContext & PbContext>(
+        async ({ formBuilder, pageBuilder }) => {
+            /**
+             * If page contains a form that has been deleted,
+             * then this hook should trigger re-render for all pages that contain that form.
+             */
+            formBuilder.onFormAfterDelete.subscribe(async ({ form }) => {
+                await pageBuilder.prerendering.render({
+                    tags: [
+                        { tag: { key: "fb-form", value: form.formId } },
```
adrians5j: Don't we just need the single `{ tag: { key: "fb-form", value: form.formId } },` entry here?

All rendered forms, no matter if it was a specific rev or latest one, will have `<ps-tag data-key="fb-form" data-value={formData.formId} />`. That means all relevant pages will be rerendered.

So, that handles deletion of the whole form. Deletion of revisions-only is handled `afterFormRevisionDelete.ts` file.

packages/app-page-builder-elements/src/renderers/form/FormRender.tsx
```diff
@@ -211,6 +212,10 @@ const FormRender: React.FC<FormRenderProps> = props => {
         <>
             <FormLayoutComponent {...layoutProps} />
             <ps-tag data-key="fb-form" data-value={formData.formId} />
+            <ps-tag
+                data-key="fb-form-revision"
+                data-value={`${formData.id}${isLatestRevision ? "#latest" : ""}`}
```
adrians5j: I thought we said we don't need the `#latest` suffix anymore. 🤔

packages/api-form-builder/src/prerendering/hooks/afterFormPublish.ts
```diff
@@ -0,0 +1,22 @@
...
+        async ({ formBuilder, pageBuilder }) => {
+            /**
+             * If page contains a form and the version of that form is "latest published revision",
+             * then it should trigger re-render for all pages that contain that form.
+             */
+            formBuilder.onFormAfterPublish.subscribe(async ({ form }) => {
+                await pageBuilder.prerendering.render({
+                    tags: [
+                        { tag: { key: "fb-form", value: form.formId } },
+                        { tag: { key: "fb-form-revision", value: `${form.id}#latest` } }
```
adrians5j: We don't need to specify both tags. Just rerendering all pages with `{ tag: { key: "fb-form", value: form.formId } },` will be enough.

This is because, again, no matter if a specific rev was rendered on a page or the latest rev, the page will contain the `<ps-tag data-key="fb-form" data-value={formData.formId} />` tag.

We did discuss that this solution might not be too optimal, because we're potentially rerendering more pages than required, but for now it's fine. Let's just have a comment that mentions this here. You write something and I'll correct if required.

packages/api-form-builder/src/prerendering/hooks/afterFormRevisionDelete.ts
```diff
@@ -0,0 +1,22 @@
...
+    return new ContextPlugin<FormBuilderContext & PbContext>(
+        async ({ formBuilder, pageBuilder }) => {
+            /**
+             * If page contains a form revision that has been unpublished or deleted,
+             * then this hook should trigger re-render for all pages that contain that form revision.
+             */
+            formBuilder.onFormRevisionAfterDelete.subscribe(async ({ form }) => {
+                await pageBuilder.prerendering.render({
+                    tags: [
+                        { tag: { key: "fb-form", value: form.formId } },
```
adrians5j: No need for `{ tag: { key: "fb-form", value: form.formId } },` entry.

adrians5j (changes requested):
Not quite there yet, some fixes on hooks need to be done.

packages/api-form-builder/src/prerendering/hooks/afterFormRevisionDelete.ts
neatbyte-ivelychko: @adrians5j fixed all comments

packages/api-form-builder/src/prerendering/hooks/afterFormRevisionDelete.ts
```diff
@@ -0,0 +1,19 @@
...
+import { PbContext } from "@webiny/api-page-builder/types";
+import { FormBuilderContext } from "~/types";
+
+export default () => {
+    return new ContextPlugin<FormBuilderContext & PbContext>(
+        async ({ formBuilder, pageBuilder }) => {
+            /**
+             * If page contains a form revision that has been unpublished or deleted,
+             * then this hook should trigger re-render for all pages that contain that form revision.
+             */
```
adrians5j:
```suggestion
            /**
             * If there are published pages that include the form revision
             * that was deleted, we need to rerender them.
             */
```

packages/api-form-builder/src/prerendering/hooks/afterFormPublish.ts
```diff
@@ -0,0 +1,19 @@
...
+import { FormBuilderContext } from "~/types";
+import { PbContext } from "@webiny/api-page-builder/types";
+
+export default () => {
+    return new ContextPlugin<FormBuilderContext & PbContext>(
+        async ({ formBuilder, pageBuilder }) => {
+            /**
+             * If page contains a form and the version of that form is "latest published revision",
+             * then it should trigger re-render for all pages that contain that form.
+             */
```
adrians5j:
```suggestion
            /**
             * After a form was published, we want to rerender all published pages that include it.
             */
```

packages/api-form-builder/src/prerendering/hooks/afterFormDelete.ts
```diff
@@ -0,0 +1,19 @@
...
+import { PbContext } from "@webiny/api-page-builder/types";
+import { FormBuilderContext } from "~/types";
+
+export default () => {
+    return new ContextPlugin<FormBuilderContext & PbContext>(
+        async ({ formBuilder, pageBuilder }) => {
+            /**
+             * If page contains a form that has been deleted,
+             * then this hook should trigger re-render for all pages that contain that form.
+             */
```
adrians5j:
```suggestion
            /**
             * After a form was deleted, we want to rerender all published pages that include it.
             */
```

packages/api-form-builder/src/prerendering/index.ts
```diff
@@ -0,0 +1,5 @@
+import hooks from "./hooks";
+
+export default () => {
+    return [...hooks()];
```
adrians5j:
```suggestion
    return hooks();
```

packages/api-form-builder/src/prerendering/hooks/index.ts
adrians5j, line 1: Let's move this into `packages/api-form-builder/src/plugins/prerenderingHooks/`

adrians5j (changes requested):

adrians5j (approved):

Issue #3135, opened 2023-03-24
Unable to submit a form in newly created locale


winglunlam:
### Version

5.29.0

### Operating System

Mac

### Browser

Chrome

### What are the steps to reproduce this bug?

1. Create a new locale
2. Create a form
3. Create a page with the form
4. Click the submit button on the form
5. No response and able to see the error message "Form Builder" settings not found!" in Network

### What is the expected behavior?

The form should be submitted successfully

### What do you see instead?

Able to see the error message "Form Builder" settings not found!" in Network

Pavel910:
Thanks, we'll have a look at those settings, to see what's going on with new locales.

PR #3546, opened 2023-09-21
fix: fix form builder settings after switching to new locale

neatbyte-vnobis:
## Changes
Resolve issue: [Unable to submit a form in newly created locale](https://github.com/webiny/webiny-js/issues/3135).

Form submission was throwing `"Form Builder" settings not found!` error, since there were no `Form Builder` settings created for the new locale. This also caused the same issue when saving `Form Builder` settings form.
Issue was resolved by triggering `Form Builder` settings creation, when a new `locale` is created.

While fixing this issue I've found that the same is happening for the `Website` settings form, since `Website` settings are also not created for new `locale`. If this fix is ok, then something similar can be done inside `api-page-builder` to create settings after a new `locale` is created.

Also while testing this fix I've noticed that after switching to a new `locale` user is unable to create any pages until a page category with `static` slug is created. Is this expected behavior?

@adrians5j @Pavel910 What do you think?

## How Has This Been Tested?
Manual

## Documentation
None


packages/api-form-builder/src/plugins/settings/createSettingsForNewLocale.ts
```diff
@@ -0,0 +1,15 @@
+import { ContextPlugin } from "@webiny/api";
+import { FormBuilderContext } from "~/types";
+
+export const createSettingsForNewLocale = new ContextPlugin<FormBuilderContext>(context => {
```
adrians5j, line 4: Once a locale has been deleted, let's also delete settings.

packages/api-form-builder/src/plugins/settings/createSettingsForNewLocale.ts
```diff
@@ -0,0 +1,15 @@
...
+    context.i18n.locales.onLocaleAfterCreate.subscribe(async ({ locale }) => {
+        const existingSettings = await context.formBuilder.getSettings({
+            auth: false,
+            locale: locale.code
```
adrians5j: Do we really need locale here? Wouldn't the current locale automatically be pulled [above](https://github.com/webiny/webiny-js/pull/3546/files#diff-d2594e3f9145cb8635149f194259ac7a98c4cf1181fe220c13ea7065424e7b65R113)?

If there's no really a need to expose locale param, let's maybe not do it.

adrians5j (changes requested):
Minor stuff.

I agree with PB-related things you mentioned. We can do them all within this PR or via separate PRs, your choice. 🙂

neatbyte-vnobis:
@adrians5j While adding similar changes to PB I've noticed that there is no delete option in PB settings crud (FB settings crud [has](https://github.com/webiny/webiny-js/blob/next/packages/api-form-builder/src/plugins/crud/settings.crud.ts#L189) it). Should we add one? Then we will be able to delete settings after the locale is deleted (just like we do in FB now).

adrians5j:
Yes, let's do it @neatbyte-vnobis . Thanks!

neatbyte-vnobis:
@adrians5j Currently, the FB/HCMS PR is focused on migrating only `forms` and `submissions`. Changes in this PR are related to FB `settings` (not sure what is current plan for FB `settings` migration to HCMS).

The FB-related changes can be included in the FB/HCMS PR, and, as you've suggested, the PB part can be moved to a separate PR.

neatbyte-vnobis:
@adrians5j Any updates on this one? Should we move PB changes to a separate PR or leave it as it is?

PR #3548, opened 2023-09-22
fix: fixed issue with border for image

neatbyte-vnobis:
## Changes
Resolve issue "Image border not applied correctly"
Task provided by @SvenAlHamad no reference to issue or Notion.

Before Changes
Знімок екрана 2023-09-22 о 16 11 32

After Changes
Знімок екрана 2023-09-22 о 16 09 56

## How Has This Been Tested?
Manually

## Documentation
None

### DIFF
--- a/packages/app-page-builder-elements/src/renderers/image.tsx
+++ b/packages/app-page-builder-elements/src/renderers/image.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useEffect } from "react";
 import styled from "@emotion/styled";
 import { createRenderer, CreateRendererOptions } from "~/createRenderer";
 import { useRenderer } from "~/hooks/useRenderer";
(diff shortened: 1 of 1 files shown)


neatbyte-ivelychko:
There are two solutions for this issue.
First of the solutions is used in this PR, and the second one I will provide in this comment.

In the second solution we would not need to use `useEffect` and loop through the styles for the element as we do in the first solution. What we need is to use
```
const borderStyles = border({ element: { ...element, type: "" }, theme });
const borderAssignedStyles = assignStyles({
   breakpoints: theme.breakpoints,
    styles: borderStyles || {}
});
const PbImg = styled.img({
  width: element.data.image.width,
  height: element.data.image.height,
  maxWidth: "100%",
  ...borderAssignedStyles
});
```
in `packages/app-page-builder-elements/src/renderers/image.tsx`.


And we should add additional if statement here `packages/app-page-builder-elements/src/modifiers/styles/border.ts`
```
if (element.type === "image") {
        return {};
}
```
And it would also work.
@adrians5j What do you think?

adrians5j:
Will need to think a bit about this. Will need to see if we can do this in a better way.

neatbyte-vnobis:
@adrians5j Any new thoughts on this PR? Or should we stick with one of the provided solutions?

PR #3550, opened 2023-09-22
feat: advanced search filter builder for HCMS 

leopuleo:
## Changes
With this PR, we introduce the new advanced search filter builder for HCMS.

The feature allows users to:
- create custom filters based on the content model fields and default fields
- group conditions and define the operation type (`AND` / `OR`)

The following default fields are included:
- `status`
- `createdOn`
- `savedOn`

The following model field types are included:
- `boolean`
- `datetime`
- `number`
- `text`

It's impossible to save a search filter configuration; this will be implemented with additional PR.

### Bonus
The PR includes some improvements to @webiny/form package to improve the error validation.

### Video
https://github.com/webiny/webiny-js/assets/2866531/a8745555-8361-4e5e-a21a-9d9d80f6f0e9

## How Has This Been Tested?

Manually + Jest


packages/app-aco/src/components/AdvancedSearch/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,47 @@
...
+import { DrawerContainer } from "./Drawer.styled";
+
+interface DrawerProps {
+    open: boolean;
+    onClose: () => void;
+    onSubmit: (data: any) => void;
+    fields: FieldRaw[];
+}
+
+export const Drawer = observer(({ open, onClose, fields, onSubmit }: DrawerProps) => {
```
Pavel910: I think you can remove the `observer` HOC, since the `Drawer` itself is never reading from an observable object.

packages/app-aco/src/components/AdvancedSearch/QueryBuilder/QueryBuilder.tsx
```diff
@@ -0,0 +1,28 @@
...
+    fields: FieldRaw[];
+    onForm: (form: FormAPI) => void;
+    onSubmit: (data: any) => void;
+}
+
+export const QueryBuilder = ({ fields, onForm, onSubmit }: QueryBuilderProps) => {
+    const [presenter, setPresenter] = useState<QueryBuilderPresenter | undefined>();
+
+    useEffect(() => {
+        if (fields) {
```
Pavel910: Is it at all possible that `fields` are undefined? Because the Props type tells me `fields` will always exist. If they do exist always, then you can remove this `useEffect` completely, and just instantiate the presenter right there in the useState, like this:

```jsx
const [presenter] = useState(() => new QueryBuilderPresenter(fields));
```

packages/app-aco/src/components/AdvancedSearch/QueryBuilder/QueryBuilder.tsx
```diff
@@ -0,0 +1,28 @@
...
+            setPresenter(new QueryBuilderPresenter(fields));
+        }
+    }, []);
+
+    if (!presenter) {
+        return null;
+    }
```
Pavel910: This also becomes redundant, if `fields` are always present.

packages/app-aco/src/components/AdvancedSearch/QueryBuilder/adapters/QueryBuilderViewModel.ts
```diff
@@ -0,0 +1,14 @@
...
+    public queryObject: QueryObjectDTO;
+    public fields: FieldDTO[];
+    public invalidFields: Record<string, { isValid: boolean; message: string }> = {};
+
+    constructor(queryObject: QueryObjectDTO, fields: FieldDTO[]) {
+        this.queryObject = queryObject;
+        this.fields = fields;
+        makeAutoObservable(this);
+    }
+}
```
Pavel910: This class is now redundant. You can simply keep the `queryObject` and `fields` in the Presenter, and assign them to the viewModel when/if necessary. This class does nothing, so you can get rid of it.

packages/app-aco/src/components/AdvancedSearch/QueryBuilder/domain/Field.ts
```diff
@@ -0,0 +1,197 @@
+import { CmsModelField } from "@webiny/app-headless-cms-common/types/model";
+
+export type FieldRaw = Pick<
+    CmsModelField,
+    "id" | "type" | "label" | "multipleValues" | "predefinedValues" | "settings"
+>;
+
+export enum TypeDTO {
```
Pavel910: I think `TypeDTO` is a misleading name. This is just an `enum`, so we can probably just call it `FieldType`?

packages/app-headless-cms/src/admin/components/ContentEntries/Filters/Filters.tsx
```diff
@@ -1,11 +1,75 @@
...
+                    type: modelField.type,
+                    label: modelField.label,
+                    multipleValues: modelField.multipleValues || false,
+                    predefinedValues: modelField?.predefinedValues || undefined,
+                    settings: modelField.settings
+                };
+            });
+
+        return [...defaultFields, ...fields];
+    }, [model.fields]);
```
Pavel910: This whole part can be extracted into:
a) a standalone hook, which lives net to the `Filters` component, just to have this outside of the component, or better...
b) create a small Presenter to handle this logic, and instantiate the Presenter with the `model` in the `useEffect` 
hook, when you have the model available

With a simple Presenter (which doesn't even have to be observable, you can use a simple callback method to get the fields) you get a better separation, and no React 😅

Pavel910 (changes requested):
Love it! So nice to review small single-purpose components. 
I left a few comments and suggestions, but this already looks better than most of our React stuff 🚀 😎

leopuleo:
Closing it in favour of #3604

PR #3554, opened 2023-09-26
fix: add support for `titleFieldId` defined inside the HCMS model

leopuleo:
## Changes
The pull request addresses the issue of sorting the `Name` column in the HCMS ACO table when the corresponding field is missing from the HCMS model.

We map the `Name` column with the `titleFieldId` from the content model but keep the table header title as it is because the column also contains the folder's title.

Closes [WEB-3212](https://webiny.atlassian.net/browse/WEB-3212)

## How Has This Been Tested?
Manually

## Documentation
Inline

### DIFF
--- a/packages/app-aco/src/contexts/acoList.tsx
+++ b/packages/app-aco/src/contexts/acoList.tsx
@@ -98,6 +98,7 @@ const getCurrentRecordList = <T = GenericSearchData,>(
 export interface AcoListProviderProps {
     children: React.ReactNode;
     own?: boolean;
+    titleFieldId: string | null;
 }
 
 export const AcoListProvider: React.VFC<AcoListProviderProps> = ({ children, ...props }) => {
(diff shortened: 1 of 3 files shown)


packages/app-aco/src/contexts/acoList.tsx
```diff
@@ -208,7 +209,11 @@ export const AcoListProvider: React.VFC<AcoListProviderProps> = ({ children, ...
      */
     useEffect(() => {
         setFolders(prev => {
-            return sortTableItems(prev, state.listSort);
+            // We might receive a different field name as `title`, here we set it back to `title` for folders.
+            const titleField = props?.titleFieldId || "title";
```
brunozoric: use the `id` field, as you did in https://github.com/webiny/webiny-js/pull/3554/files#diff-c38fb80874bd10f632b5189f7a787951b9f9bb5037802c93e0247a4d6b542a34R78

brunozoric (changes requested):

packages/app-aco/src/contexts/acoList.tsx
leopuleo: 👍 thanks Bruno, nice spot!

Pavel910 (approved):

PR #3580, opened 2023-10-06
feat: add wcp check for audit logs

neatbyte-vnobis:
## Changes
Adds ability to activate and deactivate `Audit Logs` module via `WCP`.

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/apps/admin/src/plugins/index.ts
+++ b/apps/admin/src/plugins/index.ts
@@ -2,7 +2,6 @@ import { plugins } from "@webiny/plugins";
 import pageBuilderPlugins from "./pageBuilder";
 import formBuilderPlugins from "./formBuilder";
 import headlessCmsPlugins from "./headlessCms";
-import auditLogsPlugins from "./auditLogs";
 import theme from "theme";
 
 // Imports plugins created via scaffolding utilities.
@@ -21,10 +20,6 @@ plugins.register([
      * Headless CMS app.
      */
     headlessCmsPlugins,
-    /**
-     * Audit Logs app.
-     */
-    auditLogsPlugins,
     /**
      * App theme controls page builder and form builder layouts, styles, etc.
      */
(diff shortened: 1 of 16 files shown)


packages/app-audit-logs/src/plugins/permissionRenderer/index.tsx
```diff
@@ -1,27 +1,37 @@
...
+                >
+                    <AuditLogsPermissionsComponent {...props} />
+                </AccordionItem>
+            );
+        }
+    };
 };
 
-export default plugin;
+export const AuditLogsPermissions: React.FC = () => {
```
adrians5j, line 32: Why this approach? Why not just export plugin like we have it with other?

apps/admin/src/plugins/index.ts
```diff
@@ -21,10 +20,6 @@ plugins.register([
      * Headless CMS app.
      */
     headlessCmsPlugins,
-    /**
-     * Audit Logs app.
-     */
-    auditLogsPlugins,
```
adrians5j, line 27: Why was this removed? 🤔

adrians5j (changes requested):
Left two questions.

packages/app-audit-logs/src/plugins/permissionRenderer/index.tsx
neatbyte-vnobis: I've used the same approach as APW package, that you've suggested to use as reference. [Here](https://github.com/webiny/webiny-js/blob/next/packages/app-apw/src/plugins/permissionRenderer/index.tsx#L27) is APW permission settings plugin and it is registered based on [this](https://github.com/webiny/webiny-js/blob/next/packages/app-apw/src/index.tsx#L45) condition.

apps/admin/src/plugins/index.ts
adrians5j: Question here remains.

adrians5j (changes requested):
One question remanins.

apps/admin/src/plugins/index.ts
neatbyte-vnobis: Please refer to [this](https://github.com/webiny/webiny-js/pull/3580#discussion_r1358301392) answer. Plugin is now registered inside the package based on condition.

adrians5j: Srry, thanks!

PR #3594, opened 2023-10-11
fix: fix images after node update

neatbyte-vnobis:
## Changes
After updating to `aws-sdk v3`, image transform lambdas needed some adjustments to work properly with `sharp`.

## How Has This Been Tested?
Manual

## Documentation
None


packages/api-file-manager/package.json
```diff
@@ -66,7 +66,8 @@
     ],
     "ignore": {
       "src": [
-        "sharp"
+        "sharp",
+        "stream"
```
Pavel910: Let's move the `stream` dep to a global config, since it's a node built-in library. See `.adiorc.js:24`

packages/api-file-manager/src/handlers/transform/optimizeImage.ts
```diff
@@ -3,28 +3,33 @@
  */
 // @ts-ignore
```
Pavel910: Let's get rid of this ignore, and add `"@types/sharp": "^0.32.0"` to `devDependencies`. We'll get proper autocomplete then.

Pavel910 (changes requested):
A few tiny changes to make it all cleaner, and we're done here.

brunozoric (approved):

PR #3607, opened 2023-10-18
fix: fix build after updating @types/node to v18

neatbyte-vnobis:
## Changes
Resolves issue that caused ts errors while building most of the packages.

## How Has This Been Tested?
Manual

## Documentation
None

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -6,7 +6,7 @@
     "allowSyntheticDefaultImports": true,
     "moduleResolution": "node",
     "module": "esnext",
-    "lib": ["esnext"],
+    "lib": ["esnext", "dom", "dom.iterable"],
     "esModuleInterop": true,
     "declaration": true,
     "composite": true,
--- a/package.json
+++ b/package.json
@@ -217,7 +217,6 @@
     "@emotion/react": "11.10.8",
     "@types/react": "17.0.39",
     "@types/jsdom": "link:./typings/void",
-    "@typescript/lib-dom": "npm:@types/web@*",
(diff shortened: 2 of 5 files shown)


tsconfig.build.json
```diff
@@ -6,7 +6,7 @@
     "allowSyntheticDefaultImports": true,
     "moduleResolution": "node",
     "module": "esnext",
-    "lib": ["esnext"],
+    "lib": ["esnext", "dom", "dom.iterable"],
```
Pavel910, line 9: Is it necessary to specify `dom` and `dom.iterable`? Aren't these automatically included?

If we must have this, we'll also need to update `cwp-template-aws` package. See `packages/cwp-template-aws/template/common/tsconfig.build.json`

Also, existing projects will need a project upgrade (a codemod) to apply this change, once people upgrade to 5.38.
Let's write this down as a new task, because it will be in a different repository.

Pavel910 (approved):

tsconfig.build.json
neatbyte-vnobis: It's not strictly necessary to specify `dom` and `dom.iterable` unless the `lib` option is used. When `lib` is specified, TypeScript will only include the libraries listed. So, possible solutions to fix build are:
1) Remove the `lib` option - TypeScript will revert to including the default libraries based on the `esnext` target.
2) Specify `dom` and `dom.iterable` in the `lib` option.

Pavel910: Ok so we need a codemod then. Let's keep the `lib` option, and specify those in all projects.

PR #3643, opened 2023-10-26
fix(api-headless-cms): models and group loaders

brunozoric:
## Changes
This PR adds cached model and group lists.
The model and group list method will now always return a cached promise - so it can either be running or resolved. 

## How Has This Been Tested?
Jest and manually.

### DIFF
--- a/packages/api-headless-cms/src/crud/contentModel.crud.ts
+++ b/packages/api-headless-cms/src/crud/contentModel.crud.ts
@@ -662,6 +662,8 @@ export const createModelsCrud = (params: CreateModelsCrudParams): CmsModelContex
                 );
             }
 
+            clearModelsCache();
+
             await onModelAfterDelete.publish({
                 model
             });
--- a/packages/api-headless-cms/src/crud/contentModel.crud.ts
+++ b/packages/api-headless-cms/src/crud/contentModel.crud.ts
@@ -104,10 +104,12 @@ export const createModelsCrud = (params: CreateModelsCrudParams): CmsModelContex
         })
     };
 
+    const listModelsCache = new Map<string, Promise<CmsModel[]>>();
     const clearModelsCache = (): void => {
         for (const loader of Object.values(loaders)) {
             loader.clearAll();
         }
+        listModelsCache.clear();
     };
 
     const managers = new Map<string, CmsModelManager>();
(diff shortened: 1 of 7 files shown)


packages/api-headless-cms/src/crud/contentModel.crud.ts
```diff
@@ -183,30 +185,49 @@ export const createModelsCrud = (params: CreateModelsCrudParams): CmsModelContex
...
-
-                if (!ownsModel) {
-                    return false;
-                }
+        /**
+         * Maybe we can cache based on permissions, not the identity id?
+         *
+         * TODO: @adrian please check if possible.
+         */
+        const key = JSON.stringify({
```
adrians5j, line 200: Do we really need this cache key thingie here?

I mean, for every HTTP request, there's really only one tenant / locale / identity combo right? If that's true, do we even need cache keys? Or we can just store the cache in a say `cache` variable directly and that's it?

Am I missing something here? 🤔

packages/api-headless-cms/src/crud/contentModel.crud.ts
```diff
@@ -375,7 +396,7 @@ export const createModelsCrud = (params: CreateModelsCrudParams): CmsModelContex
                 model
             });
 
-            loaders.listModels.clearAll();
+            clearModelsCache();
```
adrians5j, line 399: We should call this in the delete model method below, no?

packages/api-headless-cms/src/crud/contentModelGroup.crud.ts
```diff
@@ -191,37 +188,52 @@ export const createModelGroupsCrud = (params: CreateModelGroupsCrudParams): CmsG
...
-                locale: locale || getLocale().code
-            }
+        /**
+         * Maybe we can cache based on permissions, not the identity id?
+         *
+         * TODO: @adrian please check if possible.
+         */
+        const key = JSON.stringify({
+            tenant: tenant || getTenant().id,
+            locale: locale || getLocale().code,
```
adrians5j, line 206: The same question re cache keys.

packages/api-headless-cms/src/crud/contentModel.crud.ts
brunozoric: Just ask @Pavel910 :)
Or you can think of how to fetch all models from a few diffrent tenant/locale combinations in a single request without having this kind of cache key.

adrians5j (approved):

PR #3645, opened 2023-10-27
Form Builder - Move backend to HCMS

neatbyte-vnobis:
## Changes
Form Methods moved to HCMS:
- [x] createForm
- [x] getForm
- [x] listFormRevisions
- [x] listForms
- [x] createFormFrom
- [x] deleteForm
- [x] deleteFormRevision
- [x] publishForm
- [x] unpublishForm
- [x] updateForm

Submission Methods moved to HCMS:
- [x] listSubmissions
- [x] createSubmission
- [x] updateSubmission

Skipped Submission Methods: `deleteSubmission`, `getSubmission`. They were skipped because they are not used anywhere.

## How Has This Been Tested?
Manually.

## Documentation
[Form Builder - Move backend to HCMS](https://webiny.notion.site/Form-Builder-Move-backend-to-HCMS-1f2cc9b528b544aba56263380eacbab3)


packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormBuilderStorage.ts
```diff
@@ -0,0 +1,96 @@
...
+                    entryId: id
+                }
+            });
+        });
+
+        return [entries.map(entry => this.getFormFieldValues(entry)), meta];
+    };
+
+    listForms = async (params: FormBuilderStorageOperationsListFormsParams) => {
+        const tenant = params.where.tenant;
```
neatbyte-ivelychko: Arrow functions: `createForm`, `getForm`, `listFormRevisions` will be changed to a regular functions after I will find out the reason why the context of `this` is broken.

P.S. Fixed issue with `this` now using regular functions instead of arrow functions.

packages/api-form-builder/src/plugins/crud/index.ts
```diff
@@ -12,97 +11,96 @@ export interface CreateFormBuilderCrudParams {
     storageOperations: FormBuilderStorageOperations;
 }
 
-export default (params: CreateFormBuilderCrudParams) => {
-    const { storageOperations } = params;
+export const setupFormBuilderContext = async (params: any) => {
+    const { storageOperations, context } = params;
```
neatbyte-ivelychko, line 17: All the `any` types will be fixed.

P.S. Fixed `any` types

packages/api-headless-cms/src/crud/contentEntry.crud.ts
```diff
@@ -189,7 +189,7 @@ interface DeleteEntryParams {
 const createEntryId = (input: CreateCmsEntryInput) => {
     let entryId = mdbid();
     if (input.id) {
-        if (input.id.match(/^([a-zA-Z0-9])([a-zA-Z0-9\-]+)([a-zA-Z0-9])$/) === null) {
+        if (input.id.match(/^([a-zA-Z0-9])([a-zA-Z0-9\-#]+)([a-zA-Z0-9])$/) === null) {
             throw new WebinyError(
```
neatbyte-ivelychko: We have added `#` because `createEntryId` did not support ids with `#`, and in our case form id contains `#`.

neatbyte-ivelychko:
@adrians5j, @SvenAlHamad what should we do about these bugs?

While working on this feature we have found bug that exists for a long time.

1. When we have a form that has ONLY one revision and after deleting that revision we would see that the form that should have been deleted still being displayed in UI in the list of the forms. But after reloading the page it would not be rendered in the list of forms. I think that happens becuase of caching. Furthemore this bug doesn not appear if we delete a form instead of revision.

https://github.com/webiny/webiny-js/assets/129154672/1f0a2d5e-e62f-4e6d-98b1-df3d60713a28

2. When have created a form submission, opened `Submissions` tab in Form Builder and clicked on `reload` submissions we would see that the list of submissions has been updated. 
But `Overview` stats on the `Submissions` tab are still the same, so they has not been updated. To see a most recent data for `Overview` we would need to reload the page.
Знімок екрана 2023-11-08 о 17 17 50

3. When we choose a form revision that we had chosen previously then that revision won't be re-rendered in the `Page Builder Editor Form` and it would render the latest chosen revision.

https://github.com/webiny/webiny-js/assets/129154672/061493ce-7b39-4f45-9828-c49c7b6f4f40

packages/api-form-builder/src/cmsFormBuilderStorage/creteModelField.ts
leopuleo, line 1: typo in file name: `creteModelField.ts`

packages/app-form-builder/src/components/Form/FormRender.tsx
```diff
@@ -161,6 +161,11 @@ const FormRender: React.FC<FbFormRenderComponentProps> = props => {
         fields.forEach(field => {
             const fieldId = field.fieldId;
 
+            // Remove after form model fields id fix
```
leopuleo: Please add a more explanatory comment.

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+            version,
+            published,
+            latest,
+            tenant,
+            locale
+        } = params.where;
+        const model = this.modelWithContext({ tenant, locale });
+        const formId = initialFormId || id?.split("#").shift() || "";
+
+        const entry = await this.security.withoutAuthorization(async () => {
```
leopuleo, line 59: Missing a check here, have a look at the current implementation: 

- [ddb only](https://github.com/webiny/webiny-js/blob/next/packages/api-form-builder-so-ddb/src/operations/form/index.ts#L307)
- [ddb + es](https://github.com/webiny/webiny-js/blob/next/packages/api-form-builder-so-ddb-es/src/operations/form/index.ts#L347)

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+            id,
+            formId: initialFormId,
+            version,
+            published,
+            latest,
+            tenant,
+            locale
+        } = params.where;
+        const model = this.modelWithContext({ tenant, locale });
+        const formId = initialFormId || id?.split("#").shift() || "";
```
leopuleo: Use [parseIdentifier](https://github.com/webiny/webiny-js/blob/next/packages/utils/src/parseIdentifier.ts) helper method

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+
+        const entry = await this.security.withoutAuthorization(async () => {
+            if (latest) {
+                const [entries] = await this.cms.listLatestEntries(model, {
```
leopuleo: Here, you can use a cleaner way:

`await cms.getEntry(model, { where: { entryId: formId, latest: true } });`

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -411,21 +360,19 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
...
                     form,
                     original
                 });
-                const result = await this.storageOperations.updateForm({
-                    input: data,
+                const result = await this.storageOperations.forms.updateForm({
                     form,
-                    original
+                    input,
+                    meta: {},
```
leopuleo: We don't need to pass `input`, `meta` and `options`. In the storage operation, use the entire `form` as input to `cms.updateEntry`.

packages/api-form-builder/src/types.ts
```diff
@@ -619,9 +617,10 @@ export interface FormBuilderStorageOperationsCreateFormFromParams {
  * @category StorageOperationsParams
  */
 export interface FormBuilderStorageOperationsUpdateFormParams {
-    input?: Record<string, any>;
-    original: FbForm;
     form: FbForm;
+    input: Record<string, any>;
+    meta?: Record<string, any>;
```
leopuleo: See my comment here above; `meta` and `options` are optional and passed as empty objects. Remove them.

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+                    views: 0
+                }
+            });
+        });
+
+        return this.getFormFieldValues(entry);
+    }
+
+    async updateForm(params: FormBuilderStorageOperationsUpdateFormParams): Promise<FbForm> {
+        const { form, input, meta, options } = params;
```
leopuleo: You find a related comment here above (`forms.crud.ts`) regarding the parameters passed to this method.

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+    ): Promise<FormBuilderStorageOperationsListFormsResponse> {
+        const { id, tenant, locale, ...restWhere } = params.where;
+        const model = this.modelWithContext({ tenant, locale });
+
+        const [entries, meta] = await this.security.withoutAuthorization(async () => {
+            return await this.cms.listLatestEntries(model, {
+                after: params.after,
+                limit: params.limit,
+                sort: params.sort,
+                where: { entryId: id, ...restWhere }
```
leopuleo: It looks like `id` is never assigned. Please remove it from here and the `FormBuilderStorageOperationsListFormsParams`.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -633,12 +575,10 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
             const original = await this.getForm(id, {
                 auth: false
             });
-
             const originalFormFormId = original.formId || (original.id.split("#").pop() as string);
```
leopuleo: Use [parseIdentifier](https://github.com/webiny/webiny-js/blob/next/packages/utils/src/parseIdentifier.ts) helper method

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -763,9 +701,9 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
             };
 
             try {
-                await this.storageOperations.updateForm({
-                    original,
-                    form
+                await this.storageOperations.forms.updateForm({
```
leopuleo: See my comment here above

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -540,11 +483,11 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
         },
         async publishForm(this: FormBuilder, id) {
             await formsPermissions.ensure({ rwd: "r", pw: "p" });
-
+            const [pid, revisionNumber = "0001"] = id.split("#");
```
leopuleo: Use [parseIdentifier](https://github.com/webiny/webiny-js/blob/next/packages/utils/src/parseIdentifier.ts) helper method

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -367,10 +319,7 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
                 await onFormBeforeCreate.publish({
                     form
                 });
-                const result = await this.storageOperations.createForm({
-                    input,
-                    form
-                });
+                const result = await this.storageOperations.forms.createForm({ form, input });
```
leopuleo: Here, we are passing both `form` and `input`, but only `form` has been used by the related storage operation method. Remove `input` from the interface.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -391,7 +340,7 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
             await updateData.validate();
             const data = await updateData.toJSON({ onlyDirty: true });
```
leopuleo: This comment is valid for all methods: validation should be performed by HCMS model definition. We don't need to validate it here. 

I expect to see no manual validation performed for any HCMS-powered model.

Delete these files:
- [forms.models.ts](https://github.com/webiny/webiny-js/blob/d512ba4cd0128d85e2c9f3904efe37b2a2dc671c/packages/api-form-builder/src/plugins/crud/forms.models.ts)
- [submissions.models.ts](https://github.com/webiny/webiny-js/blob/d512ba4cd0128d85e2c9f3904efe37b2a2dc671c/packages/api-form-builder/src/plugins/crud/submissions.models.ts)

Also, we want to get rid of commodo and use [Zod](https://github.com/colinhacks/zod) instead. We are already using it in the system: you will find many examples.

Could you please refactor [settings.models.ts](https://github.com/webiny/webiny-js/blob/ace6efe9930f365d0d5b436f872cc863dd3dd528/packages/api-form-builder/src/plugins/crud/settings.models.ts) to use Zod?

packages/api-form-builder/src/plugins/crud/forms.crud.ts
leopuleo: On line 313 we are using `FormSettingsModel` method from [forms.models.ts](https://github.com/webiny/webiny-js/blob/d512ba4cd0128d85e2c9f3904efe37b2a2dc671c/packages/api-form-builder/src/plugins/crud/forms.models.ts). 

I expect this entire file to be deprecated. 

If you need to create default values, just create a method that returns those. In the future , we want to rely to default values defined into the content model.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -493,17 +440,17 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
 
             const formFormId = form.formId || form.id.split("#").pop();
```
leopuleo: Use [parseIdentifier](https://github.com/webiny/webiny-js/blob/next/packages/utils/src/parseIdentifier.ts) helper method

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -493,17 +440,17 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
...
 
-            const revisions = await this.storageOperations.listFormRevisions({
+            const revisions = await this.storageOperations.forms.listFormRevisions({
                 where: {
-                    formId: formFormId,
+                    formId: formFormId || "",
```
leopuleo: `this.storageOperations.forms.listFormRevisions` expect a valid `formId`, not an empty string.

This should be solved by using `parseIdentifier` (see previous comment).

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+            });
+        });
+
+        return [entries.map(entry => this.getFormFieldValues(entry)), meta];
+    }
+
+    async listFormRevisions(
+        params: FormBuilderStorageOperationsListFormRevisionsParams
+    ): Promise<FbForm[]> {
+        const { tenant, locale, formId } = params.where;
```
leopuleo, line 165: We are passing `sort` from `getFormRevisions`, but you are not using it in the query. Please add sort capabilities to both `getFormRevisions` query in GraphQL and implement a test for that.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -554,20 +497,19 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
...
             try {
                 await onFormBeforePublish.publish({
                     form
                 });
-                const result = await this.storageOperations.publishForm({
+                const result = await this.storageOperations.forms.publishForm({
                     original,
-                    form
+                    form,
+                    input: form
```
leopuleo: We don't need to pass `input` and `original`. In the storage operation, use the entire `form` as input to `cms.updateEntry`.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -599,17 +541,17 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
...
             try {
                 await onFormBeforeUnpublish.publish({
                     form
                 });
-                const result = await this.storageOperations.unpublishForm({
+                const result = await this.storageOperations.forms.unpublishForm({
                     original,
-                    form
+                    form,
+                    input: form
```
leopuleo: We don't need to pass `input` and `original`. In the storage operation, use the entire `form` as input to `cms.updateEntry`.

packages/api-form-builder/src/plugins/crud/submissions.crud.ts
```diff
@@ -382,7 +384,7 @@ export const createSubmissionsCrud = (params: CreateSubmissionsCrudParams): Subm
             const data = await new models.FormSubmissionUpdateDataModel().populate(input);
             data.validate();
```
leopuleo: Same here: validations should be handled by HCMS model definition.

packages/api-form-builder/src/index.ts
```diff
@@ -1,24 +1,21 @@
...
+import { FormBuilderStorageOperations, FormBuilderContext } from "~/types";
+import { FormsPermissions } from "./plugins/crud/permissions/FormsPermissions";
 
 export interface CreateFormBuilderParams {
     storageOperations: FormBuilderStorageOperations;
+    formsPermissions: FormsPermissions;
+    context: FormBuilderContext;
 }
 
 export const createFormBuilder = (params: CreateFormBuilderParams) => {
```
leopuleo: As mentioned above, `createFormBuilder` needs to return plugins created by both `createFormBuilderContext` and `createFormBuilderGraphQL` called, including: 

- triggerHandlers
- validators
- formBuilderPrerenderingPlugins

apps/api/graphql/src/index.ts
```diff
@@ -22,9 +22,12 @@ import {
     createFileModelModifier
 } from "@webiny/api-file-manager";
 import { createFileManagerStorageOperations } from "@webiny/api-file-manager-ddb";
+import {
```
leopuleo: Please, let's keep the original import  `createFormBuilder` to avoid user's project upgrade.

apps/api/pageBuilder/export/combine/src/index.ts
```diff
@@ -40,17 +45,30 @@ export const handler = createHandler({
...
+            })
+        }),
         createPageBuilderContext({
             storageOperations: createPageBuilderStorageOperations({
                 documentClient
             })
         }),
         createPageBuilderGraphQL(),
-        createFormBuilder({
+        createFormBuilderContext({
```
leopuleo: Keep the original import `createFormBuilder`

apps/api/pageBuilder/export/process/src/index.ts
```diff
@@ -61,7 +61,7 @@ export const handler = createHandler({
                 documentClient
             })
         }),
-        createFormBuilder({
+        createFormBuilderContext({
```
leopuleo: Keep the original import `createFormBuilder`

apps/api/pageBuilder/import/create/src/index.ts
```diff
@@ -38,17 +43,30 @@ export const handler = createHandler({
...
+            })
+        }),
         createPageBuilderContext({
             storageOperations: createPageBuilderStorageOperations({
                 documentClient
             })
         }),
         createPageBuilderGraphQL(),
-        createFormBuilder({
+        createFormBuilderContext({
```
leopuleo: Keep the original import `createFormBuilder`

apps/api/pageBuilder/import/process/src/index.ts
```diff
@@ -64,7 +64,7 @@ export const handler = createHandler({
                 documentClient
             })
         }),
-        createFormBuilder({
+        createFormBuilderContext({
```
leopuleo: Keep the original import `createFormBuilder`

packages/api-form-builder-so-ddb-es/src/index.ts
leopuleo, line 1: Export everything from the `api-form-builder-so-ddb` because we only have `settings` and `system` which are working only with dynamo-db.

The rest of the package (`src` and `tests`) can be removed. 

Actually, this will be a "facade" package.

packages/api-form-builder-so-ddb/src/operations/form/index.ts
leopuleo, line 1: Let's remove unused types definition and methods like `createFormPartitionKey`

packages/api-form-builder-so-ddb/src/operations/submission/index.ts
leopuleo, line 1: Let's remove unused types definition and methods like `createSubmissionPartitionKey`

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+            if (latest) {
+                const [entries] = await this.cms.listLatestEntries(model, {
+                    where: { entryId: formId }
+                });
+
+                return entries[0];
+            } else if (published && !version) {
+                const entries = (await this.cms.getEntryRevisions(model, formId))
+                    .filter(entryItem => entryItem.values.published)
+                    .sort((a, b) => b.version - a.version);
```
leopuleo: Please, extract this in a private method and give it a clear name, like `getSortedFormRevisions`

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
```diff
@@ -0,0 +1,224 @@
...
+
+                return entries[0];
+            } else if (id || version) {
+                return await this.cms.getEntryById(
+                    model,
+                    id ||
+                        createIdentifier({
+                            id: formId as string,
+                            version: version as number
+                        })
```
leopuleo: Please, remove casting. 

`version` can be undefined but it's required in `createIdentifier`: add a fallback value.

The same can be achieved in a more readable manner, like:

```ts
              const fallbackId = createIdentifier({
                  id: formId,
                  version: version || 1
              });

              return await this.cms.getEntryById(model, id || fallbackId);
```

packages/api-form-builder/src/cmsFormBuilderStorage/CmsSubmissionsStorage.ts
```diff
@@ -0,0 +1,128 @@
...
+        const model = this.modelWithContext(where);
+
+        return await this.security.withoutAuthorization(async () => {
+            const entry = await this.cms.getEntry(model, {
+                where: { entryId: where.id, latest: true }
+            });
+
+            return this.getSubmissionValues(entry);
+        });
+    }
```
leopuleo: We just reviewed this in-depth, you can safely remove this method. Thanks for adding it in the first place.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -540,11 +483,11 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
         },
         async publishForm(this: FormBuilder, id) {
             await formsPermissions.ensure({ rwd: "r", pw: "p" });
-
+            const [pid, revisionNumber = "0001"] = id.split("#");
             /**
              * getForm checks for existence of the form.
              */
-            const original = await this.getForm(id, {
+            const original = await this.getForm(`${pid}#${revisionNumber}`, {
```
leopuleo: Use [createIdentifier](https://github.com/webiny/webiny-js/blob/next/packages/utils/src/createIdentifier.ts) helper

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -730,9 +668,9 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
             };
 
             try {
-                await this.storageOperations.updateForm({
-                    original,
-                    form
+                await this.storageOperations.forms.updateForm({
+                    form,
+                    input: form
```
leopuleo: Check my comment here above.

packages/api-form-builder/src/plugins/crud/index.ts
```diff
@@ -10,99 +9,113 @@ import { SettingsPermissions } from "~/plugins/crud/permissions/SettingsPermissi
...
+        }
+    } catch (ex) {
+        throw new WebinyError(
+            ex.message || "Could not run before init in Form Builder storage operations.",
+            ex.code || "STORAGE_OPERATIONS_BEFORE_INIT_ERROR",
+            {
+                ...ex
+            }
+        );
+    }
```
leopuleo: We don't need `init` and  `beforeInit` anymore, remove them from the interface and the storage operations.

packages/api-form-builder/src/plugins/crud/index.ts
```diff
@@ -10,99 +9,113 @@ import { SettingsPermissions } from "~/plugins/crud/permissions/SettingsPermissi
...
+        }
+    } catch (ex) {
+        throw new WebinyError(
+            ex.message || "Could not run init in Form Builder storage operations.",
+            ex.code || "STORAGE_OPERATIONS_INIT_ERROR",
+            {
+                ...ex
+            }
+        );
+    }
```
leopuleo: See my comment above

packages/api-form-builder/src/plugins/graphql/formsSchema.ts
leopuleo, line 1: Let's improve the code here by adding types to resolvers arguments. Types shouldn't be the same as the crud operation but designed explicitly for the resolver.

packages/api-form-builder/src/plugins/graphql/submissionsSchema.ts
leopuleo, line 1: Same here: let's improve the code here by adding types to resolvers arguments. Types shouldn't be the same as the crud operation but designed explicitly for the resolver.

packages/api-form-builder/src/types.ts
```diff
@@ -567,7 +566,7 @@ export interface FormBuilderStorageOperationsListFormsParams {
  */
 export interface FormBuilderStorageOperationsListFormRevisionsParamsWhere {
     id?: string;
```
leopuleo: It looks like the `id` is never used, is it right? If so, please remove it.

packages/api-form-builder/src/types.ts
```diff
@@ -637,14 +636,6 @@ export interface FormBuilderStorageOperationsDeleteFormParams {
...
 export interface FormBuilderStorageOperationsDeleteFormRevisionParams {
-    /**
-     * Method always receives all the revisions of given form ordered by version_DESC.
-     */
-    revisions: FbForm[];
-    /**
-     * Previous revision of the current form. Always the first lesser available version.
-     */
-    previous: FbForm | null;
     form: FbForm;
```
leopuleo, line 618: 🙏

packages/api-headless-cms/src/crud/contentEntry.crud.ts
leopuleo: Why are you sending `#` into `createEntryId`? 

In general. `entryId` does not contain `#`. Only revision does:

Entry is the entire record with all of its revisions and has an `entryId`.
A revision is a version of an entry containing the `#0001` part.


- entryId -> `abcdefghi`
- revisionId -> `abcdefghi#0001`

packages/api-form-builder/src/cmsFormBuilderStorage/models/form.model.ts
leopuleo, line 1: Please, review how the fields are defined: some field types have been changed, some required fields are optional now and vice-versa.

Use API playground in the admin app to confront the newly generated schema with the previous one.

packages/api-form-builder/src/cmsFormBuilderStorage/models/submission.model.ts
leopuleo, line 1: Check my comment here above.

leopuleo (changes requested):
Thanks @neatbyte-vnobis and @neatbyte-ivelychko for this Pull Request. 

You will find many comments below; most are repetitive, and many others address tiny issues.

There are a couple of issues we need to address before merging:

- we want to keep the original imported method `createFormBuilder`. TLDR: we don't want to upgrade some part of user projects.
- we want to keep the GraphQL schema unchanged: some field types have been changed, some of them are not required anymore and vice-versa. Use the API playground in the admin app to confront the newly generated schema with the previous one.

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormsStorage.ts
neatbyte-ivelychko: @leopuleo you mean that we have to use `sort` in the `this.cms.getEntryRevisions` (method from the api-headless-cms)?
But as I can see method `this.cms.getEntryRevisions` does not support `sort`. 
Should we add `sort` prop support to the `this.cms.getEntryRevisions`?

packages/api-form-builder/src/types.ts
neatbyte-ivelychko: @leopuleo You want me to return these lines?

packages/api-form-builder/src/plugins/graphql/formsSchema.ts
neatbyte-ivelychko: @leopuleo 
While writing these schemas: `formsSchemas` and `submissionsSchema` we were looking at the example from [`filesSchema`](https://github.com/webiny/webiny-js/blob/next/packages/api-file-manager/src/graphql/filesSchema.ts#L41). And as we can see we do not have types for resolvers arguments, they have `any` type.
That's because `args` have `Record<string, any>` type.
Here is that part of code [`Resolvers<TContext>`](https://github.com/webiny/webiny-js/blob/next/packages/handler-graphql/src/types.ts#L37).
I was trying to look whether we have type for `args` in [`api-aco`](https://github.com/webiny/webiny-js/blob/next/packages/api-aco/src/folder/folder.gql.ts#L149) and again we do not have types there only `any`.

Should we modify [`Resolvers<TContext>`](https://github.com/webiny/webiny-js/blob/next/packages/handler-graphql/src/types.ts#L37) to achive what want or we gonna leave it as is?

packages/api-form-builder/src/types.ts
leopuleo: @neatbyte-ivelychko it's ok. Thanks for removing these 2 params

packages/api-form-builder/src/plugins/graphql/formsSchema.ts
neatbyte-ivelychko: @leopuleo what do you think about this?

packages/api-form-builder/src/plugins/crud/forms.crud.ts
neatbyte-ivelychko: @leopuleo @Pavel910

While moving [settings.models.ts](https://github.com/webiny/webiny-js/blob/ace6efe9930f365d0d5b436f872cc863dd3dd528/packages/api-form-builder/src/plugins/crud/settings.models.ts) from `commodo` to `Zod` I saw that tests started to fail. 
I think that's happening because of the `Zod` model, it does not return full `reCapthca` object on method `createSettings` instead it returns `reCapthca: null` but the test expect `this` object
```
reCaptcha: {
   enabled: null,
   siteKey: null,
   secretKey: null
}

```
Moreover the test `formInstallion.test.ts` is failing because of the issue with validation of `Zod` models.
Знімок екрана 2023-11-28 о 12 32 09

Current models for settings look like:
```
export const CreateDataModel = () => {
    return zod.object({
        domain: zod.string(),
        reCaptcha: zod.object({
            enabled: zod.boolean(),
            siteKey: zod.string().max(100),
            secretKey: zod.string().max(100)
        })
    });
};

export const UpdateDataModel = () => {
    return zod.object({
        domain: zod.string(),
        reCaptcha: zod.object({
            enabled: zod.boolean(),
            siteKey: zod.string().max(100),
            secretKey: zod.string().max(100)
        })
    });
};
```

Maybe something wrong with my models?

Pavel910: @neatbyte-ivelychko Let's tweak this a bit. Set `enabled` to have a default value of `false`. `siteKey` and `secretKey` are irrelevant if `enabled: false`. With that, you can update the test to only expect the `enabled` key to be set.

neatbyte-ivelychko: Okay, I will do test with `enabled: false`.

Pavel910: You can use the `.default(false)` to set default values right there on the schema: https://zod.dev/?id=default

neatbyte-ivelychko: I have fixed this issue with validation by adding a `.deepPartial()` to the `Zod` models. After doing that `formInstallion.test.ts` started working again, but we have started getting the `reCaptcha: null` responce in the `settings.test.ts` file instead of getting this expected responce:

```
reCaptcha: {
   enabled: null,
   siteKey: null,
   secretKey: null
}
```

And I want to know whether it is okay, because we need to change expected responces in the `settings.test.ts` file. 
And in the `API` we would change `Settings` type in `api-form-builder` because `values` from `reCaptcha` object could be `undefined` after a `.deepPartial()` change in the model.

@leopuleo @Pavel910

Pavel910: @neatbyte-ivelychko you know what.... let's do this in a safer way. Let's keep the data as it was before, to reduce the risk of breaking stuff on the UI side. So for now, just add default values into the zod schema for all 3 attributes. maybe you'll also need to set a default value for the whole `reCaptcha` attribute. But in any case, let's not use `deepPartial` because that basically allows anything, it's almost like `any` in typescript. I would rather have `null` but have actual object structure respected.

neatbyte-ivelychko:
@leopuleo I have fixed issue with settings zod model

packages/api-form-builder-so-ddb/src/operations/submission/index.ts
```diff
@@ -1,267 +1,34 @@
...
 
     return {
         createSubmission,
         deleteSubmission,
         updateSubmission,
-        listSubmissions,
-        getSubmission,
-        createSubmissionPartitionKey,
-        createSubmissionSortKey
+        listSubmissions
```
leopuleo, line 32: `getSubmission` is missing here.

packages/cwp-template-aws/template/ddb-es/apps/api/graphql/src/index.ts
```diff
@@ -98,8 +98,7 @@ export const handler = createHandler({
         }),
         createFormBuilder({
             storageOperations: createFormBuilderStorageOperations({
-                documentClient,
-                elasticsearch: elasticsearchClient
+                documentClient
```
leopuleo: Can we avoid to remove the `elasticsearch` dependency? We don't want to upgrade user project files.

packages/api-headless-cms/src/crud/contentEntry.crud.ts
```diff
@@ -826,7 +826,7 @@ export const createContentEntryCrud = (params: CreateContentEntryCrudParams): Cm
             version: nextVersion,
             savedOn: getDate(rawInput.savedOn, currentDate),
             createdOn: getDate(rawInput.createdOn, currentDate),
-            publishedOn: getDate(rawInput.publishedOn, originalEntry.publishedOn),
+            publishedOn: getDate(rawInput.publishedOn, undefined),
```
leopuleo: This part should not change. Am I missing something?

packages/api-form-builder/src/plugins/graphql/formsSchema.ts
leopuleo: We can leave it as it is right now, we will come back later to refactor it.

leopuleo (changes requested):
We are almost there; please review the comments in the code I just left.

Also, I noticed some of the GraphQL schemas generated have been changed. Below, you will find some notes:

```graphql
// FbForm
{
  createdBy: FbFormUser!
  createdOn: DateTime!
  fields: [FbForm_Fields!] -> The list should be required
  formId: String -> Should be required
  id: ID!
  locked: Boolean -> Should be required
  name: String -> Should be required
  overallStats: FbForm_OverallStats -> Should be required
  ownedBy: FbFormUser!
  published: Boolean -> Should be required
  publishedOn: DateTime
  savedOn: DateTime!
  settings: FbForm_Settings -> Should be required
  slug: String -> Should be required
  stats: FbForm_Stats -> Should be required
  status: String -> Should be required / Should be an enum
  steps: [FbForm_Steps!] -> The list should be required
  triggers: JSON
  version: Number!
}

// FbForm_Fields 
{
  _id: String -> Should be required
  fieldId: String -> Should be required
  type: String -> Should be required
  name: String -> Should be required
  label: String
  placeholderText: String
  helpText: String
  options: [FbForm_Fields_Options!] -> Should NOT be required
  validation: [FbForm_Fields_Validation!] -> Should NOT be required
  settings: JSON
}

// FbForm_Fields_Validation 
{
  name: String -> Should be required
  message: String
  settings: JSON
}

// FbForm_Steps 
{
  title: String
  layout: [JSON] -> The original was [[String]]
}
```

packages/api-headless-cms/src/crud/contentEntry.crud.ts
neatbyte-vnobis: Shouldn't a new entry revision created from a published entry have the `publishedOn` date unset, unless specifically passed through `rawInput`?

Previously, without this change, creating a new entry revision from a published entry resulted in the new revision inheriting the `publishedOn` date from the original entry, despite its `draft` status.

If this behavior is indeed expected, I can propose some solutions. One option is to add a parameter to the `createEntryRevisionFrom` function, which would specify a list of properties to omit during cloning. Alternatively, we could add a callback to remap properties and selectively remove certain ones.

brunozoric: We want to keep the originalEntry publishedOn value. For example:
* You created an article on 2010-11-30.
* You publish an article on 2010-12-01.
* 13 years later you find a typo in the article
* You create a new revision of the article on 2023-12-01 and fix the typo
* You publish that new revision of the article on 2023-12-01

Article was not published on 2023-12-01 but 2010-12-01 - we want to keep the original publishedOn date, otherwise this updated and published revision of the article would be on top of the last published list on the website.

neatbyte-vnobis:
> Also, I noticed some of the GraphQL schemas generated have been changed.

@leopuleo  I've investigated this a little bit and found that a similar issue exists in `FileManager` too. Fields marked as required in the file model are not required in the schema. The problem lies in the [renderFields](https://github.com/webiny/webiny-js/blob/next/packages/api-headless-cms/src/utils/renderFields.ts) function. This function utilizes `createTypeField`, which is not influenced by the `required` value of a field.

packages/api-form-builder-so-ddb/src/operations/submission/index.ts
neatbyte-vnobis: `getSubmissions` was removed from `CmsSubmissionsStorage` as per [this](https://github.com/webiny/webiny-js/pull/3645#discussion_r1392618771) comment.
`CmsSubmissionsStorage` and `createSubmissionStorageOperations` currently using the same interface `FormBuilderSubmissionStorageOperations` from the `api-form-builder` package. Given this, should we consider creating a separate interface for `CmsSubmissionsStorage` that excludes `getSubmissions`?

packages/api-headless-cms/src/crud/contentEntry.crud.ts
neatbyte-vnobis: It seems that the `publishedOn` property of the form revision is currently not being used in either the `api-form-builder` or the `app-form-builder`, making it unclear what exactly it suppose to represent.

Should it operate as in your example, representing the date when the first revision was published, or should it indicate the date when a particular revision was published? Based on the form tests, it appears that the `publishedOn` value of a form revision should be `null` if we create a new revision from a published one.

If we need the form revision's `publishedOn` to remain as it was previously, we might consider using another property name since the current one conflicts with that of the entry, or we could implement one of the previously suggested solutions.

packages/api-form-builder-so-ddb/src/operations/submission/index.ts
leopuleo: It's ok, thanks for noticing it. Leave it without `getSubmission`.

leopuleo:
> > Also, I noticed some of the GraphQL schemas generated have been changed.
> 
> @leopuleo I've investigated this a little bit and found that a similar issue exists in `FileManager` too. Fields marked as required in the file model are not required in the schema. The problem lies in the [renderFields](https://github.com/webiny/webiny-js/blob/next/packages/api-headless-cms/src/utils/renderFields.ts) function. This function utilizes `createTypeField`, which is not influenced by the `required` value of a field.

@neatbyte-vnobis thanks for investigating. Actually, marking a field as required does not change the generated GraphQL schema: this is happening also for UI-managed content models.

Similar behaviour happens for predefined values. So, skip my comments mentioning `Should be required` and `Should be an enum`.

Could you please have a look at the other comments?

neatbyte-vnobis:
> > > Also, I noticed some of the GraphQL schemas generated have been changed.
> > 
> > 
> > @leopuleo I've investigated this a little bit and found that a similar issue exists in `FileManager` too. Fields marked as required in the file model are not required in the schema. The problem lies in the [renderFields](https://github.com/webiny/webiny-js/blob/next/packages/api-headless-cms/src/utils/renderFields.ts) function. This function utilizes `createTypeField`, which is not influenced by the `required` value of a field.
> 
> @neatbyte-vnobis thanks for investigating. Actually, marking a field as required does not change the generated GraphQL schema: this is happening also for UI-managed content models.
> 
> Similar behaviour happens for predefined values. So, skip my comments mentioning `Should be required` and `Should be an enum`.
> 
> Could you please have a look at the other comments?

@leopuleo regarding the `layout` property in the form steps, which was originally of type `[[String]]`, the current implementation of `createTypeField` does not support two-dimensional arrays.

The possible solutions I see are: adding a separate type for this value, modifying `createTypeField` to support two-dimensional arrays, or use the `[JSON]` type as it is currently implemented (which can be simplified to just `JSON`).

leopuleo:
> > > > Also, I noticed some of the GraphQL schemas generated have been changed.
> > > 
> > > 
> > > @leopuleo I've investigated this a little bit and found that a similar issue exists in `FileManager` too. Fields marked as required in the file model are not required in the schema. The problem lies in the [renderFields](https://github.com/webiny/webiny-js/blob/next/packages/api-headless-cms/src/utils/renderFields.ts) function. This function utilizes `createTypeField`, which is not influenced by the `required` value of a field.
> > 
> > 
> > @neatbyte-vnobis thanks for investigating. Actually, marking a field as required does not change the generated GraphQL schema: this is happening also for UI-managed content models.
> > Similar behaviour happens for predefined values. So, skip my comments mentioning `Should be required` and `Should be an enum`.
> > Could you please have a look at the other comments?
> 
> @leopuleo regarding the `layout` property in the form steps, which was originally of type `[[String]]`, the current implementation of `createTypeField` does not support two-dimensional arrays.
> 
> The possible solutions I see are: adding a separate type for this value, modifying `createTypeField` to support two-dimensional arrays, or use the `[JSON]` type as it is currently implemented (which can be simplified to just `JSON`).

@neatbyte-vnobis thanks for the explanation. Let's stick to the current implementation `[JSON]`

packages/api-headless-cms/src/crud/contentEntry.crud.ts
adrians5j: Just got pinged by @Pavel910 about this thread here.

I suggest we don't do anything here @neatbyte-vnobis, because this is currently being reworked in my [PR](https://github.com/webiny/webiny-js/blob/a10d00f09c7ed245b57253a2533117c8a4a22e19/packages/api-headless-cms/src/crud/contentEntry.crud.ts#L876-L930).

We'll basically have revision and entry-level sets of these meta fields. For revision-level, yes, the published-related fields will be unset. But, for entry-level, fields values will be propagated throughout all of the revisions.

So, let's undo this change, and I guess down the road we'll see how we'll go about merging all of these changes. I'm mostly interested in figuring out how we'll get your PR up-to-date with the changes in my PR. I mean, everything will still work because of backwards compatibility, but yeah, we'd definitely want to use the new fields. 

Maybe you can just continue as is, using the old meta fields, and then, in a separate PR later, we do the transition (once both this and my PRs are merged)? 🤔

neatbyte-vnobis:
@leopuleo Is this PR ready to be merged?

neatbyte-vnobis:
@adrians5j @leopuleo As we're removing the revision selection for forms in the Page Builder, I assume we should now only display forms with a published revision in the form select. Should we create a new `listPublishedForms` API endpoint for this, or modify the existing `listForms` with an additional parameter? I think a separate endpoint would be preferable, as it would utilize `listPublishedEntries` instead of `listLatestEntries`.

Additionally, considering we now have only one published revision, should we rename `getLatestPublishedFormRevision` to `getPublishedFormRevision`?

neatbyte-vnobis:
> @adrians5j @leopuleo As we're removing the revision selection for forms in the Page Builder, I assume we should now only display forms with a published revision in the form select. Should we create a new `listPublishedForms` API endpoint for this, or modify the existing `listForms` with an additional parameter? I think a separate endpoint would be preferable, as it would utilize `listPublishedEntries` instead of `listLatestEntries`.
> 
> Additionally, considering we now have only one published revision, should we rename `getLatestPublishedFormRevision` to `getPublishedFormRevision`?

@adrians5j @leopuleo Any thoughts on this?

neatbyte-vnobis:
@adrians5j While updating tests following the recent status updates, I noticed that [this](https://github.com/webiny/webiny-js/blob/next/packages/api-form-builder/__tests__/forms.test.ts#L249) test fails. The test involves publishing a form revision and then incrementing its views. However, with our current use of published HCMS entries, we cannot update them. This is because the current implementation of HCMS includes a check to see if an entry is `locked` before updating, and it is set to `true` when an entry is published. So we cannot update already published form - which is a breaking change compared to old API and FB logic.

What should our approach be in this case? Could we consider making some changes to HCMS to address this issue? (option to override the `locked` status or something like this)

neatbyte-vnobis:
> @adrians5j While updating tests following the recent status updates, I noticed that [this](https://github.com/webiny/webiny-js/blob/next/packages/api-form-builder/__tests__/forms.test.ts#L249) test fails. The test involves publishing a form revision and then incrementing its views. However, with our current use of published HCMS entries, we cannot update them. This is because the current implementation of HCMS includes a check to see if an entry is `locked` before updating, and it is set to `true` when an entry is published. So we cannot update already published form - which is a breaking change compared to old API and FB logic.
> 
> What should our approach be in this case? Could we consider making some changes to HCMS to address this issue? (option to override the `locked` status or something like this)

@adrians5j @Pavel910  Any thoughts on this?

adrians5j:
We're still discussing this internally. We'll get back to you re this soon.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -341,7 +336,7 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
 
             if (!original) {
                 throw new NotFoundError(`Form "${id}" was not found!`);
-            } else if (original.locked) {
+            } else if (original.status === "unpublished") {
```
leopuleo: Let's create a `FORM_STATUS` enum: you can import this [enum](https://github.com/webiny/webiny-js/blob/725f6732b7061e72eec8b6a3edfe3cfabb8639ae/packages/api-headless-cms/src/types.ts#L2933) from `api-headless-cms` package and re-export it.


i.e. 

```
if (original.status === CONTENT_ENTRY_STATUS.UNPUBLISHED) {
  ...
}
```

packages/api-form-builder/src/plugins/graphql/submissionsSchema.ts
```diff
@@ -69,7 +69,7 @@ export const createSubmissionsSchema = (params: CreateSubmissionsTypeDefsParams)
                          * Get all revisions of the form.
                          */
                         const revisions = await formBuilder.getFormRevisions(form);
-                        const publishedRevisions = revisions.filter(r => r.published);
+                        const publishedRevisions = revisions.filter(r => r.status === "published");
```
leopuleo: Let's create a `FORM_STATUS` enum: you can import this [enum](https://github.com/webiny/webiny-js/blob/725f6732b7061e72eec8b6a3edfe3cfabb8639ae/packages/api-headless-cms/src/types.ts#L2933) from `api-headless-cms` package and re-export it.

packages/app-form-builder/src/admin/plugins/editor/defaultBar/Name/Name.tsx
```diff
@@ -92,7 +92,7 @@ export const Name = () => {
         <NameWrapper>
             <FormMeta>
                 <Typography use={"overline"}>{`status: ${
-                    state.data.published ? t`published` : t`draft`
+                    state.data.status === "published" ? t`published` : t`draft`
```
leopuleo: Let's create an enum with the possible status.

packages/app-form-builder/src/admin/plugins/formDetails/formRevisions/Revision.tsx
```diff
@@ -37,7 +37,7 @@ const revisionsMenu = css({
 
 const getIcon = (revision: Pick<FbFormModel, "status">) => {
     switch (revision.status) {
-        case "locked":
+        case "unpublished":
```
leopuleo: Same here, let's use an enum with the possible status.

packages/app-form-builder/src/admin/views/Forms/FormsDataList.tsx
```diff
@@ -138,7 +138,7 @@ const FormsDataList = (props: FormsDataListProps) => {
         const handlerKey = form.id + form.status;
         if (!editHandlers.current[handlerKey]) {
             editHandlers.current[handlerKey] = async () => {
-                if (!form.published) {
+                if (form.status !== "published") {
```
leopuleo: Same here, let's use an enum with the possible status.

leopuleo (changes requested):
@neatbyte-vnobis, we are almost there 💪. 

I highlighted some minor changes, plus while testing the admin app, I noticed a couple of bugs: 

1. **App Form Builder:** While trying to create a new version from a previous one, it always keeps the configuration from the latest version instead of the selected one.
2. **App Page Builder**: The application crashes while trying to insert a form on a page. It looks like an error on the `getPublishedForm` query. See the video here below:

https://github.com/webiny/webiny-js/assets/2866531/37555ac7-e90f-4ce1-b54c-9847765c7574

leopuleo:
At the moment, this PR is on hold because of an issue with using HCMS storage operation for the submission entity.

When there is a large number of submissions stored in the database (10,000 submissions), attempting to list even a small subset (20 submissions) results in a `504 Gateway Timeout` error.

This is an example of submission record: 

```
{
 "PK": "T#root#L#en-US#CMS#CME#CME#65c668267f15950008b1cb98",
 "SK": "L",
 "createdOn": "2024-02-09T18:00:06.537Z",
 "entryId": "65c668267f15950008b1cb98",
 "GSI1_PK": "T#root#L#en-US#CMS#CME#M#fbSubmission#L",
 "GSI1_SK": "65c668267f15950008b1cb98#0001",
 "id": "65c668267f15950008b1cb98",
 "locale": "en-US",
 "location": {
  "folderId": "root"
 },
 "locked": false,
 "modelId": "fbSubmission",
 "status": "draft",
 "tenant": "root",
 "TYPE": "cms.entry.l",
 "values": {
  "json@data": {
   "demoField": "Submission from Angie Hyatt"
  },
  "json@logs": [
  ],
  "object@form": {
   "object@fields": [
    {
     "json@settings": {
      "defaultValue": null
     },
     "object@options": [
     ],
     "object@validation": [
     ],
     "text@fieldId": "demoField",
     "text@helpText": null,
     "text@label": "Demo field",
     "text@name": "text",
     "text@placeholderText": null,
     "text@type": "text",
     "text@_id": "Ef86jNVJIZ"
    }
   ],
   "object@steps": [
    {
     "json@layout": [
      [
       "Ef86jNVJIZ"
      ]
     ],
     "text@title": "Step 1"
    }
   ],
   "text@id": "65c650d188458c00081fea0f#0001",
   "text@name": "Form 2",
   "text@parent": "65c650d188458c00081fea0f",
   "text@version": 1
  },
  "object@meta": {
   "datetime@submittedOn": "Sun Apr 16 2023 09:22:22 GMT+0200 (Central European Summer Time)",
   "object@url": {
    "json@query": {},
    "text@location": "https://d8t7yvqn5xojf.cloudfront.net/untitled-lseu8t15"
   },
   "text@ip": "200.22.178.28"
  }
 },
 "version": 1,
 "webinyVersion": "5.39.0",
 "_ct": "2024-02-10T15:10:40.797Z",
 "_et": "CmsEntries",
 "_md": "2024-02-10T15:10:40.797Z"
}
```

[Link to private Slack conversation](https://webiny.slack.com/archives/CUSPFP9TR/p1707721551874629)

PR #3658, opened 2023-11-01
Cypress Tests - Batch 2, 3, and 4

Asmaloth:
## Changes

Closes #ISSUE_NUMBER


cypress-tests/cypress/e2e/admin/pageBuilder/templates/pageTemplatesCrud.cy.ts
```diff
@@ -0,0 +1,65 @@
+import { customAlphabet } from "nanoid";
+
+context("Page Builder - Blocks", () => {
+    const nanoid = customAlphabet("abcdefghijklmnopqrstuvwxyz");
```
Pavel910, line 4: @Asmaloth @adrians5j we have these utils for ID generation neatly packed in our `@webiny/utils` package. We use these extensively in our Jest tests, so please use that, instead of going for raw `nanoid`. Have a look at how they're implemented: https://github.com/webiny/webiny-js/blob/next/packages/utils/src/generateId.ts

For your needs here, you would just use this:

```
import { generateAlphaLowerCaseId } from "@webiny/utils";

generateAlphaLowerCaseId(6);
```

cypress-tests/cypress/e2e/admin/pageBuilder/templates/pageTemplatesExportImport.cy.ts
Pavel910, line 1: Same comment applies to all yout test files here: let's not use `nanoid` directly, because I see dozens of usages scattered across the code. In a few days, `nanoid` will publish a broken version, or a security issue, and we'll need to replace the usage of that library, and you'll have to edit hundreds of occurrences. To avoid that, use our own package.

adrians5j: We agreed we'll resolve this in a separate PR. Thanks @Pavel910!

PR #3664, opened 2023-11-01
fix(api-aco): flp entry decorator

brunozoric:
## Changes
Fix for CMS Entry CRUD decorator:
* not all variables are passed into the original method
* move entry method not implemented
* create entry revision method not implemented
* listing entries with correct folder types

## How Has This Been Tested?
Jest and manually.


adrians5j (approved):
Thank you @brunozoric !

leopuleo (approved):

PR #3668, opened 2023-11-03
fix: re-initialize filter repository on namespace change

leopuleo:
## Changes
We addressed an issue that occurred while navigating through various HCMS entry lists. 

The issue was caused by the `FilterRepository` not being re-initialized upon a change in `namespace`. 

To fix this, we implemented a check to determine if any new `namespace` or `client` was received and used `useMemo` to memoize the returned value.

## How Has This Been Tested?
Jest


brunozoric (approved):
Looks good.

Pavel910 (approved):
Great! Nothing to add here :)

PR #3685, opened 2023-11-09
fix: advanced search filter boolean files

leopuleo:
## Changes
With this PR we fixed a bug encountered while trying to save a filter with a **boolean** field: 

- handling and saving the filter field `value` as a string - as expected by `FilterDTO`
- converting it back to `boolean` when returned back via `GraphQLInputMapper`

### How to reproduce the bug
https://github.com/webiny/webiny-js/assets/2866531/3c54606f-cdab-4aac-afd4-dce27f39d724

## How Has This Been Tested?
Jest + manually


packages/app-aco/src/components/AdvancedSearch/GraphQLInputMapper.ts
```diff
@@ -9,10 +9,22 @@ export class GraphQLInputMapper {
...
+    private static convertToBooleanOrString(value: string | boolean): string | boolean {
+        if (value === "true") {
+            return true;
+        }
+
+        if (value === "false") {
+            return false;
+        }
+
+        return value || "";
```
brunozoric: If the value is false (boolean) you will get empty string as a result.

leopuleo: Thanks @brunozoric, for pointing this out. I changed it into the nullish coalescing operator.

brunozoric (approved):

PR #3699, opened 2023-11-14
fix(api-security): load permissions immediately after authentication

Pavel910:
## Changes
This PR changes *WHEN* permissions are loaded for current identity. We want to have permissions loaded as soon as possible, *BEFORE* GraphQL resolvers kick in.

## How Has This Been Tested?
Jest tests.


brunozoric (approved):

Pavel910:
This PR is outdated and will be replaced with a different one.

PR #3701, opened 2023-11-14
refactor: change Icon Picker architecture

neatbyte-vnobis:
## Changes
Refactor Icon Picker using "Fast Test Architecture".

Example of definition:

#### IconPackProvider

```
<IconPickerConfig>
    <IconPickerConfig.IconPack
        name={"emoji-icon-pack"}
        provider={() => [{type: "emoji", name: "thumbs_up", value: "👍"}]}
    />
</IconPickerConfig>
```

#### IconType
```
const EmojiIcon = () => {
    const { icon } = useIcon<Icon>();

    return <div>{icon.value}</div>;
};
```

```
const EmojiIconTab = observer(() => {
    const presenter = useIconPicker();
    
    const onIconSelect = (icon: Icon) => {
        presenter.setIcon(icon);
        presenter.closeMenu();
    };
    
    return (
        <IconPickerTab
            label={"Emojis"}
            onChange={onIconSelect}
        />
    );
});
```

```
<IconPickerConfig.IconType name={"emoji"}>
    <IconPickerConfig.IconType.Icon element={<EmojiIcon />} />
    <IconPickerConfig.IconType.Tab element={<EmojiIconTab />} />
</IconPickerConfig.IconType>
```

## How Has This Been Tested?
Manual

## Documentation
None


packages/app-admin/src/components/IconPicker/IconPicker.tsx
```diff
@@ -1,391 +1,38 @@
...
-type TabContentProps = {
-    icons: Icon[];
-    type: string;
-    value: Icon;
-    onChange: (value: Icon, closeMenu?: boolean) => void;
-    filter: string;
-    onFilterChange: DelayedOnChangeProps["onChange"];
-    refetchCustomIcons?: () => void;
-    isLoading?: boolean;
+    return <IconPickerComponent repository={repository} {...props} />;
```
Pavel910: This place is idea to act as a `Composition Root`. It's a place where you construct all dependencies that are required to setup a Presenter. Then you just pass the `presenter` through props to your UI component. Here's an example:

```tsx
// This is your Composition Root. Ideally, you will inject a complete Presenter into the final UI component.
const IconPickerInner = (props: IconPickerProps) => {
    const config = useIconPickerConfig();
    const repository = iconRepositoryFactory.getRepository(config);

    const presenter = useMemo(() => {
        return new IconPickerPresenter(repository);
    }, [repository]);

    return <IconPickerComponent presenter={presenter} {...props} />;
};
```

This way your UI component only knows about the `presenter`, as it should. This is the IDEAL scenario you should always strive for.

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,146 @@
...
+                >
+                    {({ closeMenu }) => (
+                        <>
+                            <MenuHeader>
+                                <Typography use={"body1"}>Select an icon</Typography>
+                                <CloseIcon onClick={() => closeMenu()} />
+                            </MenuHeader>
+                            <MenuContent>
+                                {presenter.vm.isLoading && <CircularProgress />}
+                                <Tabs ref={tabsRef}>
```
Pavel910: `<Tabs>` component supports a `value` prop, which designates the active tab, and also the `onActivate` prop, which is called when you switch tabs. This way you can avoid using the `ref` and imperative API to switch tabs.

```tsx
<Tabs value={activeTab} onActivate={setActiveTab}>
```

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+        return (
+            <IconPickerWrapper>
+                {label && (
+                    <div className={iconPickerLabel}>
+                        <Typography use={"body1"}>{label}</Typography>
+                    </div>
+                )}
+
+                <Menu
```
Pavel910: Use the `onClose` prop, to reset the active tab when the Menu is closed. The callback would look something like this:

```ts
    const resetActiveTab = useCallback(() => {
        setActiveTab(value ? getActiveTab(value.type) : 0);
    }, [value?.type]);
    ```

File: packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx [hunk 4]
```diff
@@ -0,0 +1,146 @@
...
+
+export const IconPickerComponent = observer(
+    ({ repository, label, description, ...props }: IconPickerComponentProps) => {
+        const { isValid: validationIsValid, message: validationMessage } = props.validation || {};
+
+        const tabsRef = useRef<TabsImperativeApi>();
+
+        const presenter = useMemo(() => {
+            return new IconPickerPresenter(repository);
+        }, [repository]);
```
Pavel910: Following the comment I did in the previous files, the moment you inject the `presenter` through props, this part of code can be removed, and you'll have less `import` statements at the top, because Repository will no longer be a dependency at all.

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+        useEffect(() => {
+            presenter.load(props.value);
+        }, [repository, props.value]);
```
Pavel910: Here you'll be able to remove the `repository` from dependencies.

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+        useEffect(() => {
+            if (props.onChange) {
+                props.onChange(presenter.vm.selectedIcon);
+            }
+        }, [presenter.vm.selectedIcon]);
+
+        const iconPickerPlugins = plugins.byType<IconPickerPlugin>("admin-icon-picker");
```
Pavel910: In general, this type of things should all be moved to the Repository. You want to resolve plugins in the Repository (or Presenter, if your component doens't have a Repository), and transform these things into a shape that can be consumed by the Presenter. Using `plugins` in the UI is a code smell; it's simply wrong, because it makes it very difficult to add new data sources (for example configs, API, etc.).

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                <CloseIcon onClick={() => closeMenu()} />
+                            </MenuHeader>
+                            <MenuContent>
+                                {presenter.vm.isLoading && <CircularProgress />}
+                                <Tabs ref={tabsRef}>
+                                    {/* Order of tabs defined here: `packages/app-admin/components/IconPicker/plugins/index.ts`. */}
+                                    {iconPickerPlugins.map(plugin => {
+                                        const iconsByType = presenter.vm.data.find(
+                                            iconsByType => iconsByType.type === plugin.iconType
+                                        );
```
Pavel910: This should be calculated in the Presenter, and provided to the UI component through `vm`.

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,146 @@
...
+
+                                        if (!iconsByType) {
+                                            return;
+                                        }
```
Pavel910: Once the `iconsByType` are precalculated in `vm`, this `if` will become unnecessary, because you will already know which tabs should exist (you'll have an array of 1,2, or 3 items)

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                                    closeMenu();
+                                                }
+                                            },
+                                            filter: presenter.vm.filter,
+                                            onFilterChange: value => presenter.filterIcons(value),
+                                            color: presenter.vm.color,
+                                            onColorChange: color => presenter.setColor(color),
+                                            checkSkinToneSupport: icon =>
+                                                presenter.checkSkinToneSupport(icon)
+                                        });
```
Pavel910: I would like to avoid using our old `plugins` concept, and use more of Config + Composition approach. To add support for different renderers of tabs, we need to add a new component to the `IconPickerConfig`, and here you'll be able to just use your `<IconPickerTab>`. This component will then be composed via the `<IconPickerConfig.IconPickerTab decorator={myTabRenderer}/>`.

Let me know when you start working on this, and I'll provide some references, we already do similar stuff in Lexical, Filters, etc.

packages/app-admin/src/components/IconPicker/domain/IconRepository.ts
Pavel910, line 1: Feel free to put all the classes in the root of the `IconPicker` component. The `domain` folder is not required, and it will be easier to navigate and browse files. We haven't ourselves figured out the folder structure yet, so for the sake of simplicity, just put it all into the root folder.

If you have multiple React components, that's what you might want to put in a `components` subfolder.

packages/app-admin/src/components/IconPicker/domain/IconRepositoryFactory.ts
```diff
@@ -0,0 +1,21 @@
...
+import { IconPackProviderInterface as IconPackProvider } from "~/components/IconPicker/config";
+
+class IconRepositoryFactory {
+    private cache: Map<string, IconRepository> = new Map();
+
+    getRepository(iconPackProviders: IconPackProvider[]) {
+        const cacheKey = iconPackProviders
+            .map(provider => provider.name)
+            .sort()
+            .join("#");
```
Pavel910: Extract cache key generation into a private method, it will make the code a lot more readable:

```ts
class IconRepositoryFactory {
    private cache: Map<string, IconRepository> = new Map();

    getRepository(iconPackProviders: IconPackProvider[]) {
        const cacheKey = this.getCacheKey(iconPackProviders);

        if (!this.cache.has(cacheKey)) {
            this.cache.set(cacheKey, new IconRepository(iconPackProviders));
        }

        return this.cache.get(cacheKey) as IconRepository;
    }

    private getCacheKey(iconPackProviders: IconPackProvider[]) {
        return iconPackProviders
            .map(provider => provider.name)
            .sort()
            .join("#");
    }
}
```

packages/app-admin/src/components/IconPicker/useIconRepository.ts
Pavel910, line 1: I think this whole file is not needed, since this repository doesn't have a wider use case. It's just to power the IconPicker component, which we setup in the `IconPickerInner`. Maybe for the time being, remove it to reduce the clutter.

packages/app-admin/src/components/IconPicker/IconRenderer.tsx
Pavel910, line 1: Once we implement the composable renderer I mentioned in the previous comments, we'll be able to use it here as well. That way plugins will no longer be needed, and this logic of find what to render will not be copy/pasted across components.

packages/app-admin/src/components/IconPicker/config/IconPackProvider.tsx
```diff
@@ -1,64 +1,46 @@
...
-        }
+export const IconPackProvider = ({ name, provider }: IconPackProviderProps) => {
+    const getId = useIdGenerator("iconPackProvider");
 
-        setIcons(iconsData);
-        setIsInitialized(true);
-        setIsLoading(false);
+    const load = useCallback(async () => {
+        return await provider();
     }, [provider]);
```
Pavel910: Not sure why you need this, since the `provider` is already a closure, its reference will never change, and so it is safe to just pass to the `<Property>`.

Pavel910 (changes requested):

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
neatbyte-vnobis: Is [this](https://github.com/webiny/webiny-js/blob/next/packages/app-file-manager/src/modules/FileManagerRenderer/index.tsx#L21) the reference you mentioned? Please let me know if this is what you meant or if there's anything else I should look at.

neatbyte-vnobis:
@Pavel910

I've implemented the `emoji` tab following the `icon` tab implementation. However, there's an issue with the `presenter` value inside the plugins tabs components ([EmojiTab](https://github.com/webiny/webiny-js/blob/neatbyte/icon-picker-refactor/packages/app-admin/src/components/IconPicker/plugins/emojisPlugin.tsx#L117), [IconTab](https://github.com/webiny/webiny-js/blob/neatbyte/icon-picker-refactor/packages/app-admin/src/components/IconPicker/plugins/iconsPlugin.tsx#L57)). It appears that they are only rendered once, and the `presenter` value from `useIconPicker` is assigned just once as well.

This is very noticable on the `emoji` tab, as `SkinToneSelect` does not receive the actual `selectedIcon` value.

On the `icon` tab it causes the first color change not being applied (the state changes, cells get rerendered via `cellDecorator`, but the actual icon value in the `presenter` is not changed). All subsequent color changes work, as the color state gets updated, triggering a rerender of `IconTab`, and the `presenter` is assigned a new value (although `selectedIcon` is still one state off, it's not as noticeable for `IconColorPicker` as it is for `SkinToneSelect`).

Pavel910:
@neatbyte-vnobis I'll clone your branch and have a look a bit later.

Pavel910:
@neatbyte-vnobis Check out my last commit:
-  I removed the `color` from the Presenter, it's an implementation detail of the `icon` type, so it's the wrong place for that state.
- in the `emoji` type, I simplified the check for `hasSkinToneSupport`: you already have that information in the `seletedIcon`, so no need to dig through icons; this also eliminates the need for `getIcons()` on the Presenter.
- in general, Presenter should never be _asked_ to give information (means, no getters should exist on the Presenter). Presenter provides data to the View (React component), via the `vm`. The only methods on the Presenter that are allowed, are the commands (setSomething, addSomething, updateSomething)
- the reason why the Tabs were not updated is because the `presenter` in the context is an `observable` and to make those components react to the changes, you ned to wrap those components into a `observer()` HOC. Basically, the rule is, every component that is reading data from an observable object, needs to be an observer. https://mobx.js.org/react-integration.html

I also left some comments in the code itself, so please see the commit.

packages/app-admin/src/components/IconPicker/IconRepository.ts
```diff
@@ -2,47 +2,72 @@ import cloneDeep from "lodash/cloneDeep";
...
-            this.icons = icons.filter(icon => iconTypes.includes(icon.type));
-        });
+            runInAction(() => {
+                this.customIcons = customIcons.map(customIcon => ({
+                    type: "custom",
+                    name: customIcon.name,
+                    value: customIcon.src
+                }));
+            });
+        }
```
Pavel910: This whole thing should not exist in the Repository. It should be coming from an IconPackProvider. So `custom` icons are nothing more than an API call to fetch files from the File Manager. Repository already works with icon providers, so there's nothing for you to do in this file. And adding new dependencies, especially something like `Custom` should immediately ring the alarm.

packages/app-admin/src/components/IconPicker/IconRepositoryFactory.ts
Pavel910, line 1: The moment you move custom icons loading to the IconPack provider, this whole file can be reverted to its previous state. Core of the system shouldn't know about your custom loading logic.

packages/app-admin/src/components/IconPicker/plugins/customPlugin.tsx
```diff
@@ -1,19 +1,100 @@
...
+            label={"Custom"}
+            onChange={onIconSelect}
+            actions={<IconFilePicker onChange={onIconFileSelect} onUpload={onIconFileUpload} />}
+        />
+    );
+});
+
+export const CustomIconPlugin = () => {
+    return (
+        <IconPickerConfig>
```
Pavel910, line 99: Just add an IconPack provider config here to load your custom icons.

Pavel910 (changes requested):

Pavel910 (commented):
Just tested your latest commit: works nice. And I believe you can see for yourself how nicely the plugins work now, without any changes to the core 👍

neatbyte-vnobis:
@Pavel910

I've completed all the todos listed in [this PR](https://github.com/webiny/webiny-js/pull/3720), except for the last one. I'am currently waiting for @leopuleo input on that particular item.

If everything else looks good, maybe we can merge this PR, so @SvenAlHamad can test if the UI works as expected.

packages/app-admin/src/components/IconPicker/IconPickerComponent.tsx
```diff
@@ -0,0 +1,127 @@
...
+                                        className={placeholderIcon}
+                                    />
+                                )}
+                            </IconPickerInput>
+                        }
+                        onClose={closeMenu}
+                        onOpen={() => {
+                            openMenu();
+                            resetActiveTab();
+                        }}
```
Pavel910: Let's polish this even more: we can completely remove the `resetActiveTab` logic from the UI, and move it into the Presenter. Since you're only calling `resetActiveTab` in the `onOpen` handler, you can move that logic into the Presenter to be executed whenever a menu is opened. With that, you can delete the `getActiveTab`, as it becomes redundant.

In general, as a rule of thumb, you want to avoid getter methods on Presenters. If you need to "get" something, use a VM to contain that data in the first place.

packages/app-admin/src/components/IconPicker/IconPickerPresenter.test.ts
Pavel910, line 1: Love it! 🥇

packages/app-admin/src/components/IconPicker/plugins/customPlugin.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { css } from "emotion";
+
+import { ButtonSecondary } from "@webiny/ui/Button";
+
+import { FileManager, FileManagerFileItem } from "~/base/ui/FileManager";
+import { IconPickerTab } from "../IconPickerTab";
+import { useIcon } from "..";
+import { useIconPicker } from "../IconPickerPresenterProvider";
+import { IconPickerConfig } from "../config";
+import { ListCustomIconsQueryResponse, LIST_CUSTOM_ICONS } from "../config/graphql";
```
Pavel910: Please move the graphql file right next to the `customPlugin` since it is only used by the custom plugin and is not part of the core.

Pavel910 (changes requested):
Great progress, we're almost done. I left just a few more suggestions. I will also let @leopuleo and you polish out the exports of hooks.

Basically, the issue we're trying to solve is the flood of hooks being exported. Imagine if dozens of components had 5 hooks each, and you export all of them from the root of the package: there will sooner or later be name collisions. To export things in a meaningful way, we want to expose hooks and components not as top-level exports, but as properties of the component and its config. Hope this clarifies the goal.

neatbyte-vnobis:
@Pavel910 Is this PR ready to be merged?

Pavel910:
@neatbyte-vnobis we still have components and hooks exports to finish. Do you have the necessary information from @leopuleo ?

neatbyte-vnobis:
> @neatbyte-vnobis we still have components and hooks exports to finish. Do you have the necessary information from @leopuleo ?

I've already made the following export updates:

1. Exported at the package root [here](https://github.com/webiny/webiny-js/pull/3701/files#diff-1d132e85e18a0b7b1e6904c1d21b263fcdbc56292dd861a4a16f246dc1b1ecb3R44).
2. Exported IconPickerTab like [this](https://github.com/webiny/webiny-js/pull/3701/files#diff-95f5dd4cde2fbc8ad1716c4421b3c0c2bacd78c8912db6e1b57807f29c4d306eR38).
3. Exported hooks [here](https://github.com/webiny/webiny-js/pull/3701/files#diff-d74878c695d6c5d3f8c82f8250d3a7d4064522475b9507e88899c9f18b1d096eR3-R5).

@leopuleo, based on these updates, do you think it is done, or is there something I might have missed?

neatbyte-vnobis:
@leopuleo, please take a look at last question.
@Pavel910, @adrians5j, other than that, this one is ready for review.

PR #3715, opened 2023-11-21
feat: add bulk edit to File Manager

leopuleo:
## Changes
With this PR, we can bulk edit file custom registered fields.

To work, the field should be registered as  `bulkEdit: true` as follows:

```ts
modifier.addField({
  id: "customLabel",
  fieldId: "customLabel",
  label: "Custom Label",
  type: "text",
  renderer: {
    name: "text-input"
  },
  bulkEdit: true <-- Available for bulk edit actions
});
```

By clicking on the action, the user will have the possibility to select the fields to edit and:

- Override existing data
- Remove existing data
- Append new data (available for fields that accept `multipleValues`)

The file data is updated using the default bulk actions worker. 
CleanShot 2023-11-21 at 14 32 57

## How Has This Been Tested?
Jest + Manually

### DIFF
--- a/packages/api-file-manager/src/modelModifier/CmsModelModifier.ts
+++ b/packages/api-file-manager/src/modelModifier/CmsModelModifier.ts
@@ -4,7 +4,7 @@ import { FILE_MODEL_ID } from "~/cmsFileStorage/file.model";
 import { createModelField } from "~/cmsFileStorage/createModelField";
 import { CmsPrivateModelFull } from "@webiny/api-headless-cms";
 
-type CmsModelField = Omit<BaseModelField, "storageId">;
+type CmsModelField = Omit<BaseModelField, "storageId"> & { bulkEdit?: boolean };
 
 class CmsModelFieldsModifier {
     private fields: BaseModelField[];
@@ -14,8 +14,11 @@ class CmsModelFieldsModifier {
     }
 
     addField(field: CmsModelField) {
+        const { bulkEdit, tags, ...rest } = field;
+
         this.fields.push({
-            ...field,
+            ...rest,
+            tags: (tags ?? []).concat(bulkEdit ? ["field:bulk-edit"] : []),
             storageId: `${field.type}@${field.id}`
         });
     }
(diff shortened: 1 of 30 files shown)


packages/api-file-manager/src/modelModifier/CmsModelModifier.ts
```diff
@@ -14,8 +14,11 @@ class CmsModelFieldsModifier {
     }
 
     addField(field: CmsModelField) {
+        const { bulkEdit, tags, ...rest } = field;
+
         this.fields.push({
-            ...field,
+            ...rest,
+            tags: (tags ?? []).concat(bulkEdit ? ["field:bulk-edit"] : []),
```
Pavel910: Not sure about the `field:` prefix. These tags already are "field tags". Maybe let's rename `field:` to `internal:` or `system:`, to designate them as something Webiny handles, and it hsouldn't be changed.

Or, simply adding a `$` prefix, like this `$bulk-edit`? `$` usually means something is a system value.

packages/app-file-manager/src/components/BulkActions/ActionEdit/ActionEdit.tsx
```diff
@@ -0,0 +1,134 @@
...
+                worker.resetItems();
+
+                showResultsDialog({
+                    results: worker.results,
+                    title: "Edit files",
+                    message: "Finished editing files! See full report below:"
+                });
+            }
+        });
+    };
```
Pavel910, line 98: Just a comment on this: I think it would be great to create a generic `DialogPresenter` class which can be reused across other presenter implementations, to manage thee dialog interactions. That way we'll be able to remove this huge bloat of coad from the components. Something to try.

Pavel910 (changes requested):

packages/api-file-manager/src/modelModifier/CmsModelModifier.ts
leopuleo: Thanks for the suggestion. I opted for `$bulk-edit`

Pavel910 (approved):
Nice work! Let's merge this! 🚀 🎸

PR #3718, opened 2023-11-22
fix(api-headless-cms): dates in entry input

brunozoric:
## Changes
* Add possibility for users to send createdOn, savedOn and publishedOn into the create, createFrom and update entry GraphQL methods.
* Add possibility for users to send createdBy, modifiedBy and ownedBy into the create, createFrom and update entry GraphQL methods.
* Add possibility for users to send createdOn and savedOn into the create and update file GraphQL methods.
* Add possibility for users to send createdBy, modifiedBy into the create and update file GraphQL methods.

When publishing an entry, user can choose to skip updating the publishedOn and savedOn value - only via direct API calls, not the UI.

## How Has This Been Tested?
Jest and manually.


Pavel910 (commented):
Dropped you some comments on Slack, I'm posting this review so it remembers what I've viewed so far :)

Pavel910 (approved):

PR #3737, opened 2023-12-01
Rules for Condition Groups

neatbyte-vnobis:
## Changes
Added rules for condition groups.

## How Has This Been Tested?
Manually.

## Documentation
None.


SvenAlHamad:
On this one, I've got two main points of feedback: 

1. **UI**: We've recently delivered a new feature called[ Advanced Search](https://www.webiny.com/docs/release-notes/5.38.0/changelog#introducing-advanced-search-filter-builder-3604). I would like to use the look and feel of that accordion (styling, bg color) and the arrangement of the fields and conditions as well as the match all/any option. I think it will make the UI more uniformed across the apps. 

2. **Submit & redirect**: While talking to one of our customers about the conditional steps they brought forward a project they are currently working on where they need to send users to different pages depending on which step the user submitted. At the moment under Rules we have a dropdown with options: "Got to step" and "Submit", I would like to add here a 3rd option: Submit & Redirect. This would display an input where the user can define the url -> It should work the same way as the "Redirect" trigger that we have. The submit & redirect will take priority over the generic redirect trigger.

packages/app-page-builder-elements/src/renderers/form/types.ts
```diff
@@ -87,6 +105,8 @@ export type FormLayoutComponentProps<T = any> = {
     getDefaultValues: () => { [key: string]: any };
     goToNextStep: () => void;
     goToPreviousStep: () => void;
+    validateStepConditions: (formData: Record<string, any>, stepIndex: number) => void;
```
adrians5j, line 113: Let's fix `Record<string, any>`.

packages/app-form-builder/src/components/Form/functions/getNextStepIndex.ts
```diff
@@ -0,0 +1,218 @@
+import includes from "lodash/includes";
+import endsWith from "lodash/endsWith";
+import startsWith from "lodash/startsWith";
+import eq from "lodash/eq";
+import lte from "lodash/lte";
```
adrians5j: Let's do all of this without `lodash`. 

This is more important for packages/app-page-builder-elements/src/renderers/form/FormRender/functions/getNextStepIndex.ts`, because `app-page-builder-elements` is a pkg that's used on a public-facing website, and here we want to be as cheap as possible in terms of extra dependencies.

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/Styled.ts
```diff
@@ -46,6 +119,21 @@ export const Row = styled("div")({
     overflowX: "auto"
 });
 
+export const conditionGroupContainer = css({
```
adrians5j: Let's not use `css` if possible. Let's rely on `@emotion/styled` (I think I mentioned this in `onDrop` refactor PR).

packages/app-page-builder-elements/src/renderers/form/FormRender/functions/getNextStepIndex.ts
```diff
@@ -0,0 +1,218 @@
...
+    }
+
+    if (!equal) {
+        return lt(filterValue, filterValue);
+    } else {
+        return lte(fieldValue, filterValue);
+    }
+};
+
+const timeGtValidator = ({
```
adrians5j, line 81: Can we use `@webiny/validation` lib here? I see a lot of validators here that the lib already has.

adrians5j (changes requested):

packages/app-form-builder/src/admin/components/FormEditor/Context/functions/handleDeleteConditionGroup.ts
```diff
@@ -0,0 +1,55 @@
...
+    data: FbFormModel;
+    formStep: FbFormStep;
+    stepFields: FbFormModelField[];
+    conditionGroup: FbFormModelField;
+    conditionGroupFields: FbFormModelField[];
+}
+/*
+    When we delete condition group we also need to delete fields inside of it,
+    because those fields belong directly (they are being stored in the setting of the condition group) to the condition group and not the step.
+*/
```
neatbyte-ivelychko: I will change format of this comments from `/**/` to `//`

packages/app-form-builder/src/admin/components/FormEditor/Context/functions/handleMoveRow.ts
```diff
@@ -9,7 +10,13 @@ interface HandleMoveRowParams {
...
     destinationRow: number;
 }
-
+/*
+    The difference between moving row between steps, step and condition group or between two condition groups:
+    * When we move row between steps we are going to change property "layout" of those steps.
+    * When we move row between step and condition group we are going to change property "layout" of step and the property "layout" of the Condition Group.
+    * When we move row between condition groups we are going to change property "layout" of those condition groups and we don't need information about steps in which those,
+      those condition groups are being stored, because we are not affecting layout of steps in this case.
+*/
```
neatbyte-ivelychko: Will remove extra `those`

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/Styled.ts
```diff
@@ -12,7 +18,74 @@ export const StyledAccordionItem = styled(AccordionItem)`
...
+    display: flex;
+    align-items: center;
+    margin-top: 5px;
+    & > span {
+        margin-left: 5px;
+        font-size: 14px;
+    }
+`;
+
+export const ConditionSetupWrapper = styled.div``;
```
adrians5j: Can we dbl check this, do we need this?

packages/app-form-builder/src/components/Form/functions/getNextStepIndex.ts
adrians5j: This comment is still pending.

packages/app-page-builder-elements/src/renderers/form/FormRender/functions/getNextStepIndex.ts
adrians5j: This comment is still pending.

adrians5j: Also, as mentioned in one of my prev comments, let's try avoiding lodash here. The less deps `app-page-builder-elements` pkg has, the better (because it's a pkg that's used on the public website).

packages/app-page-builder-elements/src/renderers/form/types.ts
adrians5j: This comment still pending. But now I'm wondering, can we even do something about it?

adrians5j (changes requested):
Left some comments.

And while I think we can tackle it, I think it's also important we check @SvenAlHamad's [comments](https://github.com/webiny/webiny-js/pull/3737#issuecomment-1845397945) he posted some time ago.

Can we check that out? I'd guess tackling that feedback will introduce more changes into the PR.

packages/app-page-builder-elements/src/renderers/form/FormRender/functions/getNextStepIndex.ts
neatbyte-ivelychko: I am working on it.

packages/app-page-builder-elements/src/renderers/form/types.ts
neatbyte-ivelychko: We get `formData` from `Form` component and by default the type of that data is `any`.
But the `formData` contains form fields and their values, so I will investigate whether we can use `keyof typeof` in order to create a proper type.

packages/app-form-builder/src/components/Form/functions/getNextStepIndex.ts
neatbyte-ivelychko: In the progress

neatbyte-ivelychko:
> And while I think we can tackle it, I think it's also important we check @SvenAlHamad's [comments](https://github.com/webiny/webiny-js/pull/3737#issuecomment-1845397945) he posted some time ago.
> 
> Can we check that out? I'd guess tackling that feedback will introduce more changes into the PR.

I think that we can introduce requested changes in this PR.

packages/app-page-builder-elements/src/renderers/form/types.ts
neatbyte-ivelychko: After investigation I can say that this should probably stay as `Record<string, any>` because pararm `data` of property `onChange` of `Form` component has type `Record<string, any>`. So there not much we can do to improve type

packages/app-form-builder/src/components/Form/functions/getNextStepIndex.ts
neatbyte-ivelychko: @adrians5j I have one question. Some of the validators that we use for `Step` `conditions` does not exist in `@webiny/validation` maybe we should consider adding them to the `@webiny/validation`?
For instance we need `startsWith`, `endsWith`

adrians5j: Sounds good @neatbyte-ivelychko !

packages/app-page-builder-elements/src/renderers/form/types.ts
adrians5j: OK!

adrians5j:
Hey @neatbyte-ivelychko @neatbyte-vnobis , I was just checking the `lodash` / `@webiny/validation` changes. Looks good... but...... I did some more thinking here.

I went to [bundlephobia](https://bundlephobia.com/) and entered `@webiny/validation` to see what's the size of `@webiny/validation`.

As you can see in the sshot, it's 28.4kb, which is not small.

image

But, the more interesting part is that the lib actually requires `lodash` internally 😅😅😅

All in all... I know we're going back and forth on this, but.... Since here we're talking about like.... what.... 6 validators? Can we just create a `validators` folder in `page-builder-elements` package and literally have validators manually coded in there? Without any libs?

Again, if this was an admin facing pkg, we wouldn't care. But we gotta care when it comes to code that's about to get served on public website. I believe that if we coded it manually like explained above, we'd probably end up with less then 1kb of extra code.

Wdyt?

neatbyte-ivelychko:
> But, the more interesting part is that the lib actually requires `lodash` internally 😅😅😅
> 
> All in all... I know we're going back and forth on this, but.... Since here we're talking about like.... what.... 6 validators? Can we just create a `validators` folder in `page-builder-elements` package and literally have validators manually coded in there? Without any libs?
> 
> Again, if this was an admin facing pkg, we wouldn't care. But we gotta care when it comes to code that's about to get served on public website. I believe that if we coded it manually like explained above, we'd probably end up with less then 1kb of extra code.

Yeah, will do that as soon as I finish working on `Submit & Redirect` rule action

neatbyte-ivelychko:
> > But, the more interesting part is that the lib actually requires `lodash` internally 😅😅😅
> > All in all... I know we're going back and forth on this, but.... Since here we're talking about like.... what.... 6 validators? Can we just create a `validators` folder in `page-builder-elements` package and literally have validators manually coded in there? Without any libs?
> > Again, if this was an admin facing pkg, we wouldn't care. But we gotta care when it comes to code that's about to get served on public website. I believe that if we coded it manually like explained above, we'd probably end up with less then 1kb of extra code.
> 
> Yeah, will do that as soon as I finish working on `Submit & Redirect` rule action

Should we do the same steps for `app-form-builder` or should we leave `@webiny/validation` there?

adrians5j:
Yes please.

Maybe it's worth exploring `@webiny/app-form-builder-condition-groups-validation` shared package? 🤔 

Let's see how much code there will be and if there's a solid amount of code to c/p, then having a shared package might be a better idea.

neatbyte-ivelychko:
> Yes please.
> 
> Maybe it's worth exploring `@webiny/app-form-builder-condition-groups-validation` shared package? 🤔
> 
> Let's see how much code there will be and if there's a solid amount of code to c/p, then having a shared package might be a better idea.

Btw. Maybe we should consider adding unit tests for validation as we have done it in [IconPicker](https://github.com/webiny/webiny-js/blob/f851660601bfadbb43324396ee412ec6628c826b/packages/app-admin/src/components/IconPicker/IconPickerPresenter.test.ts)?
Because we would not need to manually test validation methods.

neatbyte-ivelychko:
> Btw. Maybe we should consider adding unit tests for validation as we have done it in [IconPicker](https://github.com/webiny/webiny-js/blob/f851660601bfadbb43324396ee412ec6628c826b/packages/app-admin/src/components/IconPicker/IconPickerPresenter.test.ts)? Because we would not need to manually test validation methods.

@adrians5j what do you think about this?

neatbyte-ivelychko:
@SvenAlHamad 
Hi!
I have updated UI for Rules tabs.
Looking forward to your feedback!

1 - Rules for step
Знімок екрана 2024-01-11 о 14 20 30
2 - Rules for condition group
Знімок екрана 2024-01-11 о 14 22 30
3 - Rule with `Submit & Redirect` action
Знімок екрана 2024-01-11 о 14 24 58

packages/app-form-builder/src/admin/components/FormEditor/Tabs/EditTab/Styled.ts
neatbyte-ivelychko: Will remove in the next commit

PR #3743, opened 2023-12-06
refactor: prepare react 18 upgrade - batch #3

neatbyte-vnobis:
## Changes
Third batch of changes to React component definitions before upgrading to React 18.
https://www.notion.so/webiny/Upgrade-to-React-18-ceacd9bbfcf74b39a7ceb70943c3afa7

## How Has This Been Tested?
Manually.

## Documentation
None.

### DIFF
--- a/packages/app-aco/src/components/AdvancedSearch/QueryBuilderDrawer/Header.tsx
+++ b/packages/app-aco/src/components/AdvancedSearch/QueryBuilderDrawer/Header.tsx
@@ -8,7 +8,7 @@ interface HeaderProps {
     onClose: () => void;
 }
 
-export const Header: React.VFC<HeaderProps> = ({ onClose }) => {
+export const Header = ({ onClose }: HeaderProps) => {
     return (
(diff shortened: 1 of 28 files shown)


neatbyte-ivelychko:
@brunozoric 

I have a question about this type [(Component: React.FC): React.FC](https://github.com/webiny/webiny-js/blob/next/packages/app-aco/src/Folders.tsx#L5).
What should we do in such case, should we not change type from React.FC? 
Because I was trying to find another type that we can use in this case but I did not find it.

brunozoric (approved):

brunozoric:
> @brunozoric
> 
> I have a question about this type [(Component: React.FC): React.FC](https://github.com/webiny/webiny-js/blob/next/packages/app-aco/src/Folders.tsx#L5). What should we do in such case, should we not change type from React.FC? Because I was trying to find another type that we can use in this case but I did not find it.

Leave it as is for the moment. We will see if it will create issues when we upgrade to React 18.

PR #3756, opened 2023-12-07
feat(app-aco): add table columns definition via react-properties

leopuleo:
## Changes
With this PR, we can define table columns via `react-properties`. The column definition is scoped based on the application.

## How Has This Been Tested?
Manually

### Example of definition:
The following example demonstrates how to create a new column on the Headless CMS table. Similar definitions are available for targeting Page Builder and File Manager tables.

```
<ContentEntryListConfig>
    <Browser.Table.Column
         name={"fieldId"}
         header={"Header name"}
         cell={<CellComponent />}
         sortable={true}
         resizable={false}
         visible={true}
         hideable={false}
         size={400}
         className={"custom-css-class"}
         modelIds={["article"]}
    />
</ContentEntryListConfig>
```

**Props accepted:**
- `name`: used to target the property definition and also the `fieldId` from entry  - `string`
- `header`: used to render the table header - `string | ReactElement | undefined`
- `cell`: used to render the cell content - `string | ReactElement | undefined` (if undefined, return the default value from the `fieldId` defined)
- `resizable`:  define if the column resizable  - `boolean | undefined` (default: true)
- `sortable`:  define if the column sortable  - `boolean | undefined` (default: false)
- `visible`:  define if the column visible in the list  - `boolean | undefined` (default: true)
- `hideable`:  define if users can toggle the visibility for this column  - `boolean | undefined` (default: false)
-  `size`:  define the column size - `number | undefined` (default: 200)
- `className`: additional class names - `string | undefined`
- `modelIds`: define where to show the column based on the `modelId` - `string[] | undefined`

### Example of `<CellComponent />`

```
export const CellComponent = () => {
    const { useTableRow, isFolderRow } = ContentEntryListConfig.Browser.Table.Column;
    const { row } = useTableRow();

    if (isFolderRow(row)) {
        return <>{"-"}</>;
    }

    return (
        <>{`Render any value for: ${row.fieldId}`}</>
    );
};
```

To access the entire row item, you can use the `useTableRow` hook.

Use the `isFolderRow` hook to differentiate between a folder and a record and render different content based on the content type.


packages/app-aco/src/components/Table/ColumnVisibility.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    constructor() {
+        this.state = {};
+        makeAutoObservable(this);
+    }
+
+    public load(namespace: string, columnConfigs: ColumnConfig[]) {
+        this.storage = new LocalStorage(`webiny_column_visibility_${namespace}`);
+        this.state = this.getDefault(columnConfigs);
+    }
```
Pavel910: `namespace` is only used here to pass to the `LocalStorage` constructor. This drives the following conclusions:
- `namespace` is a dependency of `LocalStorage`, and NOT of `ColumnVisibility`.
- `LocalStorage` looks like a dependency of `ColumnVisibility`, so it needs to be injected through the `Columnvisibility` constructor.
- in turn, `namespace` will be passed to the `LocalStorage` through its constructor _outside_ of the `Columnvisibility` class, meaning you can remove all references to this variable from this class

On a different topic: the class name `ColumnVisibility` is a bit strange. You're passing the entire `ColumnConfig[]` to the `load` method, which makes me think that this class can do more than just visibility. What if we had a `ColumnRepository`, which manages columns, visibility, ordering, etc.? Then you would be able to also manage security related things (like showing/hiding columns), loading visibility from the API/local storage, etc.

Pavel910 (approved):

PR #3760, opened 2023-12-11
Ensure extension exists on edited images

Pavel910:
## Changes
This PR fixes a bug where an edited image would lose its extension, if the original image name didn't contain an extension. We're now sending the original file key to the API when requesting a presigned payload. The original file key is used to generate the new file key. File name is kept the same as the original.

## How Has This Been Tested?
Jest tests on the API side, and manually in the Admin app.


adrians5j (approved):
Loved the tests. 🙏🏻

PR #3762, opened 2023-12-11
Fix graphql selection in deleteElement mutation 

Pavel910:
## Changes
This PR fixes a GraphQL mutation which is executed when saved page builder elements are deleted via a button in the toolbar. 

## How Has This Been Tested?
Manually.


leopuleo (approved):
👍

adrians5j:
Updated branch by clicking the button here, so that tests can be started (didn't see any test results). Sorry for 'Merge branch...' commit. 😅

Was a bit curious about the `packages/app-page-builder/src/pageEditor/config/blockEditing/graphql.ts` file being deleted, but now I see in my editor that it's all gray. 😁

Thanks for the cleanup!

PR #3763, opened 2023-12-11
fix(app-file-manager): render extensions next to the basic fields

Pavel910:
## Changes
With this PR we change the way how File field are laid out; we combine Basic Details and Advanced Details tabs into one, making all fields immediately visible to the user. The old layout can be turned on via a config, if you find it more appropriate for your use case.

We also added the ability to control left and right drawer panel width ratio.

```tsx
// Turn grouping in tabs on
<FileDetails.GroupFields value={true} />

// Control the width of the file details drawer
<FileDetails.Width value={"1000px"} />

// Use comma to add left and right panel "flex"value (default is 1)
<FileDetails.Width value={"1000px,1,0.5"} />
```

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/app-file-manager/src/components/FileDetails/FileDetails.tsx
+++ b/packages/app-file-manager/src/components/FileDetails/FileDetails.tsx
@@ -1,5 +1,5 @@
 import React, { useMemo, useState } from "react";
-// @ts-ignore
+// @ts-expect-error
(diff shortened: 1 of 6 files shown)


leopuleo (approved):
Thanks @Pavel910! Testing the file details `width` property, I noticed it works even with percentage width. 👍

PR #3771, opened 2023-12-13
feat: update to react 18

brunozoric:
## Changes
Umbrella PR for React 18 upgrade.

## How Has This Been Tested?
Jest, Cypress and manually.

## Upgrade
User project will, probably, need to upgraded.

## Documentation
Note that this is a breaking change and users will need to upgrade their projects.


neatbyte-ivelychko:
We are using [`react-sortable-hoc`](https://github.com/clauderic/react-sortable-hoc) library in the `app-form-builder`, and we might have issues because of that. 
Basically this library has not been supported for 2 years and it uses React v16 and because of that the build is red.
But I can see that developers of `react-sortable-hoc` have created new library [`dnd-kit`](https://github.com/clauderic/dnd-kit). And they strongly recommend to use newer library.
So what should we do?

Знімок екрана 2024-01-12 о 16 19 23

Pavel910 (approved):

PR #3775, opened 2023-12-18
feat: upgrade to react 18 - updated rmwc version

neatbyte-vnobis:
## Changes
Updated version of rmwc and material-components-web package.

Packages: ui.

## How Has This Been Tested?
Manually.

## Documentation
None.


brunozoric:
@neatbyte-vnobis does the ui package build after this change?

neatbyte-ivelychko:
@brunozoric Now it does build. We had to upgrade version of `material-components-web` because we were using an outdated version of it.

brunozoric (approved):

PR #3780, opened 2023-12-21
refactor(api-form-builder): form statistics

neatbyte-vnobis:
## Changes
Move form statistics to a separate `fbFormStat` content model.

## How Has This Been Tested?
Manually.

## Documentation
[Form Builder - Move backend to HCMS](https://webiny.notion.site/Form-Builder-Move-backend-to-HCMS-1f2cc9b528b544aba56263380eacbab3)


neatbyte-vnobis:
@adrians5j
Tests with `ddb` storage are passing, while the `ddb-es` storage tests are failing. I've looked into this issue but can't quite figure out the cause. Do you have any ideas why this might happen?

The failure occurs in `deleteForm` function, particularly with the newly added `deleteFormStats`. This function utilizes the `deleteMultipleEntries` method from the `api-headless-cms` package, which seems to be unused elsewhere in the code. I'm not entirely sure, but this could potentially be the root cause of the problem.

I want to investigate this issue further, but I'm not sure how to run tests with `ddb-es` storage locally. Is there a way to do this?

adrians5j:
Yeah @neatbyte-vnobis , it's a little bit tricky but it can be done.

I'll send some notes.

neatbyte-vnobis:
@adrians5j
The issue was inside [deleteMultipleEntries](https://github.com/webiny/webiny-js/blob/next/packages/api-headless-cms-ddb-es/src/operations/entry/index.ts#L919) method. This method was attempting to delete items that did not exist in the `ddb-es` table. Specifically, in this test, it attempted to delete both published and latest `formStats`, but no published formStats were present.

To address this, I've implemented a solution based on the existing [deleteEntry](https://github.com/webiny/webiny-js/blob/next/packages/api-headless-cms-ddb-es/src/operations/entry/index.ts#L692) method. This method first queries for all entry items before deleting them, rather than directly deleting the latest and published items for all provided ids.

The tests for the `ddb` table were functioning correctly because DynamoDB does not throw errors when deleting non-existent items.

This PR is now ready for review.

neatbyte-vnobis:
@adrians5j Please review this.

packages/api-form-builder/src/plugins/crud/formStats.crud.ts
```diff
@@ -0,0 +1,233 @@
...
+            }
+
+            if (!formStats) {
+                throw new NotFoundError("Form stats not found.");
+            }
+
+            return formStats;
+        },
+        async getFormOverallStats(this: FormBuilder, id) {
+            const [formId] = id.split("#");
```
leopuleo: Please, let's use `parseIdentifier`

packages/api-form-builder/src/cmsFormBuilderStorage/CmsFormStatsStorage.ts
```diff
@@ -0,0 +1,104 @@
...
+        });
+
+        return this.getFormStatsValues(entry);
+    }
+
+    async updateFormStats({ formStats }: FormBuilderStorageOperationsUpdateFormStatsParams) {
+        const model = this.modelWithContext(formStats);
+
+        // Append `#0001` since `formStats` always has only one revision.
+        const formStatsRevisionId = `${formStats.id}#0001`;
```
leopuleo: Please, use `createIdentifier`

packages/api-form-builder/src/plugins/crud/formStats.crud.ts
```diff
@@ -0,0 +1,233 @@
...
+    FormStatsCRUD,
+    OnFormStatsBeforeCreate,
+    OnFormStatsAfterCreate,
+    OnFormStatsBeforeUpdate,
+    OnFormStatsAfterUpdate,
+    OnFormStatsBeforeDelete,
+    OnFormStatsAfterDelete
+} from "~/types";
+
+const getFormStatsId = (formId: string) => {
```
leopuleo: Please, let's use `parseIdentifier`

packages/api-form-builder/src/plugins/crud/formStats.crud.ts
```diff
@@ -0,0 +1,233 @@
...
+        onFormStatsBeforeCreate,
+        onFormStatsAfterCreate,
+        onFormStatsBeforeUpdate,
+        onFormStatsAfterUpdate,
+        onFormStatsBeforeDelete,
+        onFormStatsAfterDelete,
+        async getFormStats(this: FormBuilder, formRevisionId) {
+            const id = getFormStatsId(formRevisionId);
+
+            let formStats;
```
leopuleo: Let's avoid using `let` when possible. This method could be refactored this way:

```.ts
        async getFormStats(this: FormBuilder, formRevisionId) {
            const id = getFormStatsId(formRevisionId);

            try {
                const formStats = await this.storageOperations.formStats.getFormStats({
                    where: { id, tenant: getTenant().id, locale: getLocale().code }
                });

                if (!formStats) {
                    throw new NotFoundError("Form stats not found.");
                }

                return formStats;
            } catch (ex) {
                throw new WebinyError(
                    ex.message || "Could not load form stats.",
                    ex.code || "GET_FORM_STATS_ERROR",
                    {
                        id
                    }
                );
            }
        }
```

packages/api-form-builder/src/plugins/crud/formStats.crud.ts
```diff
@@ -0,0 +1,233 @@
...
+            if (!formStats) {
+                throw new NotFoundError("Form stats not found.");
+            }
+
+            return formStats;
+        },
+        async getFormOverallStats(this: FormBuilder, id) {
+            const [formId] = id.split("#");
+
+            let formStats;
```
leopuleo: Same here, let's avoid using `let` when possible.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -305,15 +268,16 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
                 webinyVersion: context.WEBINY_VERSION
             };
 
+            let result;
```
leopuleo: We should create the `formStats` entry only if the `form` entry is successfully created. 

Right now, we are trying to create the `formStats` entry even when the creation fails, right?

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -543,17 +514,18 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
                 auth: false
             });
 
+            let result;
```
leopuleo: Same here, see my comment above.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
```diff
@@ -412,6 +379,10 @@ export const createFormsCrud = (params: CreateFormsCrudParams): FormsCRUD => {
                     }
                 );
             }
+
+            await this.deleteFormStats(form.formId);
```
leopuleo, line 419: Same here, we should not delete the `formStats` entry if the `form` entry deletion fails.

leopuleo (changes requested):
Thanks @neatbyte-vnobis! 
I left some comments, but overall, it looks good.

packages/api-form-builder/src/plugins/crud/forms.crud.ts
neatbyte-vnobis: The code inside the try-catch will throw an error if the form is not created. This prevents the `formStats` section from being executed. The same applies to the next two comments as well.

leopuleo: Ok, let's throw a meaningful error in case the `formStats` creation fails, something like:

```
throw new WebinyError(
    ex.message || "Could not create form.",
    ex.code || "CREATE_FORM_STATS_ERROR",
    {
        form
    }
);

neatbyte-vnobis: Errors are already implemented for all formStats CRUD operations.

In the `createForm` method, if form creation fails, it will throw a [CREATE_FORM_ERROR](https://github.com/webiny/webiny-js/blob/neatbyte/form-stats/packages/api-form-builder/src/plugins/crud/forms.crud.ts#L282). If it succeeds, the `createFormStats` method is executed, which, in case of failure, will throw a [CREATE_FORM_STATS_ERROR](https://github.com/webiny/webiny-js/blob/neatbyte/form-stats/packages/api-form-builder/src/plugins/crud/formStats.crud.ts#L154).

Similarly, in the `createFormRevision` method, if form revision creation fails, a [CREATE_FORM_FROM_ERROR](https://github.com/webiny/webiny-js/blob/neatbyte/form-stats/packages/api-form-builder/src/plugins/crud/forms.crud.ts#L530) is thrown. However, if the form revision is successfully created but `createFormStats` fails, a [CREATE_FORM_STATS_ERROR](https://github.com/webiny/webiny-js/blob/neatbyte/form-stats/packages/api-form-builder/src/plugins/crud/formStats.crud.ts#L154) is thrown.

Lastly, for `deleteForm`, if form deletion fails, it throws a [DELETE_FORM_ERROR](https://github.com/webiny/webiny-js/blob/neatbyte/form-stats/packages/api-form-builder/src/plugins/crud/forms.crud.ts#L374). Conversely, if the form is deleted successfully but `deleteFormStats` encounters an error, a [DELETE_FORM_STATS_ERROR](https://github.com/webiny/webiny-js/blob/neatbyte/form-stats/packages/api-form-builder/src/plugins/crud/formStats.crud.ts#L220) is thrown.

PR #3783, opened 2023-12-27
feat(app-aco): support reference field in advanced search filter

leopuleo:
## Changes
With this PR, we add support to `reference` fields inside the Advanced Search Filter.

The user can pick the referenced entry via the `Autocomplete` input field and search through the models set during the entry configuration.

For the value picked, we save both the `entryId` and the `modelId`: both are needed to create the filter query and get the saved entry while loading the reference field.

### Refactor: field renderer definition via react properties
To support app-specific field renderers, such as the reference field, we introduce the new `<AdvancedSearch.FieldRenderer>` config, allowing us to define field renderers outside of `app-aco`.

```jsx
<AcoConfig>
  <AdvancedSearch.FieldRenderer name={"text"} type={FieldType.TEXT} element={<Input />} />
</AcoConfig>
```

- `name`: arbitrary name used to target the config
- `type`:  `FieldType`s used by the element
- `element`: the element that will be rendered

Together with the config, the developer will have at his disposal:
- `useInputField` to access the current input field data (no external props passed to the renderer element)
- `FieldType` enum

## How Has This Been Tested?
Manually + Jest


packages/app-aco/src/components/AdvancedSearch/QueryBuilderDrawer/QueryBuilderDrawer.tsx
```diff
@@ -28,7 +27,7 @@ interface QueryBuilderDrawerProps {
 
 export const QueryBuilderDrawer = observer(({ filter, ...props }: QueryBuilderDrawerProps) => {
     const presenter = useMemo<QueryBuilderDrawerPresenter>(() => {
-        return new QueryBuilderDrawerPresenter(props.fields);
+        return new QueryBuilderDrawerPresenter();
     }, [props.fields]);
```
Pavel910: This `props.fields` dependency is no longer necessary.

packages/app-aco/src/components/AdvancedSearch/domain/Field.ts
```diff
@@ -26,43 +27,72 @@ export interface PredefinedDTO {
...
+    }
+}
+
+export class Value {
+    public readonly value: string;
+
+    static createFromField(rawData: FieldRaw) {
+        if (rawData.type === FieldType.REF) {
+            return new Value(`${rawData.id}.entryId`);
+        }
```
Pavel910, line 89: This will cause the same problem we had with that big `switch` statement before. Every time you have to add a non-standard value handling, you'll be adding `if` statements. Let's leave this for now, but this should also be redesigned in a way which allows each field to handle its own value (like that `fieldsWithElements`).

packages/app-aco/src/components/AdvancedSearch/domain/Field.ts
```diff
@@ -129,6 +159,12 @@ export class Condition {
                     { label: "is after or equal to", value: "_gte" }
                 ]);
 
+            case "ref":
+                return createConditions([
+                    { label: "contains", value: " " },
+                    { label: "doesn't contain", value: "_not" }
+                ]);
+
```
Pavel910, line 167: Same problem as my previous comment. Essentially, each field needs to provide all the implementations to your app, so your app is free of any knowledge about the specific implementation.

packages/app-headless-cms/src/admin/components/ContentEntries/Filters/RefFieldRenderer/components/RefPresenter.ts
```diff
@@ -0,0 +1,76 @@
...
+        makeAutoObservable(this);
+    }
+
+    async load(value?: string) {
+        try {
+            if (!value) {
+                return;
+            }
+
+            const { entryId, modelId } = JSON.parse(value);
```
Pavel910, line 40: Add a comment about _why_ is `value` a string, and not an object? Because it's not clear why you need to JSON.parse if you doin't know all the internals.

packages/app-headless-cms/src/admin/components/ContentEntries/Filters/RefFieldRenderer/domain/EntryRepositoryFactory.ts
```diff
@@ -0,0 +1,28 @@
+import { ApolloClient } from "apollo-client";
+import { EntriesGatewayInterface, EntriesGraphQLGateway } from "../adapters";
+import { EntryRepository } from "./EntryRepository";
+
+export class EntryRepositoryFactory {
+    private gateway: EntriesGatewayInterface | undefined;
+    private cache: Map<string, EntryRepository> = new Map();
+
+    getRepository(client: ApolloClient<any>, modelIds: string[]) {
```
Pavel910: It would be cleaner if you passed the Gateway instance instead of the `client`. Because if you look closely, you're passing a `client` just so you can pass it further down to the Gateway constructor, and nothing else. So it's prop drilling (or dependency drilling), which should be avoided. Pass a gateway, and modelIds to the factory.

Pavel910 (changes requested):

packages/app-aco/src/components/AdvancedSearch/domain/Field.ts
leopuleo: I agree, I'll open a separate PR to refactor it. We can use configs to define field conditions. Thanks for pointing this out 👍

Pavel910 (approved):
Thank you! 👍

PR #3786, opened 2023-12-29
Move Meta Fields Into Meta object

adrians5j:
## Changes
With this PR, we're moving all of the newly introduced on/by entry meta fields into the `meta` object, along with the rest of the entry level meta fields like `status`, `locked`, etc.

## How Has This Been Tested?
Manually. Jest.

## Documentation
N/A


adrians5j:
Ultimately, we decided not to proceed with these changes.

PR #3802, opened 2024-01-10
fix(tasks): wrong position of the error in the log items array

brunozoric:
## Changes
When having an error in the log items array, on the array index 2, and not array index 0 or 1, error is pushed to array index 1 due to a bug in object indexing plugin. This only affected Elasticsearch systems.

## How Has This Been Tested?
Jest and manually.


Pavel910 (approved):
Thankfully, it's a simple fix 😅

Pavel910 (approved):

PR #3803, opened 2024-01-10
Meta Fields - Remove Entry-prefixed Fields

adrians5j:
## Changes

Initially, the idea with the new CMS entry meta fields was to have revision and entry-level meta fields. The idea was also to deprecate the prefixless fields and remove them in the near future. For more info, check the initial [PR](https://github.com/webiny/webiny-js/pull/3725).

After some testing and code merging, we realized that it'd be best to not deprecate the prefixless fields, but to keep them as an alias for entry-level fields. So, we started implementing that change, but quickly realized that now there's just too much fields to deal with: 10 prefixless fields + 10 revision-level fields + 10 entry-level fields = 30 fields 🫢. And it's not just about the total count of fields and the DX, but also juggling all these fields in our internal code is just cumbersome and could introduce bugs.

After some back and forth action with the code, we decided to ultimately remove entry-prefixed fields, and simply keep only prefixless and revision-prefixed fields:

- createdOn
- savedOn
- modifiedOn
- firstPublishedOn
- lastPublishedOn
- createdBy
- savedBy
- modifiedBy
- firstPublishedBy
- lastPublishedBy
- revisionCreatedOn
- revisionSavedOn
- revisionModifiedOn
- revisionFirstPublishedOn
- revisionLastPublishedOn
- revisionCreatedBy
- revisionSavedBy
- revisionModifiedBy
- revisionFirstPublishedBy
- revisionLastPublishedBy

Despite the fact that this *probably* want cause issues for users, we need to be aware that this is a breaking change, because these prefixless fields were previously acting as revision-level fields and now they're acting as entry-level fields. We will document this in the release notes.

While this is the gist of the PR, there were some other changes, some of which are also breaking, so I'll list them in the following section.

### 1. Removed `ownedBy` Field (Breaking Change❗)
In the initial PR, we marked this field as deprecated, but now, since the prefixless fields are here to stay, we've removed the redundant `ownedBy` field. The field that is to be used instead is `createdBy`. Also note that the previous `createdBy` field, which was acting as a revision-level field, is now to be replaced with `revisionCreatedBy`.

### 2. Removed `meta.publishedOn` Field (Breaking Change❗)
In the initial PR, we marked this field as deprecated, but now, in order to keep things consistent, we've removed it. Instead, users need to use `firstPublishedOn` / `lastPublishedOn` fields, which are located in the root of the entry and not inside of the `meta` object.

### 3. Removed Options On Publish Entry Mutation (Breaking Change❗)
In the initial PR, we marked entry publishing-related options (`updateSavedOn` and `updatePublishedOn` flags) as deprecated. But, with this PR, since the intention behind these flags can be achieved more easily via a single create entry mutation, we've removed this. 

### 4. Adde `savedBy` and `modifiedOn` Meta Fields
In the initial PR, we had the following change:

> Prior to this PR, a CMS entry did not have `savedBy` but had `savedOn`. It also did not have `modifiedOn`, but did have `modifiedBy`. By adding the missing `savedBy` and `modifiedOn` fields, this inconsistency has now been resolved.

In the initial PR, this was done only on the entry-level fields. Prefixless fields were not touched because they were to be deprecated. But, since that's not the case anymore, the same thing was done to prefixless fields.

All in all, there are no more inconsistencies with these fields. 5 fields for date time and 5 fields for identity-related fields.

### 5. No More Special Mappings In HCMS-powered Apps (ACO, FM, APW)
With the initial PR, we had to do some mapping within HCMS-powered apps' code. We had to map `createdOn` to `entryCreatedOn`, and do the same with the rest of the fields.

This mapping is no longer needed, so the extra code has been removed.

## Linked PRs
https://github.com/webiny/webiny-js/pull/3782

## How Has This Been Tested?
Existing tests. Manual testing.

## Documentation
Changelog. Separate articles describing meta fields.


leopuleo (approved):
Thanks @adrians5j. 👍

PR #3823, opened 2024-01-17
feat(app-file-manager): edit content entry file field

Pavel910:
## Changes

This PR moves the CMS file field into the `app-file-manager` app, and makes it possible to edit a file directly from the content entry form.

https://github.com/webiny/webiny-js/assets/3920893/60dd090a-c1f7-4b85-98b0-775aa6462a81

## How Has This Been Tested?
Manually.


leopuleo (approved):
👍 
Thanks @Pavel910

PR #3977, opened 2024-03-04
fix: ensure PB elements have stable element IDs

Pavel910:
## Changes
This PR fixes the logic of generating and maintaining element IDs across all page builder editors.

Prior to this PR, we were regenerating element IDs every time a document was opened in one of the Page Builder editors. To the best of my recollection, since this is a very old piece of code, we did that as a quick solution to predefined (saved) blocks insertion. 

This worked "ok" until recently. As we're adding more features to the page builder, this particular logic of regenerating element IDs is creating problems, as we don't have a stable reference ID to each individual element. This PR introduces **backwards compatible** improvements to the logic.

### Block Editor
- all elements and variables added to the "block" definition, in the Block Editor, will have permanent IDs.
- blocks that are imported from our [blocks library](https://blocks.webiny.com/) will get the element IDs generated during the import (currently, those downloadable blocks don't have elements IDs, because they were exported when this was not an issue)
- within the block editor, element IDs will be a single 10-character long string, e.g. `WdAoCjNlFu`
- when elements are linked to a variable, the variable `id` attribute will match the element ID

### Page Template Editor
- adding a block (using predefined blocks) creates a new element with an auto-generated ID (the `block` element). **All child element IDs are prepended with this top-level block ID, in form of `{blockId}#{elementId}`**. Example: `WdAoCjNlFu#soKq1HDmiB`. The `#` character separates the top-level block element ID from the referenced block element ID. This means that we *do not* generate new unique IDs for all elements recursively. Instead, we merely prepend the unique top-level element ID to an existing referenced block element IDs (recursively). This way we have a unique element ID on the page level, and we can debug easier (visual inspection becomes much easier), as the second part of the element ID will match the original referenced block element ID.
- when blocks with variables are added to the template, the variable IDs also get prefixed, as described in the previous bullet point.
- cloning a block within the template will copy the source block data, and replace the source block element ID with a newly generated block ID. All elements, and references to blocks and variables will be recursively updated with this new block ID
- on the GraphQL side, since referenced blocks get their content dynamically resolved on each load from the API, we can reliably recreate the IDs of all child elements. We simply follow the rule: keep the top-level element ID, and prepend it to all child elements that belong to the predefined block.
- when a block is unlinked, we DO NOT generate new IDs for its elements. Instead, we keep the exact same IDs that are assigned at the time of unlinking. It is safe to do, because all child elements IDs depend on the top-level element ID, which is ALWAYS unique within a single page.
  - all other page elements, created using atomic elements, will keep their IDs for as long as they remain part of the page content.

### Page Editor
- when variable values are being updated, they're being stored within the content element `data` object
- on page save, variable values are copied to the `content.data.template.variables` array, so they get preserved between page loads (because referenced block data structure is not saved as page content; it's resolved on every page load)

## Other Improvements
- fixed a bug with `ResizableObserver`, added a debounce to avoid excessive callback execution (browser was throwing an error on element deactivation)
- added debug utils for development mode, inside of all PB editors: you can now open the browser console, and start typing `debug_` to see the 2 new functions: one will allow you to print the active element data structure to console, the other will print the entire page content, with all elements, recursively.
- added `useStateIfMounted` hook, which only sets the state if the component is mounted
- fixed  `setState on an unmounted component` warning in many places (PB, Headless CMS, ACO)
- fixed block variable not updating the main page content (using `dot-prop-immutable`, instead of modification by reference)
- Template Editor now has its own `clone element` action plugin, because there's some additional block-level logic that needs to be performed
- added `PageContent` and `PageElementId` value classes to be used in the `api-page-builder`
- when a blank page is created, its content is no longer `null`, but a `document` element 
- added links to Youtube tutorials on "Unlink Block" actions
- typos and grammar in UI

## How Has This Been Tested?
Manually + some Jest tests.


packages/app-page-builder/src/templateEditor/Editor.tsx
```diff
@@ -108,7 +98,10 @@ export const TemplateEditor = () => {
                 const { content, ...restOfTemplateData } = pageTemplateData;
 
                 const existingContent = content
-                    ? { ...content, elements: getBlocksWithUniqueElementIds(content.elements) }
+                    ? {
+                          ...content,
+                          elements: content.elements
+                      }
```
adrians5j:
```suggestion
                    ? content
```

P.S. Not sure if we care about making a shallow clone here.

packages/app-page-builder/src/templateEditor/config/BlockElementSidebarPlugin.tsx
```diff
@@ -61,7 +61,7 @@ const UnlinkTab = ({ permission }: UnlinkTabProps) => {
                     )}
                 </div>
                 <div className="info-wrapper">
```
adrians5j, line 63: If you ask me, I'd just remove this. Or if we have an actual link, which at this point we may actually have, maybe we can link it finally.

packages/app-page-builder/src/pageEditor/config/BlockElementSidebarPlugin.tsx
```diff
@@ -64,7 +64,7 @@ const UnlinkTab = ({ permission }: UnlinkTabProps) => {
                     )}
                 </div>
                 <div className="info-wrapper">
-                    <InfoIcon /> Click here to learn more about how block work
+                    <InfoIcon /> Click here to learn more about how blocks work.
```
adrians5j: Same comment here as my previous one.

packages/app-page-builder/src/editor/recoil/actions/cloneElement/action.ts
```diff
@@ -40,11 +43,14 @@ export const cloneElementAction: EventActionCallable<CloneElementActionArgsType>
     const parent = await state.getElementById(element.parent);
     const position = parent.elements.findIndex(el => el === element.id) + 1;
 
+    const newElementId = getNanoid();
```
adrians5j:
```suggestion
    const newElementId = randomIdGenerator();
```

Can we do this instead here?

packages/app-page-builder/src/editor/components/Editor/Content.tsx
```diff
@@ -81,24 +77,17 @@ const BaseContainer = styled(LegacyBaseContainer)`
...
-        },
-        [uiAtom]
-    );
-
     const resizeObserver = useMemo(() => {
         return new ResizeObserver((entries: ResizeObserverEntry[]) => {
             for (const entry of entries) {
                 const { width, height } = entry.contentRect;
-                setPagePreviewDimension({ width, height });
+                setPagePreviewDimensions({ width, height });
```
adrians5j: Why do we even need this resizeObserver? I know I've been removing some of these in the past, when switching to new rendering engine.

But maybe this one is still required. Can you remind me?

packages/app-page-builder/src/pageEditor/Editor.tsx
```diff
@@ -136,7 +126,10 @@ export const PageEditor = () => {
                 const { revisions = [], content, ...restOfPageData } = extractPageData(data);
 
                 const existingContent = content
-                    ? { ...content, elements: getBlocksWithUniqueElementIds(content.elements) }
+                    ? {
```
adrians5j: Check my previous comment. Not sure if applicable.

packages/app-page-builder/src/templateEditor/config/eventActions/cloneElement/action.ts
```diff
@@ -0,0 +1,91 @@
...
+    const { element } = args;
+    if (!element.parent) {
+        return {
+            actions: []
+        };
+    }
+    const parent = structuredClone(await state.getElementById(element.parent));
+    const position = parent.elements.findIndex(el => el === element.id) + 1;
+
+    const newElementId = getNanoid();
```
adrians5j: Check my previous comment. Not sure if applicable.

adrians5j (changes requested):
A couple of smaller comments, but looking good as always.

packages/app-page-builder/src/editor/recoil/actions/cloneElement/action.ts
Pavel910: Oh, good catch man, I missed that one! 👍

packages/app-page-builder/src/templateEditor/Editor.tsx
Pavel910: Actually this can go away completely. We no longer need the `existingContent` variable.

packages/app-page-builder/src/editor/components/Editor/Content.tsx
Pavel910: Looks like this one can be removed. There's still one for BC in the `useResponsiveClassName` hook, which is only there for legacy rendering engine (so it seems, because we don't use that hook any more).

packages/app-page-builder/src/hooks/useVariable.ts
```diff
@@ -1,20 +1,19 @@
...
-    const variable = element?.data?.variables?.find(
-        (variable: PbBlockVariable) => variable.id === variableId
-    );
+    const variable = element?.data?.variables?.find(variable => variable.id === variableId);
 
     // We're storing a reference to an element, so we don't have to recreate the `onChange` callback on every update.
     useEffect(() => {
         elementRef.current = element;
-    }, [element]);
+        variableIdRef.current = variableId;
```
adrians5j, line 19: Hm, could we maybe add a comment here explaining why we're storing `variableId` into a ref? 

It has to do something with recreating the `onChange` callback less frequently?

packages/api-page-builder/src/graphql/crud/pageBlocks.crud.ts
```diff
@@ -287,11 +290,18 @@ export const createPageBlocksCrud = (params: CreatePageBlocksCrudParams): PageBl
         ) {
             const blocks = [];
 
-            for (const pageBlock of content?.elements) {
-                const blockId = pageBlock.data?.blockId;
-                // If block has blockId, then it is a reference block, and we need to get elements for it.
+            /**
+             * If the content block has a `blockId`, then it is a referenced block.
```
adrians5j, line 294: BTW do we need to make similar changes in `packages/api-page-builder/src/graphql/crud/pageTemplates.crud.ts:305` ?

adrians5j (approved):
Looks good, just left a couple of questions. Liked some of the new things you tried within this PR too.

One extra question BTW, when you say:
image

This _is_ backwards compatible, right?

Pavel910:
@adrians5j 

> This is backwards compatible, right?

Yes. We do this reverse mapping from element to a variable ID, so existing content will have IDs that don't match the element, but still be able to map to the correct variable ID. All the new content will be a lot more logical and easier to read when inspecting the raw JSON, because those IDs will actually match.

packages/api-page-builder/src/graphql/crud/pageBlocks.crud.ts
Pavel910: No, because template passes block interpolation to `pageBlocks.crud`. The `pageTemplates.crud` is only responsible for loading the template definition, and assigning variable values to the `content.data`. Then it passes actual block content loading to `pageBlocks.crud`. So this logic is not duplicated.

PR #3984, opened 2024-03-08
fix(app-page-builder): optimize Image element renderer

Pavel910:
## Changes
This PR addresses an issue with the Image element renderer, that was particularly visible in the Page Builder editor,  when network cache was disabled in the browser. 

- In the `PeImage` component, the FileManager element was conditionally added to the image element, depending on the state of the element (active / inactive). This was an unnecessary logic, since the FileManager really doesn't make any difference, and it simply sits there until interacted with. Removing the conditional logic fixed the problem with remounting of the image element.
- In the image renderer itself, the styled image element was being recreated on every render, which made the whole screen flicker, since React had to unmount the previous element, and mount a new one using the new component. The styled component is now defined outside of the image renderer, and it simply accepts props from the image element settings.
- In the `Elements.tsx` file, the element key now also contains the `element.type`, to make it easier to visually inspect and find the right element in the profiler.
- the `collection` array of objects is no longer passed to the element `meta` prop, as it is useless, but makes it difficult to memoize the prop. The `meta` prop is still not memoized properly, but this makes it a lot lighter.

## How Has This Been Tested?
Manually.

### DIFF
--- a/packages/app-page-builder-elements/src/components/Elements.tsx
+++ b/packages/app-page-builder-elements/src/components/Elements.tsx
@@ -13,10 +13,15 @@ const getElementKey = (
     elementIndex: number,
     parentBlockElement?: ElementType
 ) => {
+    let parts: string[] = [element.id];
+
     if (parentBlockElement) {
-        return `${parentBlockElement.id}-${elementIndex}`;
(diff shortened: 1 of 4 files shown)


adrians5j (approved):
All good stuff, thanks! 🚀

PR #4041, opened 2024-03-13
Page Builder - Use `object` HTML Tag When Rendering SVGs

adrians5j:
## Changes
With this PR, we're allowing users to use the [`object`](https://www.w3schools.com/tags/tag_object.asp) HTML tag when rendering SVGs with the **Image** page element. 

The following screenshot shows two images, one rendered the usual way, one rendered with the new `object` HTML tag:

image

Note that, for existing users / backwards compatibility, existing SVG images on pages will still be rendered via the `img` HTML tag. But, if needed, users are able to change this via the new HTML Tag input:

image

Also note that, once a user upgrades their Webiny project, all newly selected images will have their HTML tag set to "Auto-detect". In other words, if a user selects a file that ends up with the `.svg` extension, `object` HTML tag will be used. Otherwise, `img` HTML tag will be used.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.

### DIFF
--- a/packages/app-page-builder-elements/src/renderers/image.tsx
+++ b/packages/app-page-builder-elements/src/renderers/image.tsx
@@ -5,6 +5,16 @@ import { useRenderer } from "~/hooks/useRenderer";
 import { LinkComponent as LinkComponentType } from "~/types";
 import { DefaultLinkComponent } from "~/renderers/components";
 
+declare global {
+    // eslint-disable-next-line
+    namespace JSX {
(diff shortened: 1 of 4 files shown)


Pavel910 (approved):
Good improvement! Let's go with this for the time being, and in the future we'll do some housekeeping on that image renderer, as it's becoming increasingly difficult to follow the logic, too many `if` statements and reassignments.

PR #4052, opened 2024-03-20
Add support for federated OIDC identity providers

Pavel910:
## Changes
This PR adds support for OIDC federated identity providers to be configured with Webiny.

Log in entirely through external IDP:
CleanShot 2024-03-20 at 10 13 36

Log in through the built-in user pool AND through an external IDP:
CleanShot 2024-03-20 at 10 16 34

`@webiny/app-admin-users-cognito` now exports 2 components:
1) `Cognito` - this configures the login screen as well as the Admin Users module to manage users via Webiny Admin app.
2) `CognitoLogin` - configures the login screen, without the Admin Users module. This is useful when you want to use an external IDP to manage users, and just need a login screen.

## How Has This Been Tested?
Manually.

## Documentation
A full article is to be written, but for now, this is how you use the feature:

```tsx 
// apps/admin/src/App.tsx
import React from "react";
import { Admin } from "@webiny/app-serverless-cms";
import {
  CognitoLogin,
  CreateAuthenticationConfig,
  Components,
} from "@webiny/app-admin-users-cognito";
import "./App.scss";
import { ButtonPrimary } from "@webiny/ui/Button";

const cognitoConfig: CreateAuthenticationConfig = {
  oauth: {
    domain: String(process.env.REACT_APP_USER_POOL_DOMAIN),
    redirectSignIn: "http://localhost:3001",
    redirectSignOut: "http://localhost:3001",
    scope: ["profile", "email", "openid"],
    responseType: "token",
  },
  federatedProviders: [
    {
      name: "MyIDP",
      component: ({ signIn }) => (
        <ButtonPrimary onClick={() => signIn()}>
          Sign in via My IDP
        </ButtonPrimary>
      ),
    },
  ],
};

// Decorate the SignIn view, and customize props.
// The result of this customization is show on the first screenshot above.
// This is entirely optional. If this customization is not applied, the login screen will look as show on the second screenshot.
const SignInDecorator = Components.SignIn.createDecorator((Original) => {
  return function SignIn(props) {
    return (
      <Original
        {...props}
        allowSignInWithCredentials={false}
        title={"Sign In"}
        description={
          "You will be taken to an external service to complete the sign in process."
        }
      />
    );
  };
});

export const App: React.FC = () => {
  return (
    <Admin>
      <CognitoLogin config={cognitoConfig} userMenuItems={{ signOut: true, userInfo: true }} />
      <SignInDecorator />
    </Admin>
  );
};
```

And to configure your User Pool, you need to modify the Pulumi config:

```ts
// apps/core/webiny.application.ts
import {
  createCoreApp,
  configureAdminCognitoFederation,
} from "@webiny/serverless-cms-aws";

  pulumi(app) {
    configureAdminCognitoFederation(app, {
      // Provide a name for the user pool domain.
      domain: "webiny-admin",
      // Whitelist callback URLs (can be localhost or real domain).
      callbackUrls: ["http://localhost:3001"],
      // Configure external identity providers.
      identityProviders: [
        {
          name: "MyIDP",
          type: "oidc",
          providerDetails: {
            attributes_request_method: "POST",
            authorize_scopes: "email profile openid",
            client_id: String(process.env.REMOTE_IDP_CLIENT_ID),
            client_secret: String(process.env.REMOTE_IDP_CLIENT_SECRET),
            oidc_issuer: String(process.env.REMOTE_IDP_ISSUER),
          },
        },
      ],
    });
  },
});
```

Map federated identity to a user group:
```ts
// apps/api/graphql/src/security.ts
cognitoAuthentication({
  region: String(process.env.COGNITO_REGION),
  userPoolId: String(process.env.COGNITO_USER_POOL_ID),
  identityType: "admin",
  getIdentity({ identity, token }) {
    const federatedIdentity = Boolean(token.identities);

    return {
      ...identity,
      group: federatedIdentity ? "full-access" : undefined,
    };
  },
});
```


packages/app-admin-cognito/package.json
Pavel910, line 1: I added this to trick partial builds into resolving the dependency graph correctly.

packages/app-admin-users-cognito/src/plugins/userMenu/userInfo.tsx
```diff
@@ -73,7 +77,10 @@ export const UserInfo = () => {
     const { currentTenant, defaultTenant } = security.identity;
 
     let wrapper: any = { Component: Link, props: { to: "/account" } };
-    if (currentTenant && defaultTenant && currentTenant.id !== defaultTenant.id) {
+    if (
+        !accountRoute ||
```
adrians5j: Maybe we make this a bit nicer? 🤯

adrians5j: What is this `accountRoute` anyways? Should we comment this?

adrians5j (commented):
23/34 files reviewed. Will continue in the afternoon.

packages/app-admin-users-cognito/src/plugins/userMenu/userInfo.tsx
Pavel910: Account route is a path to the user "account" page. When you use Okta, Auth0, or any external IDP, you don't have an account view, because you can't edit a user from within the Admin app. However, the `UserInfo` menu item shows details of the user (email, displayName), and optionally renders a link to the "account" view. That's what this is all about.

Now, what does external IDP have to do with Cognito: this is why Cognito is the trickiest of all setups: it supports both "in-house" users, as well as federated IDPs. This means that sometimes you do have a user you manage from within the Admin app, and sometimes you don't. This hybrid use case is something we'll need to come back to, I don't think it's polished enough, at the moment.

In case of `CognitoLogin`, when you only want a Login screen,   and no user management, you will still have the `UserInfo` component, but it will NOT have the link to the "account" view. That's what that `accountRoute` prop is for.

Pavel910: > Maybe we make this a bit nicer? 🤯

I can try extracting this into a local utility, but there's no nice, global, way of handling this.

Pavel910: @adrians5j Did some refactoring on this.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/TemplateGallery.tsx
```diff
@@ -67,29 +67,23 @@ const Gallery = makeDecoratable("TemplateGalley", (props: TemplateGalleryProps)
...
+        return true;
     });
 }
 
-export const TemplateGallery = Object.assign(withDecoratorFactory(Gallery), {
-    Container: withDecoratorFactory(GalleryContainer),
-    List: withDecoratorFactory(GalleryList),
-    Item: withDecoratorFactory(TemplateItem),
-    Close: withDecoratorFactory(CloseGallery)
+export const TemplateGallery = Object.assign(withShouldRender(Gallery), {
```
adrians5j, line 88: Can you explain this to me? Why are we wrapping all components with `withShouldRender` and not only `Gallery`? 

I probably don't understand what's going on here.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/TemplateGallery.tsx
```diff
@@ -67,29 +67,23 @@ const Gallery = makeDecoratable("TemplateGalley", (props: TemplateGalleryProps)
     );
 });
 
-function withDecoratorFactory<T extends DecoratableComponent>(Component: T) {
-    return Object.assign(Component, {
-        createDecorator: createDecoratorFactory<{ modelIds?: string[] }>()(
-            Component,
-            decoratorProps => {
-                const { model } = useModel();
+export type ShouldRender = { modelIds?: string[] };
```
adrians5j, line 70: Check this file for the `TemplateGalley` string 🙃

packages/app-admin-users-cognito/src/plugins/userMenu/accountDetails.tsx
```diff
@@ -41,7 +47,7 @@ export const AccountDetails = () => {
     }
 
     return (
-        <Link to={"/account"} className={linkStyles}>
+        <Link to={accountRoute} className={linkStyles}>
```
adrians5j, line 50: Didn't realize we don't show Account details dropdown menu item if not on default tenant. 💡

packages/app-admin-users-cognito/src/plugins/userMenu/userInfo.tsx
```diff
@@ -62,70 +63,97 @@ const linkStyles = css({
...
+}
+
+const UserInfoListItemContainer = ({
+    accountRoute,
+    className,
+    children
+}: UserInfoListItemContainer) => {
+    const isDefaultTenant = useIsDefaultTenant();
+
+    // If there's no "accountRoute", or the user is not on his default tenant, don't render the link to "Account Details".
```
adrians5j, line 120: Hm is this true? I see in current system, we do show user info item when not on non-default tenant:

image

🤔 

The only thing is that when clicking on it, it doesn't lead anywhere.

adrians5j (changes requested):
A couple of extra comments. Nothing special.

packages/app-admin-users-cognito/src/plugins/userMenu/userInfo.tsx
Pavel910: @adrians5j yes, we do show it, but it's not a link. It's a simple `div`. Because on a non-default tenant, there's nowhere to go. It's been like this forever. Hopefully the future UI redesign will address some of these  visual problems, and it will not look confusing.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/TemplateGallery.tsx
Pavel910: We're wrapping each component, because they're all decoratable, and this `withShouldRender` attaches a conditional decorator, which optionally takes a `modelIds` props, so you can control on which models that component will be decorated.

TemplateGallery consists of many components, Container, List, Item, Close, and they're all decoratable. Will C/P this explanation into the code comment.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/TemplateGallery.tsx
Pavel910: 😂 Nice find, but functionally it doesn't affect anything, it's just a displayName for components. Will fix.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/TemplateGallery.tsx
adrians5j: Ah I understand now, tnx!

packages/app-admin-users-cognito/src/plugins/userMenu/userInfo.tsx
adrians5j: Ah srry my bad, I misunderstood the code.

All good now, ty!

adrians5j (approved):

PR #4057, opened 2024-03-21
chore: remove @webiny/commodo and its typings

Pavel910:
## Changes
This PR removes the `@webiny/commodo` package which is no longer being used. WIth it, we can now also remove the types from the `typings` folder. We still use some `@commodo` packages directly, but this facade package was not used anywhere in our code base, so it is safe to be removed.

## How Has This Been Tested?
Nothing to test here :)


adrians5j (approved):
Very nice! 😊

Just ran one failed test, just in case (flakyness probably).

PR #4064, opened 2024-03-28
feat(api-headless-cms): add crud operation to restore entry from trash bin

leopuleo:
## Changes
With this PR, developers can restore an entry previously moved to the trash bin.

When an entry has been restored, we add the following fields:

- `deleted`: changed from `true` to `false` (boolean)
- `restoredOn`: the datetime when the entry has been restored (string)
- `restoredBy`: the user that restored the entry (identity object)

Similar types have been added for the `revisionRestoredOn` and `revisionRestoredBy` fields.

### Folder verification while restoring an entry
In the restoration process, we verify if the folder referenced by the `folderId` inside `wby_location` still exists. If not, we relocate the entry to the `ROOT_FOLDER`. 

This aspect of the business logic is handled by subscribing to `restoreEntry` event topics and managed in a separate package called `api-headless-cms-aco`.

### Related PR
[https://github.com/webiny/webiny-upgrades/pull/29](https://github.com/webiny/webiny-upgrades/pull/29)

## How Has This Been Tested?
Manually + Jest


scripts/listPackagesWithTests.js
```diff
@@ -90,6 +90,12 @@ const CUSTOM_HANDLERS = {
             "packages/api-headless-cms --storage=ddb-es,ddb"
         ];
     },
+    "api-headless-cms-aco": () => {
```
adrians5j, line 93: BTW, there is a new file that does this: `.github/workflows/wac/utils/listPackagesWithJestTests.ts`.

Please add this change there. This file is actually no longer needed. You can even delete it.

There's one other thing you'll need to do here I think, and that is rebuild CI workflows.

Run `yarn ghawac build` to do that, and then push alltogether.

I know I know... wthell do I need to do anything CI-related? It's something I still need to address. ☺️

packages/api-headless-cms/src/graphql/schema/resolvers/manage/resolveRestore.ts
```diff
@@ -0,0 +1,24 @@
...
+
+type ResolveRestore = ResolverFactory<any, ResolveRestoreArgs>;
+
+export const resolveRestore: ResolveRestore =
+    ({ model }) =>
+    async (_, args: any, context) => {
+        const { revision } = args;
+        try {
+            if (!revision) {
+                throw new Error(`The input value "revision" is required!`);
```
adrians5j: Hm... wouldn't the `await context.cms.restoreEntry(model, revision);` call throw an error in case `revision` was missing anyways?

packages/api-headless-cms/src/types/types.ts
```diff
@@ -1944,6 +2015,10 @@ export interface CmsEntryStorageOperations<T extends CmsStorageEntry = CmsStorag
      * Move the entry to bin.
      */
     moveToBin: (model: CmsModel, params: CmsEntryStorageOperationsMoveToBinParams) => Promise<void>;
+    /**
+     * Move the entry to bin.
```
adrians5j:
```suggestion
     * Restore the entry from bin.
```

packages/api-headless-cms/src/types/types.ts
```diff
@@ -1944,6 +2015,10 @@ export interface CmsEntryStorageOperations<T extends CmsStorageEntry = CmsStorag
...
+     */
+    restore: (model: CmsModel, params: CmsEntryStorageOperationsRestoreParams<T>) => Promise<T>;
```
adrians5j: Should we maybe have this named like `restoreFromBin`, just so we're more explicit and also consistent with `moveToBin`? 

Does it maybe get clunky with such a long name? 🤔

packages/api-headless-cms-aco/__tests__/utils/createTestWcpLicense.ts
```diff
@@ -0,0 +1,44 @@
...
+                    options: {
+                        maxCount: {
+                            type: MT_OPTIONS_MAX_COUNT_TYPE.SEAT_BASED
+                        }
+                    }
+                },
+                [PROJECT_PACKAGE_FEATURE_NAME.APW]: {
+                    enabled: false
+                },
+                [PROJECT_PACKAGE_FEATURE_NAME.AUDIT_LOGS]: {
```
adrians5j, line 32: Can you just add RECORD_LOCKING here too?
https://github.com/webiny/webiny-js/pull/4073/files#diff-28a27b0014482b8d1670cb2977ad2020082b4c703210cfce6bcd27c6adcd3635R35-R37

This was merged into `next` like yday, so it's new. But note that only after merging `next` you should be able to add it.

packages/api-headless-cms/__tests__/contentAPI/contentEntry.restore.test.ts
```diff
@@ -0,0 +1,308 @@
...
+                            type: "admin"
+                        }
+                    },
+                    error: null
+                }
+            }
+        });
+
+        /**
+         * We should NOT be able to restore an entry tha has not been moved to the trash bin.
```
adrians5j:
```suggestion
         * We should NOT be able to restore an entry that has not been moved to the trash bin.
```

packages/api-headless-cms-ddb/src/operations/entry/index.ts
```diff
@@ -600,6 +600,80 @@ export const createEntriesStorageOperations = (
...
+            storageEntry: initialStorageEntry
+        });
+
+        /**
+         * Then create the batch writes for the DynamoDB, with the updated data.
+         */
+        const items = records.map(record => {
+            return entity.putBatch({
+                ...record,
+                ...storageEntry
```
adrians5j: Am I reading this correct?

We update every retrieved record (revision) with `storagEntry`'s properties?

adrians5j (changes requested):
A couple of comments, mostly just nit picking. :)

packages/api-headless-cms/src/types/types.ts
leopuleo: Nice advice, I changed the methods, events and GraphQL mutation following `restoreFromBin` name convention.

packages/api-headless-cms-aco/__tests__/utils/createTestWcpLicense.ts
leopuleo: It should be there already; I merged next into this branch...can you have a look to see if everything is ok?

packages/api-headless-cms-aco/src/hooks/entry/onEntryBeforeRestoreFromBin.hook.ts
```diff
@@ -0,0 +1,37 @@
+import { HcmsAcoContext } from "~/types";
+import WebinyError from "@webiny/error";
+import { ROOT_FOLDER } from "@webiny/api-headless-cms/constants";
+
+export const onEntryBeforeRestoreFromBinHook = (context: HcmsAcoContext) => {
+    const { aco, cms } = context;
+
+    cms.onEntryBeforeRestoreFromBin.subscribe(async ({ entry }) => {
```
adrians5j, line 8: This hook looks like it has old code (prev logic), am I missing smt here?

Like, we're setting `entry.location.folderId` to `ROOT_FOLDER` in case the folder does not exist.

🤔

packages/api-headless-cms-ddb/src/operations/entry/index.ts
```diff
@@ -816,14 +911,12 @@ export const createEntriesStorageOperations = (
...
 
-        return items
-            .filter(item => item.deleted !== true)
-            .map(item => {
-                return convertFromStorageEntry({
-                    storageEntry: item,
-                    model
-                });
+        return items.map(item => {
+            return convertFromStorageEntry({
```
adrians5j, line 915: I see we've removed the `filter` call here. What was the reason? Just curious.

packages/api-headless-cms/src/crud/contentEntry/useCases/RestoreEntryFromBin/TransformEntryRestore.ts
```diff
@@ -0,0 +1,68 @@
+import { SecurityIdentity } from "@webiny/api-security/types";
+import { entryFromStorageTransform, entryToStorageTransform } from "~/utils/entryStorage";
+import { getDate } from "~/utils/date";
+import { getIdentity } from "~/utils/identity";
+import { validateModelEntryDataOrThrow } from "~/crud/contentEntry/entryDataValidation";
+import { CmsContext, CmsEntry, CmsEntryStorageOperationsMoveToBinParams, CmsModel } from "~/types";
+
+export class TransformEntryRestore {
```
adrians5j: `TransformEntryRestoreFromBin` maybe?

I saw you have the `RestoreFromBin` everywhere, so just dbl checking if we want to do the same here.

packages/api-headless-cms-ddb-es/src/operations/entry/index.ts
```diff
@@ -838,16 +848,18 @@ export const createEntriesStorageOperations = (
         /**
          * We update all ES records with data received.
          */
-        const updatedEntryLevelMetaFields = pickEntryMetaFields(entry, isEntryLevelEntryMetaField);
+        const updatedEntryMetaFields = pickEntryMetaFields(entry, isDeletedEntryMetaField);
```
adrians5j: I saw we have above:

 ```ts
const updatedDeletedMetaFields = pickEntryMetaFields(storageEntry, isDeletedEntryMetaField);
``` 

Is this needed? Can we use the above result?

adrians5j (changes requested):
A couple of minor things.

packages/api-headless-cms-ddb/src/operations/entry/index.ts
leopuleo: This is a leftover from a previous PR: at the moment, we are removing deleted entries decorating the default use case.

https://github.com/webiny/webiny-js/blob/next/packages/api-headless-cms/src/crud/contentEntry/useCases/GetRevisionsByEntryId/GetRevisionsByEntryIdNotDeleted.ts

packages/api-headless-cms-aco/src/hooks/entry/onEntryBeforeRestoreFromBin.hook.ts
leopuleo: Exactly! I decoupled the 2 pieces of logic. On entry restoration:

- by default we store back the original `folderId`
- after that, we check if the folder assigned to the entry still exists.

So the logic declared here (`onEntryBeforeRestoreFromBin.hook.ts`) does not change, we changed the way we store the location when we delete an entry.

packages/api-headless-cms/src/crud/contentEntry/useCases/RestoreEntryFromBin/TransformEntryRestore.ts
leopuleo: Done 👍 
Thank you!

packages/api-headless-cms-ddb-es/src/operations/entry/index.ts
leopuleo: Nice one, thank you!

adrians5j (approved):

PR #4081, opened 2024-04-10
Lambda@Edge Website Router - Use DDB Document Client Commands

adrians5j:
## Changes
Our Lambda@Edge function code was using DynamoDB command classes exported from the `@aws-sdk/client-dynamodb` package, and not `@aws-sdk/lib-dynamodb`. 

This caused a problem in the [`getTenantIdByDomain`](https://github.com/webiny/webiny-js/pull/4081/files#diff-2bddcac078c19e5e72c02412217b1aead6db3c17022962423885f496527481e2R46) function, because it was not returning a string (tenant ID), but an `{ S: TENANT_ID }` unmarshalled object.

Ultimately, this caused the website router not to work in cases where users had 2 or more tenants in their system (the above function only gets called in this case), which the log below made obvious:

image

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


adrians5j:
Closing because this fix needs to go out within the 5.39.4 patch release.

PR #4085, opened 2024-04-16
feat(api-headless-cms): empty HCMS trash bin via cron job

leopuleo:
## Changes
With this PR, we handle the deletion of HCMS entries found in the trash bin that were moved there before a specific date. 

This is the overall process description: 

1. Every day at 3:00 am UTC, an EventBridge event triggers our main api-graphql lambda.
2. a new `hcmsEntriesEmptyTrashBins` background task is triggered for each tenant registered. 

### `hcmsEntriesEmptyTrashBins`

The `hcmsEntriesEmptyTrashBins` task triggers a `hcmsEntriesEmptyTrashBinByModel` background task for each public model registered. 

### `hcmsEntriesEmptyTrashBinByModel`
This background task receives as input: 

- `modelId`: the model ID that we want to target
- `where`:  the query condition; for this use case, we are passing `deletedOn_lt` s calculated using the `WEBINY_TRASH_BIN_RETENTION_PERIOD_DAYS` env variable (default value 90).

The task lists the entries to delete according to the query received and triggers `hcmsEntriesDeleteTrashBinEntries` tasks.

### `hcmsEntriesDeleteTrashBinEntries`
This background task receives as input: 

- `modelId`: the model ID that we want to target
- `entryIds` is the list of entry IDs we want to delete.

### Related PR
https://github.com/webiny/webiny-upgrades/pull/29

## How Has This Been Tested?
Jest + manually

## Documentation
We must document how to set the trash bin retention period using the env variable (`WEBINY_TRASH_BIN_RETENTION_PERIOD_DAYS`).


packages/api-headless-cms-tasks/src/graphql.ts
```diff
@@ -0,0 +1,74 @@
...
+                                }
+                            });
+
+                            return new Response({
+                                id: response.id
+                            });
+                        }
+                    }
+                }
+            });
```
Pavel910: Do we really need to create types and mutations for every model? I think it's enough to just have one mutation that takes in a `modelId` as input: `emptyTrashBin(modelId: String)`. It will pollute the schema much less.

packages/api-headless-cms-tasks/src/tasks/entries/emptyTrashBinByModelTask.ts
```diff
@@ -0,0 +1,55 @@
...
+} from "~/types";
+
+export const createEmptyTrashBinByModelTask = () => {
+    return createPrivateTaskDefinition<
+        HcmsTasksContext,
+        IEmptyTrashBinByModelInput,
+        IEmptyTrashBinByModelOutput
+    >({
+        id: EntriesTask.EmptyTrashBinByModel,
+        title: "Headless CMS - Empty trash bin by model",
```
Pavel910, line 17: In the previous file, the title is written in "Title Case", but here you have it in "Sentence case".

packages/api-headless-cms-tasks/src/tasks/entries/emptyTrashBinsTask.ts
```diff
@@ -0,0 +1,45 @@
+import { createTaskDefinition } from "@webiny/tasks";
+import { ChildTasksCleanup } from "~/tasks/common";
+import { EntriesTask, HcmsTasksContext } from "~/types";
+
+export const createEmptyTrashBinsTask = () => {
+    return createTaskDefinition<HcmsTasksContext>({
+        id: EntriesTask.EmptyTrashBins,
+        title: "Headless CMS - Empty all trash bins",
```
Pavel910, line 8: Same here, use the same letter case for all tasks in the system.

packages/api-headless-cms-tasks/src/tasks/entries/useCases/EmptyTrashBinByModel/CreateDeleteEntriesTasks.ts
```diff
@@ -0,0 +1,96 @@
+import { ITaskResponseResult } from "@webiny/tasks";
+import { CmsEntryListParams } from "@webiny/api-headless-cms/types";
+import { EntriesTask, IDeleteTrashBinEntriesInput, IEmptyTrashBinByModelTaskParams } from "~/types";
+
+const DELETE_ENTRIES_IN_BATCH = 50;
+const DELETE_ENTRIES_WAIT_TIME = 5;
+
+export class CreateDeleteEntriesTasks {
```
Pavel910, line 8: Since these classes are all "use cases", I would recommend creating one generic interface, and use it everywhere, in all use cases, something like this:

```ts
interface IUseCase<TInput, TOutput> {
    execute(params: TInput) : Promise<TOutput>
}

class CreateDeleteEntriesTasks implements IUseCase<IEmptyTrashBinByModelTaskParams, ITaskResponseResult> {...}
```

packages/api-i18n/src/graphql/context.ts
```diff
@@ -219,6 +219,26 @@ const createBaseContextPlugin = () => {
...
+                    results.push(await cb(locale));
+                } finally {
+                    setContentLocale(initialLocale);
+                    setCurrentLocale("default", initialLocale);
+                }
+            }
+
+            return results;
+        };
+
```
Pavel910, line 241: Once this is merged, I would create a separate PR, to make this work with AsyncLocalStorage, so that this `withEachLocale` supports parallel execution. I'll be working on the same thing for tenants, so it would be nice to have this all in 5.40.

Pavel910 (changes requested):

Pavel910 (approved):

PR #4087, opened 2024-04-18
fix(api-headless-cms): rename `deleted` flag to `wbyDeleted`

leopuleo:
## Changes
With this PR, we rename the newly created `deleted` flag to `wbyDeleted`. This is necessary to avoid conflicting with user-defined 'deleted' fields by treating `wby` as a reserved field.

## How Has This Been Tested?
Jest + manually

## Documentation
We need to highlight the newly created field (`wbyDeleted`) inside the release docs.


Pavel910 (approved):
These are just renames, so I think this is ok.

PR #4101, opened 2024-04-29
fix(api-aco): verify existence of identity while emptying the trash bins

leopuleo:
## Changes
When FLP is enabled, we were encountering the error below while trying to list the entries to be deleted: 

```
Error: Cannot read properties of undefined (reading 'type')
```

We were not defining any `identity` when emptying the trash bin.

In the same PR, we corrected the context type name from `HeadlessCmsTasksContext` to `HcmsTasksContext` in the test files.

## Testing
This has been tested manually.

## Documentation
None available.


packages/api-aco/src/utils/FolderLevelPermissions.ts
```diff
@@ -74,8 +74,9 @@ export class FolderLevelPermissions {
             // At the moment, we only want FLP to be used with identities of type "admin".
             // This temporarily addresses the issue of API keys not being able to access content, because
             // FLPs doesn't work with them. Once we start adding FLPs to API keys, we can remove this check.
+            // Also, if the CRUD method is called from a background task, the identity might not exist at all.
```
adrians5j: Looks good, but let's do a quick refactor:

```ts
this.canUseFolderLevelPermissions = () => {
    const identity = this.getIdentity();

    // FLPs only work with authenticated identities (logged-in users).
    if (!identity) {
        return false;
    }

    // At the moment, we only want FLP to be used with identities of type "admin".
    // This temporarily addresses the issue of API keys not being able to access content, because
    // FLPs doesn't work with them. Once we start adding FLPs to API keys, we can remove this check.
    if (identity.type !== "admin") {
        return false;
    }

    return params.canUseFolderLevelPermissions();
};
```

adrians5j (changes requested):
Quick refactor, but LGTM.

packages/api-aco/src/utils/FolderLevelPermissions.ts
leopuleo: Thanks @adrians5j!

adrians5j (approved):

PR #4113, opened 2024-05-07
fix(ui): various RMWC and styles fixes

leopuleo:
## Changes
This PR: 

- fixes table cell alignment defined via props
- fixes table cell skeleton while loading
- fixes `<Select />` with size `medium` and `small`
- fixes `<Textarea />` label
- fixed HCMS table title skeleton
- fixed HCMS table filters button

## How Has This Been Tested?
Manually


packages/app-headless-cms/src/admin/components/ContentEntries/Header/Header.tsx
Pavel910, line 1: Why did you bring back the Grid? The goal of this change was to avoid using Grid, because it breaks on smaller widths, while `flex` really works great.

packages/app-headless-cms/src/admin/components/ContentEntries/Header/styled.tsx
Pavel910, line 1: This is related to my previous question.

Pavel910 (commented):
I just have one question about bringing back the Grid component.

Pavel910 (approved):

PR #4119, opened 2024-05-08
5.40.0 Fixes - Batch 1

adrians5j:
## Changes
This PR includes a couple of polishes for the 5.40.0 release.

#### 1. Added README.md In the New `extensions` Folder

For better UX, the `extensions` folder will now also include a short README file, just so it's easier for users to know what the folder is and how to get started with creating a new extension.

#### 2. Deployment Logs Always Included In CI

With 5.40.0, deployment logs are automatically hidden. But, most probably, this should not be happening in CI environments. So, we ensured that the logs are always shown in CI environments.

#### 3. Deployments - `allow-local-state-files` Flag

When deploying into production environment, an error will be thrown if the user is using local state files. This is simply because it's not recommended. Still, if the uses wishes to ignore the error, they can do it by passing the `allow-local-state-files` flag.

#### 4. Extensions Scaffolds - Updating Dependencies In `package.json`

When using the API / Admin scaffolds, respectively, the `apps/api/graphql/package.json` and `apps/admin/package.json` are now updated with the new package that was created in the process. Previously this wasn't the case.

#### 5. Extensions Scaffolds - Added `--dependencies` Parameter

When running the scaffold command in a non-interactive mode, users also now have the ability to pass the `--dependencies` flag in order to immediately add required dependencies in the new extension's `package.json`. For example:

```bash
yarn webiny scaffold new-extension --type admin --name customFilePreview --dependencies @webiny/app-serverless-cms
```

Multiple dependencies can also be added, by separating them with a comma, for example:

```bash
--dependencies @webiny/app-serverless-cms,@webiny/app-file-manager
```

#### 6. `@webiny/pulumi-aws` - `DEFAULT_PROD_ENV_NAMES` Constant

Throughout multiple files, we had the following line of code:

```
const productionEnvironments = app.params.create.productionEnvironments || ["prod"];
```

For easier maintenance, the `DEFAULT_PROD_ENV_NAMES` constant was introduced, and the above line has now been changed to:

```
const productionEnvironments = app.params.create.productionEnvironments || DEFAULT_PROD_ENV_NAMES;
```

#### 6. `@webiny/ui` - Removed `import "@rmwc/chip/styles";` Import

We've noticed Webpack reporting a weird error upon building both Admin and Website apps:

image

After investigation, we've detected that it's because of the `import "@rmwc/chip/styles";` statement in the `packages/ui/src/Chips/Chips.tsx` file. Removing it resolved the issue, and, most importantly, did not cause any visual regressions with the component.

#### 7. `dynamoDbElasticsearchTable` No Longer Present In DDB-Only Projects
The `dynamoDbElasticsearchTable` property would always be included when deploying the API app, both in DDB-only and DDB+ES projects. 

Although it wasn't making any serious issues, when deploying a DDB-only project, a warning from Pulumi would be printed out. 

image

With [this minor fix](https://github.com/webiny/webiny-js/pull/4119/files#diff-20da458def2cd6b0aaa7f9fd857c1aba4af337c1eb6a4b3c58b991c276b468d6R274-R280), this is no longer the case.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


adrians5j:

PR #4124, opened 2024-05-13
5.40.0 Fixes - Batch 2 (`themes/src`)

adrians5j:
## Changes
This PR moves all of the code in the `extensions/theme` folder to `extensions/theme/src`. This was made for consistency (the new scaffolds also follow this folder structure).

With that, one additional change we did here is the import of the `global.scss` file. Previously, the file was directly imported from Admin/Website React apps, but now, they are imported within the the `theme` package's code. 

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


packages/cwp-template-aws/template/ddb/dependencies.json
```diff
@@ -50,7 +50,7 @@
     "typescript": "4.7.4"
   },
   "workspaces": {
-    "packages": ["apps/admin", "apps/website", "apps/api/graphql", "extensions/theme", "packages/*"]
+    "packages": ["apps/admin", "apps/website", "apps/api/graphql", "packages/*", "extensions/theme"]
```
Pavel910, line 53: It should be `extensions/*`, just like `packages/*`, and then you never need to modify these rules again.

adrians5j: The thing is, users can also create extensions in e.g. `extensions/fm/api/xyz` folder.

Pavel910: If you think everyone will be using scaffolds, then probably this is ok.

adrians5j: Yes, every extension article we'll have will essentially start with scaffold run. Examples:

https://docs-webiny-com-git-release-5400-webiny.vercel.app/docs/file-manager/extending/customize-file-preview
https://docs-webiny-com-git-release-5400-webiny.vercel.app/docs/file-manager/extending/customize-file-actions

Tip: the Getting Started section is actually generated via [a React component](https://github.com/webiny/docs.webiny.com/pull/659/files#diff-5ea5a0a8cd8c4b551d0e1e9a1090329c96916375060cb64d5247db012f9b45b5).

adrians5j:

PR #4132, opened 2024-05-14
Admin Extension - Insert React Component Instead of Factory Function Call - Batch 4

adrians5j:
## Changes
With "Admin extension", instead of injecting a factory function call in `apps/admin/src/Extensions.tsx`, we're now injecting a regular React component.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


adrians5j:

PR #4135, opened 2024-05-15
5.40.0 Fixes - Batch 2 (`themes/src`)

adrians5j:
## Changes
This PR moves all of the code in the `extensions/theme` folder to `extensions/theme/src`. This was made for consistency (the new scaffolds also follow this folder structure).

With that, one additional change we did here is the import of the `global.scss` file. Previously, the file was directly imported from Admin/Website React apps, but now, they are imported within the the `theme` package's code. 

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


adrians5j:

PR #4137, opened 2024-05-15
fix(api-headless-cms-ddb-es): unpublish latest record [skip ci]

brunozoric:
## Changes
In DynamoDB + Elasticsearch systems not all records were properly marked as unpublished.

## How Has This Been Tested?
Jest and manual inspection of the database table.

### DIFF
--- a/packages/api-headless-cms-ddb-es/src/operations/entry/index.ts
+++ b/packages/api-headless-cms-ddb-es/src/operations/entry/index.ts
@@ -1454,7 +1454,7 @@ export const createEntriesStorageOperations = (
             })
         ];
         /**
-         * If we are unpublishing the latest revision, let's also update the latest revision entry's status in ES.
+         * If we are unpublishing the latest revision, let's also update the latest revision entry's status in both DynamoDB tables.
          */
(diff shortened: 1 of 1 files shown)


packages/api-headless-cms-ddb-es/src/operations/entry/index.ts
```diff
@@ -1461,6 +1461,15 @@ export const createEntriesStorageOperations = (
                 model
             });
 
+            items.push(
+                entity.putBatch({
+                    ...storageEntry,
+                    PK: partitionKey,
+                    SK: createLatestSortKey(),
+                    TYPE: createRecordType()
```
adrians5j:
```suggestion
                    TYPE: createLatestRecordType()
```

adrians5j: Let's also update the comment above:

```
If we are unpublishing the latest revision, let's also update the latest revision entry's status in both DynamoDB tables.
```

adrians5j (changes requested):

PR #4153, opened 2024-06-04
feat: expand object + DZ fields via settings parameter >open<

### LINKED ISSUE #4109: HCMS: Configure Dynamic Zone and Object field to be expanded by default

In our customer project, we see the requirement to always render a dynamic zone open with it's content. Right now, every time a DZ is rendered, it is collapsed and editors always have to spend an additional click to open it and see the inner content. 

To improve editor UX, it would be best to configure a DZ or an object (accordion renderer) to always be expanded when rendered initially. 

- The `AccordionItem` react component used in [dynamicZoneRenderer.tsx](https://github.com/webiny/webiny-js/blob/1d1b8721f4e0e230b4e0f261144b31850b9183b1/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/dynamicZoneRenderer.tsx#L34C14-L34C34) does indeed support an `open` property which I expect does open the accordion item by default
- I'd say, there are only two things we need to do:
  - Extend the [CmsModelFieldSettings](https://github.com/webiny/webiny-js/blob/1d1b8721f4e0e230b4e0f261144b31850b9183b1/packages/app-headless-cms-common/src/types/model.ts#L15) by an `open?: boolean` attribute or use it's definition `[key: string]: any` to support the new `open` configuration.
  - Pass the value of the `open` configuration to the `open` property of the `AccordionItem` in the [dynamic zone renderer](https://github.com/webiny/webiny-js/blob/1d1b8721f4e0e230b4e0f261144b31850b9183b1/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/dynamicZoneRenderer.tsx#L53) and the object renderers (single & multiple). Something like: 
```
<AccordionItem
  title={title}
  description={description}
  className={className || defaultClassName}
  open={field.settings?.open || false}
  >
    {children}
</AccordionItem>
```

_No response_

ma-schmidt-de:
## Changes
The new attribute `open` of interface `CmsModelFieldSettings` controls whether the accordion renderer of objects and dynamic zones are expanded by default.
BREAKING CHANGE: none
Closes #4109

Screenshot 2024-06-04 at 14 19 45
Screenshot 2024-06-04 at 14 20 41

## How Has This Been Tested?
Manually.

## Documentation
I'm planning to add a "How to" section below this on in webiny.com/docs:
https://www.webiny.com/docs/headless-cms/extending/content-models-via-code#how-to-set-a-default-value-of-a-field

### DIFF
--- a/packages/api-headless-cms/src/types/modelField.ts
+++ b/packages/api-headless-cms/src/types/modelField.ts
@@ -293,6 +293,10 @@ export interface CmsModelFieldSettings {
      * Disable full text search explicitly on this field.
      */
     disableFullTextSearch?: boolean;
+    /**
+     * Expand accordion appearances of dynamic zone and objects.
+     */
+    open?: boolean;
 
     /**
      * There are a lot of other settings that are possible to add, so we keep the type opened.
--- a/packages/app-headless-cms-common/src/types/model.ts
+++ b/packages/app-headless-cms-common/src/types/model.ts
@@ -21,6 +21,7 @@ export interface CmsModelFieldSettings<T = unknown> {
     models?: Pick<CmsModel, "modelId">[];
     templates?: CmsDynamicZoneTemplate[];
     imagesOnly?: boolean;
+    open?: boolean;
     [key: string]: any;
 }
 
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/dynamicZoneRenderer.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/dynamicZone/dynamicZoneRenderer.tsx
@@ -29,6 +29,7 @@ export type DynamicZoneContainerProps = {
     title?: string;
     description?: string;
     className?: string;
+    open?: boolean;
 };
 
 export const DynamicZoneContainer = makeDecoratable(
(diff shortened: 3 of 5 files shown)


Pavel910 (changes requested):
That's a nice PR @ma-schmidt-de, but let's make it even better. The `next` branch now has this new feature called "renderer settings". Check out the draft changelog: https://docs-webiny-com-git-release-5400-webiny.vercel.app/docs/release-notes/5.40.0/changelog#field-renderer-settings

This feature that you've implemented would make much more sense as a renderer settings, not a field setting in general. If you think about it, what does `open` mean on a `boolean` field? So let's move this flag to renderer settings for `dynamicZone` and `object` field renderers. That way each renderer can have its own very specific settings, and the general field definition is not cluttered with unrelated properties.

Here's a reference to how you would use these settings in a renderer: https://github.com/webiny/webiny-js/blob/next/packages/app-headless-cms/src/admin/plugins/fieldRenderers/ref/advanced/components/ReferencesDialog.tsx#L81

And here's how you define new renderer settings, on a renderer level:
https://github.com/webiny/webiny-js/blob/next/packages/app-headless-cms/src/admin/plugins/fieldRenderers/ref/advanced/detailedReferencesRenderer.tsx#L40-L42

The settings will be stored into the `renderer: { name: "...", settings: { ... } }`.

ma-schmidt-de:
Sounds great, @Pavel910 
I'll look into this and refactor my PR accordingly. 👍

ma-schmidt-de:
Let me actually open a new PR. I got mixed up a bit with branches in my fork. 🙂

ma-schmidt-de:
new PR: https://github.com/webiny/webiny-js/pull/4155

PR #4155, opened 2024-06-05
feat(app-headless-cms): add object and DZ renderer settings

### LINKED ISSUE #4109: HCMS: Configure Dynamic Zone and Object field to be expanded by default

ma-schmidt-de:
## Changes
Okay, this is the 2nd iteration of the new feature to expand object and dynamic zone HCMS fields by default. According to @Pavel910's [recommendation](https://github.com/webiny/webiny-js/pull/4153#pullrequestreview-2096283885), I'm using the new **renderer settings** feature to define an `open` attribute. If this attribute is set to `true`, the accordion of the object or dynamic zone field is expanded.

BREAKING CHANGE: none
Closes: #4109

## How Has This Been Tested?
Manually, with an in-code content model.
Screenshot 2024-06-05 at 20 16 10

## Documentation
We could use this implementation as an example for the section that describes the renderer settings feature.

### DIFF
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/object/multipleObjectsAccordion.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/object/multipleObjectsAccordion.tsx
@@ -23,6 +23,7 @@ import {
 } from "./StyledComponents";
 import { generateAlphaNumericLowerCaseId } from "@webiny/utils";
 import { FieldSettings } from "./FieldSettings";
+import { AccordionRenderSettings, getAccordionRenderSettings } from "../AccordionRenderSettings";
 
 const t = i18n.ns("app-headless-cms/admin/fields/text");
 
@@ -83,10 +84,11 @@ const ObjectsRenderer = (props: CmsModelFieldRendererProps) => {
     }
 
     const settings = fieldSettings.getSettings();
+    const open = getAccordionRenderSettings(field).open;
 
     return (
         <RootAccordion>
-            <AccordionItem title={field.label} description={field.helpText}>
+            <AccordionItem title={field.label} description={field.helpText} open={open}>
                 <DynamicSection
                     {...props}
                     emptyValue={{}}
@@ -140,6 +142,9 @@ const plugin: CmsModelFieldRendererPlugin = {
         },
         render(props) {
             return <ObjectsRenderer {...props} />;
+        },
+        renderSettings({ field }) {
+            return <AccordionRenderSettings field={field} />;
         }
     }
 };
--- a/packages/app-headless-cms/src/admin/plugins/fieldRenderers/object/multipleObjectsAccordion.tsx
+++ b/packages/app-headless-cms/src/admin/plugins/fieldRenderers/object/multipleObjectsAccordion.tsx
@@ -84,7 +84,7 @@ const ObjectsRenderer = (props: CmsModelFieldRendererProps) => {
     }
 
     const settings = fieldSettings.getSettings();
-    const open = getAccordionRenderSettings(field).open;
+    const { open } = getAccordionRenderSettings(field);
 
     return (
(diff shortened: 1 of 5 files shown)


ma-schmidt-de:
Testing the no-code content model editor, I saw that my implementation acts as expected. The switch is rendered and a `settings` object is added to the field's renderer.
Screenshot 2024-06-05 at 20 22 53

Yet, saving the model the `CmsUpdateContentModel` mutation fails due to the fact, that the default fields `Title`, `Description` and `Image` don't have a `settings` object attached to their field renderer. Is this a known issue? 🤔 
Screenshot 2024-06-05 at 20 25 01

When I create an empty content model and add an object, dynamic zone or arbitrary field it works just fine.

packages/app-headless-cms/src/admin/plugins/fieldRenderers/object/multipleObjectsAccordion.tsx
```diff
@@ -83,10 +84,11 @@ const ObjectsRenderer = (props: CmsModelFieldRendererProps) => {
     }
 
     const settings = fieldSettings.getSettings();
+    const open = getAccordionRenderSettings(field).open;
```
Pavel910: This is a stylistic thing, but let's use the syntax that we usually use everywhere in our code base:

```ts
const { open } = getAccordionRenderSettings(field);
```

Pavel910 (changes requested):
Just one little change here and we're good to go.

On the topic of mutation error, the settings should be either `undefined`, or an object. I will have a look at this right there in your PR, and will push a fix for this, so we expedite merging of this PR.

Thank you!

Pavel910:
@ma-schmidt-de I pushed a fix for that issue with existing fields, so `settings` will always fall back to `{}` if they don't exist on the field.

ma-schmidt-de:
@Pavel910 thanks for adding the fix! I polished the code style according to your feedback.
This looks complete to me now. 👍

Pavel910 (approved):

Pavel910:
It's approved, let's wait for CI to finish, and we'll merge it then.

Pavel910:
Thanks @ma-schmidt-de for your contribution! Looking forward to more of these!! 🚀

PR #4162, opened 2024-06-10
fix(api-form-builder): check both global and form recaptcha settings

Pavel910:
## Changes
This PR fixes a problem with Form Builder, related to reCaptcha. If you had reCaptcha enabled in your global Form Builder settings, it would apply to all forms, regardless of individual form settings. So even if you would disable recaptcha on a particular form, recaptcha checks would still kick in.

With this PR, we take per-form settings into account to determine whether recaptcha token verification is necessary.

## How Has This Been Tested?
Manually.


Pavel910:
This code is pushed to `dev` branch, and we'll merge it into `next` from there.

PR #4178, opened 2024-06-21
fix(app-admin): allow users to adjust the SplitView columns width across all apps 

leopuleo:
## Changes
With this PR, we are updating the `SplitView`, `LeftPanel` and `RightPanel` to allow users to adjust the columns' width.

There is no need to update the existing `<SplitView>`. 

We include the `autoSaveId` prop for the ACO-powered applications to store the latest column width in `localStorage`.

The function `generateAutoSaveId` is exported from `@webiny/app-admin` to create a consistent localStorage key that tracks column width. The function receives the following params: 

- tenantId: `root`
- localCode: `en-US`
- applicationId: `cms:{modelId}`

https://github.com/webiny/webiny-js/assets/2866531/f9352bed-9968-4f6a-a71e-58a10d658dc7

## How Has This Been Tested?
Manually + Cypress

## Documentation
@swapnilmmane, perhaps it would be beneficial for us to add an article to the user guide.


packages/app-page-builder/src/admin/views/Pages/Table/index.tsx
```diff
@@ -16,9 +23,17 @@ import { PageListWithConfig } from "~/admin/config/pages";
...
 const View = () => {
     const { currentFolderId } = useNavigateFolder();
+    const { getCurrentLocale } = useI18N();
+    const { tenant } = useTenancy();
+
+    const autoSaveId = useMemo(() => {
+        const localeCode = getCurrentLocale("content");
+        const applicationId = "pb:page";
+        return generateAutoSaveId(tenant, localeCode, applicationId);
+    }, [getCurrentLocale, tenant]);
```
Pavel910: This code is very repetitive, and the only thing that changes is the `applicationId`. Let's expose a hook on the `SplitView` component, like `SplitView.useLayoutId`, and then use that hook to generate the `id` by passing an application id:

```tsx
const MyView = () => {
  // If CMS....
  const { model } = useModel();
  const layoutId = SplitView.useLayoutId(`cms:${model.modelId}`);

  // If PB...
  const layoutId = SplitView.useLayoutId("pb:page");

  return <SplitView layoutId={layoutId}>{/*.....*/}</SplitView>;
};
```

Also, the `autoSaveId` is very vague in the context of Webiny, and is a very specific prop for the underlying library. What does it mean `autoSaveId`, what are you auto saving? It's unclear. I suggest we change this to `layoutId`. SplitView has a certain layout, and that's its `id`. We can then do various things with this `layoutId`, doesn't have to be related to saving at all.

Pavel910 (changes requested):
Left one request, but it applies to multiple files, you'll see what I'm talking about.

PR #4182, opened 2024-06-27
Fix/fixing windows 11 build

### LINKED ISSUE #4181: Windows 11 build fails
### Version

0.0.0

### Operating System

windows 11

### Browser

_No response_

clean clone then run 
```
yarn
yarn build
``` 

Build to be successful 

### What do you see instead?

```
D:\webiny-js\node_modules\fs-extra\lib\copy-sync\copy-sync.js:185
    throw new Error('Source and destination must not be the same.')
          ^
Error: Source and destination must not be the same.
    at checkPaths (D:\webiny-js\node_modules\fs-extra\lib\copy-sync\copy-sync.js:185:11)
    at Object.copySync (D:\webiny-js\node_modules\fs-extra\lib\copy-sync\copy-sync.js:25:20)
    at D:\webiny-js\scripts\buildPackages\src\getBatches.ts:87:16
    at Generator.next (<anonymous>)
    at fulfilled (D:\webiny-js\scripts\buildPackages\src\getBatches.ts:5:58)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
```

### Additional information

_No response_

### Possible solution

There are two issues at play. 1 is `rimraf` and the other is `fs-extra`.

### rimraf

In version 4 they disabled globing in file paths by default and that has to be enabled as an option. 

- https://github.com/webiny/webiny-js/blob/b609e37fc81bc24ef23c436d4d1849f0be6f9e70/packages/project-utils/packages/buildPackage.js:15 needs to be replaced with ` rimraf.sync(join(cwd, "*.tsbuildinfo"), {glob:true});` 
- https://github.com/webiny/webiny-js/blob/4f3a75b0b1028e42689b7ea69a3e25925b7b3689/packages/project-utils/bundling/function/buildHandler.js:13 needs to be replaced with `rimraf.sync(join(cwd, "*.tsbuildinfo"), {glob:true});` 

### fs-extra

Windows 11 has a new file system with their 'dev drives' feature. It uses ReFS. The older version of `fs-extra` that's being used for builds does some interesting things when trying to compare source and destinations can fails. Its comparing inodes of the source and destination. This is where the issues is being surfaced. 

Bumping versions of 'fs-extra' to version 11.2.0 fixes that issues. 

I would put up a pull request but I am not sure the best way to do that

will-fagerstrom:
## Changes
Builds on windows 11 dev drive fail for two reasons. 
* `rimraf` globing is not on by default in version 5
* `fs-extra` out of date not working with windows 11's dev-drive ReFS file system 

Closes #4181

## How Has This Been Tested?
`yarn build` successfully completes.

## Documentation
Should ne no change to any user or developer


Pavel910:
@will-fagerstrom thanks, I triggered tests to see if everything is 🟢

adrians5j (approved):
Looks good, let's have this for 5.40.2, just so we have some time to potentially catch undiscovered issues with the lib update.

Thanks @will-fagerstrom! 🍻 🍻 🍻

PR #4199, opened 2024-07-11
feat(handler-graphql): add support for resolver decoration

Pavel910:
## Changes
This PR adds support for GraphQL resolver decoration., and improves GraphQL related types (mostly typeDefs and resolvers).

```ts
// Decorator #1
const lowerCaseName = createResolverDecorator<any, any, Context>(
  (resolver) => async (parent, args, context, info) => {
    // Run the original resolver
    const name = await resolver(parent, args, context, info);

    // Return a modified value.
    return name.toLowerCase();
  }
);

// Decorator #2
const listBooks = createResolverDecorator(() => async () => {
  return [{ name: "Article 1" }];
});

// Target fields to decorate
const decorator1 = createGraphQLSchemaPlugin({
  resolverDecorators: {
    "Query.books": [listBooks],
    "Book.name": [lowerCaseName],
  },
});
```

## How Has This Been Tested?
Jest tests.


packages/handler-graphql/src/ResolverDecoration.ts
```diff
@@ -0,0 +1,23 @@
...
+            if (!decorators) {
+                continue;
+            }
+
+            const existingDecorators = this.decorators[key] ?? [];
+            this.decorators[key] = [...existingDecorators, ...decorators];
+        }
+    }
+
+    decorateResolvers(resolvers: any) {
```
adrians5j: `any` here, on purpose? 🤔

adrians5j (changes requested):

packages/handler-graphql/src/types.ts
```diff
@@ -23,23 +25,30 @@ export type GraphQLFieldResolver<
...
-    | GraphQLScalarType
-    | GraphQLFieldResolver<any, Record<string, any>, TContext>
-    | { [property: string]: Resolvers<TContext> };
+export type Resolvers<TContext> = IResolvers<any, TContext>;
+
+export type ResolverDecorator<TSource = any, TContext = any, TArgs = any> = ResolversComposition<
+    GraphQLFieldResolver<TSource, TContext, TArgs>
+>;
+
+export type ResolverDecorators = GenericRecord<string, ResolverDecorator[]>;
```
adrians5j, line 46: Could've we used this in L51?

adrians5j (changes requested):

packages/handler-graphql/src/ResolverDecoration.ts
Pavel910: Nice, missed by accident. Fixed.

adrians5j (approved):

PR #4230, opened 2024-08-20
feat(admin-ui): handle typography - add Heading and Text components 

leopuleo:
## Changes
With this PR, we add two new UI components to handle typography: `Heading` and `Text`.
The previously used `Typography` component will render `Heading` or `Text` component based on the value inside the `use ` prop. 

### Heading
Used to render heading elements, such as `h1` -> `h6`.

It accepts the following props: 
- **level**: it controls the heading level from 1 to 6 and how it renders (as defined in Figma);
- **as**: used to override the renderer tag, so we can make an `<h3>` to look like a `<h1>` visually (optional).
- **text**: used to pass the text to render.

### Text
Used to render text elements, such as `div` or `span`.

It accepts the following props: 
- **size**: it controls the text size as defined in Figma;
- **as**: used to specify the tag used by the component: `div` or `span` (optional).
- **text**: used to pass the text to render


packages/ui-new/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,49 @@
...
+    defaultVariants: {
+        level: 1
+    }
+});
+
+export interface HeadingProps
+    extends React.HTMLAttributes<HTMLHeadingElement>,
+        VariantProps<typeof headingVariants> {
+    as?: HeadingTags;
+    text: string | React.ReactNode;
```
adrians5j: Do we need `string` here? 🤔

packages/ui-new/src/Heading/Heading.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+export const Heading = ({ level, text, className, as }: HeadingProps) => {
+    // Ensure `level` is a valid number, or fallback to a default value 1
+    const validatedLevel = level && level in TAG_MAP ? level : 1;
+
+    // Choose the tag: prefer `as`, otherwise use `TAG_MAP[validatedLevel]`
+    const Tag = as || TAG_MAP[validatedLevel] || TAG_MAP[1];
```
adrians5j: I appreciate the validation, but since TS is here, I doubt we even need to do any validation. Especially because this is a simple prop.

But if you want to leave it, I'd address this (because `TAG_MAP[validatedLevel]` is already OK):

```suggestion
    const Tag = as || TAG_MAP[validatedLevel];
```

packages/ui-new/src/Text/Text.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { cn } from "~/utils";
+
+export type TextTags = "span" | "div";
+
+const textVariants = cva("font-sans font-normal", {
+    variants: {
+        size: {
+            xl: "text-xl leading-normal",
+            lg: "text-base leading-normal",
+            md: "text-sm leading-normal",
```
adrians5j, line 12: I guess we'll also need those variants, like `muted`? I guess you wanted to leave this after the call with Kreso?

packages/ui/src/Typography/Typography.tsx
```diff
@@ -1,21 +1,41 @@
...
+    /**
+     * @deprecated
+     */
     style?: React.CSSProperties;
     tag?: string;
 }
 
 /**
- * Use Ripple component to display a list of choices, once the handler is triggered.
+ * @deprecated This component is deprecated and will be removed in future releases.
```
adrians5j, line 17: Question: does your IDE highlight the deprecation of the component when it's used?

adrians5j (changes requested):
Looks good... just a couple of short questions. :)

packages/ui/src/Typography/Typography.tsx
leopuleo: Yes 👍 

CleanShot 2024-08-21 at 11 40 49

packages/ui-new/src/Text/Text.tsx
leopuleo: Exactly, I'm leaving all the color-related variants behind

packages/ui-new/src/Heading/Heading.tsx
leopuleo: You are right, thank you!

adrians5j (approved):

PR #4233, opened 2024-08-21
New Admin UI - Button

adrians5j:
## Changes
This PR creates the Button component. Visuals still need to be resolved though.

image

## Notes
1. Dialog cancel button looks weird. We'll address this one we get to the dialog component.
2. Icon colors different on different buttons (new page / new folder)

## How Has This Been Tested?
Manually.

## Documentation
N/A


packages/admin-ui/src/Button/Button.tsx
```diff
@@ -29,24 +30,27 @@ const buttonVariants = cva(
...
-    }
-);
-Button.displayName = "Button";
+const ButtonBase = React.forwardRef<HTMLButtonElement, ButtonProps>((props, ref) => {
+    const { className, variant, size, asChild = false, text, children, ...rest } = props;
+    const Comp = asChild ? Slot : "button";
+
+    return (
+        <Comp className={cn(buttonVariants({ variant, size, className }))} ref={ref} {...rest}>
+            {text || children}
```
leopuleo: I would remove `children` if we are going to use `text`.

packages/admin-ui/src/Button/IconButton.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { Button, ButtonProps } from "./Button";
+import { makeDecoratable } from "@webiny/react-composition";
+
+interface IconButtonProps extends ButtonProps {
+    icon: React.ReactNode;
+}
+
+const IconButtonBase = (props: IconButtonProps) => {
```
leopuleo: Do we need this component? I would rather prefer to have one single `Button` component that receives `icon` as prop.  

We have 3 possible text/icon combinations: 

- Only text
- Only icon
- Text + icon

I believe we can achieve it with one single component, but I might be wrong :) 

Also, as showed in the Figma file, we should allow developers to specify the icon position inside the button: `leading` or `trailing`.

packages/ui/src/Button/Button.tsx
```diff
@@ -1,184 +1,37 @@
...
-            ripple={ripple}
-            className={classNames("webiny-ui-button", className)}
-            data-testid={props["data-testid"]}
-        >
-            {children}
-        </RmwcButton.Button>
-    );
+// We needed this default export for backwards compatibility.
+const Button = (props: ButtonProps) => {
+    return <ButtonBase {...props} />;
```
leopuleo: We might want to handle the deprecated `small` prop, using the new `size` prop.

packages/ui/src/Button/ButtonDefault.tsx
```diff
@@ -0,0 +1,6 @@
+import React from "react";
+import { Button, ButtonProps } from "./Button";
+
+export const ButtonDefault = (props: ButtonProps) => {
```
leopuleo, line 10: I would add a `@deprecated` comment to the components too.

This is the comment I use:


```
@deprecated This component is deprecated and will be removed in future releases. 
Please use the `<Button />` component from the `@webiny/admin-ui` package instead.
```

packages/ui/src/Button/ButtonIcon.tsx
```diff
@@ -0,0 +1,7 @@
+import React from "react";
+
+export interface ButtonIconProps {
+    icon: React.ReactNode;
+}
+
+export const ButtonIcon = (props: ButtonIconProps) => <>{props.icon}</>;
```
leopuleo: As we discussed, we need the `Icon` component. I would continue to use the `<Icon />` component from `@webiny/ui`. 

Internally we will re-wire it to the new component.

leopuleo (changes requested):
Looks good! I left a couple of questions/comments.

packages/ui/src/Button/ButtonIcon.tsx
adrians5j: Fixed.

packages/ui/src/Button/ButtonDefault.tsx
adrians5j: Added.

packages/admin-ui/src/Button/Button.tsx
adrians5j: Done.

packages/admin-ui/src/Button/IconButton.tsx
adrians5j: Done, but let me know what you think about the approach I took. We can discuss alternatives too, like some libs use `leftSection` / `rightSection` props instead of mentioning the word `icon`.

packages/ui/src/Button/useMappedButtonProps.ts
```diff
@@ -0,0 +1,27 @@
+import React, { useMemo } from "react";
+import { ButtonProps as AdminUiButtonProps } from "@webiny/admin-ui";
+
+interface UseButtonPropsParams {
+    children?: React.ReactNode;
+    label?: string;
+    small?: boolean;
+}
+
+export const useMappedButtonProps = (params: UseButtonPropsParams) => {
```
leopuleo, line 10: 😍

leopuleo (approved):

PR #4239, opened 2024-08-27
feat(admin-ui): Tooltip component

leopuleo:
## Changes
With this PR, we are adding two new components: 

- `Tooltip`
- `Providers`

### Tooltip
We are exporting a props-based flat component based on the following [radix component](https://www.radix-ui.com/primitives/docs/components/tooltip). The new component is fully compatible with the old Tooltip component.

### Providers
`Tooltip` needs a Provider to work. We are creating a `Providers` component to group all these UI-based providers and use it inside the `app-admin` package.


packages/admin-ui/src/Tooltip/Tooltip.tsx
```diff
@@ -1,30 +1,65 @@
...
+    extends Omit<TooltipPrimitive.TooltipContentProps, "content" | "children"> {
+    align?: TooltipPrimitive.TooltipContentProps["align"];
+    content: React.ReactNode;
+    onOpenChange?: TooltipPrimitive.TooltipProps["onOpenChange"];
+    showArrow?: boolean;
+    side?: TooltipPrimitive.TooltipContentProps["side"];
+    trigger: React.ReactNode;
+}
+
+export const Tooltip = ({
```
adrians5j: Was this made decoratable somewhere? Did I miss it? 🤔

packages/admin-ui/src/Tooltip/Tooltip.tsx
```diff
@@ -1,30 +1,65 @@
...
+    align,
+    content,
+    onOpenChange,
+    showArrow = true,
+    side,
+    trigger,
+    ...props
+}: TooltipProps) => {
```
adrians5j, line 53: I'm having some consistency-related questions in my head... Like, here, we're using `content` and `trigger` props, and then, internally, we're using `TooltipContent` and `TooltipTrigger` components. But then again, with the tabs component you showed me today, there we're actually using `content=[<Tab .../>, <Tab .../>, ...]`. In other words, we're importing extra components from the UI package.

Are we okay with this slight inconsistency?

After that little Slack chat we had, I've actually followed the Tab component approach with the Card component I'm working on. Maybe we can discuss this further via a call.

packages/ui/src/Tooltip/Tooltip.tsx
```diff
@@ -34,38 +26,43 @@ export interface TooltipProps {
...
-    public override state = {
-        tooltipIsOpen: false
-    };
+export const Tooltip = ({
+    children,
+    content,
+    placement,
+    className = "webiny-ui-tooltip"
+}: TooltipProps) => {
+    const mapPlacementToSideAndAlign = useCallback((placement?: string) => {
```
adrians5j: This could've actually also been done with a `useMemo`, so the `side` and `align` are immediately returned. With this change:

1. there would be no need to do a function call below
2. would be also more performant

packages/ui/src/Tooltip/Tooltip.tsx
```diff
@@ -1,20 +1,12 @@
...
-    | (() => React.ReactChild)
-    | React.ReactChild
-    | React.ReactFragment
-    | React.ReactPortal;
+import React, { useCallback } from "react";
+import { Tooltip as TooltipAdmin, TooltipProps as TooltipPropsAdmin } from "@webiny/admin-ui";
 
 export interface TooltipProps {
     // A component (eg. button) which will trigger the tooltip.
     children: React.ReactNode;
```
adrians5j, line 6: I'm assuming you did not put deprecation notices on props since the whole component is marked as deprecated, right?

adrians5j (changes requested):
Looks good overall. Left a couple of questions.

packages/ui/src/Tooltip/Tooltip.tsx
leopuleo: Yes, correct 👍

packages/ui/src/Tooltip/Tooltip.tsx
leopuleo: Good point, thank you!

packages/admin-ui/src/Tooltip/Tooltip.tsx
leopuleo: Our approaches might differ based on how the components are structured.

For this particular case, passing the TooltipContent and TooltipTrigger components as props might be unnecessary since the Tooltip Root component will always have one content and one trigger.

This is different from components like Tabs or Accordion, which handle multiple children and offer more customization options.

I’m happy to chat about this on a call—there’s always a chance I could be off base here! 😊

adrians5j: Yeah, makes sense. 

One of the advantages of allowing users to pass `TooltipContent` and `TooltipTrigger` components is that this way they also can pass extra props to them, like maybe `className` or something similar. Alternative is not to have them exposed, but do it the way you did it here, in which case we'd need to expose a prop or two, so that users can pass extra props to these two components. And then I agree with you... not sure how viable of a solution is to just start adding props for these kinds of cases.

A disadvantage of having users pass these components is that now they need to know about them / import them, and then also, I saw you're adding `asChild` on `TooltipTrigger`, which means we'd need to set that for them internally, so they don't have to deal with it.

All in all, still not sure... I was also thinking maybe, on the Tab component side, instead of passing `TabsTrigger` and `TabsContent`, we pass POJOs.. but I think that's not good, and I like more what you already have with Tabs component.

adrians5j (approved):

PR #4242, opened 2024-08-28
feat(admin-ui): Tabs component

leopuleo:
## Changes
This PR adds the `Tabs` component to the `@webiny/admin-ui` package. 

We are exporting a props-based flat component based on the following [radix component](https://www.radix-ui.com/primitives/docs/components/tabs). 

The new component is fully compatible with the old `Tabs` component and it needs some refactoring when UI tokens are defined.


packages/admin-ui/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,148 @@
...
+                ))}
+            </TabsRoot>
+        );
+    }
+);
+TabsBase.displayName = TabsPrimitive.Root.displayName;
+
+const Tabs = makeDecoratable("Tabs", TabsBase);
+
+export { Tabs, TabsProps, TabsContent, TabsContentProps, TabsTrigger, TabsTriggerProps };
```
adrians5j:
```suggestion
export { Tabs, TabsProps, TabsContent, TabsContentProps, TabsTrigger, type TabsTriggerProps };
```

adrians5j (changes requested):
Quick comment.

packages/admin-ui/src/Tabs/Tabs.tsx
leopuleo: 🙏

packages/admin-ui/src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+        return (
+            <TabsRoot ref={ref} defaultValue={defaultValue} {...props}>
+                <TabsList>
+                    {triggers.map(trigger => (
+                        <React.Fragment key={`tab-trigger-${trigger.key}`}>
+                            {React.cloneElement(trigger, { size })}
+                        </React.Fragment>
+                    ))}
+                </TabsList>
```
adrians5j: Nit pick, but could be simplifed. Same could be done for `contents.map` code below.

```suggestion
                <TabsList>
                    {triggers.map(trigger =>
                        React.cloneElement(trigger, { size, key: `tab-trigger-${trigger.key}` })
                    )}
                </TabsList>
```

adrians5j (approved):
Minor nitpick, but I'd say we can merge.

PR #4243, opened 2024-08-30
feat(admin-ui): Label and Switch components

leopuleo:
## Changes
This PR adds the `Label` and `Switch` components to the `@webiny/admin-ui` package.

- **Label**: simple `<label>` component to use in combination with other form-related components
- **Switch**

The new components are fully compatible with the old ones, and they need some refactoring when UI tokens are defined.

### `generateId()` utility
New utility that receives an optional `initialId` and returns a string. Based on `generateId` from `@webiny/utils`.


packages/admin-ui/src/utils.ts
```diff
@@ -4,3 +4,10 @@ import { twMerge } from "tailwind-merge";
 export function cn(...inputs: ClassValue[]) {
     return twMerge(clsx(inputs));
 }
+
+export const generateId = (initialId?: string): string => {
```
adrians5j: Can we use ?

```
import { generateId } from "@webiny/utils/generateId";
```

adrians5j (changes requested):
Quick comment.

packages/admin-ui/src/utils.ts
leopuleo: 👍

packages/admin-ui/src/Label/Label.tsx
```diff
@@ -10,12 +9,21 @@ const labelVariants = cva(
...
+interface LabelProps
+    extends Omit<React.ComponentPropsWithoutRef<typeof LabelPrimitive.Root>, "children">,
+        VariantProps<typeof labelVariants> {
+    text: React.ReactNode;
+}
+
+const LabelBase = React.forwardRef<React.ElementRef<typeof LabelPrimitive.Root>, LabelProps>(
+    ({ className, text, ...props }, ref) => (
+        <LabelPrimitive.Root ref={ref} className={cn(labelVariants(), className)} {...props}>
+            {text}
```
adrians5j, line 21: Does it make sense to use `Text` component here?

packages/admin-ui/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+interface SwitchProps
+    extends React.ComponentPropsWithoutRef<typeof SwitchPrimitives.Root>,
+        VariantProps<typeof switchVariants> {
+    label: React.ReactNode;
+}
+
+const SwitchBase = React.forwardRef<React.ElementRef<typeof SwitchPrimitives.Root>, SwitchProps>(
+    ({ id: initialId, className, label, labelPosition, ...props }, ref) => {
+        const id = generateId(initialId);
```
adrians5j: I understand that if I pass a value to `generateId`, that it will always return that same value back to me. But, if I don't pass an `id` prop to `Switch` component, then the fn will generate an ID for me. In that case, on every re-render of the component,  we'd always get a new ID here. 

So, I'd say we memoize the result of the `generateId` call? Wdyt?

Hopefully I did not miss anything here. 🤞🏻

packages/admin-ui/src/utils.ts
```diff
@@ -1,6 +1,14 @@
...
+    if (initialId !== undefined && initialId !== null) {
+        return String(initialId);
+    }
+    return "wby-" + baseGenerateId(4);
+};
```
adrians5j: Couldn't this just be?

```suggestion
export const generateId = (initialId?: string) => {
    if (initialId) {
        return initialId;
    }
    return "wby-" + baseGenerateId(4);
};
```

The other thing is.... I'd maybe just do this `if (initialId...` logic in the Switch component and not bring this extra logic into a simple `generateId` function. Kinda makes it easer to grasp `generateId`. It generates a random ID and that's basically it.

Maybe it's just subjective, so if you'd still leave this as is, let's leave it.

packages/admin-ui/src/Switch/Switch.tsx
```diff
@@ -0,0 +1,56 @@
...
+            <div className={cn(switchVariants({ labelPosition, className }))}>
+                {label && <Label text={label} htmlFor={id} />}
+                <SwitchPrimitives.Root
+                    className={cn(
+                        "peer inline-flex h-6 w-11 shrink-0 cursor-pointer items-center rounded-full border-2 border-transparent transition-colors focus-visible:outline-none focus-visible:ring-2 focus-visible:ring-ring focus-visible:ring-offset-2 focus-visible:ring-offset-background disabled:cursor-not-allowed disabled:opacity-50 data-[state=checked]:bg-primary data-[state=unchecked]:bg-input"
+                    )}
+                    {...props}
+                    ref={ref}
+                    id={id}
+                    aria-label={label ? String(label) : "Switch"}
```
adrians5j: I can see that label can be string, but also `React.ReactNode`. So, `aria-label` would be set incorrectly if I've sent something like this:
image

Maybe we just do `typeof label === 'string' ? label || "Switch"`

P.S. In general, great idea of using `aria-label`. I'm totally unaware of ARIA stuff and how one should use these attrs.

packages/ui/src/Switch/Switch.tsx
```diff
@@ -1,36 +1,63 @@
...
+    description?: string;
+
+    /** Additional CSS classes to apply to the control. */
+    className?: string;
+} & FormComponentProps<boolean>;
 
 /**
- * Switch component can be used to store simple boolean values.
+ * @deprecated This component is deprecated and will be removed in future releases.
+ * Please use the `<Switch />` component from the `@webiny/admin-ui` package instead.
```
adrians5j: BTW something super minor I noticed before, but forgot to mention. 

```suggestion
 * Please use the `Switch` component from the `@webiny/admin-ui` package instead.
```

Basically <Switch /> is actual component written in JSX.

adrians5j (changes requested):
A couple of minor comments.

packages/admin-ui/src/Switch/Switch.tsx
leopuleo: You are right, thank you!

packages/admin-ui/src/Label/Label.tsx
leopuleo: I don't think so, it needs to be a `<label>` element that labels a form control. I would keep "dummy" `Text` component separated from `Label` because: 

- they have different semantic meaning;
- in our Figma file, labels have some distinctive properties

packages/admin-ui/src/Switch/Switch.tsx
leopuleo: You are right, thank you!

packages/ui/src/Switch/Switch.tsx
leopuleo: Ok, I will apply these changes to other components too.

adrians5j (approved):
Let's merge :)

PR #4244, opened 2024-08-30
New Admin UI - Elevation, Icon, and Card Components

adrians5j:
This PR creates the Card component.

Had to pull some fixes from [here](https://github.com/shadcn-ui/ui/pull/2439/files), and now it looks good even when there's no header/footer provided.

image


packages/admin-ui/src/Card/Card.tsx
```diff
@@ -1,56 +1,104 @@
...
+CardHeaderBase.displayName = "CardHeader";
+
+const CardHeader = makeDecoratable("CardHeader", CardHeaderBase);
 
-const CardDescription = React.forwardRef<
+const CardTitleBase = React.forwardRef<
+    HTMLParagraphElement,
+    React.HTMLAttributes<HTMLHeadingElement>
+>(({ className, ...props }, ref) => (
+    <h3
```
leopuleo: Let's use the `<Heading />` component.

packages/admin-ui/src/Card/Card.tsx
```diff
@@ -1,56 +1,104 @@
...
+));
+CardTitleBase.displayName = "CardTitle";
+
+const CardTitle = makeDecoratable("CardTitle", CardTitleBase);
+
+const CardDescriptionBase = React.forwardRef<
     HTMLParagraphElement,
     React.HTMLAttributes<HTMLParagraphElement>
 >(({ className, ...props }, ref) => (
     <p ref={ref} className={cn("text-sm text-muted-foreground", className)} {...props} />
```
leopuleo: Let's use the `<Text>` component.

packages/admin-ui/src/Card/Card.tsx
```diff
@@ -1,56 +1,104 @@
...
-        <div ref={ref} className={cn("flex items-center p-6 pt-0", className)} {...props} />
+        <div ref={ref} className={cn("flex items-center", className)} {...props} />
     )
 );
-CardFooter.displayName = "CardFooter";
+CardFooterBase.displayName = "CardFooter";
+
+const CardFooter = makeDecoratable("CardFooter", CardFooterBase);
+
+interface CardProps extends Omit<CardRootProps, "content"> {
```
leopuleo: I cannot find any reference to the close icon. Is it intentional?

CleanShot 2024-09-02 at 11 25 11

packages/ui/src/Elevation/Elevation.tsx
```diff
@@ -19,15 +19,12 @@ export type ElevationProps = {
...
  */
-const Elevation = React.forwardRef((props: ElevationProps, ref: React.Ref<any>) => {
-    return (
-        <RmwcElevation ref={ref} {...props}>
-            {props.children}
-        </RmwcElevation>
-    );
-});
+const Elevation = (props: ElevationProps) => {
+    return <Card {...props} content={<CardContent>{props.children}</CardContent>} />;
```
leopuleo: I would create an `Elevation` component in the new `admin-ui` package and reference it here and inside the `Card` component.

packages/admin-ui/src/Card/Card.tsx
```diff
@@ -1,56 +1,104 @@
...
+    headerTitle?: React.ReactElement<typeof CardTitleBase>;
+    headerDescription?: React.ReactElement<typeof CardDescriptionBase>;
+    content?: React.ReactElement<typeof CardContentBase>;
+    footer?: React.ReactElement<typeof CardFooterBase>;
+}
```
leopuleo, line 139: I don't see any references to some variants defined in the Figma file, such as: 

- Elevation
- Border Radius
- Paddings

Is it intentional?

leopuleo (changes requested):

packages/admin-ui/src/Card/Card.tsx
adrians5j: OK, although I've now noticed something else. But will post a comment when I'm done with changes.

packages/admin-ui/src/Card/Card.tsx
adrians5j: OK, although I've now noticed something else. But will post a comment when I'm done with changes.

packages/admin-ui/src/Card/Card.tsx
adrians5j: Added.

packages/admin-ui/src/Card/Card.tsx
adrians5j: Added.

packages/ui/src/Elevation/Elevation.tsx
adrians5j: This one I didn't quite understand 🤔 

What's interesting is that now there's an `elevation` prop on Card, so maybe we don't need a separate `Elevation` component?

adrians5j:
Pushed another batch of changes.

Added all of the things you mentioned, thanks for the feedback! 
TBH I didn't even look much at the Figma file, hence all of the extra stuff missing, sry about that 🙃🫠

Besides adding the missing stuff, after playing with it more that I would've wanted to, I actually converted the `footer` and `content` props back into regular `React.ReactNode` props. In other words, users won't need to import `CardFooter` / `CardContent` components and use these upon defining card footer and content. In short, it was complicating some things. This way complication doesn't exist, plus it's a simpler API to use.

> [!NOTE]
> I've also taken into consideration the fact that, when using extra components for content/footer, users can pass extra props to these sections (like `data-testid`). I've also went with the assumption that this won't be that common of a use case for now. That was one additional vote for me going with the simpler API.

One other change I did here is that I introduced also the Icon component, which is now used for the close button in the card.

adrians5j:
OK so another batch of updates here. Hoping we're finally ready to merge, but yeah, let's see what you have to say.

One final thing I did here (on top of everything we've talked about these days), is that, ultimately, I decided to go with the following props API:

```
    header: <CardHeader title="This is a card title" description="This is a card description" />,
    content: <CardContent content={"This is card content. Anything can go in here."} />,
    footer: <CardFooter content={"This is card footer. Anything can go in here."} />
```

So, no primitive values sent, just components (`CardHeader`, `CardContent`, `CardFooter`), which can actually take primitive values. 

Note that `title` and `description` props on `Header` component accept `React.ReactNode`, where, if a `string` was passed, then we use `Heading` / `Text` components internally to render texts. If not `string`, then we render just whatever the user has passed to it.

> [!NOTE]
> Have in mind that `Heading` / `Text` components only accept `string` via their `text` prop.

leopuleo (approved):
Adrian, thank you so much for all the hard work you put into this PR and the design around the APIs - it was definitely not an easy task! 😄

I think there are just two small things left to do:

- Remove @rmwc/elevation from @webiny/ui.
- Rewire the Icon component in the @webiny/ui package to the newly developed one.

These aren't blockers, but it would be great to address them before we merge.

Fantastic job, and thanks again! 🍻

adrians5j:
Hehe np Leo, glad we got this finally sorted!

And thank _you_ for all the reviewing you did here, awesome job as well!

Team work 💪

P.S. I resolved the two things you mentioned before merging.

PR #4246, opened 2024-09-03
New Admin UI - Avatar

adrians5j:
This PR creates the Avatar component.


adrians5j:
So here's the Avatar component, but there's one thing we should discuss here, and those are all of the diff variants that are shown in the Figma file. Those look to me as different fallback variations. Should we treat them as such? Wdyt?

[I've also posted a comment here.](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT?node-id=204-2537#941955250)

Unless you notice something extra, I'd say that's the last thing to discuss here.

PR #4247, opened 2024-09-03
fix(app-headless-cms): bulk action enum

leopuleo:
## Changes
This PR corrects the value assigned to the action variable required by the CmsBulkAction mutation. The action value is now properly formatted in Title Case.

CleanShot 2024-09-03 at 16 18 58

## How Has This Been Tested?
Manually


adrians5j:
Looks good to me, although I have no idea what user-facing issue this is fixing and why this is the fix. :D 
Was it just that the actions were not working because incorrect strings being passed to `processInBulk` calls?

Maybe others will have more context here.

leopuleo:
> Looks good to me, although I have no idea what user-facing issue this is fixing and why this is the fix. :D Was it just that the actions were not working because incorrect strings being passed to `processInBulk` calls?
> 
> Maybe others will have more context here.

Yes, correct!

adrians5j:
Then it's LGTM.

adrians5j (dismissed):

adrians5j (approved):

PR #4249, opened 2024-09-04
feat(admin-ui): Toast component

leopuleo:
This PR adds the `Toast` component to the `@webiny/admin-ui `package.

We are exporting a props-based flat component based on the following [radix component](https://www.radix-ui.com/primitives/docs/components/toast).

The new component is fully compatible with the old `Snackbar` component, but some refactoring is needed when UI tokens are defined.

The `useSnackbar` hook is fully compatible with the new Toast component. Refactoring the hook will be necessary; a separate PR will address it.

### Text and Heading
In this PR, we also introduce new text tokens for the `Text` and `Heading` components, which are used internally by the Toast component.


packages/app-admin/src/base/Admin.tsx
```diff
@@ -22,25 +22,28 @@ export const Admin = ({ children, createApolloClient }: AdminProps) => {
...
-                            <Provider hoc={SearchProvider} />
-                            <Provider hoc={UserMenuProvider} />
-                            <Provider hoc={NavigationProvider} />
-                            <Provider hoc={DialogsProvider} />
-                            <Base />
-                            {children}
-                        </App>
-                    </WcpProvider>
-                </UiProviders>
+                {/*<UiProviders>*/}
```
adrians5j: This needs to stay commented out?

packages/admin-ui/src/Toast/Toast.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+    args: {
+        ...Default.args,
+        variant: "accent"
+    }
+};
+
+export const WithDescription: Story = {
+    args: {
+        ...Default.args,
+        description: <ToastDescription text={'Entry "Article One" has been successfully created'} />
```
adrians5j: Noticed this here, and thinking we should do the same approach I have [here](https://github.com/webiny/webiny-js/pull/4246/files#diff-d8cb04b4f9fe7ce06447bfc9b5fa1c28f4f81000b5c5395268bf66ee15d1930fR18-R20).

So, in this case, we'd end up prolly with `Toast.Description`. Wdyt?

BTW not sure if applicable here, but I did this [`withStaticProps`](https://github.com/webiny/webiny-js/pull/4246/files) helper within the Avatar PR, just to make it a bit more humane when sticking extra components as static properties on an existing React component / class.

adrians5j (changes requested):
Left two comments. All in all, looks good.

packages/app-admin/src/base/Admin.tsx
leopuleo: You are right, thank you!

packages/admin-ui/src/Toast/Toast.stories.tsx
leopuleo: You are right, I'm bringing the [withStaticProps](https://github.com/webiny/webiny-js/pull/4246/files) util. Nice one!

PR #4259, opened 2024-09-10
fix(github): enable upload hidden files

leopuleo:
See https://github.com/actions/upload-artifact/pull/598


adrians5j:
https://github.com/actions/upload-artifact/issues/602

PR #4303, opened 2024-10-01
Programmatic Creation of Roles and Teams

adrians5j:
## Changes
This PR enables users to create roles and teams programatically. For example:
```ts
// extensions/customRolesTeams/src/index.ts
import { createSecurityRolePlugin, createSecurityTeamPlugin } from "@webiny/api-serverless-cms";

export const createExtension = () => {
    return [
        createSecurityRolePlugin({
            id: "test-role",
            name: "Test Role",
            description: "Test role defined via an extension.",
            permissions: [{ name: "cms.*" }]
        }),
        createSecurityTeamPlugin({
            id: "test-team",
            name: "Test Team",
            description: "Test team defined via an extension.",
            roles: ["test-role"]
        })
    ];
};
```

With the above code in place, In the UI, users can see the following:
image

#### 1. Copy Permissions Button

This PR also adds the "Copy Permissions as JSON" button to all users. Previously this feature was hidden behind a feature flag:

image

With this feature, users will be able to define `permissions` property in the `createSecurityRolePlugin` function much easier, because they'll just be able to select the desired permissions, copy them as JSON, and then paste them into their plugin.

#### 2. Added `createTestWcpLicense` Testing Utility

The `createTestWcpLicense` utility was added in the `@webiny/wcp` package, so that it can be used in Jest tests.

#### 3. Added Teams-related Tests

I've added basic tests for Teams functionality. They did not exist before. 🤯 

## How Has This Been Tested?
Manually / Jest.

## Documentation
Changelog.


packages/api-security/src/index.ts
```diff
@@ -42,7 +42,9 @@ export const createSecurityContext = ({ storageOperations }: SecurityConfig) =>
                 const tenant = context.tenancy.getCurrentTenant();
                 return tenant ? tenant.id : undefined;
             },
-            storageOperations
+            storageOperations,
+            listGroupsFromPluginsCallback: () => context.plugins.byType("security-role"),
+            listTeamsFromPluginsCallback: () => context.plugins.byType("security-team")
```
Pavel910: Several thoughts here:
1) Let's avoid using the `Callback` suffix, as it's really meaningless. Typescript is here to tell you that it's a function.  It can simply be `listGroupsFromPlugins`.
2) Assuming we go with `listGroupsFromPlugins` in the first bullet point, let's elaborate on that: if we talk about the _purpose_ of that parameter, it is 1) an "optional dependency", and 2) it is a "provider" because it provides you with a list of groups. 
3) Moving further, we want to code against _abstractions_. But `listGroupsFromPluginsCallback` sounds extremely specific: it tells you that this provider HAS TO load groups _from plugins_. What if I don't? How can you enforce this? You can't, all you care about is that you get back a list of Groups/Teams.

So, taking all these points into account, our code can be improved by doing a few simple changes:
1) make it more abstract, by renaming `listGroupsFromPluginsCallback` to `groupsProvider` (very abstract, can be implemented in many ways).
2) Make sure providers are async, so `groupsProvider: () => Promise<Group[]>`.
3) Same logic applies for teams, so `teamsProvider: () => Promise<Team[]>`.

packages/api-security/src/createSecurity/groupsTeamsPlugins/listGroupsFromPlugins.ts
```diff
@@ -0,0 +1,45 @@
...
+
+export const listGroupsFromPlugins = (params: ListGroupsFromPluginsParams) => {
+    const { listGroupsFromPluginsCallback, where } = params;
+    if (!listGroupsFromPluginsCallback) {
+        return [];
+    }
+
+    const groups = listGroupsFromPluginsCallback().map(plugin => {
+        return plugin.securityRole;
+    });
```
Pavel910: This would be part of the `groupsProvider` implementation, since `plugin.securityRole` is an implementation detail.

packages/app-security-access-management/src/ui/views/Groups/GroupsForm.tsx
```diff
@@ -155,6 +156,26 @@ export const GroupsForm = () => {
...
+                                        </Alert>
+                                    </Cell>
+                                </Grid>
+                            )}
+                            {pluginGroup && (
+                                <Grid>
+                                    <Cell span={12}>
+                                        <Alert type={"info"} title={"Permissions are locked"}>
+                                            This role is registered via an extension. It cannot be
+                                            modified via the Admin Area.
```
Pavel910: Suggesting a simpler alternative:
> This role is registered via an extension, and cannot be modified.

It's shorter, and I would really like to avoid this `Admin Area`, since we don't have a standard for this even in our docs. Hurts less by simply avoiding that in the first place.

Pavel910 (commented):
@adrians5j I did a review, and posted some thoughts. We can talk this over via a call, if you'd like to discuss the suggestions.

packages/app-security-access-management/src/ui/views/Groups/GroupsForm.tsx
adrians5j: Done.

packages/api-security/src/createSecurity/groupsTeamsPlugins/listGroupsFromPlugins.ts
adrians5j: Done.

packages/api-security/src/index.ts
adrians5j: Done.

Pavel910 (approved):
@adrians5j I'm approving this now, because the feature works. I'll submit a separate PR with improvements to the code itself, so we can do it in parallel with the release preparations.

PR #4306, opened 2024-10-01
feat(pulumi-aws): add shared indexes param

brunozoric:
## Changes
Add sharedIndex param to OpenSearch and Elasticsearch Pulumi config.

## How Has This Been Tested?
Jest and manually.

## Usage
In both `apps/core/webiny.application.ts` and `apps/api/webiny.application.ts` user needs to set:
```
    openSearch: {
        sharedIndexes: true
    }
```

Example for `apps/core/webiny.application.ts`:

```typescript
import { createCoreApp } from "@webiny/serverless-cms-aws";

    openSearch: {
        sharedIndexes: true
    }
});

```

Example for `apps/api/webiny.application.ts`:

```typescript
import { createApiApp } from "@webiny/serverless-cms-aws";

export default createApiApp({
    openSearch: {
        sharedIndexes: true
    }
});



```

### DIFF
--- a/packages/api-elasticsearch-tasks/src/settings/EnableIndexing.ts
+++ b/packages/api-elasticsearch-tasks/src/settings/EnableIndexing.ts
@@ -11,10 +11,11 @@ export class EnableIndexing {
 
     public async exec(index: string, settings: IIndexSettingsValues): Promise<void> {
         try {
+            const refreshInterval = parseInt(settings.refreshInterval || "", 10) || 0;
             await this.settings.setSettings(index, {
                 ...settings,
                 numberOfReplicas: settings.numberOfReplicas < 1 ? 1 : settings.numberOfReplicas,
-                refreshInterval: settings.refreshInterval === "-1" ? "1s" : settings.refreshInterval
+                refreshInterval: refreshInterval <= 0 ? "1s" : settings.refreshInterval
             });
         } catch (ex) {
             throw new IndexingEnableError(ex);
--- a/packages/api-form-builder-so-ddb-es/src/configurations.ts
+++ b/packages/api-form-builder-so-ddb-es/src/configurations.ts
@@ -1,5 +1,5 @@
 import WebinyError from "@webiny/error";
-import { getLastAddedIndexPlugin } from "@webiny/api-elasticsearch";
+import { getLastAddedIndexPlugin, isSharedElasticsearchIndex } from "@webiny/api-elasticsearch";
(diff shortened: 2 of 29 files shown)


Pavel910 (commented):
This looks good to me. Once you're done with the PR, maybe @adrians5j can have a second look.

packages/api-form-builder-so-ddb-es/src/configurations.ts
```diff
@@ -54,7 +54,10 @@ export const configurations: Configurations = {
             .join("-")
             .toLowerCase();
 
-        const prefix = process.env.ELASTIC_SEARCH_INDEX_PREFIX || "";
+        const prefix =
```
adrians5j, line 57: Maybe we do the same thing e did here?

image

packages/api-form-builder-so-ddb-es/src/configurations.ts
```diff
@@ -54,7 +54,10 @@ export const configurations: Configurations = {
...
+            process.env.ELASTIC_SEARCH_INDEX_PREFIX ||
+            process.env.WEBINY_ELASTIC_SEARCH_INDEX_PREFIX ||
+            "";
```
adrians5j:
```suggestion
        const prefix = getElasticsearchIndexPrefix();
```

Maybe we could also do the same here? I like what you did with `isSharedElasticsearchIndex`. 👍🏻

packages/api-elasticsearch-tasks/src/settings/EnableIndexing.ts
```diff
@@ -11,10 +11,11 @@ export class EnableIndexing {
 
     public async exec(index: string, settings: IIndexSettingsValues): Promise<void> {
         try {
+            const refreshInterval = parseInt(settings.refreshInterval || "", 10) || 0;
             await this.settings.setSettings(index, {
                 ...settings,
                 numberOfReplicas: settings.numberOfReplicas < 1 ? 1 : settings.numberOfReplicas,
```
adrians5j, line 17: Just a reminder that, for the 5.39 data migration, we actually ignored the `numberOfReplicas` param when disabling/enabling indexing it all worked just fine:

https://github.com/webiny/webiny-js/pull/4187/files#diff-69c88efb045520b15afd8acdcb24c33eb0877e333c2592a5d10f2981047eae76

adrians5j (approved):
Left two small notes , but it's definitely good to merge. Thanks!

adrians5j (approved):

PR #4348, opened 2024-10-22
feat(admin-ui): Slider / Range Slider components

leopuleo:
## Changes

This PR introduces two new components to the `@webiny/admin-ui` package:

- `Slider`
- `RangeSlider`

along with their form variants.

### Basic Component (`<Slider />`)
For the base slider, two components are exported:

- `<ComposableSlider />`: Accepts subcomponent VMs (ViewModels) and renders Radix UI components. This component serves as the foundation for both the simple and form variants.
- `<Slider />`: Receives custom-designed props, retrieves the necessary VMs via the presenter, and passes them to `<ComposableSlider />`.

The component state is managed by a presenter and linked to the main component through a dedicated hook, such as `useSlider()`. This hook acts as the composition root in React, establishing a three-layered architecture:

- **Presenter (`SliderPresenter.ts`)**: Initializes with the component props, manages the component’s state, and exposes the VMs.
- **Hook (`useSlider.ts`)**: Serves as the composition root. It receives the component props, integrates with React, and returns the VMs.
- **Component (`Slider.ts`)**: Renders the Radix UI components.

### Form Component (`<Slider />`)
The form variant may need to manage its own state and use the simple component with others like `<Label />`. In this scenario, the component state is managed by a presenter and linked to the main component through a dedicated hook.

## How Has This Been Tested?
Manually with Storybook + Presenter tested via Jest


packages/admin-ui/src/Form/RangeSlider/useRangeSlider.ts
```diff
@@ -0,0 +1,29 @@
...
+import omit from "lodash/omit";
+import { autorun } from "mobx";
+import { RangeSliderProps } from "./RangeSlider";
+import { RangeSliderPresenter } from "~/RangeSlider";
+import { FormRangeSliderPresenter } from "./RangeSliderPresenter";
+
+export const useRangeSlider = (props: RangeSliderProps) => {
+    const presenter = useMemo(() => {
+        const rangeSliderPresenter = new RangeSliderPresenter(
+            omit(props, ["label", "labelPosition"])
```
Pavel910: Is it possible that your props change, but you want to maintain the state of the Presenter?

packages/admin-ui/src/Form/RangeSlider/useRangeSlider.ts
```diff
@@ -0,0 +1,29 @@
...
+        );
+        return new FormRangeSliderPresenter(props, rangeSliderPresenter);
+    }, [JSON.stringify(props)]);
+    const [sliderVm, setSliderVm] = useState(presenter.sliderVm);
+    const [thumbsVm, setThumbsVm] = useState(presenter.thumbsVm);
+    const [labelsVm, setLabelsVm] = useState(presenter.labelsVm);
```
Pavel910: This is one of the reasons why it's better to expose a single `presenter.vm`, and have those VMs be the children of a single VM. Example: `vm.sliderVm`, `vm.thumbsVm`... Especially if you look at the useEffect below; you handle all three VMs at once, and never independently.

packages/admin-ui/src/Form/RangeSlider/RangeSliderPresenter.ts
```diff
@@ -0,0 +1,47 @@
+import { makeAutoObservable } from "mobx";
+import { IRangeSliderPresenter } from "~/RangeSlider";
+import { RangeSliderProps } from "~/Form";
+
+class FormRangeSliderPresenter implements IRangeSliderPresenter {
+    private rangeSliderPresenter: IRangeSliderPresenter;
+    private localValues: number[];
```
Pavel910: Not sure you even need this property. You can always derive labels from the value of rangeSliderPresenter.

packages/admin-ui/src/Form/RangeSlider/RangeSliderPresenter.ts
```diff
@@ -0,0 +1,47 @@
...
+        const { defaultValue, value, min = 0, max = 100 } = props;
+        this.rangeSliderPresenter = rangeSliderPresenter;
+        this.localValues = defaultValue ?? value ?? [min, max];
+        makeAutoObservable(this);
+    }
+
+    get sliderVm() {
+        return {
+            ...this.rangeSliderPresenter.sliderVm,
+            onValueChange: this.onValueChange
```
Pavel910: I'd keep methods outside of VM. Keep them on the presenter itself, and in this case, simply make your `onValueChange` public, and forward the call to the `rangeSliderPresenter` like you're doing now, but remove the `vm` part.

packages/admin-ui/src/Form/Slider/SliderPresenter.test.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    it("should update `localValue` and `labelVm.value` when `onValueChange` is called", () => {
+        const presenter = new SliderPresenter({});
+        const formSliderPresenter = new FormSliderPresenter(
+            { label: "Label", defaultValue: 20 },
+            presenter
+        );
+
+        formSliderPresenter.sliderVm.onValueChange([40]);
```
Pavel910: Again, just to confirm my previous thoughts: make it `formSliderPresenter.setValue([40]);`

packages/admin-ui/src/Form/RangeSlider/RangeSliderPresenter.ts
```diff
@@ -0,0 +1,47 @@
...
+        };
+    }
+
+    get labelsVm() {
+        return {
+            values: this.labelValues
+        };
+    }
+
+    private onValueChange = (values: number[]): void => {
```
Pavel910: A thought on naming: a presenter is an object with methods, and it doesn't know who's calling those methods. So it would make more sense to call this `setValue`, not `onChange`. `onSomething` is more of a UI component/event-based type of naming. A presenter is a standalone, self-sufficient object, and we can remove this notion of "event-driven" from method names.

packages/admin-ui/src/RangeSlider/useRangeSlider.ts
```diff
@@ -0,0 +1,20 @@
...
+
+    useEffect(() => {
+        const dispose = autorun(() => {
+            setSliderVm(presenter.sliderVm);
+            setThumbsVm(presenter.thumbsVm);
+        });
+        return () => dispose(); // Cleanup on unmount
+    }, [presenter]);
+
+    return { sliderVm, thumbsVm };
```
Pavel910: Here you would also expose methods, not just VMs. For example, we have a `setValue` (ex. onValueChange) method, so the return value would look like `{ vm, setValue }`, which you then forward to the presenter.

packages/ui/src/Slider/Slider.tsx
```diff
@@ -15,10 +14,10 @@ type Props = FormComponentProps & {
     description?: string;
 
     // The minimum value of the Slider.
-    min: number | string;
+    min?: number | string;
```
Pavel910, line 17: Why is `string` supported?

packages/admin-ui/src/Form/Slider/Slider.tsx
```diff
@@ -0,0 +1,100 @@
...
+const SliderWithSideValue = makeDecoratable(
+    "SliderWithSideValue",
+    DecoratorableSliderWithSideValue
+);
+
+/**
+ * Slider
+ */
+const DecoratableFormSlider = ({ labelPosition, ...props }: SliderProps) => {
+    const { sliderVm, thumbVm, labelVm } = useSlider(props);
```
Pavel910: Here I'd expose a single `vm`, but also expose methods, like `setValue`, etc. and then pass that to components.

Pavel910 (changes requested):

packages/admin-ui/src/Form/RangeSlider/RangeSliderPresenter.ts
leopuleo: Before the `admin-ui` component supported both controlled and uncontrolled states, like the Radix UI base components. ([Radix UI docs](https://www.radix-ui.com/primitives/docs/components/slider#root), see `value` and `defaultValue`).

Now the `admin-ui` component does not support the `defaultValue` component, developers must provide externally controlled `value` and `onValueChange` callback. Doing so we can derive labels from values.

packages/ui/src/Slider/Slider.tsx
leopuleo: No idea 😄 it's coming from the previous implementation

PR #4365, opened 2024-11-05
feat(admin-ui): Select component

leopuleo:
## Changes
This pull request adds the `Select` component to the `@webiny/admin-ui` package. Subsequent pull requests will develop the `Select` form variant and `Autocomplete` components.

## How Has This Been Tested?
Manually - Jest


packages/admin-ui/src/Select/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./Select";
+export * from "./SelectPresenter";
```
adrians5j, line 2: Question - why are we exporting `SelectPresenter` also here? Isn't this internal? 🤔

packages/admin-ui/src/Select/SelectPresenter.ts
```diff
@@ -0,0 +1,75 @@
...
+    init(props: SelectProps) {
+        this.props = props;
+        this.options = this.transformOptions(props.options);
+    }
+
+    get vm() {
+        return {
+            selectVm: {
+                ...omit(this.props, ["placeholder", "options"])
+            },
```
adrians5j:
```suggestion
            selectVm: omit(this.props, ["placeholder", "options"])
```

adrians5j: Extra question here might be... wouldn't it be 'more controlled' if we were to explicitly pick props that are relevant to `selectVm` , instead of omitting ?

packages/admin-ui/src/Select/SelectPresenter.ts
```diff
@@ -0,0 +1,75 @@
...
+            selectTriggerVm: {
+                placeholder: this.props?.placeholder || "Select an option",
+                hasValue: !!this.props?.value,
+                size: this.props?.size,
+                variant: this.props?.variant,
+                startIcon: this.props?.startIcon,
+                endIcon: this.props?.endIcon,
+                invalid: this.props?.invalid
+            },
```
adrians5j: Do these things need to go into `vm`? Is it basically just so props can easily be spread in the `DecoratableSelectRenderer` component?

Just asking b/c I know we discussed with Pavel at one point in time that some things don't need to necessarily end up in vm. But maybe this way is easier?

packages/admin-ui/src/Select/useSelect.ts
```diff
@@ -0,0 +1,22 @@
...
+import { SelectPresenter } from "./SelectPresenter";
+
+export const useSelect = (props: SelectProps) => {
+    const presenter = useMemo(() => new SelectPresenter(), []);
+
+    const [vm, setVm] = useState(presenter.vm);
+
+    useEffect(() => {
+        presenter.init(props);
+    }, [props]);
```
adrians5j: I saw this file in other components and it always kinda got me thinking. For now, my questions would be:

1. `useEffect` cb is run after React has done mutating the DOM, so I'm wondering if this initial `init` is too late here? Should we do `init` call within the `useMemo` and then return the class instance? 🤔 

2. Kinda thinking about:
```
// This can do the `useEffect` calls for us
const presenter = usePresenter(() => new SelectPresenter())
```

But probably it's premature optimization of mine. 😊

3. I can also see below we run `autorun`. I have read online that `autorun` is _smart_ and will only trigger the callback when related observables change, in our case `presenter`. Is that the case? Do we share the same understanding? 

P.S. This is my first time working with mobx so yeah, I might ask a question or two along the way.

packages/admin-ui/src/Select/SelectPresenter.ts
```diff
@@ -0,0 +1,75 @@
+import { makeAutoObservable } from "mobx";
+import omit from "lodash/omit";
+import { SelectOptionsVm, SelectProps, SelectTriggerVm, SelectVm } from "./Select";
+import { SelectOption } from "./SelectOption";
+import { SelectOptionMapper } from "~/Select/SelectOptionMapper";
+
+interface ISelectPresenter<TProps extends SelectProps = SelectProps> {
+    vm: {
+        selectVm: SelectVm;
```
adrians5j: Kinda on the fence whether the `Vm` suffix is needed when these already within the `vm` object. Like, would it just be enough to have `vm.select`, instead of `vm.selectVm`. But doesn't matter probably in the grand scheme of things.

packages/admin-ui/src/Select/Select.tsx
```diff
@@ -0,0 +1,503 @@
...
+/**
+ * Icon
+ */
+interface IconWrapperProps {
+    icon: React.ReactElement;
+}
+
+const Icon = ({ icon }: IconWrapperProps) => {
+    return (
+        <SelectPrimitive.Icon asChild className={"h-4 w-4"}>
```
adrians5j, line 29: We'll probably need to back here, and use tokens instead of h-4 w-4.

I see the Icon component is not ready for dev in Figma, so [I'll ping](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT?node-id=81-10117&m=dev#1034826747) Kreso about it and see if I can wrap that one up.

packages/admin-ui/src/Select/SelectPresenter.ts
```diff
@@ -0,0 +1,75 @@
...
+        selectTriggerVm: SelectTriggerVm;
+        selectOptionsVm: SelectOptionsVm;
+    };
+    init: (props: TProps) => void;
+    changeValue: (value: string) => void;
+    resetValue: () => void;
+}
+
+class SelectPresenter implements ISelectPresenter {
```
adrians5j, line 24: I've been actually reading a small bit... And I'm a bit confused when it comes to terminology. Kinda have a feeling we're mixing terms maybe? 🤔 

Like... we are using the word Presenter, so that tells me we're using  the [model-view-presenter  pattern](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93presenter). But then again, we're also using view model, which now brings me to the [MVVM](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93viewmodel).

I know all of these patterns are similar / basically the same at some point.. But still, did you maybe notice the same?

adrians5j (approved):
I wrote some feedback, but it's basically just questions / trying to understand all bits and pieces. I didn't see anything that would block the merge, so I also approved.

There are some things that I wrote based on intuition, not on exp / knowledge, so that might be off. And I totally get that we probably don't want to spend too much time on chasing the "perfect" arch / code and having dubious conversations which at the end of the day don't bring any real value. 🙃

packages/admin-ui/src/Select/index.ts
leopuleo: You are right, but the `SelectPresenter` will be a dependency for the `FormSelectPresenter`, as we did for the `Slider` and the `FormSlider`. (https://github.com/webiny/webiny-js/blob/feat/new-admin-ui/packages/admin-ui/src/Form/Slider/useSlider.ts#L9)

packages/admin-ui/src/Select/SelectPresenter.ts
leopuleo: I refactored this part: as we discussed with Pavel, We don't need to pass these properties to the presenter to receive them back in the ViewModel.
At the moment, I removed the `selectVm` entirely and forwarded the wanted `props` directly to the component.

packages/admin-ui/src/Select/SelectPresenter.ts
leopuleo: Also in this case, I removed these properties from the `selectTriggerVm` and passed them directly within the component.

packages/admin-ui/src/Select/SelectPresenter.ts
leopuleo: Thank you for pointing that out! 🙌
You’re totally right: there’s a mix of MVP and MVVM going on here, and it could get confusing.

For now, I’m considering leaving the code and experimenting with it more before deciding whether to refactor to stick with MVP or fully embrace MVVM. It feels like an excellent opportunity to test things and see what works best in practice.

packages/admin-ui/src/Select/SelectPresenter.ts
leopuleo: Thank you, it makes sense 👍

packages/admin-ui/src/Select/useSelect.ts
leopuleo: 1. Initialize the presenter within the `useMemo` by passing only the required parameters.
2. It would be helpful to create a wrapper for this. Let's explore it further before implementing a wrapper for this repetitive code.
3. Yes, you are right! 👍

PR #4389, opened 2024-11-14
feat(di-container): add a DI container

Pavel910:
## Changes
We're introducing a proper DI container for internal development purposes. 

> [!IMPORTANT]  
> This is strictly for the Webiny core team. If you're building plugins for your Webiny project, you should not concern yourself with this. At least not yet.

## Background

I've developed our own dependency injection container, because existing libraries, most notably Inversify and TSyringe, are a bit too "basic" in terms of what they offer. Both use simple string-based tokens, or Symbols, to refer to an abstraction, which leads to a disconnect between the token and the actual TS type that token represents. This makes it difficult to enforce types, and have proper developer experience, where type checker guides a developer and warns of wrong types early in the development phase. There are other critical features lacking in both libraries, which I highlight further below.

My goal is to have a DI container which operates around abstractions, which carry both the runtime identifier, and a TS type. The implementation of an `Abstraction` class, and its usage, is shown below:

```ts
class Abstraction<T> {
    public readonly token: symbol;

    constructor(name: string) {
        this.token = Symbol.for(name);
    }

    toString(): string {
        return this.token.description || this.token.toString();
    }
}

// Define an abstraction intefrace
interface ILogger {
    log(...args: unknown[]): void;
}

// Define an abstraction within your application.
const LoggerAbstraction = new Abstraction<ILogger>("Logger");
```

The `LoggerAbstraction` is now an object which contains its identifier and a type. Now, my application can export all of its abstractions, and let developers create implementations of these abstractions, following the DI patterns from other, more mature languages, like C#, and Java. One of the inspirations for this DI container is the  `Autofac` library. 

Here’s an example of implementing an abstraction:
```ts
// `GetInterface` is utility type to extract an implementation type from the abstraction object.
type LoggerInterface = GetInterface<typeof LoggerAbstraction>;

// Create an implementation class for the abstraction.
class Logger implements LoggerInterface {
    log(...args: unknown[]): void {
        console.log(...args);
    }
}

// Create an implementation object, which annotates the implementation class with metadata, using the `reflect-metadata` library.
// This also checks types of the provided `dependencies` against the class constructor.
const Implementation = createImplementation({
    abstraction: LoggerAbstraction,
    implementation: Logger,
    dependencies: []
});

// Register the implementation.
container.register(Implementation);
```

In the code above, we call a `createImplementation` factory, which takes the abstraction object to implement, the implementation, and an array of dependencies that need to be injected into the implementation constructor. The dependencies are configured as follows:

```ts
// Imagine we have a `DebuggerAbstraction` abstraction, we provide an implementation class, but that class takes in some dependencies.
// The dependencies are referenced to via other abstractions. We don't want to allow any primitive types as tokens.
// `multiple: true` means it expects the container to resolve to an array of implementations for the given abstraction. 
// `optional: true` means the dependency is optional.
const DebuggerImpl = createImplementation({
    abstraction: DebuggerAbstraction,
    implementation: Debugger,
    dependencies: [
        Abstractions.Logger,
        [Abstractions.Logger, { multiple: true }],
        [Abstractions.Logger, { multiple: true, optional: true }]
    ]
});
```

### Advanced  Must-Have Feature
To make a DI container actually useful, and get the ability to use proper SOLID principles, we must also support at least the Decorator Pattern, on the abstraction level. I want to be able to create decorators for the abstractions. An example usage is shown below:

```ts
// Fetch abstraction type.
type LoggerInterface = GetInterface<typeof Logger>;

// Implement a decorator. The convention is that the decoratee is always the last argument in the constructor.
class LoggerDecorator implements LoggerInterface {
    private decoratee: LoggerInterface;
    
    constructor(decoratee: LoggerInterface) {
        this.decoratee = decoratee;
    }
    
    log(...args: unknown[]): void {
        this.decoratee.log("decorated log");
    }
}

// Create a decorator object, which we can then register into the container.
const decorator = createDecorator({
	abstraction: Logger,
	decorator: LoggerDecorator,
	dependencies: []
})

container.registerDecorator(decorator);
```

Once the `Logger` is resolved at runtime, the DI container should not only resolve the implementation, but also automatically apply all the decorators, and return the decorated object. As a reference, check out https://autofac.readthedocs.io/en/latest/advanced/adapters-decorators.html#decorators.

## How Has This Been Tested?
Jest tests.


adrians5j:
On top of the call today, I also read the PR body here, and gotta say again, this looks awesome. 🚀 🚀 👏🏻 👏🏻 

Can't wait to put this to use.

PR #4398, opened 2024-11-19
New Admin UI - Container

adrians5j:
## Changes
This PR introduces the Container component.

Note that the component is listed in its own section in Storybook, simply because it can be used with diff components. It's not just Grid that can be rendered with the container.

Still, because Container component is frequently used with Grid, I added two examples in SB:
image

And I basically also did the same in the Grid section:
image

## Extra Changes
I refactored `packages/admin-ui/src/Grid/Grid.stories.tsx` file a bit:
1. cleaned up the code a bit by extracting the `StyledColumn` component into its own file
2. simplifed `StyledColumn` component's its usage (ultimately story file got a bit cleaner, but nothing spectacular)

## How Has This Been Tested?
Manually.


adrians5j:
Closing this for now so that it doesn't create noise. It's noted in Notion doc so we won't miss it.

PR #4400, opened 2024-11-19
feat(admin-ui): Radio & RadioGroup component

leopuleo:
## Changes
This pull request adds the `Radio` and `RadioGroup` components to the `@webiny/admin-ui` package. Subsequent pull requests will develop the form variant and update the components within the `@webiny/ui`.

## How Has This Been Tested?
Manually - Jest


packages/admin-ui/src/Radio/RadioItem.ts
```diff
@@ -0,0 +1,41 @@
...
+    static create(data: RadioItemDto) {
+        return new RadioItem({
+            id: generateId(data.id),
+            label: data.label,
+            value: data.value,
+            disabled: data.disabled ?? false
+        });
+    }
+
+    get id() {
```
adrians5j, line 26: Sometimes I wonder do we really need all these getters when they're just returning prop value. 🤔 
I guess they're needed because we have private underscored versions of props.

packages/admin-ui/src/Radio/RadioItemDto.ts
```diff
@@ -0,0 +1,8 @@
+import React from "react";
```
adrians5j: I saw in other file this import is not there, but here we use it. 

Just mentioning this for consistency... but yeah.. nitpick.

packages/admin-ui/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,58 @@
...
+    RadioProps
+>(({ className, label, id, ...props }, ref) => {
+    return (
+        <div className="flex items-start space-x-sm-extra">
+            <RadioGroupPrimitive.Item
+                ref={ref}
+                id={id}
+                className={cn(
+                    [
+                        "group peer aspect-square h-4 w-4 rounded-xl mt-xxs",
```
adrians5j: `h-4 w-4` used here.

packages/admin-ui/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,58 @@
...
+                        "disabled:cursor-not-allowed disabled:border-neutral-muted disabled:bg-neutral-disabled"
+                    ],
+                    className
+                )}
+                {...props}
+            >
+                <RadioGroupPrimitive.Indicator className="flex items-center justify-center">
+                    <span
+                        className={cn([
+                            "h-2 w-2 rounded-xl",
```
adrians5j: `h-2 w-2` used here.

packages/admin-ui/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,58 @@
...
+ * RadioItem
+ */
+type RadioProps = React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Item> & {
+    label: string | React.ReactNode;
+};
+
+const DecoratableRadio = React.forwardRef<
+    React.ElementRef<typeof RadioGroupPrimitive.Item>,
+    RadioProps
+>(({ className, label, id, ...props }, ref) => {
```
adrians5j, line 16: Will the users need to pass `id` all the time in order to have label also clickable / act as if a user clicked on the radio button? 🤔 

Can we generate this maybe in cases the user did not pass the ID?

packages/admin-ui/src/Radio/Radio.tsx
```diff
@@ -0,0 +1,58 @@
+import * as React from "react";
+import * as RadioGroupPrimitive from "@radix-ui/react-radio-group";
+import { cn, makeDecoratable } from "~/utils";
+
+/**
+ * RadioItem
+ */
+type RadioProps = React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Item> & {
```
adrians5j: I noticed you're using `type` in some of these. 

Which is fine, but I believe we're generally leaning more towards using `interface` when possible. So in this case, we could do:

```ts
interface RadioProps extends React.ComponentPropsWithoutRef<typeof RadioGroupPrimitive.Item> {
   label: string | React.ReactNode;
}
```

Just mentioning it, it's a nitpick. Bu feel free to leave it if you want.

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+class RadioGroupPresenter<TValue = string> implements IRadioGroupPresenter<TValue> {
+    private params?: RadioGroupPresenterParams<TValue> = undefined;
+
+    constructor() {
+        makeAutoObservable(this);
+    }
+
+    public init = (params: RadioGroupPresenterParams<TValue>) => {
```
adrians5j, line 25: Saw this before and wanted to share an alternative to having to instantiate a class and then call `init` on it. It would include having a static `create` class, which takes params and instantiates / inits. So basically:

```ts
// Instead of this...
const rgp = new RadioGroupPresenter();
rgp.init(params);

// ...this (the `static` create method creates a new instance of the class):
const rgp = RadioGroupPresenter.create(params);
```

Maybe I missed something though, and this is probably a bit subjective. Still wanted to share, can't hurt. :)

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
```diff
@@ -0,0 +1,48 @@
...
+        this.params = params;
+    };
+
+    get vm() {
+        return {
+            items: this.items.map(item => RadioItemMapper.toFormatted(item))
```
adrians5j: Question I had here before... why is `RadioItemMapper` having the `toFormatted` fn. If it's a "mapper", shouldn't it have `toMapped` method instead?

Don't know what's the correct term/verb here: are we talking about a mapper or a formatter here? 🤔

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
```diff
@@ -0,0 +1,48 @@
...
+        };
+    }
+
+    public changeValue = (value: TValue) => {
+        this.params?.onValueChange(value);
+    };
+
+    private get items() {
```
adrians5j: I'd kinda want to see a more explicit name here because this is not just a simple getter. Maybe something like `mapItemsToRadioItems` or maybe `getRadioItemsFromParams`. 

Jus something that crossed my mind. But yeah, more of a nitpick, especially since this is a private method.

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
```diff
@@ -0,0 +1,48 @@
+import { makeAutoObservable } from "mobx";
+import { RadioGroupVm } from "./RadioGroup";
+import { RadioItem } from "~/Radio/RadioItem";
+import { RadioItemMapper } from "~/Radio/RadioItemMapper";
+import { RadioItemDto } from "~/Radio/RadioItemDto";
+
+interface RadioGroupPresenterParams<TValue = string> {
+    items: RadioItemDto[];
```
adrians5j: Still on the fence whether this is really `RadioItemDto`. :D 

Maybe `RadioItemParams`? 

IDK, I guess LR videos effed my brain and hard-wired the DTO concept to gateway layer 😵 .

adrians5j (changes requested):
Left some nitpick level comments. 

But since we could always discuss these things / we have a lot work left to do, I guess just ignore whatever you feel is not worth discussing.

packages/admin-ui/src/Radio/Radio.tsx
leopuleo: Thank you, I have switched to `interface` in this component. I will implement the same change in other components as well.

packages/admin-ui/src/Radio/RadioItem.ts
leopuleo: Although it might seem excessive, having a `RadioItem` model is very beneficial, and the `id` getter serves as a good example. 

While an `id` is not required when defining a radio item, it's crucial for consistency in various situations. Ensuring that each `RadioItem` has a valid `id` helps us avoid issues when interacting with or displaying items.

If an `id` isn't provided externally, the model can generate one internally, keeping everything uniform across the application.

packages/admin-ui/src/Radio/Radio.tsx
leopuleo: Thank you! 👍

packages/admin-ui/src/Radio/Radio.tsx
leopuleo: Thank you! 👍

packages/admin-ui/src/Radio/Radio.tsx
leopuleo: This component is quite simple; it just receives props and renders the radio button. I modified the props interface to make the `id` property mandatory.

The `id` is passed from the `RadioItem` model, and we know that it will always be a string, even if the user does not provide it.

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
leopuleo: It makes sense: the original idea was to have a `RadioItemMapper` class with multiple methods (`toFormatted`, `toDto`...).

I changed it to `RadioItemFormatter` with a single `format` method.

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
leopuleo: I'll spend some time experimenting with it; thank you for the suggestion. 😃

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
leopuleo: It makes sense, I changed it to `getRadioItems`

packages/admin-ui/src/Radio/RadioGroupPresenter.ts
leopuleo: DTOs encapsulate and transfer data between application layers or components, so they are not wired to the gateway layer only.

Anyway, I changed `RadioItemDto` into `RadioItemParams` as you suggest. It makes sense 👍

PR #4405, opened 2024-11-21
New Admin UI - Card Component

adrians5j:
## Changes

This PR introduces the Card component.

image

## Extra Changes

With Kreso, we introduced an additional button variant. Basically an icon-only button, with a larger icon in it. This is utilized in the Card component, here:

image

Note that I've added an additional story within the Button section in SB.

## How Has This Been Tested?
Manually.


adrians5j:
Quick discuss:
More and more I see we're using `Button` for icon-only button. But, didn't we want to have like `IconButton` or something like that?

Maybe it was just a miss from my side? 🤔

leopuleo (dismissed):
It looks good to me. Thank you, Adrian!

PR #4406, opened 2024-11-22
feat(admin-ui): Checkbox & CheckboxGroup component

leopuleo:
## Changes
This pull request adds the `Checkbox` and `CheckboxGroup` components to the `@webiny/admin-ui` package. Subsequent pull requests will develop the form variant and update the components within the `@webiny/ui`.

## How Has This Been Tested?
Manually + Jest


packages/admin-ui/src/Checkbox/CheckboxPresenter.ts
```diff
@@ -0,0 +1,50 @@
...
+import { CheckboxItemDto } from "~/Checkbox/CheckboxItemDto";
+
+type CheckboxPresenterParams = CheckboxItemDto & {
+    onCheckedChange: (checked: boolean) => void;
+};
+
+interface ICheckboxPresenter<TParams extends CheckboxPresenterParams = CheckboxPresenterParams> {
+    vm: CheckboxVm;
+    init: (params: TParams) => void;
+    changeChecked: (checked: boolean) => void;
```
adrians5j, line 14: Maybe it's because of FTA, maybe it's subjective, but gotta say `changeChecked` sounds a bit weird. 

Can this be called `setChecked` ?

packages/admin-ui/src/Checkbox/CheckboxGroupPresenter.ts
```diff
@@ -0,0 +1,74 @@
...
+    changeChecked: (value: TValue) => void;
+}
+
+class CheckboxGroupPresenter<TValue = string | number> implements ICheckboxGroupPresenter<TValue> {
+    private params?: CheckboxGroupPresenterParams<TValue>;
+    private items: CheckboxItem[] = [];
+
+    constructor() {
+        this.params = undefined;
+        makeAutoObservable(this);
```
adrians5j: I see we are initializing `items` to `[]` in L21. Can't we do the same immediately with `params`?

packages/admin-ui/src/Checkbox/CheckboxItemFormatted.ts
```diff
@@ -0,0 +1,8 @@
+export interface CheckboxItemFormatted {
+    id: string;
+    label: string | React.ReactNode;
```
adrians5j, line 5: This works without importing React at the top of the file? 🤔

packages/admin-ui/src/Checkbox/useCheckbox.ts
```diff
@@ -0,0 +1,46 @@
...
+        const presenter = new CheckboxPresenter();
+        presenter.init(params);
+        return presenter;
+    }, []);
+
+    const [vm, setVm] = useState(presenter.vm);
+
+    useEffect(() => {
+        presenter.init(props);
+    }, [params, presenter]);
```
adrians5j: Is `presenter` needed here? from the looks of it, it never changes because cache key in L31?

```suggestion
    }, [params]);
```

packages/admin-ui/src/Checkbox/CheckboxGroupPresenter.ts
```diff
@@ -0,0 +1,74 @@
+import { makeAutoObservable } from "mobx";
+import { CheckboxGroupVm } from "./CheckboxGroup";
+import { CheckboxItem } from "./CheckboxItem";
+import { CheckboxItemMapper } from "./CheckboxItemMapper";
+import { CheckboxItemDto } from "./CheckboxItemDto";
+
+interface CheckboxGroupPresenterParams<TValue = string | number> {
+    items: CheckboxItemDto[];
```
adrians5j, line 8: I was watching the FTA videos a bit, and saw that DTO was being mentioned within the gateway layer. Maybe I need to watch a bit more videos but... it seems that DTO can be used anywhere?

adrians5j (approved):
Was looking at this PR for like 45mins, where I'd say I spent most of the time just looking at diff layers and how the code is architected.

Can't provide any bold feedback about it yet though. I mean, it all looks good, it works, it's layered, it's testable. 

Fine.

Now, is it possible to simplify something or make it "better"? That I don't know as I don't have hands-on exp with this yet. I did start to watch the LR's course, so I hope that'll make me a bit more competent, but still, hands-on will be needed in order to be good at it.

So, yeah, for now, I just tried to detect any illogicalities, but there were none I'd say, so gonna just approve this, and let's move on.

P.S. As some of the ideas and approaches consolidate, we might want to do another call with Pavel just so he sees where we're at (if anything has even changed since our last call).

packages/admin-ui/src/Checkbox/CheckboxGroupPresenter.ts
```diff
@@ -0,0 +1,74 @@
...
+            return;
+        }
+
+        const currentValues = [...this.params.values];
+        const newValues = this.updateValues(currentValues, value);
+        this.items = this.getItems(this.params.items, newValues);
+        this.params.onCheckedChange(newValues);
+    };
+
+    private updateValues(currentValues: TValue[], value: TValue): TValue[] {
```
adrians5j: Found the `updateValues` name a bit odd. 🤔 

It's a nitpick ofc, but renaming the fn to maybe `getUpdatedValues` clarifies the intent and makes it evident that the function doesn't mutate currentValues.

packages/admin-ui/src/Checkbox/CheckboxPresenter.ts
leopuleo: It's completely subjective. I would proceed to create all the input components and return here to see if there is a better name.

packages/admin-ui/src/Checkbox/CheckboxGroupPresenter.ts
leopuleo: Yes, thanks for pointing this out 👍

packages/admin-ui/src/Checkbox/CheckboxItemFormatted.ts
leopuleo: Yes, it's working because we have `@types/react` package installed. I imported it, to avoid potential future issues. Thank you!

packages/admin-ui/src/Checkbox/useCheckbox.ts
leopuleo: 👍🙏

packages/admin-ui/src/Checkbox/CheckboxGroupPresenter.ts
leopuleo: We can discuss better naming here as well, but generally, we can use a "DTO" structure for communication between different layers of the architecture.

packages/admin-ui/src/Checkbox/CheckboxGroupPresenter.ts
leopuleo: 👍

packages/admin-ui/src/Checkbox/CheckboxGroupPresenter.ts
adrians5j: Sounds good. Tnx!

packages/admin-ui/src/Checkbox/CheckboxItemFormatted.ts
adrians5j: Thanks1

packages/admin-ui/src/Checkbox/CheckboxPresenter.ts
adrians5j: Cool!

PR #4407, opened 2024-11-22
feat(admin-ui): Textarea component

leopuleo:
## Changes
This pull request adds the `Textarea` component to the @webiny/admin-ui package. A subsequent pull request will develop the `Textarea` form component.

Next to the new `Textarea` component, we are also:

-  applying the `font-sans` and `antialiased` utility classes to the body element via the `theme.scss` file and it's base template
- applying new tokens to `Input` component (i.e. `bg-neutral-dimmed-a1`)


packages/admin-ui/src/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import { cva, type VariantProps } from "class-variance-authority";
+import { cn, makeDecoratable } from "~/utils";
+
+const textareaVariants = cva(
+    [
+        "flex min-h-[80px] w-full border-sm px-3 py-2 text-md focus-visible:outline-none disabled:cursor-not-allowed"
```
adrians5j: Hm `px-3 py-2`, can we use tokens here? 🤔

packages/admin-ui/src/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from "react";
+import { cva, type VariantProps } from "class-variance-authority";
```
adrians5j: Nitpick, but these can be imported from "~/utils".

packages/admin-ui/src/Textarea/Textarea.tsx
```diff
@@ -0,0 +1,81 @@
...
+    }
+);
+
+type TextareaProps = React.ComponentProps<"textarea"> & VariantProps<typeof textareaVariants>;
+
+const DecoratableTextarea = React.forwardRef<HTMLTextAreaElement, TextareaProps>(
+    ({ className, variant, invalid, size, ...props }, ref) => {
+        return (
+            <textarea
+                className={cn(textareaVariants({ variant, invalid, size, className }))}
```
adrians5j: I realized some time ago that `cn` is not really needed in cases like these. Just because `textareaVariants` already received `className`, and it'll also do the string/class concatenation.

Nitpick but just putting it out there.

packages/admin-ui/src/Input/Input.tsx
```diff
@@ -108,19 +108,19 @@ const inputVariants = cva(
...
+                    "focus:bg-neutral-base focus:border-neutral-black",
                     "disabled:bg-neutral-disabled disabled:text-neutral-disabled disabled:placeholder:text-neutral-disabled"
                 ],
                 ghost: [
-                    "bg-neutral-base border-transparent text-neutral-strong placeholder:text-neutral-dimmed",
-                    "hover:bg-neutral-light",
-                    "focus:bg-neutral-light",
-                    "disabled:bg-neutral-disabled disabled:text-neutral-disabled disabled:placeholder:text-neutral-disabled"
+                    "bg-transparent border-transparent text-neutral-strong placeholder:text-neutral-dimmed",
+                    "hover:bg-neutral-dimmed-a1",
```
adrians5j: So.... re these `-a1` and similar tokens..

As you probably already know, these are basically just the original token, with applied alpha or in other words, color opacity.

Which can also be controlled with TW. In this case, we can just use `bg-neutral-dimmed/10`.

I also needed the same thing in one of the prev PRs, and this is where we initially introduced these. And what I did in that PR is.... I did not rely on these tokens from DS, but just did it the TW way, as mentioned above.

In the import script, I also basically skipped these `-a{x}` suffixed tokens, just to maybe try to make things a bit cleaner. But now I see I did not cover all categories, as you were able to use the `bg-neutral-dimmed-a1` here. After looking into the script, I see I only [did it for bg color](https://github.com/webiny/webiny-js/blob/e93b67c30c5624a9abb48371c21b2e95a022c99c/packages/admin-ui/scripts/importFromFigma/createTailwindConfigTheme.js#L10).

In any case... what would you say if we did the same here? 

I think it makes sense to use TW's utils instead of a separate token here. But maybe I'm issing something :)

adrians5j (changes requested):
A couple of nitpicks, nothing special.

At the very least, maybe we check the padding tokens and quickly discuss `-a{x}` tokens?

adrians5j (approved):

PR #4421, opened 2024-11-29
feat: create Icon component

adrians5j:
This PR adds tokens to the Icon component.

### DIFF
--- a/packages/admin-ui/src/Icon/Icon.stories.tsx
+++ b/packages/admin-ui/src/Icon/Icon.stories.tsx
@@ -27,7 +27,7 @@ export const Default: Story = {
         },
         color: {
             control: "select",
-            options: ["accent", "neutral", "neutral-strong", "neutral-xstrong"]
+            options: ["accent", "neutral", "neutral-strong", "neutral-light"]
         }
     }
 };
@@ -63,7 +63,7 @@ export const ColorNeutralStrong: Story = {
 export const ColorNeutralXStrong: Story = {
     args: {
         ...Default.args,
-        color: "neutral-xstrong"
+        color: "neutral-strong"
     }
 };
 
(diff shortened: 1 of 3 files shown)


adrians5j:
For next week...

Please let me know what you think about the [`color`](https://github.com/webiny/webiny-js/blob/925798ab706084114601b6f931ee1c29688dcfb3/packages/admin-ui/src/Icon/Icon.tsx#L14-L19) prop and the values it takes: "accent", "neutral", "neutral-strong", "neutral-xstrong".

Kinda not happy about this. Maybe we dbl check with Kreso about alternative naming here? Maybe something similar to Button, where we had `variant` and "primary", "secondary", "tertiary", "ghost", "ghost-negative" values.

@leopuleo

leopuleo:
> For next week...
> 
> Please let me know what you think about the [`color`](https://github.com/webiny/webiny-js/blob/925798ab706084114601b6f931ee1c29688dcfb3/packages/admin-ui/src/Icon/Icon.tsx#L14-L19) prop and the values it takes: "accent", "neutral", "neutral-strong", "neutral-xstrong".
> 
> Kinda not happy about this. Maybe we dbl check with Kreso about alternative naming here? Maybe something similar to Button, where we had `variant` and "primary", "secondary", "tertiary", "ghost", "ghost-negative" values.
> 
> @leopuleo

Hey @adrians5j, I would use the same naming convention that we are going to use for the [Tag component](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT/Webiny-Design-System?node-id=63-12745&node-type=symbol&m=dev).

Here's the mapping:

CleanShot 2024-12-02 at 07 27 52@2x

leopuleo (approved):

PR #4423, opened 2024-12-02
feat(admin-ui): Tag + Icon Button components

leopuleo:
## Changes
This pull request adds the `Tag` and `IconButton` components to the `@webiny/admin-ui` package. 

### IconButton
This component is used when we want to show a simple button with an Icon. 
In this PR, we are also:

- Remove the `iconSize` prop from the `Button` component.
- Add the `IconButton` to several other components, such as `Tag`, `Card`, `Select`, `Alert`, and `Toast`.
- Remove deprecated props from the `@webiny/ui/Button` component; some (i.e. `flat` or `small`) were passed to the new `@webiny/admin-ui/Button` component, causing console warnings.

### Tag
A visual element used to display brief, contextual information or status. The developers can:

- Make the component interactable by providing an `onClick` callback.
- Make the component dismissible by providing an `onDismiss` callback.
- Customise the "dismiss" icon via `dismissIconElement` and `dismissIconLabel` props.

### Icon
The `Icon` component should be used internally whenever an icon is displayed. Sometimes, we want to set the `Icon` fill colour to the parent for more granular control. For this reason, the default colour prop is `inherit`.


packages/admin-ui/src/Button/Button.tsx
```diff
@@ -5,7 +5,7 @@ import { cn, cva, VariantProps, makeDecoratable } from "~/utils";
 const buttonVariants = cva(
     [
         "border-transparent rounded font-sans inline-flex items-center justify-center whitespace-nowrap ring-offset-background transition-colors",
-        "disabled:pointer-events-none",
+        "disabled:cursor-not-allowed",
```
adrians5j: I love the not allowed cursor.. but do we need to remove pointer-events-none here? Wouldn't the user still be able to click on the button now?

adrians5j (approved):
Awesome man, a bunch of good stuff in here.

Thanks!

PR #4440, opened 2024-12-12
New Admin UI - Link Component

adrians5j:
## Changes
This PR introduces the Link component.

#### 1. Default Text/Link Styles
I updated the `theme.scss` file and added default styles for all text and links that were not rendered via Text/Link components, respectively. I tested this manually then in Admin area.

image

Previously, we had this:
image

#### 2. Alert Links
Once I've added default styles for links, link in Alert boxes did not look as it's defined in Figma (links basically have the same styles as the text in the alert, except they're also bolded and underlined). So, I have to add a couple classes in there as well to preserve this.

image

#### 3. Updated Tokens
I reimported tokens from Figma.

#### 4. Colors With Alpha Were Still Imported
There were a couple color categories where ignoring colors with alpha/transparency was not added. 

Those categories were: `fill`, `ringColor`, and `textColor`.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


adrians5j:
This one is ready for review. The only thing I need to [adjust is the [3px] ring size](https://github.com/webiny/webiny-js/blob/c9be0b77acfd0e11a990f239c87ac0a20305abc5/packages/admin-ui/src/Link/Link.tsx#L18-L33). Don't have a token for this, gonna get it probably next week since Kreso is absent tomo.

packages/admin-ui/src/Link/Link.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { makeDecoratable } from "@webiny/react-composition";
+import { cva, type VariantProps } from "~/utils";
+import {
+    Link as WebinyReactRouterLink,
+    type LinkProps as WebinyReactRouterLinkProps
+} from "@webiny/react-router";
+
+const linkVariants = cva("font-sans rounded-xs", {
+    variants: {
+        size: {
```
leopuleo, line 11: I'm not sure we need a `size` variant: the font size of a link should be inherited from its parent; otherwise, we would end up in this situation: 

`<Text size="lg">Lorem ipsum with <Link size="lg">link</Link> text.</Text>`

declaring the `size` prop twice.

leopuleo (approved):

PR #4443, opened 2024-12-12
feat(app-headless-cms): enable full-screen editor for content entries 

leopuleo:
## Changes
This PR introduces a new experimental full-screen editor for headless CMS entries. The feature is under the feature flag: `allowCmsFullScreenEditor: true`.

As we work on this feature, we enable the EntryForm Header to be decorated. This capability is essential for its development.
CleanShot 2024-12-12 at 18 34 31@2x

## How Has This Been Tested?
Manually


packages/app-headless-cms/src/admin/views/contentEntries/ContentEntry/FullScreenContentEntry/FullScreenContentEntryHeaderLeft.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const { entry } = useContentEntry();
+    const { navigateToFolder, currentFolderId } = useNavigateFolder();
+
+    const title = useMemo(
+        () => entry?.meta?.title || `New ${model.name}`,
+        [entry.meta, model.name]
+    );
+    const isNewEntry = useMemo(() => !entry.meta?.title, [entry.meta]);
+    const version = useMemo(() => entry.meta?.version ?? null, [entry.meta]);
+    const status = useMemo(() => entry.meta?.status ?? null, [entry.meta]);
```
Pavel910: All three hooks will always run, if `entry.meta` changes. You might as well combine all 3 into 1 hook, which returns an object `{ isNewEntry, version, status }`. Otherwise, use specific dependencies on each hook, like `entry.meta?.title`, `entry.meta?.version`, etc.

packages/app-headless-cms/src/admin/views/contentEntries/ContentEntry/FullScreenContentEntry/FullScreenContentEntryHeaderLeft.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const modelName = useMemo(() => model.name, [model.name]);
```
Pavel910: This `useMemo` is completely pointless. You'll spend more processing power to execute the hook, than gain anything. You're memoizing that same string value, which you have in the dependencies array :)

packages/app-headless-cms/src/admin/views/contentEntries/ContentEntry/FullScreenContentEntry/FullScreenContentEntryHeaderLeft.tsx
```diff
@@ -0,0 +1,46 @@
...
+import {
+    EntryMeta,
+    EntryTitle,
+    EntryVersion,
+    TitleWrapper,
+    EntryName
+} from "./FullScreenContentEntry.styled";
+
+export const FullScreenContentEntryHeaderLeft = () => {
+    const { model } = useModel();
```
Pavel910: This hook is redundant, because `useContentEntry()` hook below, has access to that same model, just add a `contentModel` key.

packages/app-headless-cms/src/admin/views/contentEntries/ContentEntry/FullScreenContentEntry/FullScreenContentEntryHeaderLeft.tsx
```diff
@@ -0,0 +1,46 @@
...
+    const { entry } = useContentEntry();
+    const { navigateToFolder, currentFolderId } = useNavigateFolder();
+
+    const title = useMemo(
+        () => entry?.meta?.title || `New ${model.name}`,
+        [entry.meta, model.name]
+    );
```
Pavel910: These type of memoizations are adding more calculations than they're optimizing. These simple assingments can be executed on each render, especially since the output is a primitive `string`.

packages/app-headless-cms/src/admin/views/contentEntries/ContentEntry/FullScreenContentEntry/FullScreenContentEntry.tsx
```diff
@@ -0,0 +1,101 @@
...
+            if (!headerRightElement) {
+                return <Original />;
+            }
+
+            return createPortal(<Original />, headerRightElement);
+        };
+    });
+
+export const FullScreenContentEntry = () => {
+    return (
```
Pavel910, line 86: Add feature flag here, and remove it from all other components:

```
if (!featureFlags.allowCmsFullScreenEditor) {
    return null;
}
````

If the feature is disabled, you simply don't decorate anything.

Pavel910 (changes requested):

Pavel910 (approved):

PR #4468, opened 2024-12-31
New Admin UI - Dialog Component

adrians5j:
## Changes
This PR creates the Dialog component. Here are a couple of screenshots.

image
image
image
image

#### 1. Alerts - Default Variant Change
I [changed](https://github.com/webiny/webiny-js/pull/4468/files#diff-12a24d9f461c477d278909019dc9c9a3074e48ed9fb0e187d3dce6f1fd632560) the default variant from 'subtle' to 'strong'.

image

Felt it was just a bit too much seeing those flashy colors. Let's see along the way how that will look like and revert if needed.

#### 2. Larger Buttons

I removed `small={true}` from these. Looks nice IMO.

image

#### 3. Minor Cleanups on Dialog Components

Removed usage of `emotion` and its `css` util in a couple of places:

https://github.com/webiny/webiny-js/pull/4468/files#diff-81d1c874d39cd9f3f7695ec45c0dc7207d0e8eb80c1d8db4f4e56888623b15c5

https://github.com/webiny/webiny-js/pull/4468/files#diff-6b146a1a388811e719571c4ea24d939474e57ad1efc1535b4133df0a17311c1c

https://github.com/webiny/webiny-js/pull/4468/files#diff-564e985ae1ec27ccae95bb9fa00f2d24823065e198418d0d8799a82c0bb5ad0a

#### 4. `packages/ui/src/Grid/Grid.tsx`

[Updated deprecation comments](https://github.com/webiny/webiny-js/pull/4468/files#diff-9a536db3996a3e04e952363d375527a32a410a4a9d1f3aca5711550da75870d9).

#### 5. Updated Text Sizes
Spoke to Kreso about a couple of missing tokens. Now, text sizes are [fully tokenized](https://github.com/webiny/webiny-js/pull/4468/files#diff-d387a0e53f22492f20f7c58ee547603a6d075f8c18d82498962fb9c3b84d4d16R217-R226) (previously there were a couple of hardcoded pixels in there).

#### 6. Added `inline-block` CSS Class To Dropdown Menu Trigger Wrapper `div`

This is required because otherwise, we end up with the following issue:
image

It's not an ideal solution, but this will all not be needed when we finally resolve that issue where ref forwarding doesn't work on `makeDecoratable` components.

#### 7. Updated `@radix-ui/react-dialog` / `@radix-ui/react-dropdown` Packages

Updating these packages fixed important dropdown menu rendering-related issues that would occur when showing the menu in a dialog and drawer components.

#### 1. Tabs in Dialog

We had a discussion on tabs being rendered in an opened dialog. There were a couple of options here, but, ultimately, I've done it in a way where you pass `bodyPadding: false` prop to the `Dialog` component. That way, the body has no padding, and tabs can be used within the dialog as imagined in Figma designs (they taking the full width of the dialog). 

We still do need to revisit tokens on Tabs, but overall, I think it'll do.

image


packages/admin-ui/src/Dialog/components/CancelButton.tsx
```diff
@@ -0,0 +1,12 @@
...
+import { DialogClose } from "./DialogClose";
+import { Button, ButtonProps } from "~/Button";
+
+const CancelButtonBase = (props: ButtonProps) => (
+    <DialogClose asChild>
+        <Button text={"Cancel"} {...props} variant="secondary" />
+    </DialogClose>
+);
+
+export const CancelButton = makeDecoratable("CancelButton", CancelButtonBase);
```
leopuleo, line 12: I would avoid using `makeDecoratable` for "primitive" components. For now, let’s only allow decoration on “public” components.

packages/admin-ui/src/Dialog/components/ConfirmButton.tsx
```diff
@@ -0,0 +1,9 @@
+import * as React from "react";
+import { makeDecoratable } from "~/utils";
+import { Button, ButtonProps } from "~/Button";
+
+const ConfirmButtonBase = (props: ButtonProps) => (
+    <Button text={"Confirm"} {...props} variant="primary" />
+);
+
+export const ConfirmButton = makeDecoratable("ConfirmButton", ConfirmButtonBase);
```
leopuleo, line 9: Same here, let's remove `makeDecoratable`

packages/admin-ui/src/Dialog/components/DialogContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+import * as DialogPrimitive from "@radix-ui/react-dialog";
+import { cn } from "~/utils";
+
+export interface DialogContentProps
+    extends React.ComponentPropsWithoutRef<typeof DialogPrimitive.Content> {
+    preventOutsideDismiss?: boolean;
+}
+
+const DialogContent = React.forwardRef<
+    React.ElementRef<typeof DialogPrimitive.Content>,
```
leopuleo, line 11: Do we need the `ref` here? If not, let's remove it.

packages/admin-ui/src/Dialog/components/DialogContent.tsx
```diff
@@ -0,0 +1,45 @@
...
+    DialogContentProps
+>(({ className, preventOutsideDismiss, children, ...props }, ref) => {
+    let preventOutsideDismissProps: Pick<
```
leopuleo: Shall we use `useMemo` here? The `preventOutsideDismiss` won't change so it might have sense.

packages/admin-ui/src/Dialog/components/DialogDescription.tsx
```diff
@@ -0,0 +1,16 @@
...
+>;
+
+export const DialogDescription = ({ className, ...props }: DialogDescriptionProps) => (
+    <DialogPrimitive.Description
+        {...props}
+        className={cn("text-sm text-neutral-strong", className)}
+    />
+);
+
+DialogDescription.displayName = DialogPrimitive.Description.displayName;
```
leopuleo: I would remove this: we are not setting `displayName` for other components (without `ref`). Let's keep it consistent.

packages/admin-ui/src/Dialog/components/DialogFooter.tsx
```diff
@@ -0,0 +1,24 @@
...
+    info?: React.ReactNode;
+}
+
+export const DialogFooter = ({ actions, info, className, ...props }: DialogFooterProps) => {
+    if (!actions && !info) {
+        return null;
+    }
+
+    return (
+        <div {...props} className={cn("flex justify-between p-lg pt-md-extra", className)}>
```
leopuleo, line 15: I would implement a check to ensure that `info` and `actions` exist before rendering the outer divs.

packages/admin-ui/src/Dialog/components/DialogHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const DialogHeader = ({
+    title,
+    titleIcon,
+    description,
+    showCloseButton,
+    className,
+    ...props
+}: DialogHeaderProps) => {
+    let renderedTitle = (
```
leopuleo: I don't understand this part: why should I render `VisuallyHidden.Root` should wrap a text hidden from the screen but announced by the screen reader. 

In case of a falsy `title`, there is nothing to announce. Am I missing something?

packages/admin-ui/src/Dialog/components/DialogHeader.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return (
+        <div
+            {...props}
+            className={cn(
+                "flex flex-col gap-sm px-lg py-md text-center sm:text-left text-neutral-primary",
+                className
+            )}
+        >
+            <div className={"relative"}>
+                {showCloseButton !== false && (
```
leopuleo, line 34: I would set `showCloseButton` prop to default `true` instead of` showCloseButton !== false`. But this is a personal choice.

packages/admin-ui/src/Dialog/components/DialogOverlay.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from "react";
+import * as DialogPrimitive from "@radix-ui/react-dialog";
+import { cn } from "~/utils";
+
+export const DialogOverlay = React.forwardRef<
+    React.ElementRef<typeof DialogPrimitive.Overlay>,
+    React.ComponentPropsWithoutRef<typeof DialogPrimitive.Overlay>
+>(({ className, ...props }, ref) => (
+    <DialogPrimitive.Overlay
+        ref={ref}
```
leopuleo, line 10: Do we need the ref here? If not, let's remove it.

packages/admin-ui/src/Dialog/components/DialogTitle.tsx
```diff
@@ -0,0 +1,11 @@
...
+import * as DialogPrimitive from "@radix-ui/react-dialog";
+import { cn } from "~/utils";
+
+export type DialogTitleProps = React.ComponentPropsWithoutRef<typeof DialogPrimitive.Title>;
+
+export const DialogTitle = ({ className, ...props }: DialogTitleProps) => (
+    <DialogPrimitive.Title {...props} className={cn("text-h4 flex gap-sm", className)} />
+);
+
+DialogTitle.displayName = DialogPrimitive.Title.displayName;
```
leopuleo: I would remove this: we are not setting `displayName` for other components (without `ref`). Let's keep it consistent.

packages/admin-ui/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+import { TitleIcon } from "./components/TitleIcon";
+import { ConfirmButton } from "./components/ConfirmButton";
+import { CancelButton } from "./components/CancelButton";
+
+export interface DialogProps
+    extends React.ComponentPropsWithoutRef<typeof DialogRoot>,
+        Omit<React.ComponentPropsWithoutRef<typeof DialogContent>, "title"> {
+    trigger?: React.ReactNode;
+    title?: React.ReactNode | string;
+    titleIcon?: React.ReactElement;
```
leopuleo: Shall we use `icon` instead of `titleIcon`? Since don't have multiple icons to identify.

packages/admin-ui/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+    showCloseButton?: boolean;
+    preventOutsideDismiss?: boolean;
```
leopuleo: Shall we use `disableDismiss` instead of `preventOutsideDismiss`?

packages/admin-ui/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+    bodyPadding?: boolean;
```
leopuleo, line 23: Not sure about this prop, but I don't have a strong opinion on it.

packages/admin-ui/src/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+    description?: React.ReactNode | string;
```
leopuleo: `React.ReactNode` already includes `string`, right? No need to add it.

packages/admin-ui/src/Heading/Heading.tsx
```diff
@@ -3,7 +3,7 @@ import { makeDecoratable } from "@webiny/react-composition";
 import { cva, type VariantProps } from "class-variance-authority";
 import { cn } from "~/utils";
 
-export type HeadingTags = "h1" | "h2" | "h3" | "h4" | "h5" | "h6";
+export type HeadingTags = "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "p" | "div";
```
leopuleo: Why do we need to add `p` and `div` tags?

packages/app-admin-rmwc/src/modules/Overlays/Dialog.tsx
```diff
@@ -47,23 +53,30 @@ export const DialogContainer = () => {
...
                 {actions.cancel && (
-                    <DialogAccept onClick={actions.cancel.onClick}>
+                    <DialogCancel onClick={actions.cancel.onClick}>
                         {actions.cancel.label}
-                    </DialogAccept>
+                    </DialogCancel>
                 )}
                 {actions.accept && (
-                    <ButtonPrimary
+                    <DialogAccept
```
leopuleo, line 73: :heart:

leopuleo (changes requested):
Nice one, Adrian! 

There are a few minor comments here and there; otherwise, it looks good. Regarding the naming of the props, it’s just my personal opinion, but we can discuss it if you wish.

packages/admin-ui/src/Dialog/Dialog.tsx
adrians5j: Makes sense...  

Although still.. `icon` prop and `Alert.Icon` sounds a bit vague to me when compared to `titleIcon` and `Alert.TitleIcon`. I mean... the question that pops up in my head: "Is this an icon that can be used anywhere in the dialog or is it a specific location? 🤔".

In other words, it's not clear that it's the icon that's shown in the title of the dialog. 

IDK, might be subjective ofc. But if you'd rather have it set to just `icon`, let me know and I'll make the adjustment. No problem at all.

packages/admin-ui/src/Dialog/Dialog.tsx
adrians5j: What about just `canDismiss` or `dismissable` ? I'm like always trying to name props in a positive way, not as a negation. 

So basically, instead of doing `disableDismiss={true}`, users would do `dismissable={false}`. Wdyt?

packages/admin-ui/src/Dialog/Dialog.tsx
adrians5j: Not sure as well. But added it without much thinking TBF / to just move on. It _is_ essentially the implementation of the idea we discussed with Kreso.

packages/admin-ui/src/Dialog/Dialog.tsx
adrians5j: Nice one, ty!

packages/admin-ui/src/Dialog/components/CancelButton.tsx
adrians5j: Actually, I followed the rule "whatever is exposed to the user, we can make it decoratable". Did I get the rule wrong then? 😀

packages/admin-ui/src/Dialog/components/ConfirmButton.tsx
adrians5j: Same as with Cancel. :)

packages/admin-ui/src/Dialog/components/DialogContent.tsx
adrians5j: Good point!

packages/admin-ui/src/Dialog/components/DialogContent.tsx
adrians5j: Good point.

packages/admin-ui/src/Dialog/components/DialogContent.tsx
adrians5j: Actually, it's needed. Without it, I start receiving:

> Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Looks like Radix needs it.

packages/admin-ui/src/Dialog/components/DialogDescription.tsx
adrians5j: Ah cool.

packages/admin-ui/src/Dialog/components/DialogFooter.tsx
adrians5j: Sounds good.

packages/admin-ui/src/Dialog/components/DialogHeader.tsx
adrians5j: This way I'm explicitly not-rendering only if `false` was passed. Every other value is considered as `true` and close button is shown. 

In other words, ths saves me the trouble of defining default values in props in multiple components (parent Dialog and this one). I know it's not like this component will be used standalone, but still, less to think about. :)

And it's probably also just a personal choice as you said. ☺️

packages/admin-ui/src/Dialog/components/DialogHeader.tsx
adrians5j: Hey man.... I don't know what happened, but I remember without all of these weird things I did, Radix would complain about "description" / "title" missing. And to use VisuallyHidden in this case. Something along those lines.

In any case, I did more testing, and realized this new version of the code also works without any warnings in the console. 🤷‍

I'm thinking maybe it was the package upgrade I did that fixed this. In any case, thanks for noticing this, I believe now it should be much clearer.

packages/admin-ui/src/Dialog/components/DialogOverlay.tsx
adrians5j: Radix complains when ref is ommited: 

> Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

packages/admin-ui/src/Dialog/components/DialogTitle.tsx
adrians5j: Nice one.

packages/admin-ui/src/Heading/Heading.tsx
adrians5j: Removed it. I think I needed it at one point, but ultimately looks like this is no longer the case. 

Thanks!

packages/admin-ui/src/Dialog/Dialog.tsx
leopuleo: Let's go ahead and leave this as it is for now.

packages/admin-ui/src/Dialog/Dialog.tsx
leopuleo: Ok, I like the positive-naming approach. Let's proceed with `canDismiss` or `dismissable`

packages/admin-ui/src/Dialog/Dialog.tsx
leopuleo: As I see it, it's similar to the `description` prop: they are very similar in how they are used inside the component, and we are not referring to it as `titleDescription`.

packages/admin-ui/src/Dialog/components/CancelButton.tsx
leopuleo: You are right, sorry...I missed the fact the component is "public" 😄

packages/admin-ui/src/Dialog/components/DialogHeader.tsx
leopuleo: Ok, let's leave it like it is 👍

packages/admin-ui/src/Dialog/Dialog.tsx
adrians5j: Went with just `icon`.

PR #4470, opened 2025-01-03
fix: aco small improvements

leopuleo:
## Changes
This PR aims to fix some minor yet annoying bugs found in ACO-powered apps (Page Builder, CMS, File Manager):

- **Folder Tree:** if an error occurs while moving a folder across the tree, it returns to its original position.
- **Dialogs**: the `useDialogs` hook now supports multiple dialogs simultaneously. This functionality is essential when creating a new folder while transferring an entry into another folder.
- **File Manager**: the displayed number of loaded files appears as a string instead of a placeholder (`currentCount`).
- **File Manager**: When users upload one or multiple files, the File Manager drawer unintentionally closed. This issue has been resolved, thanks to @Pavel910. 
- **ACO Table**: the total count of entries present in a list is updated any time a user creates, deletes or moves entries.
-  **Bulk actions icons**: the z-index for the icon wrapper has been adjusted to align with its container. 

## How Has This Been Tested?
Manually


packages/app-aco/src/contexts/records.tsx
```diff
@@ -279,6 +287,10 @@ export const SearchRecordsProvider = ({ children }: Props) => {
                     setRecords(prev => {
                         return prev.filter(record => record.id !== recordId);
                     });
+                    setMeta(meta => ({
```
adrians5j, line 290: Removing/adding from meta within the `getRecord` method is something I don't quite understand, but I guess it's fine / I might be missing some context :)

packages/app-aco/src/dialogs/DialogSetPermissions/UsersTeamsSelection/ListItemMeta.tsx
```diff
@@ -122,6 +122,7 @@ export const ListItemMeta = ({
                     disabled={!!disabledReason}
                     // Should prevent first item from being autofocused, but it doesn't. 🤷‍
                     focusOnOpen={false}
+                    portalZIndex={101}
```
adrians5j, line 126: Why this change? Maybe a comment here?

adrians5j (changes requested):
The code itself looks fine I guess.

UPDATE: I had this text below here, but ignore it. Left another comment [below](https://github.com/webiny/webiny-js/pull/4470#pullrequestreview-2538788002).
> Although, while testing, I did stumble upon moving an entry into a new folder:
> https://github.com/user-attachments/assets/4c0641e7-3781-45e8-a048-e15ac1b17747
> 
> If it's not already resolved in another PR, maybe we check it here?
>

adrians5j (approved):
Actually you know what, scratch the video I attached above.

That one was from `dev` 🙂 . So I guess we'll have it fixed with 5.42.0 (the bug is not present in your branch).

But one other thing I did notice is the following:

https://github.com/user-attachments/assets/8afd6da8-3766-4401-a805-8e21bc8cb26e

Basically I clicked on an entry's Move action, and then, in the dialog, did not select anything and just submitted the form, and then the entry was moved to somewhere, who knows where. :D 

If you want, for now, we could maybe just backlog this one. It was probably always there. Or if you want, I can take a look at it.

But yeah, I'm approving this PR now, since all of the things that were covered with it are 🟢 .

packages/app-aco/src/dialogs/DialogSetPermissions/UsersTeamsSelection/ListItemMeta.tsx
leopuleo: Done! Thank you for noticing this. 👍

packages/app-aco/src/contexts/records.tsx
leopuleo: This file needs a review, and some methods should be revisited. Currently, `getRecord` updates the record in the cache like the other methods above:

- addRecordToCache
- updateRecordInCache
- removeRecordFromCache

We'll get to that.

adrians5j (approved):

PR #4471, opened 2025-01-03
feat: add PB page migration

Pavel910:
## Changes
This PR adds a PB Pages migration, which moves page attributes from the root of the dynamo entity to the `data` envelope. This makes adding new attributes much easier, and will not require updates to the Entity definition in the future.

In the storage operations, all actions performed against Dynamo DB now include a compatibility layer, which ensures that the data is written to both the root of the record and the new `data` attribute. Once we get to 5.43., this compatibility layer can be removed.

TODO: PB Blocks migration to achieve the same goal.

## How Has This Been Tested?
Jest tests and manually.


Pavel910:
I decided to not go through with this PR, as in some mid-term future we want to migrate Page Builder to use CMS as a storage layer, so this migration is just a waste of time and an unnecessary risk (no matter how low the risk is).

PR #4473, opened 2025-01-03
fix(MultiAutoComplete): use CSS variable instead of hardcoded values

### LINKED ISSUE #4472: Replace hardcoded colors with variables
# Description
While styling the admin app, I encountered these hardcoded values that should be changed in the core to avoid using a complex CSS targeting technique.

## Code to change
- https://github.com/webiny/webiny-js/blob/552f914547aac9978f83ee86ef486ab3335fb724/packages/ui/src/AutoComplete/MultiAutoComplete.tsx#L40
- https://github.com/webiny/webiny-js/blob/552f914547aac9978f83ee86ef486ab3335fb724/packages/ui/src/AutoComplete/MultiAutoComplete.tsx#L131

lopadz:
## Changes
Replace hardcoded color value with theme variable, using default Webiny color as a fallback.
Closes #4472 

## How Has This Been Tested?
Visually

### DIFF
--- a/packages/ui/src/AutoComplete/MultiAutoComplete.tsx
+++ b/packages/ui/src/AutoComplete/MultiAutoComplete.tsx
@@ -37,7 +37,7 @@ const style = {
             display: "flex",
             justifyContent: "space-between",
             alignItems: "center",
-            borderBottom: "2px solid #fa5723",
+            borderBottom: "2px solid var(--mdc-theme-primary, #fa5723)",
             padding: "6px 0"
         }),
         pages: css({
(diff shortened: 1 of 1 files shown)


Pavel910:
@lopadz thanks for the fix. I can't merge this PR, because `dev` is already 5.41.x, and you need this in 5.40.x. I moved your changes manually to [this commit](https://github.com/webiny/webiny-js/commit/105f3d8acf9fab9c63d285e496c224904d0d8f62), and we'll ship it as part of the 5.40.8.

PR #4475, opened 2025-01-07
feat: create Drawer component

adrians5j:
## Changes

This PR creates the Drawer component. Here is a pair of sshots.

The first is File Manager, where `modal` prop is set to `false` (default value actually). This basically means the user can also interact with the UI in the background and that the drawer is only closed when pressing Esc or clicking on the **X** icon. This was done as discussed with Kreso.

image

This is the main menu, where `modal` prop is set to `true`, meaning no background interactions are possible.

image

Another random sshot:

image

#### 1. Updated File Manager Drawer Code
[Had to accommodate the file details drawer in File Manager a bit](https://github.com/webiny/webiny-js/pull/4475/files#diff-775d9ff701d7990dc639f91953d72661c6bb8192b6cb107ad5d172bb212a80ab).

Did some cleanups along the way too (some styled components are no longer needed).

#### 2. File Manager - Increased Size of Buttons
I made these buttons larger:

image

#### 3. Updated Query Builder Drawer Code
[Had to accommodate the file details drawer in ACO's Query Builder a bit](https://github.com/webiny/webiny-js/pull/4475/files#diff-74aa3a9808356b19ac42c036895b8179b45619885d7493b5eba3074a273c4027).

Did some cleanups along the way too (some styled components are no longer needed).

#### 1. Tabs in Dialog

We had a discussion on tabs being rendered in an opened dialog. There were a couple of options here, but, ultimately, I've done it in a way where you pass `bodyPadding: false` prop to the `Dialog` component. That way, the body has no padding, and tabs can be used within the dialog as imagined in Figma designs (they taking the full width of the dialog).


packages/admin-ui/src/Drawer/components/DrawerContent.tsx
```diff
@@ -0,0 +1,70 @@
...
+        defaultVariants: {
+            side: "right"
+        }
+    }
+);
+
+const DrawerContent = React.forwardRef<
+    React.ElementRef<typeof DrawerPrimitive.Content>,
+    DrawerContentProps
+>(({ side = "right", modal, width, className, style = {}, children, ...props }, ref) => {
```
leopuleo: We can remove the default assigned to `right` prop: the default variant is already assigned here above.

packages/admin-ui/src/Drawer/components/DrawerHeader.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+export const DrawerHeader = ({
+    title,
+    icon,
+    description,
+    showCloseButton,
+    className,
+    ...props
+}: DrawerHeaderProps) => {
+    if (!title && !description) {
```
leopuleo: We might need to display the `close` button even without `title` and `description` values. If we can I would move this check down below.

packages/admin-ui/src/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,111 @@
...
+import { DrawerOverlay } from "./components/DrawerOverlay";
+import { DrawerPortal } from "./components/DrawerPortal";
+import { DrawerRoot } from "./components/DrawerRoot";
+import { DrawerTrigger } from "./components/DrawerTrigger";
+import { DrawerBody } from "~/Drawer/components/DrawerBody";
+import { Icon } from "./components/Icon";
+import { ConfirmButton } from "./components/ConfirmButton";
+import { CancelButton } from "./components/CancelButton";
+
+export interface DrawerProps
```
leopuleo, line 15: Shall we export this at the end of the file? It's more of a cosmetic change rather than a functional one.

leopuleo (changes requested):

packages/admin-ui/src/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,111 @@
...
+    icon?: React.ReactElement;
+    modal?: boolean;
+    showCloseButton?: boolean;
+    bodyPadding?: boolean;
+    description?: React.ReactNode;
+    children: React.ReactNode;
+    actions?: React.ReactNode;
+    info?: React.ReactNode;
+    width?: React.CSSProperties["width"];
+    tabs?: boolean;
```
leopuleo: This prop is not used in the code.

leopuleo:
The PR looks lovely, thank you Adrian 🙏 
I left some comments here and there, and these are minor change requests.

I have a couple of general comments:

#### 1) Scrollable content
The content inside the drawer is not scrollable. Check out the video:

https://github.com/user-attachments/assets/8eda55d9-03af-4898-8ad5-32e5a7aff262

Shall we try to prevent this behaviour by default or leave the developer to handle it when needed?

#### 2) Internal components naming
I've noticed that the internal components have the `Drawer` prefix. I usually don't use it, so `DrawerBody` -> `Body`. 

It's debatable, and after we finish developing all the components, we should go through the library and adjust these minor aspects.

#### 3) Use of `children` as `Drawer` prop
Similar to the `Dialog` component, I've noticed we are using the default `children` prop to provide the content to the drawer. Shall we use dedicated props like `content`, similar to what we do with `title`, `description` etc.? 

Also in this case, we can adjust this later.

packages/admin-ui/src/Drawer/Drawer.tsx
adrians5j: LOL good catch. I have no idea how this ended up in here. :D

packages/admin-ui/src/Drawer/components/DrawerContent.tsx
adrians5j: Nice.

packages/admin-ui/src/Drawer/components/DrawerHeader.tsx
adrians5j: Cool.

packages/admin-ui/src/Drawer/Drawer.tsx
adrians5j: It sounds OK, although I've been doing this across all components so far, so I'm thinking not to do anything at this point.

Maybe an ESLint rule or something could help make these things more consistent.

adrians5j:
Resolved scrollable content issue by adding overflow-auto to the drawer body section. 

So, scrollbars will show if the height of the body is not enough to show all content in it. The main menu is also showing scrollbars now.

Re other points... we can discuss them later.

PR #4480, opened 2025-01-07
fix(app-headless-cms): various  full screen editor

leopuleo:
## Changes
This PR introduces a couple of fixes for the full-screen editor in the Headless CMS:

1. Now, the full-screen editor is the default. Developers can disable it by using the legacy two-column setup, setting the `cmsLegacyEntryEditor` feature flag.
2. The full-screen editor container has been expanded by 100px to more effectively accommodate a particular use case: the `ref` field with the `Detailed view with modal search` renderer in a two-column layout. Refer to the screenshots for visual reference.

### Screenshots
CleanShot 2025-01-07 at 15 38 43@2x
CleanShot 2025-01-07 at 15 38 13@2x

## How Has This Been Tested?
Manually + Cypress


adrians5j (changes requested):
Looks good. Maybe we slightly rename the flag here?

Maybe from `allowCmsLegacyEntryEditor` to just `cmsLegacyEntryEditor`?

IDK, don't see the point in having the "allow" verb. :)

adrians5j (approved):

adrians5j (approved):

PR #4492, opened 2025-01-15
fix(admin-ui): various UI fixes

leopuleo:
## Changes
With this PR introduces multiple changes:

- Add `xs` icon size.
- Fix the form components spacing.
- Fix the form components `ghost` variant when `disabled: true`.
- Cast `Tooltip` to `font-weight: normal`.
- Improve `extendTailwindMerge` configs.
- Style and refactor the `Tabs` component's public props (it now accepts a `tags` prop with the child tabs definition).
- Fix `Avatar` component while displaying images.
- Remove unused `Chip` props.
- Remove `Popover` portal from Autocomplete components.
- Include `Tooltip.Portal`; there were instances when the tooltip content became invisible.

## How Has This Been Tested?
Manually


packages/admin-ui/src/AutoComplete/primitives/AutoCompletePrimitive.tsx
```diff
@@ -116,22 +120,20 @@ const AutoCompletePrimitive = (props: AutoCompletePrimitiveProps) => {
...
-                            onOptionSelect={handleSelectOption}
-                            isEmpty={vm.optionsListVm.isEmpty}
-                            isLoading={props.isLoading}
-                            loadingMessage={vm.optionsListVm.loadingMessage}
-                            emptyMessage={vm.optionsListVm.emptyMessage}
-                            optionRenderer={props.optionRenderer}
-                        />
-                    </Popover.Content>
-                </Popover.Portal>
+                <Popover.Content
```
adrians5j, line 125: It says here popover was removed:
image

But here it's actually here.

What did I get wrong? 😇

packages/admin-ui/src/Avatar/Avatar.tsx
```diff
@@ -40,13 +40,13 @@ interface AvatarProps
...
 
-const avatarVariants = cva("relative flex shrink-0 overflow-hidden border-sm border-transparent", {
+const avatarVariants = cva("relative flex shrink-0 overflow-hidden", {
     variants: {
         size: {
-            sm: "text-h6 rounded-sm p-[calc(theme(padding.xs)-theme(borderWidth.sm))] [&>*]:size-md",
-            md: "text-h6 rounded-md p-[calc(theme(padding.xs)-theme(borderWidth.sm))] [&>*]:size-lg",
-            lg: "text-h6 rounded-md p-[calc(theme(padding.sm)-theme(borderWidth.sm))] [&>*]:size-lg",
-            xl: "text-h4 rounded-lg p-[calc(theme(padding.sm)-theme(borderWidth.sm))] [&>*]:size-xl"
+            sm: "text-h6 rounded-sm size-[24px] [&_svg]:size-[16px]",
```
adrians5j, line 46: Can we utilize [spacing tokens](https://github.com/webiny/webiny-js/blob/9707cc256d258ec3195352e4a95f26877dd07466/packages/admin-ui/src/theme.scss#L168-L181) here?

adrians5j (changes requested):
#### 1. Orange Border Height
I noticed this is 3px in Figma, and in code it's 2px.
image

But I think you already posted a comment to K about this, right?

#### 2. Not sure if this should be in line. 
image

#### 3. Active States
Should we discuss these with K maybe?
image

image

#### 4. Tabs Moving?
Isn't it weird how the right tab is moving by 1px as the left one is switching between active/inactive?

https://github.com/user-attachments/assets/84a175ce-05c0-431d-9412-b436794dd388

adrians5j:
> [!NOTE]  
> At this point, this is just a discussion.

BTW what do you think about the following API here?

```tsx
<Tabs>
    <Tabs.Tab id="account" title="Account" icon={<Tabs.Tab.Icon icon={...} label={...}/>}>Account content</Tabs.Tab>
    <Tabs.Tab id="password" title="Password">Password content</Tabs.Tab>
</Tabs>
```

Some notes...
1. The current `Tabs` component uses `tabs` prop and has no `children` being used. Since both of these are just regular props, and both can be affected via a decorator, maybe might as well use `children` here? It'll also be a bit more in the spirit of React.
2. In this example, I'm using `Tabs.Tab` instead of POJO. Note that both `Tabs` and `Tabs.Tab` can also be made as decoratable, giving users full control.
3. Using `Tabs.Tab.Icon` not only allows the user to specify the icon, but also other props. Like in this example, a label for the icon, if needed. This is how I also did it with Dialog and it's `icon` prop. You pass `Dialog.Icon` element to it.
4. BTW I used `id` instead of `value`. `value` sounds a bit weird to me... but hey, if it's how it works in the React/components world, I guess we leave it as is.

The only potentially tricky thing I can see here is the fact that the `title` needs to be extracted from the child components, by the `Tabs` parent component, in order to construct triggers. [I can see where doing this already](https://github.com/webiny/webiny-js/blob/765738aef0e41317df2a2bd0f46287148cdb105c/packages/admin-ui/src/Tabs/Tabs.tsx#L34-L52), but I can also see it's easier to do this when you have `tabs` POJO prop, and not these child components (although I believe it's still possible to read `props` from these child components, I think I did it in the past).

packages/admin-ui/src/Avatar/Avatar.tsx
leopuleo: I would love to, but we don't have a `40px` token 😢 
Let's leave it as it is for now...what do you think?

packages/admin-ui/src/AutoComplete/primitives/AutoCompletePrimitive.tsx
leopuleo: I've checked; it has been removed from the Autocomplete component because it's already placed inside `Popover.Content`.

packages/admin-ui/src/Avatar/Avatar.tsx
adrians5j: OK

leopuleo:
All the styling issues highlighted above have been solved or discussed with K.

### `tabs` prop
I've changed from  POJO to the `Tabs.Tab` component. We need some little magic under the hood to extract all the valuable props for the underlying `Trigger` and `Content` components. The exact mechanism is used today.

This component is a good candidate for improvements via FTA; I noted it in the GH project.

Regarding the `tabs` vs `children` props: let's revisit them together with the other components.

### `tab` icon
Currently, the `icon` prop receives any `ReactElement`, aka any SVG. We use the `Icon` component inside the Trigger where we pass the SVG received as a prop with fixed size, color and label. 

I would keep it like this: I don't want to repeat all these fixed props while using this component around the admin.

### `id` vs `value` prop
I chose to use `value` instead of `id` because this is what Radix UI uses, together with `onValueChance` callback. Also, the `id` prop is used internally by Radix UI to couple the content and the trigger.

CleanShot 2025-01-22 at 15 35 14@2x

PR #4505, opened 2025-01-26
New Admin UI - Accordion Component

adrians5j:
## TODO
- [x] Check with Kreso
  - [x] [focus-active state (overall design + rounded borders check)](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT?node-id=3245-2246&m=dev#1117934731)
  - [x] [hover (overall design, watch out for rounded borders, icon button hover missing)](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT?node-id=3245-2246&m=dev#1119868453)
  - [x] [with handle... border not around handle icon](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT?node-id=3245-2246&m=dev#1121090049)
- [ ] should we skip header title - [bolded / regular variants](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT/Webiny-Design-System?node-id=3245-6447&m=dev) for now? Not sure how to name the prop for this and thought maybe we just wait for the usecase to arrive, then we deal with it.

## Changes
This PR creates the Accordion component.

Some screenshots.

image
image
image

#### 1. Icon Component - `label` Prop Optional
A lot of times, I needed to use the Icon component, without the need to provide the label for the icon. So, I made the `label` prop optional.

#### 2. Removed Redundant Comments
While working on the `wby-` prefix, code generation left comments in a couple of places. For example:

```
// No Tailwind classes to modify in this file
```

I removed these.

#### 3. Reordered `keyframes` and `animation` In `tailwind.config.js`

I had to add `keyframes` / `animation` into `tailwind.config.js`, and then I noticed these already got here via the Skeleton PR. All good, but I still just alphabetically ordered the properties.

#### 4. Form Builder - Minor Tweaks
Had to do a bit of tweaking there. Nothing special. But it is included in this PR.


packages/admin-ui/src/Accordion/components/AccordionContent.tsx
```diff
@@ -0,0 +1,49 @@
...
+const accordionContentVariants = cva(
+    [
+        "wby-overflow-hidden wby-text-md",
+        "wby-transition-all data-[state=closed]:wby-animate-accordion-up data-[state=open]:wby-animate-accordion-down"
+    ],
+    {
+        // Using pixel values here because of non-existing design tokens.
+        variants: {
+            withIcon: {
+                true: "wby-pl-[36px]"
```
leopuleo, line 14: Shall we use `wby-pl-5` and `wby-pl-9` instead? 
Generally, we don't have tokens for every possible spacing value, shall we use Tailwind classnames when possible?

I leave it as comment, not a blocker.

packages/admin-ui/src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,52 @@
...
+        }
+        return rootProps;
+    }, [baseRootProps]);
+
+    return (
+        <AccordionRoot
+            {...rootProps}
+            className={cn(
+                "w-full",
+                { "wby-gap-xs wby-flex wby-flex-col": variant === "container" },
```
leopuleo: Why not use the `cva` utility?

packages/admin-ui/src/Accordion/components/AccordionItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+    return (
+        <AccordionPrimitive.Item
+            {...itemProps}
+            className={cn(
+                "wby-border-b-sm wby-border-b-neutral-dimmed data-[state=open]:wby-rounded-bl-lg data-[state=open]:wby-rounded-br-lg",
+                {
+                    "wby-rounded-lg": variant === "container",
+                    "wby-bg-neutral-base": background === "base",
+                    "wby-bg-neutral-light": background === "light"
+                }
```
leopuleo: I would use `cva` util.

packages/admin-ui/src/Accordion/components/AccordionItem.tsx
```diff
@@ -0,0 +1,73 @@
...
+            )}
+        >
+            <AccordionTrigger {...triggerProps} />
+            <AccordionContent {...contentProps} />
+        </AccordionPrimitive.Item>
+    );
+};
+
+AccordionItemBase.displayName = "AccordionItem";
```
leopuleo: Do we need this? Maybe not, since we are not using `ref`

packages/admin-ui/src/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,52 @@
...
+            )}
+        >
+            <AccordionContext.Provider value={{ variant, background }}>
+                {children}
+            </AccordionContext.Provider>
+        </AccordionRoot>
+    );
+};
+
+AccordionBase.displayName = "Accordion";
```
leopuleo: Do we need this? Maybe not, since we are not using `ref`

packages/admin-ui/scripts/importFromFigma/createTailwindConfigTheme.js
```diff
@@ -8,6 +8,11 @@ const isColorWithAlpha = variantName => {
 
 const createTailwindConfigTheme = normalizedFigmaExport => {
     return {
+        animation: {
+            "accordion-down": "accordion-down 0.2s ease-out",
+            "accordion-up": "accordion-up 0.2s ease-out",
+            skeletonPulse: "skeletonPulse 1400ms ease-in-out infinite"
```
leopuleo: Shall we stick to a camelCase or kebab-case naming convention? The kebab-case looks good, but I would change `skeletonPulse` :pray:

packages/admin-ui/src/Accordion/components/AccordionContext.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+import { type AccordionProps } from "../Accordion";
+
+const AccordionContext = React.createContext<Pick<AccordionProps, "variant" | "background">>({});
+
+const useAccordion = () => {
```
leopuleo: The `AccordionContext` holds variant and background values accessible to inner components.

I had a similar situation with the `Toast` component, and I managed to style the inner components by adding `wby-group` and the variation class names to the root component -> https://github.com/webiny/webiny-js/blob/feat/new-admin-ui/packages/admin-ui/src/Toast/components/Root.tsx#L12-L15

Then, I can target the parent class name from the inner component and style it differently based on the variation class name -> https://github.com/webiny/webiny-js/blob/feat/new-admin-ui/packages/admin-ui/src/Toast/components/Title.tsx#L15

What do you think?

packages/admin-ui/src/Icon/Icon.tsx
```diff
@@ -29,12 +29,12 @@ const iconVariants = cva("", {
 interface IconProps
     extends Omit<React.HTMLAttributes<HTMLOrSVGElement>, "color">,
         VariantProps<typeof iconVariants> {
-    label: string;
+    label?: string; // We don't want to force the label to be required.
```
leopuleo: I understand we don't want to force this property, but it's actually quite useful if we aim to have an accessible application. It's okay to make it optional, but providing a label should be encouraged as a best practice.

leopuleo (changes requested):
Thank you Adrian, I left some comments here and there but the overall PR looks good!

packages/admin-ui/scripts/importFromFigma/createTailwindConfigTheme.js
adrians5j: Done.

packages/admin-ui/src/Icon/Icon.tsx
```diff
@@ -29,12 +29,12 @@ const iconVariants = cva("", {
 interface IconProps
     extends Omit<React.HTMLAttributes<HTMLOrSVGElement>, "color">,
         VariantProps<typeof iconVariants> {
-    label?: string; // We don't want to force the label to be required.
+    label: string; // We don't want to force the label to be required.
```
leopuleo: The inline comment is outdated.

leopuleo (approved):

PR #4508, opened 2025-01-27
feat(admin-ui): Loader component (CircularProgress)

leopuleo:
## Changes
With this PR we are refactoring the `Loader` component - previously known as `CircularProgress` and applying some fixes to `Autocomplete` and `MultiAutocomplete`.

### Loader
This component, previously known as `CircularProgress`, has been refactored to receive a controlled `value` (number) or the `indeterminate` flag (boolean).

If set to `indeterminate` (default: `true`), the component will show a spinning circular progress bar. 

Otherwise, the component can accept the controlled `value`, along with `min` and `max`. In this case, the Loader will display the current state while loading.

The component has been styled following the design specifications provided by K.

### OverlayLoader
This component is a variant of `Loader` designed to overlay content, making it ideal for scenarios where we need to indicate that the entire page or a specific section is loading. 

It serves as the equivalent of the `CircularProgress` component from the old UI package.

### Autocomplete and MultiAutocomplete
These two components has been improved: 

- displaying the loading status using the `Loader` component within the input
- introducing a new prop called `initialMessage`, which is utilized when the AutoComplete options list is empty and the user has not yet performed a search for an option.
- removing `is` prefix to boolean props (`isLoading` -> `loading`)
- improving the selected options state management for MultiAutocomplete
- handling the icon size by referring to the current input size.

## How Has This Been Tested?
Manually + Jest


packages/admin-ui/src/MultiAutoComplete/primitives/presenters/MultiAutoCompleteListOptionsPresenter.ts
```diff
@@ -8,18 +8,20 @@ export interface IMultiAutoCompleteListOptionsPresenterParams {
...
 
 export interface IMultiAutoCompleteListOptionsPresenter {
     vm: {
         options: CommandOptionFormatted[];
-        emptyMessage: string;
-        loadingMessage: string;
-        isOpen: boolean;
-        isEmpty: boolean;
+        emptyMessage: any;
+        loadingMessage: any;
```
adrians5j, line 18: Should we fix this? 🤔

adrians5j (commented):
Two minor questions but let's merge 💣

adrians5j (approved):
Left a question, but let's merge 💣

packages/admin-ui/src/MultiAutoComplete/primitives/presenters/MultiAutoCompleteListOptionsPresenter.ts
leopuleo: I would not do that right now. Both `emptyMessage` and `loadingMessage` could be a `React.ReactNode`, but I'm not sure I want to tie this type into the presenter.

Let's maintain it this way for the time being.

PR #4519, opened 2025-02-03
feat(admin-ui): IconPicker component

leopuleo:
## Changes
With this PR, we are upgrading the IconPicker component to ensure it complies with the new design system UI.

You'll find 2 `IconPicker` components: 

### `webiny/admin-ui`
CleanShot 2025-02-05 at 14 46 33@2x
This component was previously integrated into the `@webiny/app-headless-cms` and `@webiny/app-page-builder` packages. Both of these components now utilize the generic version exported by `@webiny/admin-ui`.

This component only renders icons from `@fortawesome/react-fontawesome` package.

### `webiny/app-admin`
CleanShot 2025-02-05 at 14 46 13@2x
This component is utilised by `@webiny/app-page-builder` and is more flexible regarding which icons can be handled. 

### `<DelayedOnChange />`
This component has been transferred from `@webiny/ui` to the fresh `@webiny/admin-ui`.

### Add `forwardEventOnChange` prop to `<Input />` and `<Textarea />`
We introduce the `forwardEventOnChange?: boolean` prop; if set to true, the full event will be forwarded to the provided `onChange` callback. This prop was previously named `rawOnChange`.

### `<Popover />`
We set the `collisionPadding` prop to prevent the component from colliding with the edge of the viewport.

## How Has This Been Tested?
Manually


packages/admin-ui/src/IconPicker/primitives/IconPickerPrimitive.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { Popover } from "~/Popover";
+import { cn, cva, type VariantProps } from "~/utils";
+import { IconPickerGrid, IconPickerInput, IconPickerTrigger } from "./components";
+import { IconDto } from "../domains";
+import { useIconPicker } from "./useIconPicker";
+
+const iconPickerVariants = cva("wby-cursor-pointer wby-text-neutral-strong", {
+    variants: {
+        size: {
+            md: "wby-w-[64px]",
```
adrians5j: I guess we also use regular number here instead of px? Just for consistency maybe? 🤔

packages/app-headless-cms/src/admin/components/IconPicker.tsx
adrians5j, line 1: ❤️

packages/app-headless-cms-common/src/types/index.ts
```diff
@@ -17,6 +17,7 @@ import {
 import { CmsModel, CmsModelField } from "./model";
 import { CmsIdentity } from "~/types/shared";
 import type { SourceType } from "dnd-core";
+import { IconDto } from "@webiny/admin-ui";
```
adrians5j: Naming.... constant pain :( 

It's just hard to know what's this `IconDto` related to. 

Like... if it was maybe `IconPickerIconDto`, that would be more understandable. But then ofc that's a looooong name. 

I'm not saying we do anything about it now, but we might want to note this down somewhere / see how we want to deal with these things.

packages/app-admin/src/components/IconPicker/components/IconPickerTrigger.tsx
```diff
@@ -0,0 +1,62 @@
...
+import { ReactComponent as EmptyIcon } from "@material-design-icons/svg/outlined/remove.svg";
+import { ReactComponent as ChevronDown } from "@material-design-icons/svg/outlined/keyboard_arrow_down.svg";
+import { cn, cva, Icon as IconComponent, inputVariants, type VariantProps } from "@webiny/admin-ui";
+import { IconProvider, IconRenderer } from "~/components/IconPicker/IconRenderer";
+import { Icon } from "~/components/IconPicker/types";
+
+const iconPickerTriggerVariants = cva("wby-cursor-pointer fill-neutral-xstrong", {
+    variants: {
+        size: {
+            md: "wby-w-[64px]",
```
adrians5j, line 11: Pixels here, don't know if you want to change it to TW numbers. 🤷🏻

adrians5j (changes requested):
Initial review done. Overall, looks good ofc!

Extra questions below. Note: nr2 is probably good.

1. Should these be a bit darker? 
image

2. This input is more gray in Figma.
image

3. Ah I guess it's because in Figma it's not focused. Still, found this placeholder to be a bit darker in Figma:
image

Maybe I just don't see it right. Can you dbl check?

4. Can you dbl check these paddings?
image

5. Should we ask Kreso to provide this? At least note it down somewhere for him?
image

6. Do we have an example where a component does not already have a value assigned to it? Like [this](https://www.figma.com/design/f0QUDWX37Kt5X53eltTRiT/Webiny-Design-System?node-id=3342-2101&m=dev)?

PR #4523, opened 2025-02-04
feat(cli-plugin-deploy-pulumi): add region to pulumi cli arguments

brunozoric:
## Changes
Add an option for user to deploy to a certain region, which is not necessarily the one in .env files.
Also removed the handler-logs usage from all packages.
Simple handler-logs package was left, otherwise the `yarn up "@webiny/*@5.43.x"` would not work.

## How Has This Been Tested?
Manually.


packages/pulumi-aws/src/components/tenantRouter/WebsiteTenantRouter.ts
adrians5j, line 1: This whole file can be deleted I believe. @Pavel910 mentioned this to me some time ago.

packages/pulumi-aws/src/env/base.ts
```diff
@@ -0,0 +1,33 @@
...
+    name: string;
+    defaultValue: T;
+}
+
+/**
+ * TODO @adrian @pavel verify please
+ *
+ *
+ * We do not allow env with name "undefined" or "null". It is considered as an empty string.
+ * Sometimes, when setting env variable, a user error might set it to "undefined" or "null" value by mistake.
```
adrians5j, line 11: Sounds good.... hopefully nobody will need these specific values in their env vars, but we'll see. :)

packages/pulumi-aws/src/env/logsForwardUrl.ts
```diff
@@ -0,0 +1,6 @@
+import { createGetEnvOptional } from "~/env/base";
+
+export const getEnvVariableLogsForwardUrl = createGetEnvOptional<string>({
+    name: "WEBINY_LOGS_FORWARD_URL",
```
adrians5j: No longer needed. This was the logs forwarding feature related.

packages/handler-aws/src/execute.ts
```diff
@@ -71,12 +71,20 @@ export const execute = (params: ExecuteParams): Promise<any> => {
     });
     const cookies = getPayloadProperty(payload, "cookies", {});
 
+    // console.log({
+    //     executing: true,
+    //     payload,
```
adrians5j: Leftover here.

packages/cli-plugin-deploy-pulumi/src/utils/env/withLogsForwardUrl.ts
adrians5j, line 1: Can be deleted.

packages/cli-plugin-deploy-pulumi/src/utils/env/withEnvVariant.ts
```diff
@@ -0,0 +1,17 @@
+import { createConfiguration } from "./configuration";
+
+export interface IWithEnvVariantParams {
+    variant: string | undefined;
+}
+
+export const withEnvVariant = (params: IWithEnvVariantParams) => {
+    return createConfiguration(() => {
+        const variant = (params.variant || "").trim();
+        if (!variant || variant.length < 3) {
```
adrians5j: Does this mean that if user set `a` as variant name, that it would not be in the config?

Would this silently happen?

Thinking if we need the `length < 3` check here. 🤔

packages/cli-plugin-deploy-pulumi/src/utils/aws/uploadFolderToS3.ts
```diff
@@ -51,6 +51,10 @@ export interface IUploadFolderToS3Params {
     cacheControl?: PutObjectCommandInput["CacheControl"] | CacheControlMap;
 }
 
+/**
+ * TODO @adrian
```
adrians5j, line 55: Dbl check with @Pavel910 if he's maybe using it somewhere or some user of his is maybe using it. Is it possible that this was just exposed for user consumption?

But yeah, could be a leftover.

adrians5j (changes requested):
Looks good, just a couple of minor comments.

PR #4534, opened 2025-02-12
fix: ensure Admin User entry creation with Cognito Federation

adrians5j:
## Changes
This PR ensures Admin User entries are created when logging into Webiny via a federated identity, when achieved via the [Cognito Federation](https://www.webiny.com/docs/enterprise/cognito-federation) feature.

Note that in order to do that, I've changed the logic that determines whether:
1. the `getIdentity` callback needs to be run
2. groups/teams authorizer needs to be added
3. Admin user hooks need to be executed (Admin user entry creation needs to happen)

Now, we basically do all of the above only if the identity is federated, which is determined via the `identities` property in the received JWT token. [Check this change here](https://github.com/webiny/webiny-js/pull/4534/files#diff-21927ddb94357c55384a3452dd96f87fea11dcccd07f4ec489b82365f0d130f1R76).

> [!NOTE]  
> Let's just double check / confirm this over a call @Pavel910.

#### 1. Separated `onLogin` Event Subscription and Context Plugin Creation

In [`packages/api-admin-users/src/createExternalIdpAdminUserHooks.ts`](https://github.com/webiny/webiny-js/pull/4534/files#diff-0b8a30efc9aaca77a365ea3a35a5e349a73b745470682b1b2afdca5b346c635a), I had to make a slight refactor, where I had to extract the actual `onLogin` event subscription and context plugin creation into two separate functions: `createExternalIdpAdminUserHooks` and `createExternalIdpAdminUserHooksPlugin`. This was needed when trying to achieve the point number 3 in the above list.

#### 2. Removed `createAdminUsersHooks.ts` Files
Both Okta and Auth0 packages had `createAdminUsersHooks.ts` file, which was just reexporting the `createExternalIdpAdminUserHooks` fn from `@webiny/api-admin-users/createExternalIdpAdminUserHooks` file. 

I deleted these two files and directly imported the new `createExternalIdpAdminUserHooksPlugin` from `@webiny/api-admin-users/createExternalIdpAdminUserHooks`.

## How Has This Been Tested?
Manually.

## Documentation
Changelog.


packages/api-security-cognito/src/createCognito.ts
```diff
@@ -88,20 +73,27 @@ export const createCognito = <
...
                 } as unknown as TIdentity;
 
-                if (typeof getIdentity === "function") {
-                    const customIdentity = getIdentity({
+                const federatedIdentity = Boolean(tokenObj.identities);
+                if (!federatedIdentity) {
+                    return defaultIdentity;
+                }
+
+                const authorizer = createGroupsTeamsAuthorizerHandler(config, context);
```
adrians5j, line 91: Always execute `getIdentity`, no matter if we're using federation or now.

PR #4542, opened 2025-02-18
feat(admin-ui): List and DataList components

leopuleo:
## Changes
With this pull request, we introduce two new components: `List` and `DataList`.

These two components are utilised to present data across various admin app views. The `List` component serves as the foundational element and is used internally by `DataList`, which displays data along with additional props necessary for functionalities such as filtering and ordering, etc.

With this PR, we refactored the views that are using `DataList` to:

- Show the data correctly; see `Breaking changes`. 
- Use the button icons exported by the `DataList` component: now the icons used are imported directly from `@material-design-icons/svg/outline` set.

### Breaking changes
Previously, developers could use any component as direct `children` of `List.ListItem`. This is no longer possible; to ensure the correct positioning, the appropriate `List.ListItem` child tag must now be used (i.e.: `<ListItemTextPrimary />`). 

Example:
```diff
<ListItemText>
-  {item.firstName} {item.lastName}
+  <ListItemTextPrimary>
+    {item.firstName} {item.lastName}
+  </ListItemTextPrimary>
   <ListItemTextSecondary>{item.email}</ListItemTextSecondary>
</ListItemText>
```

### Missing from the design

- [ ] ListItem `active` and `selected` state.

## Other improvements
- **Accordion**: various minor fixes to the component; most of which are cosmetic. 
- **FormComponentProps**: `FormComponentProps` now accept `TValue` generic for the `onChange` prop.
- **IconPicker**:  any domain entity name has been prefixed with `IconPicker` namespace to not collide with other components such as `Icon`.
- **Heading** and **Text:**  incorporate text class names into `extendTailwindMerge` for enhanced merging.

CleanShot 2025-02-18 at 16 00 27@2x
CleanShot 2025-02-18 at 16 00 13@2x
CleanShot 2025-02-18 at 15 59 57@2x
CleanShot 2025-02-18 at 15 59 43@2x
CleanShot 2025-02-18 at 16 11 20@2x


adrians5j (approved):
Looks good.

A ton of files, but it wasn't that bad. I mean, it's basically a couple of different changes being repeated, so the 107 number of files wasn't that big of a deal.

I noticed these two, but feel free to merge.

Disabled state incorrect here?
image

And also.... I see there's focus state on list item, which is good. The thing I don't understand is what should actually happen when trying to interact with it via keyboard? 🤷🏻 

Something for backlog I'd say.
image

PR #4544, opened 2025-02-19
feat(admin-ui): add RichTextEditor component

leopuleo:
## Changes
This pull request introduces the `RichTextEditor` component to the @webiny/admin-ui package.

## How Has This Been Tested?
Manually


adrians5j (approved):
Saw a couple of these glitches, but I don't think we need to concern ourselves. It's an editor that will slowly just disappear if you ask me. And in Admin UI, all looks good.

Let's merge this!

Icons going out of box.
image

Editor text not on the far left, but centered in the box.
image

In Admin it looks good.
image

PR #4555, opened 2025-02-24
fix: control HCMS entry editor width using `Width` config

leopuleo:
## Changes
With this PR, we are introducing a new `ContentEntryEditorConfig` called `Width`.

The `Width` component in the `ContentEntryEditorConfig` accepts the following props:
- `value`: A string representing the width value (e.g., "100%", "960px").
- `modelIds`: An optional array of strings specifying the model IDs to which this width configuration applies. The width configuration will not be applied if the current model's ID is not in this array.

### Bonus
The `IsApplicableToCurrentModel` is also used for the `Action` config, rather than manually checking the provided model IDs.

## How Has This Been Tested?
Manually

## Documentation
[This extension](https://github.com/webiny/webiny-examples/blob/master/cms-live-preview/5.42.x/README.md) must be updated to use the newly created config instead of customising the editor width via CSS.


Pavel910 (approved):
Very nice! 👍

PR #4566, opened 2025-03-11
Sidebar / Header Bar in Admin UI 

adrians5j:
Big PR, but addresses a ton of things. Mainly, it integrates the new Sidebar component into Admin UI via the new `app-admin-ui` package, which is basically the old `app-admin-rmwc` package.

image

This PR also introduces new Menu/Route config components, which will be the way to go when it comes to adding new route/menus in Admin UI. For example:

```tsx
<AdminConfig>
    <IsRootTenant>
        <Menu
            name={"tenantManager"}
            element={
                <Menu.Link
                    icon={<TenantManagerIcon/>}
                    label={"Tenant Manager"}
                    path={"/tenants"}
                />
            }
        />

        <Route
            name={"tenantManager.tenants"}
            exact
            path={"/tenants"}
            element={
                <Layout title={"Tenant Manager - Tenants"}>
                    <TenantsView/>
                </Layout>
            }
        />
    </IsRootTenant>
</AdminConfig>
```

## Additional Changes

1. `@webiny/app-admin-rmwc` removed (`@webiny/app-admin-ui` introdouced)
2. Removed OmniSearch - we'll resurrect it later.
3. `HeaderBar` component introduced in `@webiny/admin-ui`
4. `Sidebar` minor fixes (removed extra classes in root, fixed collapse/expand chevron positioning)
5. Updated menus/routes/top-right user menu code across multiple packages
6. `@webiny/react-properties`'s `$first` and `$last` now work correctly (thanks goes to @Pavel910)
7. `DropdownMenu` component - added `readOnly` option to `DropdownMenu.Item` component
8. Removed code that was no longer needed (some styled components, SVGs, utilities, ...)

## More Work To Be Done
There's a solid amount of work yet to be done. Wanted to do it in this PR, but no time. 

So, the TODO for my next PR looks something like this:
- [ ] sometimes sidebar fully rerenders when navigating across pages (I know what this is)
- [ ] introduce `Tenant.Name` and `Tenant.Logo` components (`AdminConfig`)
- [ ] check `No routes matched location "/" Error Component Stack` error in console
- [ ] check `- Password field is not contained in a form...` warning in console
- [ ] recheck HCMS sidebar menu (when adding a new model, it gets added into the menu incorrectly)
- [ ] introduce `Menu.Support` so that users can add items into the Support nav item (into the dropdown)
- [ ] enable users to also add items into the top-right dropdown menu via config API (atm it's still using the old way)
- [ ] with Kreso, dbl check the locale selector and also the top-right user menu's dropdown menu (both are missing in Figma)
- [ ] gotta see what's this for [`packages/app-apw/src/plugins/cms/MenuGroupRenderer.tsx`](https://github.com/webiny/webiny-js/blob/next/packages/app-apw/src/plugins/cms/MenuGroupRenderer.tsx#L11) (I deleted it in this PR, but still gotta check if this was a mistake)
- revisit the removal of `AddMenu` and `AddRoute` components.... still thinking if this was too hasty, and thinking to resurrect these and have them internally rewired to simply use new Config APIs.. most probably, I will do this

Plus a couple of other smaller things, no need to write them down here.


packages/app-admin/src/components/SplitView/SplitView.tsx
```diff
@@ -26,7 +26,7 @@ const grid = css({
 const RightPanelWrapper = styled("div")({
     backgroundColor: "var(--mdc-theme-background)",
     overflow: "auto",
-    height: "calc(100vh - 64px)"
+    height: "calc(100vh - 45px)"
```
leopuleo, line 29: Shall we create a variable to hold this value, as shown below, and use it instead of the arbitrary value?

```
module.exports = {
  theme: {
    extend: {
      spacing: {
        'header-bar': '45px',
      }
    }
  }
}
```

This is merely a suggestion; let me know what you think and if this makes sense.

packages/admin-ui/src/HeaderBar/HeaderBar.tsx
```diff
@@ -0,0 +1,28 @@
+import React from "react";
+import { makeDecoratable } from "~/utils";
+import { Separator } from "~/Separator";
+
+interface HeaderBarProps {
+    left?: React.ReactNode;
+    right?: React.ReactNode;
+}
```
leopuleo: Shall we use `start` and `end` instead of `left` and `right`?

leopuleo: Maybe we need something similar to the current `TopAppBarSection` component that receives an `align` prop. 

The new component will be used together with `HeaderBar`:

```tsx
<HeaderBar>
    <HeaderBar.Section align="start">Left component</HeaderBar.Section>
    <HeaderBar.Section>Center component</HeaderBar.Section>
    <HeaderBar.Section align="end">Right component</HeaderBar.Section>
</HeaderBar>
```

packages/admin-ui/src/HeaderBar/HeaderBar.tsx
leopuleo, line 1: The component is not re-exported from `@webiny/ui` package.

packages/app-admin/src/components/SplitView/SplitView.tsx
adrians5j: Your proposal makes sense. But, what it does not make so much sense is this whole approach I'd say.

I think having this kind of calculation should be avoided, and we should find another way to ensure the content below header fills the whole remaining vert space of the screen. 

All in all, I believe there must be a better way to accomplish this. I'll check it out in my upcoming PR.

Thanks for mentioning this!

packages/admin-ui/src/HeaderBar/HeaderBar.tsx
adrians5j: Awesome, making this change now!

packages/admin-ui/src/HeaderBar/HeaderBar.tsx
adrians5j: Will take a look at this one in the next PR, thanks for reporting!

PR #4570, opened 2025-03-19
Config APIs - Footer Menus / Support Menus / Other Fixes

adrians5j:
## Changes
The main goal of this PR was to further improve config APIs that enable users to add:
1. menu items in the footer section of the sidebar
2. menu items in the support dropdown

Adding this code...

```tsx
// These two were already implement as part of the https://github.com/webiny/webiny-js/pull/4566 PR.
// Included them just for context.
<Menu
    name={"home"}
    pin={"start"}
    element={<Menu.Link text={"Home"} icon={<DashboardIcon />} path={"/"} />}
/>
<Menu
    name={"settings"}
    pin={"end"}
    element={
        <Menu.Link
            text={"Settings"}
            icon={<SettingsIcon />}
            path={"/access-management/roles"}
        />
    }
/>

// 🆕🆕🆕👇 Below are the new APIs. 👇🆕🆕🆕
<Menu.Footer
    name={"test-item"}
    element={<Menu.Item icon={<InfoIcon/>} text={"Test Item"}/>}
/>

<Menu.Support
    name={"test-support-item"}
    element={
        <Menu.Support.Item
            text={"Test support item"}
            icon={<InfoIcon/>}
            onClick={() => console.log("Test support item clicked!")}
        />
    }
/>

<Menu.Support
    name={"test-support-link-item"}
    element={
        <Menu.Support.Link
            text={"Test support item"}
            icon={<InfoIcon/>}
            to={"https://www.webiny.com/docs"}
            rel={"noopener noreferrer"}
            target={"_blank"}
        />
    }
/>
```

... will result in the following:

image
image

##### 1.1 `label` to `text`  
To make things more consistent, I renamed `label` to `text` for the three `AdminConfig.Menu.{Link|Item|Group}` components. The newly added  `Menu.Support` and `Menu.Footer` are also consistent with this.

I did this because I basically started noticing headaches while creating `AdminConfig` components, where I used `label` prop, but, internally, the `admin-ui` components were using `text` ([example](https://github.com/webiny/webiny-js/pull/4570/files#diff-c4d900f83d12d8b6028205a91f147697c66033e4bf45aeceae55b4c6b6942882)). This was a bit confusing, so I decided to make everything consistent.

##### 1.2 `content` to `text`
I then also noticed the same thing while creating the new `Menu.Support`, which is using `DropdownMenu.Item` behind the scenes. And again, here I needed to perform internal remapping, this time `label` -> `content`. So, ultimately, I did the `content` to `text` rename for the `DropdownMenu.{Item|Link|Label|Group|CheckboxItem}` components, and also the `label` to `text` rename on the `Menu.Support` components.

##### 1.3. `path` to `to`
Finally, the `AdminConfig.Menu.Link` component had its own `path` prop, which was again being internally remapped to `to`. In order to not do remapping, here I actually did not do any prop renaming, but instead, ensured that the component basically extends the `Link` component from `@webiny/react-router` package, which uses `to` prop and also has the same API as the `Link` component from `@webiny/components` package. So now users can user things like `rel`, `target` or any other prop that the `Link` component from `@webiny/react-router` package supports. Example:

```tsx
 <Menu.Support.Link
    text={"Documentation"}
    icon={<Menu.Support.Link.Icon label="Docs" element={<DocsIcon />} />}
    to={"https://www.webiny.com/docs"}
    rel={"noopener noreferrer"}
    target={"_blank"}
/>
```

#### 2. Further Refreshed Modules In `@webiny/app-admin-ui` Package
I refreshed the code in the `@webiny/app-admin-ui` package:

1. removed usage of `lodash`
2. used `createDecorator` instead of `compose` (consulted Pavel about this)
3. restructured folders
4. `userMenu` is still pending because I have a bit of extra work on this one (coming in next PR)

#### 3. TenantSelector

Added `TenantSelector` placeholder component. Previously this did not exist and the tenant selector rendered itself by decorating the `LocaleSelector` component, within the `@webiny/app-tenant-manager` package. We can see it being used in the main [`Layout.tsx`](https://github.com/webiny/webiny-js/pull/4570/files#diff-c80853f80cfe4557123f2a9da2c331605a80b844284aa339300868b941ee3a35).

#### 4. Wrapped Locale Selector Implementation
With the new design updates fron Kreso, I've now finalized the `LocaleSelector` component.

#### 5. `displayResetAction` on Select Component
Initially, the `displayResetAction` was only available on the Autocomplete component. Now this is also available on the Select component. Thanks goes to Leo!

#### 6. User Menu - Icons
User menu had old icons being used. I've updated them to the new ones (followed updated designs from Kreso).

#### 7. Refactored Base.tsx
Nothing special, but since the file was getting too big, I extracted menus/routes into separate `Menus.tsx` and `Routes.tsx` files.

#### 8. Added `DropdownMenu.Link` Component
Prior to this PR, we did not have the `DropdownMenu.Link` component and it was a bit tricky to add a link to the dropdown menu. Now, with this new component, users can easily add links to the dropdown menu.

This makes APIs more consistent also with the `AdminConfig.Menu.Link` component and `Sidebar.Link` component.

#### 9. Added `DropdownMenu.{Item|Link}.Icon` Components

For consistency, I've added the `Icon` component to the `DropdownMenu.Item` and `DropdownMenu.Link` components. 

#### 10. Login Screen Centered
The login screen was not centered. I've now centered it.


Pavel910:
@adrians5j No suggestions from my end. The public API is exactly how we discussed it, and it looks great!

I have an idea though, related to the original `Menu` component from the previous PR. The `before="$first"` and `after="$last"` are very technical, low-level terms. Maybe we can add a `pin` prop? Like `pin="top"` or `pin="bottom"` to make it a bit more friendly, and then internally you convert that prop into the low-level properties API. What do you think? Btw. the `before` and `after` props would still work, because you still need an ability to reorder menus. If you think it's not necessary, I'm fine with leaving it as it is. Curious to hear your thoughts.

adrians5j:
The $first/$last was something I was also thinking about, just did not have it high enough on the priority list heh.

But I totally agree and now that you’ve mentioned, I’ll definitely do it.

Thanks!

adrians5j:
Updated the cod above. So, it's `pin={"start"}` or `pin={"end"}`.

Used start/end, since we used this way of defining order in other components during the new Admin UI journey.

In general, I'm liking these words since they're not tied to vert/horizontal layouts. Whatever the layout is, the words work and user _should_ 🤞🏻  be able to understand what start/end means in the component they're using.

PR #4609, opened 2025-04-22
Fix listPackages and watch command

### LINKED ISSUE #4608: Watch command fails due to Yarn logs
### Version

5.42.2

### Operating System

MacOS 15.4.1

### Browser

Safari

Yarn version 1.22.22
Running `yarn webiny watch admin --env dev` in my newly created and deployed Webiny app

To start the admin app locally

### What do you see instead?

 I got an error like below.

Image

### Additional information

_No response_

### Possible solution

I've traced it down to the listCommands command in file `node_modules/@webiny/cli-plugin-deploy-pulumi/commands/watch/listPackages.js`

Adding --silent option to the yarn commands in this file on line 10 and 23 solved my issue for me.

I'm not sure this issue is specific to my yarn installation or not, but I think having the --silent option in the listPackages.js file wouldn't hurt.

cyberst:
## Changes
Added --silent option to yarn commands so the output doesn't have unnecessary lines that starts with $
image

Closes #4608

## How Has This Been Tested?
Manually

### DIFF
--- a/packages/cli-plugin-deploy-pulumi/commands/newWatch/listPackages.js
+++ b/packages/cli-plugin-deploy-pulumi/commands/newWatch/listPackages.js
@@ -19,6 +19,7 @@ const listPackages = async ({ inputs }) => {
         packagesList.push(...webinyPrefixedPackagesToAdd);
     } else {
         packagesList = await execa("yarn", [
+            "--silent",
             "webiny",
             "workspaces",
             "tree",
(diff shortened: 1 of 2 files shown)


adrians5j:
Thanks for the PR @cyberst , I'll take a look! 🙏

adrians5j:
Sry for the delay here :|

My question here would be... Does this really fix the issue? Because I can see the `--silent` was not passed to a `yarn` cmd, but to `yarn webiny workspaces tree ...`.

Am I missing smt here? 🤔 

image

PR #4616, opened 2025-05-07
fix: message on network error

brunozoric:
## Changes
When a network error happens during the GraphQL mutation to create/update/delete a CMS entry, we show an error via snackbar.

When env variable `DEBUG` is set to `true`, an error overlay will show as well with detailed error.

Also, now by default, fastify `reply` has a 200 code.

## How Has This Been Tested?
Manually.


ma-schmidt-de:
Sounds great. Thanks Bruno!

PR #4623, opened 2025-05-12
feat(migrations): add FLP 5.43.0 migration

leopuleo:
## Changes
This PR introduces a migration for version 5.43.0 to support the ACO FLP integration. The migration ensures that folder synchronisation is handled correctly, creating new FLP records. It includes the following changes:

A specifically designed BG Task handles the actual Folder -> FLP synchronisation called `acoSyncFlp`. The migration script calls this task for each `tenant` / `locale` in the system.

## How Has This Been Tested?
Jest


packages/migrations/src/migrations/5.43.0/001/ddb-es/index.ts
```diff
@@ -0,0 +1,279 @@
...
+                    })
+                })
+            ];
+
+            await batchWriteAll({
+                table: this.ddbEntryEntity.table,
+                items: ddbItems
+            });
+
+            batchWriteAll({
```
brunozoric: Missing await here.

brunozoric (changes requested):
Only one comment. Cant find anything else. Didnt test it though.

packages/migrations/src/migrations/5.43.0/001/ddb-es/index.ts
adrians5j: Not sure why my comment got lost but just wanted to say thanks for this!

PR #4630, opened 2025-05-19
FLP - Improved Permissions Inheritance

adrians5j:
## Changes
With this PR, we're improving FLP folder permissions inheritance.

First of all, once all of the permissions for a folder have been collected, then, the newly introduced `DefaultPermissionsMerger` class will inspect them and decide which permission is the winning one.

Some of the use cases the merger tackles:
- permissions inherited from parent can be overridden by new permissions (except `no-access`)
- `no-access` permission cannot be overridden, even if an owner level permission has been assigned

Second, we've also revisited the `Permissions.ts` class, which is responsible for merging folder permissions during the creation of FLP records. Nothing changed here drastically, except the fact that, if parent folder has the `no-access` permission for a particular target, then the permissions for that same target will also be no-access for all child folders. This cannot be overridden.

While changing the `Permissions.ts` file, I've also slightly refactored the file.

## How Has This Been Tested?
Jest.

## Documentation
Changelog.

### DIFF
--- a/packages/api-aco/src/flp/FolderLevelPermissions/useCases/GetDefaultPermissions/DefaultPermissionsMerger.ts
+++ b/packages/api-aco/src/flp/FolderLevelPermissions/useCases/GetDefaultPermissions/DefaultPermissionsMerger.ts
@@ -1,0 +1,108 @@
+import type { FolderAccessLevel, FolderPermission } from "~/flp/flp.types";
+import { Identity } from "@webiny/api-authentication/types";
+import { SecurityPermission } from "@webiny/api-security/types";
+
+export class DefaultPermissionsMerger {
+    static merge(
+        identity: Identity,
+        identityPermissions: SecurityPermission[],
+        folderPermissions: FolderPermission[]
+    ): FolderPermission[] {
+        // If the user has full access permission, add a specific "owner" permission to the list.
+        // This ensures the user has complete control over the folder.
+        const hasFullAccess = identityPermissions.some(p => p.name === "*");
+        if (hasFullAccess) {
+            return [
+                // Remove any permissions related to the full access user,
+                // as these are always superseded by the "owner" permission defined above.
+                ...folderPermissions.filter(p => p.target !== `admin:${identity.id}`),
+                {
+                    target: `admin:${identity.id}`,
+                    level: "owner" as FolderAccessLevel,
+                    inheritedFrom: "role:full-access"
+                }
+            ];
+        }
+
+        if (folderPermissions.length === 0) {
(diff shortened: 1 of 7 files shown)


packages/api-aco/src/flp/useCases/Permissions.ts
```diff
@@ -8,33 +8,52 @@ export class Permissions {
...
-        const folderPermissions = permissions ?? [];
+        const parentFolderPermissions = parentFlp?.permissions || [];
+        const currentFolderPermissions =
+            permissions?.filter(p => p.inheritedFrom !== `parent:${parentFlp?.id}`) || [];
 
-        // No permissions from the parent, let's return the permissions provided by the folder.
-        if (!parentFlp || !parentFlp.permissions?.length) {
-            return folderPermissions;
+        // If there are no parent folder permissions, we can return the current folder permissions.
+        if (!parentFolderPermissions.length) {
```
leopuleo, line 12: I found a bug here when all permissions have been removed from the parent and the change propagates to the children. 

In this particular use case, the parent permissions are not removed from the permissions found in the children because of the `!parentFlp.permissions?.length` check. It's a bug introduced by the last refactor. 

We can remove this branch entirely and test the use case (I can provide it to you).

packages/api-aco/src/flp/FolderLevelPermissions/useCases/GetDefaultPermissions/DefaultPermissionsMerger.ts
```diff
@@ -0,0 +1,110 @@
...
+                resultPermission = currentPermission;
+                continue;
+            }
+
+            if (currentPermission.level === "editor") {
+                if (resultPermission.level === "viewer") {
+                    resultPermission = currentPermission;
+                }
+            }
+        }
```
leopuleo: What about using an `Array.reduce` here?

```
// resultPermission: the winning permission after reduction
const resultPermission = restAdminPermissions.reduce((winner, current) => {
    const winnerInherits = winner.inheritedFrom?.startsWith("parent:");
    const currentInherits = current.inheritedFrom?.startsWith("parent:");

    if (winnerInherits && !currentInherits) {
        return current;
    }
    if (currentInherits && !winnerInherits) {
        return winner;
    }

    // Now compare access levels: owner > editor > viewer
    if (current.level === "owner") {
        return current;
    }
    if (current.level === "editor" && winner.level === "viewer") {
        return current;
    }
    return winner;
}, firstAdminPermission);
```

leopuleo (changes requested):

leopuleo (approved):

PR #4640, opened 2025-05-23
fix: export `createModelFieldPlugin` from `@webiny/api-serverless-cms` for consistency

swapnilmmane:
## Changes
In `v5.42.1` (PR #4502), we consolidated Headless CMS backend plugins and fixed naming issues. `createModelField` was not aligned with the new pattern. This commit deprecates `createModelField`, adds `createModelFieldPlugin`, and exports it from `@webiny/api-serverless-cms` for consistency.

## How Has This Been Tested?
Manually.

## Documentation
We will mention this in the changelog of the next patch release.

### DIFF
--- a/packages/api-headless-cms/src/utils/createModelField.ts
+++ b/packages/api-headless-cms/src/utils/createModelField.ts
@@ -8,7 +8,53 @@ export interface CreateModelFieldParams
     storageId?: string;
 }
 
+/**
+ * @deprecated Use createModelFieldPlugin instead.
+ */
 export const createModelField = (params: CreateModelFieldParams): CmsModelField => {
+    // We are not just spreading `params` because we want to ensure
+    // a default value is set for every property of the model field.
+    const {
+        id,
+        label,
+        fieldId: initialFieldId,
+        storageId,
+        type,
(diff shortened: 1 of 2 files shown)


packages/api-headless-cms/src/utils/createModelField.ts
```diff
@@ -50,3 +53,46 @@ export const createModelField = (params: CreateModelFieldParams): CmsModelField
         renderer
     };
 };
+
+export const createModelFieldPlugin = (params: CreateModelFieldParams): CmsModelField => {
```
adrians5j, line 57: Actually, I would leave out the `Plugin` suffix, because the fn result is not a plugin. It's just model field object.

Which is different from other utilities, which actually _do_ return a plugin object / instance.

adrians5j (changes requested):
Looks good, although I would revisit the rename. Check my comment!

packages/api-headless-cms/src/utils/createModelField.ts
swapnilmmane: sounds good, thank you @adrians5j!

PR #4645, opened 2025-05-27
fix(app-aco): various admin FLP related fixes

leopuleo:
## Changes
With this PR, we are addressing several FLP-related issues, making the experience smoother and more predictable for users. 

### Rename "Remove access" menu item + Disable it when the permission is inherited.
With the 5.43.0 version, owners can change inherited permissions levels (User 1: `Viewer` -> `Owner`), but not delete them. The folder will always inherit permissions from its parent if not overwritten by a more specific one.

To avoid confusion in the admin UI, we implemented two fixes: 
- Rename the menu item from "Remove access" to "Remove permission".
- Disable the menu item in case of an inherited permission

CleanShot 2025-05-27 at 13 29 24@2x

### Update `permissions` for child folders found in cache
When users update the `permissions` of a folder, the permissions of the folder's children are refreshed by a background task. With this PR, we optimistically refresh all child folders in the admin app cache.

### Update `path` for child folders found in cache
Similar to `permissions`, the folder's children `path` is refreshed by a background task. With this PR, we optimistically refresh all the child folder paths in the admin app cache.

## How Has This Been Tested?
Jest + Manually

To test the folder `path`, you can add a new column to CMS tables using the code here below.

```diff tsx
import React from "react";
import { Admin } from "@webiny/app-serverless-cms";
import { Cognito } from "@webiny/app-admin-users-cognito";
+ import { ContentEntryListConfig } from "@webiny/app-headless-cms";
import { Extensions } from "./Extensions";
import "./App.scss";

+ const { Browser } = ContentEntryListConfig;

export const App = () => {
    return (
        <Admin>
            <Cognito />
            <Extensions />
+           <ContentEntryListConfig>
+              <Browser.Table.Column name="path" header="Path" />
+          </ContentEntryListConfig>
        </Admin>
    );
};

```


packages/shared-aco/src/flp/Permissions.ts
```diff
@@ -3,7 +3,7 @@ import type { FolderLevelPermission, FolderPermission } from "~/flp/flp.types";
 export class Permissions {
     public static create(
         permissions?: FolderPermission[],
-        parentFlp?: FolderLevelPermission | null
```
adrians5j, line 6: Didn't we do `FolderLevelPermission | null` in our of our last reviews, just to replace what's on the right here?

packages/api-aco/src/flp/flp.types.ts
```diff
@@ -5,7 +5,7 @@ import { type AcoContext, type Folder } from "~/types";
 export type FolderAccessLevel = "owner" | "viewer" | "editor" | "public" | "no-access";
 
 export interface FolderPermission {
-    target: string;
+    target: `admin:${string}` | `team:${string}`;
```
adrians5j: Can we reuse types from `shared-aco` package here? And I have the same question for `packages/app-aco/src/types.ts`.

adrians5j (approved):
A couple of minor things, nothing special.

All in all, nice one!

packages/shared-aco/src/flp/Permissions.ts
leopuleo: Yes, you are correct. I had to revert and select certain fields because the `FolderLevelPermission` entity is utilised solely by the backend. In the frontend, we are using `Folder`. 

See: `packages/app-aco/src/features/folders/updateFolder/UpdateFolderRepositoryWithPermissionsChange.ts`

adrians5j (approved):

PR #4662, opened 2025-06-19
fix: prevent folder move without permissions and add move confirmation dialog

leopuleo:
## Changes
This PR introduces a new feature and fixes an existing bug, both of which are related to folders.

### Confirmation dialog on folder drag & drop
We are introducing a new config called `DropConfirmation` under the `Browser.Folder` namespace. The config is available for the following apps: 

- CMS - `ContentEntryListConfig`
- Page Builder - `PageListConfig`
- File Manager - `FileManagerViewConfig `

The default value for this config is `false`.  When the value is set to `true`, the editor must confirm the action when moving a folder from one parent to another.

For example, this config enables a confirmation dialog when the user attempts to move a folder within a model of `type: "article"`:

```
<ContentEntryListConfig>
    <Browser.Folder.DropConfirmation value={true} modelIds={["article"]} />
</ContentEntryListConfig>
```
CleanShot 2025-06-19 at 12 13 23

### Check folder permission on folder move 
This PR also fixes a bug: users should not be able to move a folder into a parent folder they do not have access to, based on their permission level (`viewer`).

## How Has This Been Tested?
Manually + Jest

### DIFF
--- a/packages/api-aco/__tests__/folder.flp.security.test.ts
+++ b/packages/api-aco/__tests__/folder.flp.security.test.ts
@@ -371,6 +371,119 @@ describe("Folder Level Permissions - Security Checks", () => {
         });
     });
 
+    it("should not allow moving a folder to an inaccessible folder, based on permission level (owner vs editor vs viewer)", async () => {
+        const folderA = await acoIdentityA
+            .createFolder({
+                data: {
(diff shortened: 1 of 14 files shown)


adrians5j (approved):
Nice one man!

PR #4667, opened 2025-07-15
fix(api-aco): check security authorization while decorating cms methods

leopuleo:
## Changes
With this PR, we are disabling FLP decoration applied to CMS CRUD methods when authorisation is disabled, such as when using `context.security.withoutAuthorization`.

## How Has This Been Tested?
Manually


brunozoric (approved):
Code looks ok. If the logic `if (!context.security.isAuthorizationEnabled()) {` is correct, everything is fine.

PR #4737, opened 2025-10-13
Remove `wby-` Prefix From Tailwind Classes

adrians5j:
As the title itself states, this PR is removing the `wby-` prefix in Tailwind CSS classes.

This should make our code more readable and easier to use.


adrians5j:
No longer needed.

PR #4753, opened 2025-10-23
Make the HCMS Content Entry grey

ku5ic:
## Changes
This pull request introduces UI improvements and minor refactoring to Tabs components, along with a visual update to the FullScreenContentEntry container. The most significant changes are grouped below by theme.

**Tabs component styling:**

* Changed the default background of tab content to `wby-bg-neutral-base` and added margin and rounded corners for a more polished look. (`packages/admin-ui/src/Tabs/components/Content.tsx`)

**FullScreenContentEntry container update:**

* Updated the container background from `wby-bg-neutral-base` to `wby-bg-neutral-light` for a lighter appearance. (`packages/app-headless-cms/src/admin/views/contentEntries/ContentEntry/FullScreenContentEntry/FullScreenContentEntry.styled.tsx`)

<details>
<summary>SCREENSHOTS</summary>

Screenshot 2025-10-23 at 12 02 35

</details>

Closes [WEB-5373](https://webiny.atlassian.net/browse/WEB-5373)
## How Has This Been Tested?
n/a

## Documentation
n/a

### DIFF
--- a/packages/admin-ui/src/Tabs/components/Content.tsx
+++ b/packages/admin-ui/src/Tabs/components/Content.tsx
@@ -4,7 +4,8 @@ import { cn, cva, type VariantProps } from "~/utils.js";
 
 const tabContentVariants = cva(
     [
-        "wby-bg-transparent",
+        "wby-bg-neutral-base",
+        "wby-my-lg wby-rounded-lg",
         "focus-visible:wby-outline-none focus-visible:wby-ring-lg focus-visible:wby-ring-primary-dimmed",
         // By default, the inner content is removed by the DOM when the tab becomes inactive.
         // This is a problem when we need to keep track of the state inside a tab content, such as forms.
--- a/packages/admin-ui/src/Dialog/components/DialogBody.tsx
+++ b/packages/admin-ui/src/Dialog/components/DialogBody.tsx
@@ -14,7 +14,7 @@ const dialogBodyVariants = cva("wby-flex-1", {
         },
         bodyPadding: {
             true: "",
-            false: "!wby-px-none"
+            false: "!wby-px-none !wby-py-none"
         }
     },
     defaultVariants: {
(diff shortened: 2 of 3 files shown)


packages/admin-ui/src/Tabs/components/Content.tsx
```diff
@@ -4,7 +4,8 @@ import { cn, cva, type VariantProps } from "~/utils.js";
 
 const tabContentVariants = cva(
     [
-        "wby-bg-transparent",
+        "wby-bg-neutral-base",
+        "wby-my-lg wby-rounded-lg",
```
adrians5j, line 8: Not sure about `my-lg`. 

One thing: we now have more spacing here, in the dialog:

Image

This story:
```
http://localhost:6006/?path=/story/components-dialog--with-tabs
```

If I recall correctly, for tabs-in-dialog combination (which we discussed multiple times in the past), we decided to just add `bodyPadding: false`:
https://github.com/webiny/webiny-js/blob/8a0cadc5b69f045ce467316ac22a1223af5a49f9/packages/admin-ui/src/Dialog/components/DialogBody.tsx#L17

We can see it only removes x padding. Maybe it should also remove y padding as well?

IDK... 

What's also interesting is that... I was looking at this dialog here:

Image

I was wondering.. how the hell did Pavel manage to do it without bodyPadding prop.

Then I realized this impl of his is wrong, because, in dialogs, you need to have a separator like this:

Image

And if you were to use the prop (on the Tabs comp) that adds this separator, then it would not work, because the separator would not go all the way to the edges of the dialog. It would stop where x padding starts. 

What I'm trying to say is... with the change you did, and proper impl in this dialog, we'd have more y spacing then needed. 

And maybe this is the answer:
> We can see it only removes x padding. Maybe it should also remove y padding as well?

...but yeah, I'm not sure, we should test it a bit. Maybe add it and do some visual testing?

P.S. I'm not asking you to go into WB and fix the dialog with bodyPadding: false. Leave it if you want, it's not high prioirity. Looks fine. I might even tackle it during the tw4 upgrade.

adrians5j (changes requested):
It's complicated 😬

adrians5j (approved):
Here we go!

PR #4754, opened 2025-10-23
Fix Accordion Component

ku5ic:
## Changes

This pull request enhances the Accordion component in the admin UI by introducing more flexible background-styling logic, refining the nested-accordion behavior, and cleaning up the trigger component for better render logic and maintainability.

### Accordion background configuration and nested logic

- Updated the internal `getBackgroundByDepth` logic to accept three parameters: `depth`, `background`, and a new `shouldAlternate` boolean. When `shouldAlternate` is true, nested accordions now alternate backgrounds depending on depth; when false, the explicitly provided background value is used without alternation.
- Added a helper `hasNestedAccordion` function to detect when children contain nested `AccordionItem` components, determining whether alternation should apply.
- Adjusted the `AccordionBase` component to adopt the new logic: it now reads `background` from props and computes `bg` via `getBackgroundByDepth(depth, background, shouldAlternate)`.
- Updated documentation (`Accordion.mdx`) and Storybook stories (`Accordion.stories.tsx`) to include the `background` prop control and show examples of both alternating and fixed-background behavior.

### Component refactor and visual improvements

- In `AccordionTrigger.tsx`, streamlined render logic: removed redundant branches, unified icon rendering, and added actions separator logic only when both `actions` and `interactive` are present—improving clarity and reducing dead code.
- In `AccordionItem.tsx`, refactored how props are destructured: instead of the previous `splitAccordionProps`, now using a `useMemo` block to extract `itemProps`, `triggerProps`, and `contentProps`, resulting in cleaner prop-distribution and easier maintenance.

## How Has This Been Tested?

(Still n/a – no explicit test cases were added in this PR.)

## Documentation

n/a


packages/admin-ui/src/Accordion/Accordion.mdx
```diff
@@ -19,7 +19,7 @@ import { ReactComponent as TrashIcon } from "@webiny/icons/delete.svg";
 
 const AccordionExample = () => {
     return (
-        <Accordion variant="underline">
+        <Accordion variant="underline" background="base">
```
adrians5j, line 22: Can we rather invert the logic here?

Instead of having the default bg being as 'alternating' (or whatever the most appropriate name would be here), can we have base  (white) as the default, and then have the user specify 'alternating' explicitly, if required for the use case they are buildig?

Because it's like..... in most cases, you don't need this alternating bg. So far we only have it for obj fields. The rest of usages throughout the Admin app are basic, with base bg.

Wdyt?

adrians5j (changes requested):
Did not have much time for this one today, but wdyt about the question I posted below?

packages/admin-ui/src/Accordion/Accordion.mdx
ku5ic: >can we have base (white) as the default, and then have the user specify 'alternating' explicitly, if required for the use case they are buildig?

But the base background is already default. What I did here is that I only reverted to how it was before.

ku5ic: >Because it's like..... in most cases, you don't need this alternating bg. So far we only have it for obj fields. The rest of usages throughout the Admin app are basic, with base bg.

>Wdyt?

Alternating background works independently from background prop. If background prop is given, alternating backgrounds are skipped.

packages/admin-ui/src/Accordion/Accordion.tsx
```diff
@@ -22,16 +28,30 @@ type AccordionProps = React.ComponentPropsWithoutRef<typeof AccordionRoot> &
...
+        return background as "base" | "light" | "transparent";
+    }
+
+    // For depth 0, return undefined to let defaultVariants apply
+    if (depth === 0) {
+        return undefined;
+    }
+
+    // For nested levels, alternate between light and base
+    return depth % 2 === 0 ? "light" : "base";
```
adrians5j: I started with suggesting the below refactor. But then, I noticed another thing. So, let's ignore the refactor for now.

My question here is.... 

Seems to me that, if `depth === 0`, then, the returning bg value will be `base`.

Then, for the next depth, which is 1, the resulting modulo result will be 1, in which case the `return depth % 2 === 0 ? "light" : "base";` statement again returns `base` as the bg value.

Which means, we would have `base` bg for to depths in a row.

Maybe I got something wrong, but, is this correct?

Tnx!

--
> [!NOTE]  
> We can discuss this nitpick refactor later. Let's focus on the above first.

What would you say about this refactor?

```ts
const getBackgroundByDepth = (
    depth: number,
    background: AccordionProps["background"]
): AccordionProps["background"] => {
    // If background is explicitly provided, use it at any depth
    if (background) {
        return background
    }

    // For nested levels, alternate between light and base
    return depth % 2 === 0 ? "light" : "base";
};
```


Basically, instead of copying background values, we reuse the `AccordionProps["background"]` type. Easier to maintain.
Second, instead of partially relying on the default bg set via cva above (the  // For depth 0, return undefined to let defaultVariants apply if statement), we contain the value to be fully determined by the function itself. This way, the logic is more localised.

It's a nitpick, not blocking anything. Just providing an alternative way to impl it.

packages/admin-ui/src/Accordion/components/AccordionTrigger.tsx
```diff
@@ -11,16 +12,18 @@ type AccordionTriggerProps = Pick<
     "title" | "description" | "icon" | "handle" | "interactive" | "actions" | "draggable"
 >;
 
-const OpenCloseIndicator = () => {
+const OpenCloseIndicator = ({ className }: { className?: string }) => {
```
adrians5j: Do we need the `div` wrapper here? I think not, but just dbl checking.

ku5ic:
>My question here is....
>Seems to me that, if depth === 0, then, the returning bg value will be base.
>Then, for the next depth, which is 1, the resulting modulo result will be 1, in which case the return depth % 2 === 0 ? "light" : "base"; statement again returns base as the bg value.
>Which means, we would have base bg for to depths in a row.
>Maybe I got something wrong, but, is this correct?

Even depths (0, 2, 4...): Returns `"light"`
Odd depths (1, 3, 5...): Returns `"base"`

- `getBackgroundByDepth(0, undefined)` → `"light"` (even depth)
- `getBackgroundByDepth(1, undefined)` → `"base"` (odd depth)
- `getBackgroundByDepth(2, undefined)` → `"light"` (even depth)
- `getBackgroundByDepth(5, "dark")` → `"dark"` (explicit override)

packages/admin-ui/src/Accordion/components/AccordionTrigger.tsx
ku5ic: Addressed with [509dede](https://github.com/webiny/webiny-js/pull/4754/commits/509deded8debb5884de9a38cecce05d6f3b93f87)

packages/admin-ui/src/Accordion/Accordion.tsx
ku5ic: Addressed with [fe2d5b6](https://github.com/webiny/webiny-js/pull/4754/commits/fe2d5b649b6d1bf2a2a86025cf60f690108c969b)

adrians5j (approved):

PR #4902, opened 2026-01-28
Add `render` Option To `defineExtensions` 

adrians5j:
A way to define compound extensions, check sshot.

image


adrians5j:
@copilot 

Please refactor packages/cognito/src/Cognito.tsx packages/okta/src/Okta.tsx and packages/auth0/src/Auth0.tsx, the same thing we did with OpenSearch.tsx.

PR #4935, opened 2026-02-12
fix(api-headless-cms): various

brunozoric:
## Changes
Add input params when creating private or public models. Models should not be created without some initial values, as they cannot function without those (modelId, name, etc...).
Add function to take first possible text field as titleFieldId, if none is defined.
A bit more informative UI error when model field renderer is not set.
Must strictly enable database logs for tasks - was strict disable before.

## How Has This Been Tested?
Jest and manually.


Pavel910 (approved):
Models make much more sense now. The only thing I'd change is the `enableDatabaseLogs`: since it's a property, and not a method, I'd change the name to simply `databaseLogs: boolean`.

PR #4970, opened 2026-03-03
set-project-id fix

adrians5j:
Two bugs in the project linking flow:

1. `SetProjectIdService` incorrectly matched `<Project.*>` elements nested inside conditional expressions when determining the JSX insertion point — now restricted to direct children of the outermost JSX fragment.

2. `LinkProjectCommand` was passing only `projectId` to `setProjectId`; it now passes the fully-qualified `{orgId}/{projectId}` composite ID.

```ts
// Before
await projectSdk.setProjectId(selectedProject.id, { force: true });

// After
await projectSdk.setProjectId(`${selectedOrg.id}/${selectedProject.id}`, { force: true });
```


adrians5j:
@copilot update PR body

PR #5013, opened 2026-03-16
fix(app-headless-cms-common): exclude live field from singleton entry queries

### LINKED ISSUE #5012: Singleton model Admin UI crashes with "Cannot query field live"
### Version

6.0.0

### Operating System

macOS Darwin 25.3.0

### Browser

Chrome (latest)

1. Create a singleton content model using `ModelFactory` with `.singleEntry()`:

```typescript
import { ModelFactory } from "webiny/api/cms/model";

class AdminSettingsModelImpl implements ModelFactory.Interface {
    async execute(builder: ModelFactory.Builder) {
        return [
            builder
                .public({
                    modelId: "adminSettings",
                    name: "Admin Settings",
                    group: "ungrouped"
                })
                .fields(fields => ({
                    moduleCore: fields.boolean().renderer("switch").label("Core")
                }))
                .layout([["moduleCore"]])
                .titleFieldId("moduleCore")
                .singleEntry()
                .singularApiName("AdminSettings")
                .pluralApiName("AdminSettings")
        ];
    }
}

export const AdminSettingsModel = ModelFactory.createImplementation({
    implementation: AdminSettingsModelImpl,
    dependencies: []
});
```

2. Register it in `webiny.config.tsx`:

```tsx
<Api.Extension src={"@/extensions/siteConfig/AdminSettingsModel.ts"} />
```

3. Deploy the API: `yarn webiny deploy api`
4. Open the Admin UI and navigate to the singleton model

Alternatively create a new model in the admin UI as a singleton with a couple of basic fields. Save and navigate to the new singleton document content entry.

The singleton entry editor should load and display the form fields.

### What do you see instead?

The Admin throws an unhandled error:

```
Uncaught (in promise) Error: GraphQL error: Cannot query field "live" on type "AdminSettings".
```

The error originates from `SingletonContentEntryContext.js` calling `getSingletonEntry`, which builds a GraphQL query that includes `live { version }` as a system field.

### Additional information

The root cause appears to be a mismatch between the API-side schema and the Admin client query generation:

- **API side**: Singleton models use `createSingularSDL.js` (not `createManageSDL.js`) to generate their GraphQL schema. The `createSingularSDL` type definition does **not** include the `live` field:

```javascript
// node_modules/@webiny/api-headless-cms/graphql/schema/createSingularSDL.js, line 47-53
type ${singularName} {
    id: ID!
    entryId: String!
    ${onByMetaGqlFields}
    values: ${singularName}Values
}
```

- **Client side**: `createEntrySystemFields` in `node_modules/@webiny/app-headless-cms-common/entries.graphql.js` conditionally excludes `wbyAco_location` and `meta` for singleton models (line 14-24), but `live { version }` (line 102-104) is **unconditionally** included outside the conditional block.

The schema selection happens in `node_modules/@webiny/api-headless-cms/graphql/schema/schemaPlugins.js` line 42:

```javascript
if (model.tags?.includes(CMS_MODEL_SINGLETON_TAG)) {
    const singularType = type === "manage" ? "manage" : "read";
    const plugin = createCmsGraphQLSchemaPlugin({
        typeDefs: createSingularSDL({ ... }),
        // ...
    });
}
```

### Possible solution

Either:

1. **Client fix** — in `app-headless-cms-common/entries.graphql.js`, move `live { version }` inside the `if (!isSingletonModel)` conditional block alongside `wbyAco_location` and `meta`, OR

2. **API fix** — in `api-headless-cms/graphql/schema/createSingularSDL.js`, add `live: CmsEntryLive` to the singleton type definition to match what the client expects.

Option 2 is probably more correct since `live` is a useful system field for singleton entries too.

brick-pixel:
## What

Moved `live { version }` from the unconditional system fields into the `!isSingletonModel` conditional block in `createEntrySystemFields`.

## Why

Singleton models use `createSingularSDL` to generate their GraphQL schema, which does **not** include the `live` field. However, the client-side query builder unconditionally included `live { version }` for all models, causing:

```
Uncaught (in promise) Error: GraphQL error: Cannot query field "live" on type "AdminSettings".
```

The `wbyAco_location` and `meta` fields were already correctly excluded for singleton models, but `live` was placed outside that conditional block.

## Changes

**File:** `packages/app-headless-cms-common/src/entries.graphql.ts`

- Moved `live { version }` inside the existing `if (!isSingletonModel)` block
- Removed the standalone `live { version }` from the unconditional return template

### Before

```typescript
if (!isSingletonModel) {
    optionalFields = \`
        wbyAco_location { folderId }
        meta { ... }
    \`;                          // live was NOT here
}

return \`
    ...
    ${optionalFields}
    live { version }              // ❌ queried for ALL models
\`;
```

### After

```typescript
if (!isSingletonModel) {
    optionalFields = \`
        wbyAco_location { folderId }
        meta { ... }
        live { version }          // ✅ only for non-singleton
    \`;
}

return \`
    ...
    ${optionalFields}            // ✅ singleton models skip live
\`;
```

## How verified

- Confirmed `createSingularSDL.ts` does not include `live` in the singleton type definition
- Confirmed `wbyAco_location` and `meta` already follow the same exclusion pattern
- Non-singleton models are unaffected since `optionalFields` includes `live` for them

Fixes #5012

### DIFF
--- a/packages/app-headless-cms-common/src/entries.graphql.ts
+++ b/packages/app-headless-cms-common/src/entries.graphql.ts
@@ -34,6 +34,9 @@ const createEntrySystemFields = (model: CmsModel) => {
             meta  {
                 ${CONTENT_META_FIELDS}
             }
+            live {
+                version
+            }
         `;
     }
 
@@ -113,9 +116,6 @@ const createEntrySystemFields = (model: CmsModel) => {
             displayName
         }
         ${optionalFields}
-        live {
-            version
-        }
     `;
 };
 


brunozoric:
hi @brick-pixel 

Issue was fixed, is in our next branch and is ready to be shipped with 6.1. Sorry for the inconvenience!

brick-pixel:
Thanks for the heads up @brunozoric! Closing this since it's already fixed on your end.

PR #5107, opened 2026-04-16
fix(skills): correct renderer names and input prop types in user skills

SvenAlHamad:
Two fixes surfaced in #dev-team on 2026-04-16 while building a real user project against v6.2:

- content-models: Field Types table listed renderer names that don't exist in the framework (`lexicalTextInput`, `fileInput`, `objectInput`, `boolean`), causing generated models to silently misbehave. Rebuilt the table from `DataFieldBuilder.ts` and split into singular/plural columns. Added a "List fields and renderer pluralization" subsection covering the `.list()` + plural-renderer rule. Added a dedicated Layout section with nested-layout examples for `object` and `dynamicZone` fields.
- website-builder: Input Types section documented the factory functions but not the TypeScript prop shape the component receives at runtime. Added a "TypeScript prop types for each input" table (verified against `Image.tsx` and `Lexical.tsx`) explicitly calling out that file/image inputs are objects with { id, name, src, width, height, mimeType, size }, not strings, plus a worked FeatureCard example.

Skill descriptions (frontmatter) updated so the skills trigger on the prompts that previously missed.

## Changes

Closes #ISSUE_NUMBER


skills/user-skills/content-models/SKILL.md
```diff
@@ -74,19 +78,156 @@ Register in `webiny.config.tsx`:
...
-| `fields.file()`     | File/image attachment         | `"fileInput"`                                |
-| `fields.ref()`      | Reference to another model    | `"refDialogSingle"`, `"refAdvancedMultiple"` |
-| `fields.object()`   | Nested object with sub-fields | `"objectInput"`                              |
+The renderer passed to `.renderer("...")` **must** come from this table. The name is a
+camelCase key that the framework maps internally to the real frontend renderer. Using a
+made-up name (e.g. `"fileInput"`, `"lexicalTextInput"`, `"objectInput"`) will silently
+misbehave in the Admin UI — the field will render with a fallback or not save values.
+
+The authoritative source for this list is
+`packages/api-headless-cms/src/features/modelBuilder/fields/DataFieldBuilder.ts` — if
```
Pavel910: SInce this skill will be used from a user project, this path will not exist. Change it to `@webiny/api-headless-cms/features/modelBuilder/fields/DataFieldBuilder.d.ts`

PR #5156, opened 2026-05-04
fix(tailwind): prevent @source directive leak into production CSS

### LINKED ISSUE #5155: Admin build leaks absolute filesystem path into production CSS
### Version

6.2.0

### Operating System

macOS 14.6.1

### Browser

Google Chrome (any modern browser; the issue is in the produced CSS, not the renderer).

1. Scaffold a new Webiny v6.2.0 project (or use any existing one).
2. Run `yarn webiny deploy --env dev`.
3. Open the deployed Admin app's network panel and download the CSS chunk that bundles pre-existing component CSS (typically named `<chunkId>.<hash>.css`, the one that includes Lexical editor styles).
4. Search inside that file for the string `@source`.

No absolute filesystem path from the build machine should appear in the production CSS bundle. The Tailwind v4 `@source` directive is a build-time hint that should be consumed by the Tailwind PostCSS plugin and stripped before output.

### What do you see instead?

The CSS chunk contains four occurrences of an absolute path pointing to the developer's home directory:

```css
@source "/Users/<username>/path/to/project/apps/webiny/extensions";
```

The directive appears prepended at the start of the file and inserted between rule blocks throughout. The affected chunk is the one that bundles pre-existing component CSS imports (e.g., from `@webiny/lexical-editor`). Other CSS files that go through Tailwind processing (like the chunk produced from `@webiny/admin-ui/theme.css`) are clean.

Anyone fetching the affected file from CloudFront sees the developer's home directory and project location, exposing internal machine details in publicly-accessible production assets.

### Additional information

The leak originates in `packages/build-tools/bundling/admin/createRsbuildConfig.js` (shipped as `node_modules/@webiny/build-tools/bundling/admin/createRsbuildConfig.js`).

Lines 121–126 — the PostCSS plugin:

```js
const createInjectTailwindSourcePlugin = sourcePath => ({
    postcssPlugin: "inject-tailwind-source",
    Once(root) {
        root.prepend(`@source "${sourcePath}";`);
    }
});
```

Lines 36–43 — how it is invoked:

```js
addPlugins([
    createInjectTailwindSourcePlugin(
        path.join(paths.projectRootFolder, "extensions")
    ),
    tailwindcss({
        base: getTailwindBasePath(paths.projectRootFolder)
    })
]);
```

The plugin prepends an `@source` directive to every CSS file processed by PostCSS. The `@tailwindcss/postcss` plugin running afterwards uses a bail-out check that only processes a file if it contains at least one of these at-rules: `@import`, `@reference`, `@theme`, `@variant`, `@config`, `@plugin`, `@apply`, or `@tailwind` ([source](https://github.com/tailwindlabs/tailwindcss/blob/HEAD/packages/@tailwindcss-postcss/src/index.ts)). For files matching that check (e.g., `@webiny/admin-ui/theme.css`, which contains `@import "tailwindcss"`), the Tailwind plugin consumes and strips the injected directive — those outputs are clean. For files that don't match (e.g., pre-bundled component CSS imported through JS, such as `@webiny/lexical-editor`'s styles, which contains plain CSS rules with no Tailwind at-rules), the Tailwind plugin bails out and the injected directive survives into the final CSS bundle.

The directive value is `path.join(paths.projectRootFolder, "extensions")`, where `paths.projectRootFolder = process.cwd()`. The resulting absolute path bakes the developer's home directory into production output.

### Possible solution

Two complementary changes:

1. **Strip any leftover `@source` from the output** by adding a small PostCSS plugin that runs after `@tailwindcss/postcss` and removes any remaining `@source` at-rule. Tailwind v4 already removes the directive from CSS files it processes, so this only affects files where the directive would otherwise leak into the final bundle (those that don't trigger Tailwind processing). The fix is independent of which at-rules trigger Tailwind processing, so it remains correct if `@tailwindcss/postcss`'s bail-out heuristic changes.

2. **Use a relative path** instead of an absolute one in the injected directive. Tailwind v4 resolves `@source` paths relative to the CSS file containing the directive, so computing a relative form (e.g., from the CSS file location to the project's `extensions` folder) would avoid exposing the developer machine's filesystem layout. Even if a single canonical relative path can't be computed for all input files, using `path.relative(...)` from the project root rather than an absolute path would still prevent leaking the user's home directory.

rickserraglia:
## Changes

The PostCSS plugin in `admin/createRsbuildConfig.js` prepended an `@source` directive — containing an absolute filesystem path — to every CSS file processed. Tailwind v4 removes that directive from CSS files it processes, but `@tailwindcss/postcss` only processes files containing one of its trigger at-rules; for other files (e.g., the pre-bundled component CSS chunk that includes the Lexical editor styles), the directive survived into the production CSS bundle, exposing the build machine's home directory in publicly-served assets.

This change adds a small PostCSS plugin that runs after `@tailwindcss/postcss` and removes any remaining `@source` at-rule from the output. The fix is independent of which at-rules trigger Tailwind processing — the leak cannot escape regardless of which files match the bail-out check.

Closes #5155

## How Has This Been Tested?

Applied the same patch in a v6.2.0 install and re-deployed the Admin app to verify the fix and check for regressions:

- Tailwind-processed bundle (`index.<hash>.css`): same hash as before the patch; utilities still present.
- Pre-bundled component CSS bundle (`234.<hash>.css`): `@source` occurrences went from 4 to 0; no absolute paths remain.

No automated tests added — `packages/build-tools/bundling/` does not currently have a test setup, and the closest precedent (#5041) modifying the same file was merged without one. Happy to add a test if there's a preferred location for it.

## Documentation

No documentation changes needed. The fix is internal to the build pipeline and does not change any public API or user-facing behavior.

### DIFF
--- a/packages/build-tools/bundling/admin/createRsbuildConfig.js
+++ b/packages/build-tools/bundling/admin/createRsbuildConfig.js
@@ -39,7 +39,8 @@ export const createRsbuildConfig = ({ cwd }) => {
                     ),
                     tailwindcss({
                         base: getTailwindBasePath(paths.projectRootFolder)
-                    })
+                    }),
+                    createStripTailwindSourceLeftoverPlugin()
                 ]);
             },
             rspack: {
@@ -122,6 +123,20 @@ const createInjectTailwindSourcePlugin = sourcePath => ({
     postcssPlugin: "inject-tailwind-source",
     Once(root) {
         root.prepend(`@source "${sourcePath}";`);
+    }
+});
+
+/*
+    Removes any leftover `@source` at-rule from the output. Tailwind v4 strips the
+    directive from files it processes, but `@tailwindcss/postcss` only processes
+    files containing one of its trigger at-rules; for other files (e.g., pre-bundled
+    component CSS imported through JS), the injected directive would otherwise
+    survive into the production bundle, exposing the absolute build-machine path.
+*/
+const createStripTailwindSourceLeftoverPlugin = () => ({
+    postcssPlugin: "strip-tailwind-source-leftover",
+    Once(root) {
+        root.walkAtRules("source", node => node.remove());
     }
 });
 


Pavel910:
Thank you @rickserraglia, good find! Thanks for the PR! 🙌

PR #5304, opened 2026-06-18
fix: detect remote Pulumi backend for configured production environments

adrians5j:
> ⚠️ **Stacked on #5302** (Node 24 templates). Base is `fix/gh-actions-templates-node-24` so the diff stays scoped. Once #5302 merges, this will be retargeted to `next`.

## Summary

Fixes two related issues that made deploying to a **custom production environment** (e.g. `staging` registered via `Infra.ProductionEnvironments`) fail with `Cannot deploy to production with local state files.` even when a remote Pulumi backend was correctly configured (state was visibly being written to S3).

### 1. Hard-coded production environments
`ValidateProductionPulumiState` hard-coded the production env list as `["prod", "production"]`, ignoring environments registered via `Infra.ProductionEnvironments`. It now reads the configured list through the existing `GetProductionEnvironments` feature (which already merges the defaults with the user-registered environments).

This is why `prod` triggered the check but `staging` silently skipped it — they behaved inconsistently.

### 2. Inconsistent Pulumi backend env var
`PulumiLoginService` reads `WEBINY_PULUMI_BACKEND` (what actually configures the backend and writes state to S3), but `IsRemotePulumiBackendService` (the validation gate) only checked `WEBINY_CLI_PULUMI_BACKEND` / `WEBINY_CLI_PULUMI_BACKEND_URL` / `PULUMI_LOGIN` — none of which the templates or docs ever set. So the gate always concluded "no remote backend" and threw.

`WEBINY_CLI_PULUMI_BACKEND` is now the **canonical** variable across both services, with `WEBINY_PULUMI_BACKEND` kept as a **legacy fallback** for backwards compatibility:
- `PulumiLoginService`: `WEBINY_CLI_PULUMI_BACKEND || WEBINY_PULUMI_BACKEND`
- `IsRemotePulumiBackendService`: now also detects `WEBINY_PULUMI_BACKEND`

The scaffolded GitHub Actions templates are updated to set the canonical `WEBINY_CLI_PULUMI_BACKEND` (secret names left unchanged, so users don't need to rename secrets).

## Context

Reported by a user (Webiny v6.4.2) whose `prod` push deploy failed while `staging` worked, despite both being declared production envs and a remote backend being configured. They traced both root causes in `node_modules`.

## Follow-up (not in this PR)
- The public docs (webiny.com/docs) still reference `WEBINY_PULUMI_BACKEND` and should be updated to the canonical name.
- The error message links to `https://webiny.link/state-files-production`, which is currently dead.

## Test plan

- [x] `yarn build -p @webiny/project` passes (typecheck + build)
- [x] `oxfmt` + `oxlint --deny-warnings` clean on changed files
- [x] No tests construct `ValidateProductionPulumiStateImpl` directly; `getProductionEnvironments` is registered in `createProjectSdkContainer`


adrians5j:
Superseded by #5305, which targets `release/6.4.3` so this fix can ship in the upcoming 6.4.3 patch (rather than being stacked on the Node 24 PR against `next`).

PR #5349, opened 2026-07-01
Feat/wb ab testing

SvenAlHamad:
Summary

Adds end-to-end A/B testing to the Website Builder: editors define an experiment on a page, create full-page variants, set a traffic split, activate/deactivate, and edit each variant on the normal editor canvas. Variants are served server-side by the SDK with deterministic bucketing, a stable split, an instant kill-switch, and a provider seam for an exposure event (PostHog wiring is a follow-up). Built into core (not an extension).

---
1. Storage & serving mechanics

Data model — two private Headless CMS models (never in the page's revision timeline):
- wbyWbExperiment — belongs to a page (pageEntryId, baselineRevisionId), holds status (draft → running → stopped), name, trafficSplit ({ control, variants: { <variantEntryId>: weight } }), targeting, analytics, timestamps, winningVariantId.
- wbyWbVariant — belongs to an experiment; a full page-like content snapshot (properties/elements/bindings/metadata/extensions) copied from the baseline, with its own draft/ready status.

Approval / governance — variants never serve unpublished. Serving reads published experiment/variant content by entryId; a PageAfterPublish handler cascades a publish to the page's running experiment + ready variants, so the page's existing approval workflow is the single gate.

How a variant is chosen (per request, in the SDK) — getPageWithExperiment(path):
1. Fetch the control page + the active experiment for the path (cached per path). No running experiment → control.
2. Kill-switch: an uncached getExperimentPaused check — a paused experiment instantly reverts to control (content stays cached).
3. Resolve a visitor context (visitor id from cookie, country from CDN geo headers, device from UA).
4. Bucketing (assignVariant): targeting gate → inclusion gate (hash(visitorId:experimentId:inclusion) vs trafficPercentage) → weighted pick over control + ready variants using cyrb53(visitorId:experimentId).
5. Overlay the chosen variant's content onto the control; forced ?wb-variant=<entryId> bypasses bucketing for QA.

Persistence & stickiness — no assignment is stored. Only a wb_ab_vid cookie (random UUID, 1-year, no PII, set in middleware) is persisted; the same visitor id always hashes to the same bucket, so assignment is sticky and stateless.

Split correctness — cyrb53 is uniform across visitor ids, so weight configured split; inclusion and bucket selection use independenthash seeds.

Caching — reads cache and key only on path / variant id, never on visitor id, so the CDN cache stays shareable; only the kill-switch read is no-store.

Permissions — experiments/variants have no separate permission entities; every read/write/publish/delete checks the corresponding wb.page action. The read-only frontend key (wb.page:r / $wb.readonly) can serve and preview them
                                                                                                                                                                  ---
2. SDK changes (website-builder-sdk / -react / -nextjs)                                                                                                           
- ContentSdk exposes getPageExperiment, getVariantContent, getExperimentPaused, and a bound getPageWithExperiment(path, options) — bucketing, targeting, overlay, exposure, caching, and the kill-switch are all encapsulated. optiones the forced-variant id) or an explicit forcedVariantId, plusvisitor overrides.
- Kill-switch added: getExperimentPaused on IDataProvider/DefaultDaRIMENT_PAUSED query + a noStore path on ApiClient.query;getPageWithExperiment reverts to control for a non-running or paused experiment.
- @webiny/website-builder-react re-exports the experiment surface (eVisitorContext, assignVariant, analytics-provider registration,FORCED_VARIANT_PARAM, DEFAULT_VISITOR_COOKIE, and all experiment types).
- @webiny/website-builder-nextjs ships a middleware helper at @webidleware (createWebsiteBuilderMiddleware + default middleware/config)handling preview/draft mode and the wb_ab_vid cookie.

---
3. Implementing the SDK in a frontend (to test this PR)

▎ These packages are unpublished on the branch, so link/build them n link the four @webiny/website-builder-* packages, or point the appat the built dist). The demo app in this workspace is already refactored to the shape below.

middleware.ts
export { middleware, config } from "@webiny/website-builder-nextjs/

app/[[...slug]]/page.tsx — the whole A/B path is one call:
import { contentSdk } from "@webiny/website-builder-nextjs";
import { initializeContentSdk, getTenant } from "@/src/contentSdk";

const { isEnabled: isPreview } = await draftMode();
initializeContentSdk({ preview: isPreview, tenantId: await getTenant() });

const { page } = await contentSdk.getPageWithExperiment(path, { searchParams: search });
return <DocumentRenderer document={page} />;

Verify:
1. In Admin, open a page → Experiments → create one with two variants and a split → edit each variant on the canvas → Activate → publish the page.
2. Split/stickiness: load the page across several fresh sessions (cn tracks the weights; within one session it stays put.
3. Forced preview: ?wb-variant=<variantEntryId> renders that variant (uncached, QA only).
4. Kill-switch: pause the experiment → the page reverts to control

---
Known limitations / follow-ups

- Exposure event not emitted yet — the provider seam exists; wiring the PostHog exposure call is the next step.
- Draft-variant "preview in a new tab" stays app-side (reads an unph public serving deliberately doesn't expose).
- Variant add/remove is locked when editing an existing experiment; no analytics/results UI yet.
- Feature should be only available behind a feature flag, later this should be behind a WCP license


packages/api-website-builder/src/features/experiments/ExperimentPause/ExperimentPauseUseCases.ts
Pavel910, line 1: Split this file into separate files per use-case. 1 file per class.

packages/api-website-builder/src/features/experiments/GetActiveExperimentForPath/abstractions.ts
```diff
@@ -0,0 +1,68 @@
...
+    );
+
+export namespace GetActiveExperimentForPathRepository {
+    export type Interface = IGetActiveExperimentForPathRepository;
+}
+
+// Use case.
+
+export interface IGetActiveExperimentForPathUseCase {
+    execute(path: string): Promise<Result<ActiveExperimentForPath | null, UseCaseError>>;
```
Pavel910: Use cases should not return `null`. If a record is not found, it should return a `Result.fail(new PageNotFoundError())`. So, the general rule of thumbs is: use case returns a record or fails. Example: `packages/api-website-builder/src/features/pages/GetPageById/abstractions.ts:38`

packages/api-website-builder/src/features/experiments/GetActiveExperimentForPath/abstractions.ts
```diff
@@ -0,0 +1,68 @@
...
+    pageEntryId: string;
+    path: string;
+}
+
+// Repository: reads the published, running experiment for a page.
+
+export interface IGetActiveExperimentForPathRepository {
+    getPublishedRunningExperiment(
+        pageEntryId: string
+    ): Promise<Result<WbExperiment | null, ExperimentPersistenceError>>;
```
Pavel910: Same as use cases, either it finds a WbExperiment, or fails with `Result.fail()`. Example: `packages/api-website-builder/src/features/pages/GetPageById/abstractions.ts:14`

packages/api-website-builder/src/features/experiments/GetActiveExperimentForPath/abstractions.ts
```diff
@@ -0,0 +1,68 @@
...
+    experiment: WbExperiment;
+    revisionId: string;
+    pageEntryId: string;
+    path: string;
+}
+
+// Repository: reads the published, running experiment for a page.
+
+export interface IGetActiveExperimentForPathRepository {
+    getPublishedRunningExperiment(
```
Pavel910: The one and only public method should be called `execute`.

packages/api-website-builder/src/features/experiments/GetActiveExperimentForPath/GetActiveExperimentForPathUseCase.ts
```diff
@@ -0,0 +1,56 @@
...
+        }
+        const experiment = experimentResult.value;
+        if (!experiment) {
+            return Result.ok(null);
+        }
+
+        // Honour the instant kill-switch: a paused experiment serves the control.
+        const pausedResult = await this.isPaused.execute(experiment.entryId);
+        if (pausedResult.isOk() && pausedResult.value) {
+            return Result.ok(null);
```
Pavel910: Create a proper error class for paused experiments, and return as `Result.fail(new ExperimentIsPausedError())`, for example.

packages/api-website-builder/src/index.ts
```diff
@@ -102,6 +134,25 @@ const createContext = () => {
...
+            DeleteExperimentFeature.register(container);
+            GraduateVariantFeature.register(container);
+            GetActiveExperimentForPathFeature.register(container);
+            EndExperimentOnPublishFeature.register(container);
+            ExperimentPauseFeature.register(container);
+            CreateVariantFeature.register(container);
+            UpdateVariantFeature.register(container);
+            DeleteVariantFeature.register(container);
+            GetVariantByIdFeature.register(container);
+            ListVariantsFeature.register(container);
```
Pavel910: Instead of registering individual features, create a top level feature, like `ExperimentFeature`, and `VariantFeature` in their respective feature root folders (e.g., `features/experiments/feature.ts`). Then register only those 2 top level features here.

Pavel910 (changes requested):
I've reviewed the api-* part of the PR. In general the only important thing to change is that we never return `null` from use cases. It's a general rule - use cases either return a record, or throw an error. The errors are nicely typed, and can be easily used for flow control. Adding more edge cases is easy, you simply add more error types.

packages/api-website-builder/src/domain/experiment/experiment.model.ts
```diff
@@ -0,0 +1,43 @@
+import { ModelFactory } from "@webiny/api-headless-cms/features/modelBuilder/index.js";
+
+export const EXPERIMENT_MODEL_ID = process.env.WEBINY_API_LEGACY_MODELS
```
brunozoric: What is this? This is a new model, there is no legacy model.
Just use `wbyWbExperiment`.

SvenAlHamad:
(c/p from my Claude Code :) )
Thanks Pavel — all addressed in 8d3f9a85e3:
 - No null from use cases/repositories: GetActiveExperimentForPath and GetActiveExperimentForRevision now return a record or Result.fail(...). Added NoActiveExperimentError and ExperimentPausedError; the two "get active" repositories fail with NoActiveExperimentError instead of Result.ok(null).
 - execute naming: renamed the repository method to execute.
 - Paused error class: the use case now returns Result.fail(new ExperimentPausedError(...)).
 - Split file: ExperimentPause is now one file per use case (Pause/Resume/IsPaused).
 - Top-level features: added ExperimentFeature and VariantFeature; index.ts registers only those two.

 One thing to flag: the serving query getPageExperiment legitimately has a "no experiment on this page" outcome. To keep that from being an error the SDK logs on every plain page, the resolver maps NoActiveExperiment/Paused to a null data response (still typed failures inside the use case, just translated at the GraphQL boundary). Happy to adjust if you'd prefer a different convention there.

packages/api-website-builder/src/domain/experiment/experiment.model.ts
```diff
@@ -0,0 +1,43 @@
...
+    : "wbyWbExperiment";
+
+class ExperimentModelFactory implements ModelFactory.Interface {
+    async execute(builder: ModelFactory.Builder) {
+        const model = builder.private({
+            modelId: EXPERIMENT_MODEL_ID,
+            name: "Website Builder - Experiment"
+        });
+
+        model.fields(fields => ({
```
brunozoric, line 12: Do you expect any of the fields to grow large? Note that max size of ALL fields in the DynamoDB record is 400kb.
Use `.compress()` to automatically compress the field before storing. Note that its not searchable at that point.

Also, if any of the fields expect to contain some credentials of any kind, use `.encrypt()`.

packages/api-website-builder/src/domain/variant/variant.model.ts
```diff
@@ -0,0 +1,33 @@
+import { ModelFactory } from "@webiny/api-headless-cms/features/modelBuilder/index.js";
+
+export const VARIANT_MODEL_ID = process.env.WEBINY_API_LEGACY_MODELS ? "wbVariant" : "wbyWbVariant";
```
brunozoric: Same as in other model, no legacy one, just use `wbyWbVariant`.

packages/api-website-builder/src/features/experiments/CreateExperiment/abstractions.ts
```diff
@@ -0,0 +1,105 @@
+import { createAbstraction, type Result } from "@webiny/feature/api";
```
brunozoric: I always tell my claude to split abstractions into separate files. I dont allow it to have more than one abstraction per file. Convention is: feature/Something/abstractions/Something.ts

packages/api-website-builder/src/features/experiments/ExperimentPause/abstractions.ts
```diff
@@ -0,0 +1,51 @@
+import { createAbstraction, type Result } from "@webiny/feature/api";
```
brunozoric: Same abstractions.ts comment as before.

packages/api-website-builder/src/features/experiments/GetActiveExperimentForPath/abstractions/GetActiveExperimentForPathUseCase.ts
```diff
@@ -0,0 +1,81 @@
+import { createAbstraction, type Result } from "@webiny/feature/api";
```
brunozoric, line 1: Same abstractions.ts comment.

packages/api-website-builder/src/features/experiments/GetActiveExperimentForRevision/abstractions/GetActiveExperimentForRevisionUseCase.ts
```diff
@@ -0,0 +1,56 @@
+import { createAbstraction, type Result } from "@webiny/feature/api";
```
brunozoric, line 1: Same abstractions.ts comment.

packages/api-website-builder/src/features/experiments/GetExperimentById/abstractions/GetExperimentByIdUseCase.ts
```diff
@@ -0,0 +1,52 @@
+import { createAbstraction, type Result } from "@webiny/feature/api";
```
brunozoric, line 1: Same abstractions.ts comment.

packages/api-website-builder/src/features/experiments/GraduateVariant/abstractions/GraduateVariantUseCase.ts
```diff
@@ -0,0 +1,87 @@
+import { createAbstraction, type Result } from "@webiny/feature/api";
```
brunozoric, line 1: Same abstractions.ts comment.

packages/api-website-builder/src/features/experiments/ListExperiments/abstractions/ListExperimentsUseCase.ts
```diff
@@ -0,0 +1,54 @@
+import { createAbstraction, type Result } from "@webiny/feature/api";
```
brunozoric, line 1: Same abstractions.ts comment.

packages/app-website-builder/src/presentation/experiments/ExperimentsDrawer.tsx
```diff
@@ -0,0 +1,251 @@
...
+    onClose: () => void;
+    experiments: ExperimentDto[];
+    pageEntryId: string;
+    baselineRevisionId: string;
+    // When set, the drawer opens straight into the edit view for this experiment.
+    initialEdit?: ExperimentDto | null;
+    onChanged: () => void | Promise<void>;
+}
+
+const slugify = (value: string): string =>
```
brunozoric: There is a create a slug library in our repo so use that one (slugify or something).

packages/app-website-builder/src/presentation/experiments/ExperimentsEditorContext.tsx
```diff
@@ -0,0 +1,230 @@
...
+    return context;
+};
+
+interface ProviderProps {
+    // Revision id of the page being edited (e.g. "<entryId>#0001"); the baseline for its experiments.
+    pageRevisionId: string;
+    children: React.ReactNode;
+}
+
+export const ExperimentsEditorProvider = ({ pageRevisionId, children }: ProviderProps) => {
```
brunozoric: Tell your claude to rewrite this with presenter. Tell it to use codegraph to explore `packages/webhooks/src/admin` and it will figure it out

brunozoric (commented):
I wrote few comments, but these stand out:
* Most abstractions.ts have same problem - more than one abstraction defined.
* Too much useEffect in your components. Move to presenters.

SvenAlHamad:
Thanks @brunozoric — all addressed. Rundown:

Models & keys — 783ace2b97
- Dropped the WEBINY_API_LEGACY_MODELS conditional; models are declared directly as wbyWbExperiment / wbyWbVariant.
- Keys now go through the repo's slugify package rather than the hand-rolled helper.
- On .compress() / .encrypt(): left off deliberately — the stored fields are small config JSON (traffic split, targeting rules, analytics id + key), so there's nothing large enough to warrant compression and no secrets/credentials to encrypt. Happy to add them if you'd still prefer.

One abstraction per file — d547cbce4f
- Every experiment/variant abstractions.ts is split into abstractions/<Name>.ts (one interface/type per file), matching the api-opensearch layout you pointed to. Each feature's index.ts stays the public barrel.

Presenters (webhooks pattern) — ff6d7ed31f, refined in cf584183d0
- Rewrote the admin experiments UI to the webhooks/src/admin shape: MobX Presenter + ViewModel + DataSource + feature (resolve in singleton scope) + observer views. The React context and useEffect-based data loading are gone — a hub presenter plus a mobx reaction handle experiments/selection and the per-card variant summaries, with init() / dispose() on the view lifecycle.
- Child components take the presenter and drive it directly (no split callback props), and all display data (bucket colours, current bucket, switcher items) is derived in the ViewModel rather than in the components. The only remaining effects are editor selectors (isReadOnly) and the autosave subscription, mirroring the framework's own PageAutoSave.

One intentional deviation — presenter scoping. The ExperimentsManager (drawer) and ExperimentForm presenters use the standard child-container + DiContainerProvider scoping. The ExperimentsEditor hub presenter, though, is an app-wide singleton (registered in Extension.tsx). Unlike a webhooks page — where one WebhookListView hosts the whole feature under a single provider — the experiments UI mounts in disjoint page-editor slots with no common React ancestor: the switcher/indicator in a top-bar Ui.TopBar.Action, the toolbar in a Ui.Content.Element, and the surface swap in PageEditor. They all have to observe the same selection/paused state, so a shared singleton is the only way to bridge them. It's initialised per page via presenter.init(page.id), and only one editor is open at a time, so it's effectively page-scoped. Happy to wrap the whole editor in an editor-level DiContainerProvider instead if you'd rather keep it uniform.

packages/api-website-builder/src/domain/experiment/experiment.model.ts
SvenAlHamad: Good callout. Checked both models against the 400KB limit:

- Experiment model (this file): every field is small and bounded — text() scalars (ids/status/name) plus small json() config (traffic split, targeting, goals, analytics). Nothing here is expected to grow, so no .compress() needed. Also worth noting .compress()/.encrypt() are only exposed on the text field type, not json()/searchableJson(), so they wouldn't apply to the config fields anyway.
- Variant model is where large content actually lives (a full page snapshot). It intentionally mirrors the core page.model.ts field-for-field — elements/bindings as json(), properties/metadata/extensions as searchableJson() — so it inherits the exact same storage behaviour and size envelope as a page. Since core pages don't compress these, keeping the variant identical seemed safer than diverging.
- Credentials: the analytics field only carries a provider id + a client-exposed/publishable key (forwarded to the browser for exposure tracking), so there's nothing secret to .encrypt(). If we add a server-side secret later, it'll go in a dedicated text().encrypt() field.

PR #5420, opened 2026-07-16
fix(background-tasks-server): make background tasks run on the server flavour

adrians5j:
Every background task on the self-hosted (server) flavour stayed stuck `pending` — the worker never completed. Two root causes, both confirmed via runtime logging:

## 1. Worker asset path
`WorkerService` spawned the worker via `new Worker(new URL("../worker/workerEntry.js", import.meta.url))`. The app bundler rewrites that `new URL(...)` to a publicPath asset (`/static/assets/workerEntry.<hash>.js`), which `new Worker()` then can't load as a filesystem path → `Cannot find module '/static/assets/...'`.

`import.meta.url` resolves to this module inside its own **dist** (the bundler preserves it and doesn't inline the package), and the compiled worker + deps live beside it under `dist/worker/` (relative + node-builtin imports only). So resolve it with `path.join` off `import.meta.url` — an expression the bundler leaves untouched — pointing at the real dist worker.

## 2. Task-event validation (AWS leak in the shared runner)
Past the worker, the shared `TaskRunner`'s `TaskEventValidation` **requires AWS Step Functions fields** — `endpoint`, `executionName`, `stateMachineId`:
```
VALIDATION_FAILED_INVALID_FIELDS — endpoint/executionName/stateMachineId: expected string, received undefined
```
AWS's Lambda passes them from the SFN execution context; the server worker's event didn't. Supply them from the server transport: `executionName` = the task id (a stable run id the runner uses for log entries); `endpoint`/`stateMachineId` are validate-only off-AWS.

Also surfaces the worker's error (previously discarded) so a failed task logs instead of silently staying `pending`.

## Verified
End-to-end on the server flavour: image upload → worker spawns → POSTs `/background-task` → validation passes → task runs to completion, AI SDK invoked. AWS flavour unaffected (uses `StepFunctionService`, not this worker).

> Note: AI *image enrichment* separately fails in local dev because it sends the external AI a localhost image URL (`https://api2.localhost/...`) it can't fetch — a deployment/ai-powerups concern, not a bg-tasks issue.

### DIFF
--- a/packages/background-tasks-server/src/service/WorkerTaskService.ts
+++ b/packages/background-tasks-server/src/service/WorkerTaskService.ts
@@ -1,3 +1,5 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
 import { Worker } from "node:worker_threads";
 import type { WorkerToParentMessage } from "~/worker/TaskOrchestratorMessage.js";
 import { TaskService } from "@webiny/background-tasks/api/domain/TaskService.js";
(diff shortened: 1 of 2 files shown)


adrians5j:
Folded into #5414 (collapsed the server-flavour follow-ups into a single PR).

PR #5473, opened 2026-07-22
feat: CMS entry collaboration (comments) + notifications inbox

SvenAlHamad:
## CMS entry collaboration (comments) + notifications inbox

Two related features in one PR (two commits): threaded **comments** on Headless CMS entries/fields, and a per-user **notifications** inbox fed by comments + the Advanced Publishing Workflow.

### 1. Collaboration — comments (`feat(collaboration)` commit)
Threaded comments anchored to a CMS entry or a specific field: replies, resolve/reopen, soft-delete, edit, @mentions. Stored via a private CMS model (`wbyCollabThread`, messages nested). Admin UI: a Comments toggle in the entry editor header + an animated side panel with per-field comment markers.
- `@webiny/api-collaboration` (core) + `@webiny/app-collaboration` (admin).
- Field-scoped filtering (click a field's marker → panel filters to that field), "Copy link to thread", and URL-query-param deep-links (`commentThread` / `commentField`) that open the panel, highlight the thread, and scroll to the field (consumed once, then stripped from the URL — shareable).

### 2. Notifications — inbox (`feat(notifications)` commit)
A per-user notifications inbox surfaced in the admin.
- **`@webiny/api-notifications`** (core) — private CMS model `wbyNotification` (recipient-scoped). Repository wraps CMS access in `withoutAuthorization`; use cases enforce access via `recipientId === current identity`. Use cases: Create, List, Counts, MarkRead/MarkAllRead, Archive/Unarchive. GraphQL under the `notifications` namespace.
- **`@webiny/api-notifications-integrations`** (event-driven bridge) — subscribes to collaboration domain events (thread created → mention; reply added → reply + mention) and workflow events (approved / rejected). Single bridge for both sources.
- **`@webiny/app-notifications`** (admin UI) — bell registered into the top bar via a `UserMenu` decorator (no fixed positioning) with an unread badge, plus a slide-in inbox panel (Inbox/Archive tabs, unread-only filter, time grouping, typed items, mark-all-read). Items deep-link to the entry + comment thread.
- Notification titles reference the **entry title** (`CmsLocatorResolver` resolves it via `getEntryTitle`), not the anchored field.

**Wiring** — registers `createCollaboration()`, `createNotifications()` + `createNotificationsIntegrations()` in the API GraphQL templates (base + OpenSearch + sqlite overlays); mounts the admin apps in `app-serverless-cms`.

### Reviewer notes
- ⚠️ **`TODO(before release)`**: collaboration + notifications are gated to tenant-only for development — the WCP/APW capability gate is ungated (search `TODO(before release)`). Re-wrap before shipping.
- **Deploy**: API packages need a redeploy; the admin UI is client-side.
- **Deferred**: `reviewRequested` notification (needs workflow step team→member resolution — only approved/rejected wired so far); tasks/assignee inbox; no admin tests yet; notification rows store the title at creation time (existing rows keep old titles).


packages/api-collaboration/src/features/thread/ListThreads/ListThreadsUseCase.ts
```diff
@@ -0,0 +1,48 @@
...
+        if (listResult.isFail()) {
+            return Result.fail(listResult.error);
+        }
+
+        const { items, meta } = listResult.value;
+        const views: ICollabThreadView[] = [];
+
+        for (const thread of items) {
+            // Soft-deleted threads never surface in the list.
+            if (thread.deleted) {
```
brunozoric: This is not good, because it might output 0 items - but there are results and possibly there are more items to be loaded.
For example, you have 50 threads, and you want first 25 threads.
When you receive those first 25, code might filter all of them out and it might get buggy on auto load more when you reach bottom of the screen - there is no scroll to bottom of the screen as there are no items in the UI so you cant scroll.

Use filtering when querying.
On line 13, something like (not sure if its correct, figure it out):

````typescript
const listResult = await this.repository.execute({
  ...params,
  where: {
    ...params.where,
    deleted: false,
  }
});
````

File: packages/api-collaboration/src/graphql/collaboration.ts [hunk 2]
```diff
@@ -0,0 +1,371 @@
...
+    replyToCollabThreadValidation,
+    updateCollabMessageValidation
+} from "./validation.js";
+
+const toGqlThread = (view: ICollabThreadView) => {
+    return { ...view.thread, anchor: view.anchor };
+};
+
+export const createCollaborationSchema = () => {
+    return new GraphQLSchemaPlugin({
```
brunozoric: This is old way of creating GraphQL schema, check `packages/webhooks/src/api/graphql/WebhookDeliverySchema.ts`.
And even there, its wrong. It should be `WebhookDeliverySchemaImpl`, not `WebhookDeliverySchema_`.

packages/api-collaboration/src/utils/cmsContentId.ts
```diff
@@ -0,0 +1,30 @@
...
+    return `${modelId}:${entryId}`;
+};
+
+export interface ParsedCmsContentId {
+    modelId: string | null;
+    entryId: string | null;
+}
+
+export const parseCmsContentId = (contentId: string): ParsedCmsContentId => {
+    const separatorIndex = contentId.indexOf(":");
```
brunozoric: Overcomplication.

```
const [modelId, ...rest] = contentId.split(":");
// this is because entryId can have : in it (not at the moment, but safeguard )
const entryId = rest.filter(item => !!item).join(":");
if (!modelId || !entryId) {
    return { modelId: null, entryId: null };
}
return { modelId, entryId };
```

packages/api-collaboration/src/utils/identity.ts
```diff
@@ -0,0 +1,16 @@
+import type { ICollabIdentity } from "~/domain/thread/abstractions.js";
+
+/**
+ * Projects any Webiny identity (request identity or CMS identity) to the stored shape.
+ */
+export const toCollabIdentity = (identity: {
```
brunozoric: No inline types.

packages/api-collaboration/src/index.ts
```diff
@@ -0,0 +1,70 @@
...
+export const createCollaboration = () => {
+    const modelsPlugin = createRegisterExtensionPlugin(context => {
+        context.container.register(CollabThreadPrivateModel);
+    });
+
+    const collaborationContextPlugin = new ContextPlugin(async context => {
+        const tenantContext = context.container.resolve(TenantContext);
+        const identityContext = context.container.resolve(IdentityContext);
+
+        // TODO(before release): collaboration ships in the APW tier — re-gate on the
```
brunozoric, line 30: Note this down somewhere, we will 100% forget about this. A todo in the code does not mean it will automatically be picked up by someone.

packages/api-notifications/src/graphql/notifications.ts
```diff
@@ -0,0 +1,224 @@
...
+    MarkNotificationReadUseCase
+} from "~/features/MarkNotifications/index.js";
+import {
+    ArchiveNotificationUseCase,
+    UnarchiveNotificationUseCase
+} from "~/features/ArchiveNotifications/index.js";
+import { listNotificationsValidation, idValidation } from "./validation.js";
+
+export const createNotificationsSchema = () => {
+    return new GraphQLSchemaPlugin({
```
brunozoric: Same comment as for previous usage of `GraphQLSchemaPlugin`.

packages/app-admin/src/features/formModel/abstractions.ts
```diff
@@ -142,6 +142,8 @@ export interface IFieldValidation {
 
 export interface IFieldVM {
     name: string;
+    /** Full dotted path of the field within the form (e.g. "author.address.street"). */
```
brunozoric, line 145: What happens in case:
`author.phone.type` - where type is mobile/home,etc..
`author.phone.value`

and phone is actually an array, since author can have more than one phone?

packages/app-collaboration/src/cms/CommentFieldMarker.tsx
```diff
@@ -0,0 +1,48 @@
...
+import type { IFieldVM } from "@webiny/app-admin/features/formModel/abstractions.js";
+import { useCommentsPresenter } from "~/presentation/comments/useComments.js";
+import "~/presentation/comments/styles.js";
+
+/**
+ * Per-field affordance in the CMS entry form: a comment-count badge when the field has open
+ * threads, otherwise a hover-revealed "Comment" pill. Clicking opens the panel focused on this
+ * field. Renders nothing outside the entry editor (no contentId loaded).
+ */
+export const CommentFieldMarker = observer(({ field }: { field: IFieldVM }) => {
```
brunozoric: Avoid inline types.

packages/app-collaboration/src/presentation/comments/components/AutoTextarea.tsx
```diff
@@ -0,0 +1,56 @@
...
+    /** Height (px) at which the textarea stops growing and starts scrolling. */
+    maxHeight?: number;
+    onKeyDown?: (event: React.KeyboardEvent<HTMLTextAreaElement>) => void;
+}
+
+/**
+ * A controlled textarea that grows with its content (starting at one line) up to `maxHeight`,
+ * then scrolls. Keeps long comments and replies readable while typing.
+ */
+export const AutoTextarea = ({
```
brunozoric, line 19: Maybe this can be a part of our admin-ui package, with correct styles?

packages/app-collaboration/src/presentation/comments/components/CommentsPanel.tsx
```diff
@@ -0,0 +1,159 @@
...
+    }
+    return (
+        <div className="wby-collab-group">
+            <button className="wby-collab-section" onClick={() => setOpen(current => !current)}>
+                {open ? <ExpandMoreIcon /> : <ExpandLessIcon />}
+                <span className="wby-collab-section__label">
+                    {label} · {count}
+                </span>
+            </button>
+            {open
```
brunozoric: When I have a list of something, which I do not want to displayed in a certain case, I add a new component which represents the list and then return null if I dont want list shown. Something like:
```tsx
const ThreadCards = ({open, threads}: IThreadCardsProps) => {
    if(!open || threads.length === 0) {
        return null;
    }
    return <>{threads.map(thread => (
        <ThreadCard
            key={thread.id}
            presenter={presenter}
            thread={thread}
            onJumpToField={onJumpToField}
        />
    ))}</>
}
```

packages/app-collaboration/src/presentation/comments/components/Composer.tsx
```diff
@@ -0,0 +1,116 @@
...
+    const authorName = identity?.displayName || "You";
+    const [body, setBody] = useState("");
+    const [mentions, setMentions] = useState<string[]>([]);
+    const [busy, setBusy] = useState(false);
+
+    const addMention = (userId: string) => {
+        setMentions(current => (current.includes(userId) ? current : [...current, userId]));
+    };
+
+    const submit = async () => {
```
brunozoric: I would say that whole submit logic can be on presenter? @Pavel910 
You would expose `presenter.submit()` and internally handle everything.
There would be a `presenter.vm.busy` for you to show loader or whatever you want.

packages/app-collaboration/src/presentation/comments/components/MentionTextarea.tsx
```diff
@@ -0,0 +1,213 @@
...
+    onMention,
+    users,
+    excludeUserId,
+    placeholder,
+    className,
+    autoFocus,
+    maxHeight = 220,
+    onKeyDown
+}: Props) => {
+    const ref = useRef<HTMLTextAreaElement>(null);
```
brunozoric, line 40: Too much logic in React...?

packages/app-collaboration/src/presentation/comments/components/ThreadCard.tsx
```diff
@@ -0,0 +1,427 @@
...
+
+const fieldLabel = (thread: CollabThread): string => {
+    const path = thread.anchor.path || [];
+    const label = thread.anchor.label || thread.locator;
+    return [...path, label].join(" › ");
+};
+
+// Highlights @mentions. Known display names (matched longest-first) are highlighted in full,
+// e.g. "@John Doe"; anything else falls back to the single `@token`.
+const renderBody = (body: string, names: string[]) => {
```
brunozoric, line 31: This looks like it can be simplified?

packages/app-collaboration/src/presentation/comments/components/ThreadCard.tsx
```diff
@@ -0,0 +1,427 @@
...
+        .filter(Boolean)
+        .join(" ");
+
+    const startEdit = () => {
+        setDraft(message.body);
+        setEditing(true);
+    };
+
+    const saveEdit = async () => {
+        if (!draft.trim() || busy) {
```
brunozoric, line 106: Same comment as in few files before: logic into presenter and just use method in react.

packages/app-collaboration/src/presentation/comments/CommentsPresenter.ts
```diff
@@ -0,0 +1,220 @@
...
+    private loading = false;
+    private error: string | null = null;
+    private isOpen = false;
+    private activeLocator: string | null = null;
+    private filterLocator: string | null = null;
+    private highlightThreadId: string | null = null;
+    private users: CollabUser[] = [];
+
+    constructor(private api: CollaborationApi.Interface) {
+        makeAutoObservable<CommentsPresenterImpl, "api">(this, { api: false });
```
brunozoric, line 42: Isn't `CommentsPresenterImpl` actually a `PresenterAbstraction.Interface`?

packages/app-notifications/src/presentation/notifications/components/NotificationItem.tsx
```diff
@@ -0,0 +1,158 @@
...
+/**
+ * Resolves a Headless CMS entry route from a notification link. Collaboration links encode
+ * `contentId` as "<modelId>:<entryId>". Returns null for links we can't navigate to yet
+ * (e.g. workflow links that carry only a bare targetId without a model).
+ */
+const cmsEntryTarget = (link?: NotificationLink | null) => {
+    if (!link || link.app !== "cms" || !link.contentId) {
+        return null;
+    }
+    const separatorIndex = link.contentId.indexOf(":");
```
brunozoric: I left a comment with code in the API side of the same function.

brunozoric (changes requested):
Generally API looks ok. I would say that there is a bit too much logic/methods in React side.

packages/app-collaboration/src/presentation/comments/components/AutoTextarea.tsx
adrians5j: 100%. We want to rely on admin-ui pkg fully. In case a feature is missing, we'll want to add it.

One of the examples to check is this component here:
Image

In this particular case, if possible, I'd rather use what we already have in DS and not invent new components.

SvenAlHamad:
@brunozoric - thanks for the review. Here's the summary of changes:

On "too much logic/methods on the React side" — agreed. I moved the submit orchestration into the presenter:
- Composer submit + its draft state now live on the presenter (submitComposer(), setComposerBody, addComposerMention, resetComposer); the component is now a thin view.
- Reply and message-edit submits go through presenter.submitReply() / presenter.submitMessageEdit().
- I kept the transient input/caret state (and the @mention detection inside the textarea) local, since that's genuinely view/DOM concern — happy to revisit if you'd draw the line differently.

GraphQL (collaboration.ts, notifications.ts) — migrated both off GraphQLSchemaPlugin to the GraphQLSchemaFactory pattern per the WebhookDeliverySchema reference: use cases are DI-injected via dependencies, args are named interfaces, registration is now context.container.register(...). SDL is unchanged.

admin-ui textareas (AutoTextarea, MentionTextarea) — both now render @webiny/admin-ui's TextareaPrimitive instead of a raw <textarea> (cc @adrians5j — this addresses your "rely on the DS" point too). Mention autocomplete + auto-grow are layered on top of the primitive rather than being a bespoke input.

ListThreadsUseCase (0 items but more to load) — good catch, real bug. Soft-deleted threads were filtered in JS after the CMS limit, so meta and the page could disagree. Now deleted: false is pushed into the CMS where, so the limit applies to already-filtered results.

author.phone.type where phone is an array — you're right, and it turned out to be the deepest one. Resolution works (numeric segments are tolerated), but anchoring doesn't: a field's qualifiedName has no array index, so phone[0].type, phone[1].type, … all share one locator and can't be uniquely targeted. Since a proper fix needs a stable per-item id (index-based anchoring detaches on reorder), and that's a core form-model concern, I've done a scoped guard for now — comment markers are hidden on fields nested inside array/list fields so no ambiguous comment can be created — and we'll wire up real per-element anchoring once the core team lands a stable item id. Everything else (top-level fields, single/non-repeatable nested objects, entry-level) is unaffected.

Smaller items:
- cmsContentId — simplified to split(":"); aligned the client-side parse in NotificationItem too.
- Inline types removed — identity.ts → WebinyIdentityLike, CommentFieldMarker → CommentFieldMarkerProps.
- CommentsPanel — extracted a ThreadList component that returns null when empty; the sections use it.
- The TODO(before release) (WCP re-gate) — agreed a code TODO isn't enough; filing a tracked issue and linking it here.

CommentsPresenterImpl / PresenterAbstraction.Interface — it already implements PresenterAbstraction.Interface; the one spot referencing the concrete type is makeAutoObservable<CommentsPresenterImpl, "api">(...), where the generic has to be the concrete class so mobx can exclude the private api field. Let me know if you meant somewhere else.

ThreadCard renderBody — left the @mention highlighter as-is for now; if you had a specific simplification in mind I'm happy to take it.

SvenAlHamad:
Issue regarding the WCP gating created: https://github.com/webiny/webiny-js/issues/5491

packages/api-collaboration/src/graphql/collaboration.ts
```diff
@@ -0,0 +1,480 @@
...
+    messageId: string;
+    body: string;
+}
+
+interface IDeleteCollabMessageArgs {
+    threadId: string;
+    messageId: string;
+}
+
+class CollaborationSchema_ implements GraphQLSchemaFactory.Interface {
```
brunozoric, line 78: Claude likes to invent new ways to go round its constraints.
It should be `CollaborationSchemaImpl`, not `CollaborationSchema_`.

Not a big issue, but we have convention for this. Who knows why Claude is not following it - and there are a lot of examples.

packages/api-notifications/src/graphql/notifications.ts
```diff
@@ -0,0 +1,282 @@
...
+interface IListNotificationsArgs {
+    where?: {
+        archived?: boolean;
+        read?: boolean;
+    };
+    limit?: number;
+    after?: string;
+}
+
+class NotificationsSchema_ implements GraphQLSchemaFactory.Interface {
```
brunozoric, line 29: Same `_` as in other GraphQLSchemaFactory implementation.

packages/api-notifications/src/index.ts
```diff
@@ -0,0 +1,54 @@
...
+import { MarkNotificationsFeature } from "./features/MarkNotifications/feature.js";
+import { ArchiveNotificationsFeature } from "./features/ArchiveNotifications/feature.js";
+import { NotificationsSchema } from "./graphql/notifications.js";
+
+export const createNotifications = () => {
+    const modelsPlugin = createRegisterExtensionPlugin(context => {
+        context.container.register(NotificationPrivateModel);
+    });
+
+    const notificationsContextPlugin = new ContextPlugin(async context => {
```
brunozoric, line 22: I think all of this can go into `createRegisterExtensionPlugin`. No need to have context and registerExtension plugins.

packages/app-collaboration/src/presentation/comments/components/MentionTextarea.tsx
brunozoric: Still too much logic in react.

packages/app-collaboration/src/presentation/comments/components/ThreadCard.tsx
brunozoric: Still not simplified. Resolve comment if you think it cant be simplified.

brunozoric (dismissed):

SvenAlHamad:
Superseded by #5619  — same feature, now with the stable-_id nested-field anchoring, the admin-ui migration, and all review feedback from here addressed.

PR #5506, opened 2026-07-28
fix: reload scheduled actions when scheduler overlay closes

adrians5j:
## Problem

Cancelling a scheduled publish/unpublish from the **Scheduler overlay** left a stale **Scheduled** tag in the entries-list *Live* column. Reported by Sven ([screenshot](https://webiny.slack.com/archives/D03UVFX8TEU/p1784806447027699)): a Draft entry scheduled to publish, then cancelled via the Scheduler overlay, still showed *Scheduled* in the parent list.

## Root cause

The Scheduler overlay opens on top of the content-entries list, so `ListView` stays mounted and the singleton `ScheduledActionsPresenter` is never re-initialised. Its only refetch trigger is the list-presenter decorator's reaction keyed on each row's `status`/`live` version — and cancelling a scheduled publish on a Draft changes neither, so no refetch happens and the cached (now-deleted) scheduled action lingers.

## Fix

`SchedulerMenuItem` passes `onClose` to `Scheduler`, reloading the current model's scheduled actions when the overlay closes. The *Live* column then reconciles against fresh data before the user sees the list again.

## Testing

- `yarn check -p @webiny/app-headless-cms-scheduler` ✓
- `yarn lint` / `yarn format` / `yarn adio` ✓

### DIFF
--- a/packages/app-headless-cms-scheduler/src/components/Browser/SchedulerMenuItem.tsx
+++ b/packages/app-headless-cms-scheduler/src/components/Browser/SchedulerMenuItem.tsx
@@ -3,11 +3,13 @@ import { useModel } from "@webiny/app-headless-cms/exports/admin/cms.js";
 import { usePermissions } from "~/hooks/usePermissions.js";
 import { Scheduler as BaseScheduler } from "@webiny/app-scheduler";
 import { SchedulerButton } from "./SchedulerButton.js";
+import { useScheduledActionsPresenter } from "~/hooks/useScheduledActionsPresenter.js";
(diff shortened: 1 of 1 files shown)


adrians5j:
Follow-up: this fix reloads via a full network refetch on overlay close. We'd rather reconcile through a **cache update** instead — the Scheduler overlay already knows which action was cancelled, so it should surgically drop that entry from the shared scheduled-actions store (optimistic cache write) rather than re-listing from the API. Tracking as a follow-up to avoid the extra round-trip.

PR #5553, opened 2026-08-04
fix(api-websockets-sql): store connection timestamps in text columns

adrians5j:
## Summary

On SQL-backed (self-hosted / server flavour) deployments, `SendToIdentity` silently matched **zero** connections, so server→client WebSocket pushes never arrived, even though the connection was registered under the correct identity. Symptom: file-manager AI image enrichment completes and updates the file, but no "Image enriched" notification appears in Admin.

## Root cause

The registry writes UTC ISO strings everywhere: `register` stores `event.connectedOn` plus `lastSeen: new Date().toISOString()`, `updateLastSeen` writes the same, and `ConnectionRegistry.Data.connectedOn` is declared `string`.

The schema, however, declared `connectedOn` and `lastSeen` as `datetime`. That leaves it to the driver to decide what comes back out, and they disagree: node-postgres and some sqlite clients return a `Date`, others a `"2026-08-04 17:02:06"` with no `T` or `Z`.

`ListConnectionsUseCase` then dropped stale connections by comparing `connectedOn` against an ISO cutoff string. A `Date` coerces to `"Wed Aug 04 2026 ..."`, which sorts below `"2026-..."`, and in the space form the space (`0x20`) sorts before `T` (`0x54`). Either way every live connection read as expired and was dropped. `SendToIdentity` sent to nobody, threw nothing, and the task still reported success.

The mismatch reached further than that one filter. `toData()` passed the raw driver value through, so it also fed `WebsocketsGraphQLFactory`'s `connectedOn: DateTime!` field, and a test asserting `expect.any(String)` failed on the PGlite lane.

## Fix

Declare both timestamp columns as text, so the value read back is the value written, on every driver.

This settles the registry on a single type rather than accepting either. A string is the right one, because the DynamoDB registry writes `Data` straight into the document client, where a `Date` lands as an empty object. Keeping `Date` out of the domain means the two registries agree and that failure mode is unreachable.

With the column fixed at the source, the ISO normalization helper an earlier revision of this PR needed is gone, and `ListConnectionsUseCase` is back to a plain string comparison. ISO-8601 UTC strings sort chronologically, so the recency filter and the stale sweep are unaffected.

Tables created before this change are converted by a new migration. Values written earlier keep whatever shape the driver rendered them as, so those rows fall outside the recency window and get swept on the next pass, while clients re-register on reconnect. No backfill needed.

The repeated `ensureTable`/`migrate` prelude is now a single `ensureSchema()`, since there are four steps and eight call sites.

## Verification

- `yarn test packages/api-websockets-sql` and `yarn test:pglite packages/api-websockets-sql`: 14/14 pass, including four new tests covering the column type, string reads, the legacy `datetime` conversion, and readability of rows written before the migration
- `yarn test packages/api-websockets` and `yarn test:pglite packages/api-websockets`: 25/25 pass (the PGlite lane had 1 failing before, "should properly list connections")
- Manually reproduced end to end on a self-hosted SQL project: upload image, AI enrichment task runs, notification renders in Admin

The migration is covered against sqlite. The Postgres conversion path is not exercised by a test, only the fresh-table path on the PGlite lane.

## Not a WS-wiring regression

The 6.5.0 merges did **not** revert any WebSocket source. The wiring (register, sql registry, `SendToIdentity`, transport, admin handler) is intact. This is a latent bug in a newer code path. AWS/DynamoDB never hit it, since ddb stores `connectedOn` as a plain string attribute that round-trips as ISO.

## Follow-up: recency filter keys off the wrong field

Out of scope here, worth its own PR.

`ListConnectionsUseCase` judges "recent" by **`connectedOn`** (first-connect time), while `listStale` correctly uses **`lastSeen`** (heartbeat). So an Admin session open longer than 3 hours, alive and heartbeating with a fresh `lastSeen`, has an old `connectedOn`, gets filtered out, and silently loses pushes. The heartbeat and `lastSeen` machinery exists for exactly this. The fix would be to filter on `lastSeen`, falling back to `connectedOn` when the connection has never been heard from.


adrians5j:
Closing in favour of #5681.

This PR moved the websockets timestamp columns to text and normalized per read site. Review feedback was to keep `datetime` columns instead, and #5681 does the conversion globally: `withKnexDefaults` installs a Knex `postProcessResponse` hook that turns any `Date` in a response into a UTC ISO string, so one mechanism covers every table and no migration is needed.

#5681 fixes the bug this PR was opened for, verified rather than assumed. That branch carries no other change (`WebsocketsConnectionRegistry.ts:222` still declares `datetime` columns, `ListConnectionsUseCase.ts:31` still does the raw lexicographic compare), and `should properly list connections` on the PGlite lane passes there purely because of the hook. That failure is what this PR was tracking.

Nothing here needs porting. The migration tests only mattered for the column change, and the regression guard already lives in `api-websockets/__tests__/context/websocketsContext.test.ts`.

The `lastSeen` follow-up noted in the description is still open and unrelated to either approach: `ListConnectionsUseCase` judges recency by `connectedOn` (first connect) while `listStale` uses `lastSeen` (heartbeat), so a session open longer than 3 hours silently stops receiving pushes even though it is alive.

PR #5577, opened 2026-08-18
fix(api-headless-cms-bulk-actions): stop stripping version from entry IDs in bulk actions

Pavel910:
## Summary
- Remove `parseIdentifier()` calls from all 6 bulk action `processData()` methods that were stripping the `#version` suffix from entry IDs
- This fixes Publish and MoveToFolder bulk actions which silently failed because `getRevisionById` requires the version suffix
- Remove unused `@webiny/utils` dependency from the package
- Add regression tests covering ID passthrough for all 6 bulk actions

## Root Cause
Entry IDs stored in background task inputs are full revision IDs (e.g., `"abc123#0001"`). All 6 `processData()` methods called `parseIdentifier()` which stripped the `#0001` suffix, passing only `"abc123"` to the underlying use case.

For `PublishEntryUseCase` and `MoveEntryUseCase`, which call `getRevisionById`, the DDB dataloader silently skips IDs without a version (`if (version === null) { continue; }`). The entry is never fetched, the use case returns `Result.fail()`, and `ProcessTask` catches the error and marks the entry as "done" — the batch completes "successfully" with nothing actually processed.

The other 4 use cases (Delete, MoveToTrash, Restore, Unpublish) use storage methods that call `parseIdentifier` internally, so stripping was redundant but harmless.

## Test plan
- [x] New unit tests verify all 6 bulk actions pass full revision IDs to their use cases
- [x] All 9 tests pass (3 existing + 6 new)
- [x] Package builds successfully
- [x] Lint passes


packages/api-headless-cms-bulk-actions/src/features/DeleteEntriesBulkAction/DeleteEntriesBulkAction.ts
```diff
@@ -24,9 +23,7 @@ class DeleteEntriesBulkActionImpl implements EntriesBulkAction.Interface {
         model: EntriesBulkAction.Model,
         params: EntriesBulkAction.ProcessParams
     ): Promise<void> {
-        const { id: entryId } = parseIdentifier(params.id);
-
-        await this.deleteEntry.execute(model, entryId, {
+        await this.deleteEntry.execute(model, params.id, {
```
Pavel910: Shouldn't we strip the version here, and use plain entryId withou the version number here?

PR #5599, opened 2026-08-26
feat(cypress): authenticate against the self-hosted identity provider

adrians5j:
### What changed

The E2E suite could only authenticate through Cognito, so every spec calling `cy.login()` was unusable on the self-hosted ("server") hosting type — **15 of the 16 specs**.

Adds `authenticateWithSelfHosted`, the counterpart to `authenticateWithCognito`. Self-hosted exposes a deliberately unauthenticated login mutation:

```graphql
selfHostedAuthLogin(email: String!, password: String!)  # → { token, expiresIn }
```

It returns the same `{ idToken: { jwtToken } }` shape the Cognito helper does, so callers keep using `user.idToken.jwtToken` unchanged — `securityCreateApiKey`, `createGqlQuery` and friends needed no edits.

### The asymmetry worth knowing

The Cognito SDK writes its own `CognitoIdentityServiceProvider.*` keys as a side effect of authenticating, so the browser session becomes authenticated for free. This provider expects the **login screen** to have stored the token — so the helper seeds `webiny_self_hosted_auth_token` itself.

Without that, `cy.login()` would hand back a perfectly usable token for API calls while the Admin app still considered the session anonymous, and every spec that authenticates and then navigates would fail in a confusing way. `hasLoginDataInLocalStorage` now checks whichever key applies to the active provider.

### Choosing the provider

A new `HOSTING_TYPE` value in the Cypress config, written by `setup-cypress` (`"aws"` / `"server"`). Explicit rather than inferred from whether the Cognito values happen to be blank — the same reason the setup scripts were split in #5586.

### Spec coverage

The `/e2e` server job now runs the **account** spec in addition to the installation wizard. That spec does `beforeEach(() => cy.login())` and then navigates, so it exercises the new path end to end — programmatic auth *and* the seeded browser session.

It is a **separate step** from the wizard, not a widened glob: Cypress orders specs alphabetically and `admin/...` sorts before `adminInstallation/...`, so one glob would try to log in before the wizard had created the user.

**Deliberately not enabling the remaining specs yet.** They cover features whose behaviour on the server hosting type is simply unknown — file manager against local storage, page builder, and so on. Turning them all on at once would mix genuine product gaps in with anything wrong in this seam, which is exactly the debugging cost this series has been avoiding one step at a time.

Also note `admin/login/login.cy.js` will need work before it can run here: it drives the Cognito login UI (`/your e-mail/i`, `/sign in/i`), while the self-hosted screen uses `Email` / `Password` / `Submit`.

### Verification

- `tsc --noEmit` on `cypress-tests` passes
- `oxfmt --check` and `oxlint` clean
- Generated a server Cypress config and confirmed `HOSTING_TYPE: "server"` alongside blank Cognito values
- AWS path untouched: same helper, same config keys, `HOSTING_TYPE: "aws"` added

Not yet exercised in CI — like everything in this series, the first real signal comes from an `/e2e` run after merge.

### Changelog

**Self-hosted end-to-end authentication**

End-to-end tests can now sign in against the self-hosted identity provider, not just AWS Cognito, so self-hosted projects can be covered by the same test suite.

### Squash Merge Commit

```
feat(cypress): authenticate against the self-hosted IdP (#5599)
```

### DIFF
--- a/.github/workflows/pullRequestsCommandE2e.yml
+++ b/.github/workflows/pullRequestsCommandE2e.yml
@@ -825,6 +825,11 @@ jobs:
         run: >-
           yarn cy:run --browser chrome --spec
           "cypress/e2e/adminInstallation/**/*.cy.js"
+      - name: Cypress - run account test
+        working-directory: ${{ needs.baseBranch.outputs.base-branch }}
+        run: >-
+          yarn cy:run --browser chrome --spec
+          "cypress/e2e/admin/account/**/*.cy.js"
       - name: Update PR comment - Server (SQLite) passed
         if: success()
         env:
--- a/.github/workflows/wac/e2e/createServerJobs.ts
+++ b/.github/workflows/wac/e2e/createServerJobs.ts
@@ -221,12 +221,26 @@ export const createServerJobs = (storageOps: ServerStorageOps) => {
                     run: "cd cypress-tests && yarn cypress install"
                 },
                 {
-                    // Only the installation wizard for now. It is pure UI (no cy.login), so it does
-                    // not need the Cognito-based auth helper that the other specs use - porting that
-                    // to the self-hosted identity provider comes next.
+                    // Must run first: it completes the wizard, which is what creates the admin user
+                    // everything below authenticates as. Kept a separate step from the specs below
+                    // because Cypress orders specs alphabetically, and "admin/..." sorts before
+                    // "adminInstallation/..." - a single glob would run them in the wrong order.
                     name: "Cypress - run installation wizard test",
                     "working-directory": DIR_WEBINY_JS,
                     run: 'yarn cy:run --browser chrome --spec "cypress/e2e/adminInstallation/**/*.cy.js"'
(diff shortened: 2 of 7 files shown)


adrians5j:
Closing unmerged. The account spec this PR added was the only thing exercising the new auth helper, and we've decided the server variant should run exactly what the AWS variants run — the installation wizard smoke test, which authenticates through the UI and never calls `cy.login()`.

That leaves `authenticateWithSelfHosted` and the `HOSTING_TYPE` switch unexercised, so rather than land auth code nobody runs, it gets rebuilt when coverage actually expands — at which point it can be validated immediately. The GraphQL shape it targets is `selfHostedAuthLogin(email, password) -> { token, expiresIn }`, and the Admin reads the JWT from `localStorage` under `webiny_self_hosted_auth_token`.

The genuinely needed part of this line of work already merged separately as #5598 (lazy Cognito pool construction), which is what let the suite load at all on a setup without Cognito.

PR #5608, opened 2026-08-27
ci: make Full Release the headline "📦 Release", disable the old one

adrians5j:
### What changed

`🚀 Full Release` is the workflow you actually start a release with — it cuts the release branch and opens the PR — so it takes the plain name:

| File | Before | After |
|---|---|---|
| `fullRelease.yml` | `🚀 Full Release` | `📦 Release` |
| `release.yml` | `📦 Release` | **deleted** |

### Why the old one goes

It published to NPM by hand: `yarn release --type=beta --tag=<dist-tag>` followed by `--type=latest`, from a chosen branch.

The `/beta` slash command runs that same beta-then-latest pair, but versioned and driven from a release PR — so the manual, branch-driven variant is unused. Its `.yml` is deleted, so it no longer appears in the Actions list and cannot be dispatched.

### Disabled, not deleted

The source is kept as `release.wac.disabled.ts`. `ghawac` globs `**/*.wac.ts`, so that suffix keeps it out of the build — which now emits **14 files instead of 15** — while leaving the steps readable if it ever needs to come back:

```
rename to release.wac.ts  →  yarn ci-workflows:build
```

The frozen v5 workflows use the same trick with `.wac.frozen.ts`; the difference is that those still have their generated `.yml` checked in and this one does not. A header on the file records what it did, why it is off, and how to restore it.

### Verification

- Build output dropped 15 → 14, confirming the suffix excludes it.
- Nothing referenced the removed workflow: checked for `gh workflow run` callers, `workflow_run` triggers (none exist in the repo) and docs.
- Every remaining workflow name is unique.
- `oxfmt --check` and `oxlint` clean.

Run history for the renamed workflow is preserved — a workflow's identity in the Actions UI is its file path, not its `name`.

### Changelog

**Clearer release workflows**

The workflow that starts a release is now simply "Release". An unused manual NPM-publishing workflow has been removed from the Actions list.

### Squash Merge Commit

```
ci: make Full Release the headline "Release", disable the old one (#5608)
```


adrians5j:
Closing until the release queue is clear — reopen or re-cut when ready.

Three release PRs are open right now (#5571 6.4.9, #5567 6.4.8, #5520 6.6.0). The rename half of this is harmless at any time, but deleting the old `release.yml` removes the manual NPM publisher (`--type=beta --tag=<dist-tag>` then `--type=latest`, from any branch) while releases are in flight. `/beta` covers that flow normally, but if one of those releases needs a hand-run publish, that escape hatch would be gone exactly when it's wanted.

What this PR did, for whoever picks it up:

1. `fullRelease.yml` renamed to `release.yml` — the workflow that starts a release takes the plain name. Note the *file* rename orphans the existing "🚀 Full Release" run history, since a workflow's identity in the Actions UI is its file path.
2. Old `release.yml` (manual NPM publish) deleted, its source kept as `release.wac.disabled.ts`. `ghawac` globs `**/*.wac.ts`, so that suffix keeps it out of the build (14 files emitted instead of 15) while leaving it readable. Restore by renaming to `release.wac.ts` and rebuilding.

Also note: this branch was built on a stale `next` and predates #5605, so it needs rebuilding on the current tip rather than reusing as-is.

PR #5613, opened 2026-08-27
feat(feature): per-request setup/afterSetup phases on createFeature

adrians5j:
Implements the phased per-request lifecycle agreed with Pavel. Supersedes the `Module` prototype this PR originally carried.

## The change

**No new authoring API.** `createFeature` gains two optional callbacks — all ~1543 existing call sites are untouched.

```ts
export const AcoFeature = createFeature({
    name: "Aco",
    register(container) { /* phase 1 — DI wiring, unchanged */ },
    async setup(ctx) { /* phase 2 — per-request, tenant known */ },
    async afterSetup(ctx) { /* phase 3 — depends on OTHER features' setup */ }
});
```

| Phase | When | ≈ today |
|---|---|---|
| `register` | immediate, sync | `Feature.register()` |
| `setup` | per-request, post-tenant | `RequestContextInitializer` |
| `afterSetup` | after **all** `setup` | *missing today* — why bulk actions / background tasks break |

The runner enforces a **phase barrier**: every feature's `setup` completes before any feature's `afterSetup` starts. Cross-feature dependencies become safe without registration-order comments or `initialized` flags.

## Mechanism

- **`FeatureLifecycle`** token + `RequestContext` / `FeaturePhases` types in `@webiny/feature`. Internal plumbing — no feature author imports it.
- **`createFeature`'s generated `register()` stashes** declared phases under that token. Two properties fall out:
  - *Uniformity* — nested features get `register(container)` with the **same container** as their parent, so a feature at any depth works. A static walk over `registerApiRequestStack`'s list would silently no-op for the (majority) nested call sites.
  - *Gating for free* — a feature gated behind a license check never registers, so it never contributes phases. Today's early-return pattern keeps working; no `enabled` predicate needed.
- **`runFeaturePhases`** in `api-event-handler-core`, next to the feature list it serves. Takes `context` (tenant / flags / identity — injected by the caller so the base package stays free of `api-core` types) and `continueOnError` (HTTP fails fast; background tasks and scheduled actions do not, matching today's per-path behaviour).

## Why not `createModule`

Two front doors ("feature or module?"), a mixed registration list, and — since the honest answer would be "module, always" — a 1543-site rename. It only earns its keep if `Module` means more than phases (scope declaration, definition/dep-graph separation). Phases are the whole change.

## Scope

**Additive — nothing calls `runFeaturePhases` yet, so this is a runtime no-op.** Removes the `Module` / `runModules` prototype it supersedes.

Neither `@webiny/feature` nor `api-event-handler-core` had test infrastructure; this adds vitest config to both, plus 14 tests. `api-event-handler-core` also gains real `@webiny/di` + `@webiny/feature` deps (it previously had only a type-only `di` import).

## Verification

- `@webiny/feature` 7 ✓ · `api-event-handler-core` 7 ✓ · `event-handler-core` 44 ✓
- **`api-core` 171 ✓** — confirms the `createFeature` change is backward compatible
- builds ✓ · `yarn adio` ✓ · oxlint clean · formatted

## Next

1. Decide trigger placement, wire `runFeaturePhases`, drop the `HttpRouter` decorator + 4 manual calls → fixes the verified double-fire
2. Migrate the 4 `RequestContextInitializer` impls, one PR at a time
3. Hoist `withoutAuthorization` into the runner
4. Delete `RequestContextInitializer` and its machinery
5. Un-comment bulk actions into `afterSetup` — the case that motivated the third phase

Design note: `plans/tenant-lifecycle-module.md`.


adrians5j:
Closing — an audit showed this solves a misdiagnosed problem. Superseded by #5632 and `plans/retire-request-context-initializer.md`.

## What this PR assumed

That per-request setup needed a **phase system**: `setup` / `afterSetup` with a barrier, so work depending on other features' setup could run late enough.

## What the audit found

The problem was never scheduling. It was **an async value injected synchronously**:

```ts
export const FileModel = createAbstraction<CmsModel>("FileModel");   // never implemented
```

Nothing implements that abstraction — a hook deposits a value into it before anything can read it. DI resolves constructor arguments synchronously, fetching a model is async, so something had to run first. Every piece of machinery (5 firing sites, `initialized` flags, `getTenant()` guards, registration-order comments, the per-path error-mode split) exists to answer *"did the setup step happen yet?"*

Make the value pull-based and the question stops being askable. **Constructors cannot await; methods can.**

Specific findings that killed the design:

- **`afterSetup`'s motivating case is already fixed.** `HcmsBulkActionsFeature` registers everything at `register()` time and is live via `HcmsTasksFeature`. No current code needs a third phase.
- **The one documented ordering dependency is stale.** `HcmsTasksFeature` claims crud must precede gql because "its resolvers resolve `DeleteModelOperations`" — but they do so *inside resolver bodies*, at execution time, long after both initializers finish.
- **The pattern already exists in the codebase.** Extension models (`extensions/models/*`) are `ModelFactory` impls with an `async execute()`, consumed lazily via `ModelsProvider`. Built-in models were the odd ones out. `GraphQLContextualSchema` has the same shape too.
- **My contributor count here was wrong.** 9, not 12 — the earlier figure came from grepping by name pattern and conflating three separate abstractions.

## What ships instead

Nine slices converting nine abstractions to providers, then deleting `RequestContextInitializer`, its runner, the `HttpRouter` decorator and the 4 manual firing calls — which also fixes the verified server double-fire. No phases, no `Module`, no `createFeature` change.

The `Module` / `runFeaturePhases` code here is well-tested but solves the wrong layer, so it goes rather than sitting unused. The design note lives on in the plan file.

PR #5675, opened 2026-09-09
fix(website-builder-sdk): make breakpoint init callable from Server Components

adrians5j:
Fixes the RSC initialization failure reported against 6.4.10.

## Problem

`ContentSdk.ts` imports the viewport singleton directly:

```ts
import { viewportManager } from "./ViewportManager.js";
```

`ViewportManager.ts` is marked `"use client"`. That directive is a module boundary, so in the server graph the bundler replaces the module's exports with client references. When the SDK is initialized inside an App Router Server Component, `viewportManager` is therefore a reference placeholder rather than the singleton, and calling a method on it throws:

```
TypeError: viewportManager.setBreakpoints is not a function
```

It throws during `init()`, before any API request, so page loading and redirect resolution both fail from RSC.

Minimal trigger:

```tsx
import { ContentSdk } from "@webiny/website-builder-sdk";

export const dynamic = "force-dynamic";

export default function Page() {
  const sdk = new ContentSdk();
  sdk.init({ apiHost: "https://api.example.com", apiKey: "test-key", apiTenant: "root" });
  return <div>SDK initialized</div>;
}
```

This is a regression from #5628, which removed the `environment.isClient()` guard so that Grid could emit responsive CSS during SSR. The import itself predates that PR and was harmless, because importing a client reference only fails when you call a method on it, and the guard meant nothing ever did on the server. Removing the guard started calling it. Restoring the guard would fix the crash but lose the SSR CSS, so it is not the fix here.

Why it was missed: the change was never run through a production build of a starter app against the actually built packages. Verification went through hand-patched `node_modules`, which is where a build-time module-boundary error like this hides. One `next build` surfaces it immediately, as the check below shows.

## Fix

Breakpoints are plain configuration that the server needs too, so they move out of the client-only `ViewportManager` into a server-safe store. This follows the separation suggested in the report.

- New `BreakpointsStore` owns the breakpoint list and the desktop fallback, with no `"use client"` directive, so it is a shared module usable from either graph.
- `ViewportManager` keeps browser viewport observation and reads breakpoints from the store. `setBreakpoints` remains as a deprecated delegate so existing callers, including the reporter's compatibility guard, keep working.
- `ContentSdk` no longer imports `ViewportManager` at all, so `init()` runs in any environment.
- `createGridStackingCss` reads widths from the store, removing the client-only dependency from the SSR CSS path.

Note that the store is a module usable on both sides, not state shared across the boundary. Each graph gets its own instance, populated by its own `init()` call, and nothing is serialized between them.

This addresses all three requirements from the report: content and redirect loading work from RSC without touching client-only exports, breakpoint initialization still happens during client-component SSR, and browser and editor behavior are unchanged. The compatibility guard can be removed.

## Verification

Reproduced and confirmed against the reporter's scenario in a real Next.js app, using the built packages:

- Before: `next build` fails with `TypeError: dE.viewportManager.setBreakpoints is not a function`.
- After: build succeeds, the Server Component route renders, and `/grid-page` still emits `@media (max-width: ...)` in the raw HTTP response, so Grid stacking is intact.

Also added the unit tests that were missing on #5628:

- `BreakpointsStore.test.ts` covers the fallback, lookup by name, and that `ViewportManager` reports what the store holds, including through the deprecated setter.
- `gridStyles.test.ts` covers class scoping and the generated media query, including per-breakpoint widths and the reverse case.

## Follow-up, not in this PR

`next` has neither #5628 nor this fix, so the Grid CSS work and this repair both still need forward-porting or 6.5 loses them.

### DIFF
--- a/packages/website-builder-react/src/editorComponents/gridStyles.test.ts
+++ b/packages/website-builder-react/src/editorComponents/gridStyles.test.ts
@@ -1,0 +1,82 @@
+import { describe, it, expect, beforeEach } from "vitest";
+import { breakpointsStore } from "@webiny/website-builder-sdk";
+import { createGridClass, createGridStackingCss } from "./gridStyles.js";
+
+const makeBreakpoint = (name: string, maxWidth: number) => ({
+    name,
+    title: "",
+    description: "",
+    icon: "",
+    minWidth: 0,
+    maxWidth
+});
+
+describe("createGridClass", () => {
+    it("scopes the class to the element id", () => {
+        expect(createGridClass("51ymnzews569243uf6wy1")).toBe("wb-grid-51ymnzews569243uf6wy1");
+    });
+
+    it("replaces characters that are not valid in a class name", () => {
+        expect(createGridClass("abc#0001.x")).toBe("wb-grid-abc-0001-x");
+    });
+});
+
+describe("createGridStackingCss", () => {
+    beforeEach(() => {
+        breakpointsStore.setBreakpoints([
+            makeBreakpoint("desktop", 4000),
+            makeBreakpoint("tablet", 991),
+            makeBreakpoint("mobile", 430)
+        ]);
+    });
+
+    it("returns nothing when no stacking breakpoint is configured", () => {
+        expect(createGridStackingCss({ gridClass: "wb-grid-a" })).toBe("");
+    });
+
+    it("returns nothing when the breakpoint is not part of the theme", () => {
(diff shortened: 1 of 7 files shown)


adrians5j:
Closing in favour of reverting #5628 on this release branch. The CSS-first Grid work needs a design that is RSC-safe from the start, so it should not ride along in a patch. Branch `claude/fix-rsc-safe-breakpoints` stays available, the `BreakpointsStore` approach here is likely still the right shape when we revisit.

PR #5682, opened 2026-09-10
feat(activity-log): Headless CMS entry activity log

SvenAlHamad:
An append-only timeline of who changed a Headless CMS entry, when, and which fields — recorded per
save, never per revision, and never holding a content value.

Two new packages, `@webiny/api-activity-log` and `@webiny/app-activity-log`, plus a `description`
field on the permission renderer so a permission can state its own scope. Both are wired into the
default bundles — `registerApiRequestStack` on the API side, `app-serverless-cms`'s `Admin` on the
admin side — so the feature is on, subject to item 1 below.

## The four properties, and where each one lives

**Per save, not per revision.** Capture hangs off the CMS's own after-events — twelve of them, one
thin handler each — so a record is written every time a save lands, not once per version. A single
revision routinely holds many saves by several people over days, which is the whole reason the
admin timeline groups by revision rather than listing revisions.

**No content values, ever.** A changeset entry is `{ path, label, operation? }`. Editors compare
versions to see values; the timeline says only what moved. `hasNote` records that a reviewer
attached an approval or rejection comment without recording the comment, which is author-written
free text about the content and the field most likely to hold something sensitive.

**Deep field paths.** `author.name` for nested objects, `sections#a1b2c3d4e5f6` for an item in a
repeatable object, `[n]` for a positional one. The differ descends into objects, lists and dynamic
zones, keying on stable item ids where they exist and falling back to a three-pass match where they
do not. Beyond a cap the changeset rolls up to a common parent and sets `truncated`.

**Permanent retention.** Records are immutable and there is no TTL. `append`, `list` and
`deleteAllForTarget` are the whole storage surface; there is no update. The one deletion path is
cleanup after an entry is permanently destroyed, run as a background task so that recording and
cleanup fail independently.

## Shape

- **`core/`** — the record shape, the storage interface, the differ, path encoding. No CMS
  concepts. Hashing runs inside the differ and is never persisted.
- **`storage/privateModel/`** — the only implementation, behind `ActivityLogStorage`. A parameterised
  conformance suite written against the interface and nothing else proves a replacement satisfies
  the same contract; it runs green on `ddb` and `sql`.
- **`cms/recorder/`** — one `ActivityWriter` owning actor resolution, source labelling, correlation
  ids, persistence and failure containment. Both recorders sit on it.
- **`cms/capture/`, `cms/review/`** — the event handlers. Coverage guards assert the handler set
  against the CMS and APW sources, so a new after-event upstream fails a test rather than silently
  going unrecorded.
- **`features/listActivity/`** — the read path, with authorisation ordered so it cannot leak
  existence and the target's real model validated against the supplied `modelId` after resolution.
- **`app-activity-log/`** — the admin timeline, with every shaping decision in pure functions and
  hooks and the components reduced to presentation, for a design handover.

## Captured

Create, update, revision create, revision delete, revision describe, publish, unpublish, republish,
move, trash, restore, delete. Plus publishing workflow activity: review submitted, step started,
step approved, step rejected, step taken over, review cancelled, review approved, review rejected,
review deleted.

Actor comes from the ambient identity only. Every `*By` meta field on an entry is settable through
the manage API, so an entry's own account of who saved it is client-controlled and cannot underpin
an audit trail. `source` distinguishes a hand-clicked publish from a scheduled one, which identity
alone cannot: background tasks impersonate the user who started them.

## Read authorisation

Two global permissions. `activityLog.timeline` grants the timeline; `activityLog.actor` grants actor
identity, and without it actors are redacted rather than the records withheld. Both are additional
to entry read access, never implied by it. `plans/entry-activity-log-read-authorisation.md` carries
the full proposal and the reasoning behind the revision join.

## Tests

523 tests. 434 in `api-activity-log` — green on both `ddb` and `sql` — and 89 in `app-activity-log`.

The differ carries the bulk of it, because that is where the subtle failures live. Two things are
worth pointing at specifically:

- **The capture integration tests were verified by breaking the code they cover.** Removing the
  private-model filter makes twelve of the fourteen hang rather than fail, which is the recursion
  the filter prevents: a record is itself an entry in a private model, so recording one produces
  another. Removing failure containment from both the writer and the recorder fails exactly the two
  tests that claim a save survives a broken recorder, and nothing else.
- **The eight UI states are executable acceptance criteria**, asserting only that each renders and
  that one distinguishing statement reaches the reader — nothing about layout, which the design
  handover will replace. If a design pass silently drops a state, it fails here rather than leaving
  a gap nobody notices until a customer hits it.

One deliberate rule in there is easy to mistake for a bug and then "simplify" away: **consecutive
saves by two redacted actors are never collapsed into one row.** Merging them would assert the two
saves were by the same person, which is exactly what the redaction withholds.

---

# Accepted costs and open items

All of these are decisions rather than defects. They are in one place because a reviewer needs the
accepted costs together, not scattered through the diff.

### 1. The tier gate is not entitlement-driven, at two sites

The feature is switched on, and both switches are hard-coded:
`registerApiRequestStack` calls `ActivityLogAppFeature.register(container, { enabled: true })`, and
`app-serverless-cms` mounts `<ActivityLog />` with no flag check. Those two lines are the sites a
real entitlement replaces.

Neither reads a feature flag, and the reason changed under us mid-build. It used to be that a lookup
*could not* work: a name absent from `LICENSE_CHECKS` resolved to **true** for anyone holding any
licence, so `isEnabled("activityLog")` would have gated nothing while reading exactly like a gate
that works. #5695 removed that in `api-core` — an unrecognised name is now the project's own flag
and answers false unless the config sets it true — so a lookup on the API side would now be safe.

It was only fixed in one place. `packages/project` carries two more copies of the same decorator
(`GetFeatureFlagsWithLicense` and `GetProjectConfigService/LicenseDecoratedFeatureFlags`), and both
still return `!isExplicitlyDisabled(name)` for any name outside `LICENSE_CHECKS` whenever a licence
exists. On that path an undeclared flag still reads as enabled. Worth knowing for whoever closes
this gate: the runtime and the CLI would currently disagree about the same flag name.

It still would not be an entitlement. A project-configured flag makes the feature opt-in per
installation; it does not make it something Webiny can sell, because nothing on the licence governs
it. Reading one here would trade an explicit decision at the call site for an implicit "off" that
looks like a tier gate and is not one. So `enabled` stays a required parameter with no default: the
feature cannot switch itself on, and the missing entitlement stays visible.

Closing it needs `activityLog` in `IFeatureFlagsDto`, `KnownFeatureFlag` and `FeatureFlags.toDto()`;
`canUseActivityLog()` on `ILicense`, `License`, `NullLicense` and `ReactLicense`; an `ACTIVITY_LOG`
entry in `PROJECT_PACKAGE_FEATURE_NAME`, `WCP_FEATURE_LABEL` and `ProjectPackageFeatures`; and the
flag in all three `LICENSE_CHECKS` maps. Five packages. It also cannot evaluate true until WCP issues
`features.activityLog`, which is outside this repository. Never `canUseAuditLogs` — reusing that
entitlement would tie this feature to enterprise permanently, and business tier is intended.

### 2. Stable item ids are a declared dependency

Stable item `_id`s do not exist on `next`; they arrive with the forward-merge from
`release/6.5.0`.

- **Fully tested now:** id matching, id churn, reordering and the longest-order-preserving
  subsequence, as differ unit tests over synthetic values carrying `_id`. The differ is a pure
  function over two value trees, so none of it waits on the server minting ids.
- **Not integration testable until the forward-merge:** capture of a change to an item *inside a
  repeatable object* through a real entry write. Entries on `next` carry no ids for the differ to
  match on, so the integration suite asserts the id-less path — a nested object path
  (`author.name`) is covered end to end, an id-keyed item path is not.
- **Re-run once it lands:** the entry-write integration suite, plus one case that only becomes
  reachable then — an entry saved before the change and re-saved after it, exercising the
  absent-to-present transition against real data.

Absent ids are not a temporary condition to be worked around and deleted later. Every entry written
before the change carries no ids until its next save, so the install stays a mixture indefinitely.
**The absent-to-present transition must read as identity being established, never as every item
being replaced** — if it read as replacement, the first save of every pre-upgrade entry would emit
a junk record claiming the whole entry was rewritten. That case has its own named differ test.

### 3. The private model's read cost

The per-target query reads the entire model on every page view: measured at 100k items and 202 MB to
return 50 records, with page position making no difference, against a Lambda configured with 1 GB
and 30 seconds. Accepted knowingly. The narrow storage interface is the mitigation and a lighter
mechanism is the fix.

### 4. The OpenSearch write cost

On a DynamoDB-and-OpenSearch deployment every activity record additionally becomes a stream-table
write, a stream event, a Lambda invocation and an indexed document — permanently, for records never
queried through OpenSearch. Per-model indexing cannot be disabled: the `CmsModelOpenSearchIndex`
hook returns only settings and a sharing flag with no way to decline, and the tenant index factory
opts private models *in* by default. This sits beside the read cost rather than below it: unlike the
read cost it appears on a customer's bill every month rather than only under load.

### 5. The storage swap — what is *not* behind the abstraction

The interface makes the data path swappable and the conformance suite proves a replacement satisfies
it. These parts sit outside it and will need work at swap time, listed so whoever does it inherits
the list rather than rediscovering it.

- **Private model registration.** `ActivityRecordModel` is a CMS `ModelFactory` registered by
  `ActivityLogAppFeature`. A different store makes it dead weight, but removing it has to be
  sequenced against the records already in it — dropping the registration first orphans them where
  nothing can reach them, because the reader is already pointed elsewhere.
- **The adapter-private ordering key.** `sequence` exists because this backend needs a unique,
  lexicographically ordered sort key. A replacement will have its own notion of order and its own
  cursor format; neither is part of the contract and neither transfers.
- **Deployment and upgrade.** A dedicated table or index does not exist: it needs Pulumi
  infrastructure, and existing installations need it created on upgrade. There is no migration
  framework on `next`, so records already in the private model must either be moved by a one-off
  task or knowingly abandoned — and permanent retention means abandoning them discards the very
  history the feature exists to keep.
- **Chunk sizing in the purge task.** `MAX_DELETE_PASSES` is tuned to a backend whose every page
  costs a full model read. A store with real pagination wants a different bound, and the task's
  timeout budget was chosen against the current cost.
- **Read authorisation.** Not a storage concern in any implementation. Private models set
  `authorization: false`, so the current store enforces nothing, and the read path must not assume a
  future store will either.

### 6. The conformance suite is a requirement for any future adapter, not a nicety

`__tests__/conformance/storageConformance.ts` is written against the interface and knows no CMS
concept whatsoever. Two rules keep it honest as a portability check: never inspect a cursor, and
never assume a write is visible before `append` resolves. A second implementation must run this file
unchanged and green before it replaces the first — that is what makes the swap verified rather than
assumed. It already found two defects in the current adapter, including a sort key that was silently
non-monotonic.

### 7. The workflow premises in the brief were void, and were corrected

The brief described APW as it was at 5.44. `packages/api-apw` no longer exists and
`ApwContentReviewContent` has zero occurrences on this branch; APW is now `api-workflows` plus
`api-headless-cms-workflows` over a `wbyWorkflowState` private CMS model. A reviewer comparing code
to brief would otherwise conclude the implementation drifted — **the brief predated the rewrite.**
Four corrections, each verified in source:

1. **One hop, not two.** `targetId` and `targetRevisionId` sit directly on the workflow state.
2. **There are events for the review actions**, which the brief said there were not.
3. **Capture subscribes to the action events, not to `afterUpdate`** — the opposite of the brief's
   mechanism. `afterUpdate` is published only by `UpdateWorkflowStateUseCase`, and approve, reject,
   start and take-over all inject the repository directly and bypass it. Diffing `afterUpdate`
   would have missed every step transition, which is essentially the entire review.
4. **Change requests and sign-off are dropped, not approximated.** Both were `api-apw` concepts with
   nothing on this branch to map them onto.

### 8. Website Builder review targets are a product decision, not a defect

A workflow can be attached to a Website Builder page. Capture skips those targets and **logs that it
did**, rather than passing over them silently, so the gap reads as a decision. Recording them needs
a target type beyond `cms-entry`, which the record shape already carries from the start precisely so
that adding one needs no data migration. Whether to record them is a product call.

### 9. The frequency-and-timing caveat on changed paths

For a reader who can already read the entry, the changed-path list mostly discloses nothing new:
model structure is fully visible through GraphQL introspection, and version compare already shows
both paths *and* values between revisions. One thing is genuinely new and worth naming — the
timeline aggregates change **frequency and timing across revisions in a single view**, which no
existing surface offers. "Salary changed three times this week, twice by the same person" is not
derivable from the schema and is tedious to reconstruct from revision compare. That is a behavioural
signal about people, not about content.

Mitigated in the order that matters: `activityLog.timeline` is a separate grant, so the aggregation
is not handed to everyone who can read content; the changeset filter closes the specific case once
AACL implements field permissions; and "no values, ever" already means the timeline cannot say what
the value became. Not a reason to hold the feature — but the changed-path list should not be
described as "information already available" without this caveat.

### 10. `ForceDeleteDecorator` is a capture gap

When `force: true` is passed and the entry is already gone, the decorator calls storage operations
directly to clean up orphaned records, bypassing the use case and therefore the delete event. Those
deletions are not recorded. Deliberate: the path exists to clean up after a failed delete, and
recording a deletion whose entry no longer exists would put a record on a timeline nothing can
reach.

### 11. The churn rule's boundary arrives with the forward-merge

The churn rule recognises a new id whose content survived. When content changed too **and both sides
carry ids**, nothing is left to match on and the block reports as removed plus added. This is not
current behaviour: it takes effect only once stable ids exist on both sides, so it arrives with the
forward-merge. Until then the positional zip covers the same case, because at least one side always
lacks an id.

### 12. No persisted hash per changed field — a departure from the brief

The reasoning, so it can be revisited if revert detection is ever specified: nothing reads a stored
hash across records, because an empty changeset already signals a no-op save; a stored hash leaks a
low-cardinality field's new value, which the changeset entry leaks anyway for the lowest-cardinality
fields (recording that a boolean changed states its new value); and the only consumer that needs
persistence — detecting a revert to a previous value — is unspecified. Hashing itself is unaffected
and still runs inside the differ.

**This is a one-way door for records written in the meantime.** A hash cannot be computed
retroactively from a record that never stored one, so adding hashes later would give revert
detection only from that point onward. Accepted.

### 13. Two interpreters of model structure

`toFieldDescriptors` reads the model directly rather than through `ModelToAstConverter`, which needs
the GraphQL field-type registry that capture cannot rely on. The drift risk is covered by an
agreement test pinning both to the same field structure on a representative model, skipped with a
warning if the registry ever stops building standalone.

### 14. The cursor fix is a worked example of an already-filed defect

The offset cursor replaced in this adapter is the same defect class as the pre-existing entry-list
cursor issue, which is out of scope here — `encodeCursor(start + limit)` against
`slice(start, end)`, emitted even when the result set is exhausted. On an append-only log read newest-first it produces a
duplicate on nearly every paged read. The keyset approach here — a unique monotonic sort key plus
`_lt`, and a cursor returned only when another page exists — is a worked example for anyone taking
that on, on a dataset where the instability is unmissable rather than intermittent.

---

## Out of scope

Content values in any form. Model-level change tracking. Website Builder pages and templates, File
Manager versions. Comments. Entry views, locks, exports. A source registry or plugin system.
`api-headless-cms-pg-os` — not benchmarked, not supported, not considered.

Two pre-existing defects were surfaced and are **not** fixed here, per the brief: the entry-list
offset cursor described in item 14, and `EmptyTrashBinTaskDefinition` swallowing a persistently
failing delete so its cleanup loop re-reads the same page until the timeout check breaks it. The
second is why `deleteAllForTarget` reports progress rather than only success or failure — a caller
has to be able to tell finished from more-remain from broken.

**Deferred pending a separate decision:** capturing *scheduling* an entry and *cancelling* a
schedule. A scheduled publish executing is in scope and covered; the two scheduling actions
themselves are not, because neither reaches the target entry as an event.

## Summary by CodeRabbit

* **New Features**
  * Added an activity timeline to CMS entry editing, covering saves, revisions, publishing, reviews, deletions, restorations, and other entry activity.
  * Added revision and person filters, pagination, refresh handling, change summaries, actor attribution, and comparison guidance without exposing field values.
  * Added activity-log read permissions with optional actor visibility and anonymized results.
  * Added automatic cleanup of activity records when entries are permanently deleted.
  * Added descriptions to permission sections to clarify permission scope.


SvenAlHamad:
Superseded by #5743, which now targets `next` and carries the whole feature.

This PR's commits are all present there — `feat/entry-activity-log-ai-summaries` was stacked on this
branch, so its history contains this one in full. Its description has been folded into #5743 as well:
the four properties, the shape, what is captured, read authorisation, out-of-scope, and all fourteen
accepted-cost items, which are now items 1–14 of the combined list.

Closing rather than merging because reviewing half a feature against the other half is harder than
reviewing the whole, and nothing here had been reviewed by a person yet. The branch
`feat/entry-activity-log` is left in place.

PR #5691, opened 2026-09-11
feat(ai-powerups): resolve AI models per capability instead of taking the first provider

adrians5j:
The `next` counterpart of #5693, which is merged into `release/6.5.0`.

AI features each read `providers.presets[0]` and used whatever model happened to be first. This replaces that with three layers:

- **Connections** — a credential for a vendor, carrying no model. The old `providers` section bundled a key with one model, so using two models from the same vendor meant pasting the key twice.
- **Model roles** — `fast`, `standard`, `vision`. A closed set of three named slots a project fills once.
- **Capabilities** — one per AI feature, registered by the feature itself. Empty by default; a project can move a feature to a different role, pin an exact connection and model, or append instructions to its prompt.

`ResolveAiCapabilityUseCase` is the single place that knows the precedence: a pinned connection and model, else the chosen role, else the capability's declared default role, else `standard`. Six use cases previously answered "which model runs this feature" independently and only happened to agree.

Prompt customisation is append-only. Replacing a prompt outright was prototyped and removed: for most capabilities the prompt carries the output contract the surrounding code parses, so replacing it means replacing part of the implementation.

## Upgrade path

A project that already uses AI Power-Ups keeps working without opening the settings screen. `ModelRoles` seeds `standard` from `providers.presets[0]` and `Connections` derives a connection from the same section, carrying the encrypted key across. `fast` and `vision` stay empty and fall back to `standard`, so image work keeps running on the model it ran on before. `__tests__/features/upgradeFromLegacyProviders.test.ts` pins this, including the id agreement between the two handlers that nothing else enforces.

## Difference from #5693

Same work, ported onto `next` with two adaptations:

- `next` refactored image enrichment into `Prepare`/`Apply` use cases with a streaming route. Capability resolution lives in `PrepareImageEnrichmentUseCase` here rather than in the task.
- `handler-graphql` is `api-graphql` on `next`, so the `multiple: true` resolver fix (#5698 on 6.5.0) is included here as its own commit rather than arriving with the forward merge.

## Verified

Manually tested against a deployed API: settings screen renders, saves with untouched fields, API keys survive a save without being re-entered, model lists filter by the selected connection's vendor, and CMS entry generation resolves and reaches the provider end to end.

`yarn build` clean across the workspace, `adio` and lint clean, 57 tests in `ai-powerups` and 45 in `api-graphql`.

## Follow-up

#5720 covers two places that stay silent: image enrichment swallowing a config failure, and the role fallback not being surfaced anywhere.

Workspace: wby-next9 · AI Powerups settings

## Summary by CodeRabbit

* **New Features**
  * Added configurable AI connections, model roles, and per-capability overrides for models, connections, and instructions.
  * Added capability discovery in AI Power-Ups settings and through the GraphQL API.
  * AI-powered content generation, translation, revision comparison, and image enrichment now use capability-specific configuration and instructions.
* **Improvements**
  * Existing provider settings migrate to the new connection and role configuration.
  * Generation controls appear only when a usable standard model is configured.
  * Image enrichment provides clearer configuration errors and customized prompts.

### DIFF
--- a/packages/ai-powerups/__tests__/features/AiImageEnrichment/AiImageEnrichmentStreamRoute.test.ts
+++ b/packages/ai-powerups/__tests__/features/AiImageEnrichment/AiImageEnrichmentStreamRoute.test.ts
@@ -24,6 +24,7 @@ const prepared: IPreparedImageEnrichment = {
     imageBase64: "aGVsbG8=",
     imageMediaType: "image/png",
     model: "anthropic/claude-sonnet-4-5",
+    prompt: "Describe this image.",
     connection: { sdkName: "anthropic", apiKey: "key" }
 };
 
--- a/packages/ai-powerups/__tests__/features/Capabilities/CapabilitiesHandler.test.ts
+++ b/packages/ai-powerups/__tests__/features/Capabilities/CapabilitiesHandler.test.ts
@@ -1,39 +1,10 @@
 import { describe, it, expect } from "vitest";
 import { Container } from "@webiny/di";
 import CapabilitiesHandler from "~/api/features/Capabilities/CapabilitiesHandler.js";
-import { AiCapability } from "~/api/features/Capabilities/abstractions.js";
 import { AiPowerUpsSettingsGroupHandler } from "~/api/features/shared/index.js";
-
-const OUR_PROMPT = "Webiny's own comparison prompt.";
-
-/** Mirrors the real capability: fixed prompt, so the replace switch is offered. */
-class ComparisonCapability implements AiCapability.Interface {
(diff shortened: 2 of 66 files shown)


adrians5j:
Superseded for now by #5693, which targets `release/6.5.0` — that is where this ships first, and 6.5.0 merges forward into `next`.

Keeping this open as a draft rather than closing it: the 6.5.0 version differs (image enrichment is an older implementation there and keeps a soft skip, and the GraphQL fix lives in `handler-graphql`), so when 6.5.0 merges forward this branch is the reference for what the `next` side should look like.

PR #5693, opened 2026-09-11
feat(ai-powerups): resolve AI models per capability instead of taking the first provider

adrians5j:
Ships in 6.5.0. A `next` counterpart exists as a draft (#5691) and will be reconciled after this lands, since `release/6.5.0` merges forward into `next`.

> [!NOTE]
> **No longer depends on #5698.** `listCapabilities` used to be the one `addResolver` call using the `[abstraction, { multiple: true }]` dependency form, which `handler-graphql` ignores. It now goes through `ListAiCapabilitiesUseCase`, which takes that dependency via `createImplementation` where it has always worked. #5698 is still a real bug worth fixing, but the two PRs are independent and can merge in either order.

## What

Every AI feature read `providers.presets[0]`, with the same lines copy-pasted into five places: re-deriving the SDK name from the model string and decrypting the key each time. So "which model runs this feature" had five answers that only happened to agree. The settings screen was never finished; picking the first model was the placeholder that shipped.

This replaces it with three layers and one resolver.

**Connections** are a vendor and a credential, no model. The section they replace welded a key to a single model, so two Anthropic models meant pasting the same key twice.

**Model roles** are the indirection between a feature and a model: `fast`, `standard`, `vision`. A feature names a role, the project decides which model fills it. For most projects this is the whole AI configuration. `vision` exists because image enrichment needs a capability, not a price tier.

**Capabilities** are the per-feature exceptions, all empty by default. Features register their own next to their own code, so the settings screen has no list to maintain and an extension that adds an AI feature gets a settings row for free.

`ResolveAiCapabilityUseCase` is the only place that knows the precedence rules: a pinned connection and model, else the chosen role, else the declared default role, else `standard`. No environment-variable fallback; settings saying one thing while requests use another is the confusion this replaces.

## Prompts are append-only

There is no "replace the prompt" option, deliberately. For most capabilities the prompt carries the output contract the surrounding code parses. Revision comparison regex-matches the HTML table its prompt specifies, including the literal "No differences detected" sentence and a `<tr>` count. Page translation specifies a JSON shape and then parses it, handing back the untranslated page when that fails. Replacing either means replacing part of the implementation, with silence as the failure mode.

`additionalInstructions` appends instead, so the contract stays in the prompt. Projects needing different behaviour decorate the use case in code, which is how the AI translation itself is built.

## Upgrades

`connections` and `modelRoles` seed themselves from the legacy `providers` section on first read: each preset becomes a connection, and preset 0's model becomes `standard`. Since preset 0 is the only model the old code ever used, this reproduces current behaviour exactly. Keys are already encrypted, so it is a pure read. An unfilled role falls back to `standard` for the same reason.

`providers` still round-trips through storage so seeding keeps working. It goes away with the next breaking release.

## Incidental fixes

- `UpdateSettingsRepository` built its "existing" snapshot without the full settings blob. Since `connections` derives itself from `providers` until it has its own section, and that snapshot is what `mapToStorage` diffs against to recognise a masked key, the first save after an upgrade would have **silently destroyed the project's stored API key**.
- The AI buttons tested `presets.length > 0`, which passes for a row with no key in it, so the button rendered and the request then failed. They now check the whole chain the api walks.
- A skipped page translation was indistinguishable from a misconfigured one; it logs the reason now.

## Deliberate difference from the `next` version

**Image enrichment keeps its soft skip.** On `next` a missing model fails the enrichment task. That is safe there because the feature is licence-gated and unconfigured projects never register it. There is no such gate on this branch, so erroring would fail a task on every image upload for any project without AI configured. It logs the reason and returns `done`, which is this branch's current behaviour plus a reason in the log.

The enrichment prompt also moved from `AiImageEnrichmentTask` into the capability declaration, since that is where the resolver reads it from. The task's image-by-URL handling is unchanged.

## Testing

40 tests in `ai-powerups`, green. `adio`, `format` and `lint` clean.

Design doc, with screenshots of each tab and the open decisions: https://app.notion.com/p/3d5523905585811885aac3da4c3eee52

Note that the screenshots there were taken against the `next` build, so the Connections vendor list shows two vendors. This branch also has a `GoogleSdkFactory`, so it will show three.

### DIFF
--- a/packages/ai-powerups/__tests__/features/Capabilities/CapabilitiesHandler.test.ts
+++ b/packages/ai-powerups/__tests__/features/Capabilities/CapabilitiesHandler.test.ts
@@ -1,0 +1,118 @@
+import { describe, it, expect } from "vitest";
+import { Container } from "@webiny/di";
+import CapabilitiesHandler from "~/api/features/Capabilities/CapabilitiesHandler.js";
+import { AiPowerUpsSettingsGroupHandler } from "~/api/features/shared/index.js";
+
+function buildHandler(): AiPowerUpsSettingsGroupHandler.Interface {
+    const container = new Container();
+    container.register(CapabilitiesHandler);
+
+    return container.resolve(AiPowerUpsSettingsGroupHandler);
+}
+
+const handler = buildHandler();
+
+describe("CapabilitiesHandler", () => {
+    /*
+     * The admin form sends `null` for a field nobody has touched, and every field in an override is
+     * untouched by default. The first version of this schema required strings, so saving the
+     * settings screen without configuring a single capability failed with four copies of
+     * "Invalid input: expected string, received null" and no indication of which section was at
+     * fault.
+     */
+    it("accepts the nulls the form sends for untouched fields", () => {
+        const result = handler.inputSchema.safeParse({
+            overrides: {
+                "cms.generateEntry": {
+                    roleId: null,
+                    connectionId: null,
+                    model: null,
+                    additionalInstructions: null
+                }
+            }
+        });
+
+        expect(result.success).toBe(true);
+    });
+
+    it("accepts a fully configured override", () => {
+        const result = handler.inputSchema.safeParse({
+            overrides: {
+                "cms.generateEntry": {
+                    roleId: "fast",
+                    connectionId: "conn-1",
+                    model: "anthropic/claude-haiku-4-5",
+                    additionalInstructions: "Be brief."
+                }
+            }
+        });
+
+        expect(result.success).toBe(true);
+    });
+
+    it("still rejects a role that does not exist", () => {
+        const result = handler.inputSchema.safeParse({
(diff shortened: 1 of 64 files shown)


adrians5j:
Split the `handler-graphql` fix out into #5698, per review — it is a change to shared infrastructure rather than part of this feature.

**#5698 has to merge first.** `listCapabilities` here uses the `multiple: true` dependency form, so the Capabilities tab does not render without it. This branch no longer touches `packages/handler-graphql` at all.

adrians5j:
Reworked per review: `listCapabilities` now goes through `ListAiCapabilitiesUseCase` instead of mapping inline in the resolver.

Side effect — **this PR no longer depends on #5698.** That resolver was the only `addResolver` using the `multiple: true` dependency form; the use case takes it via `createImplementation`, where it always worked. The two PRs are now independent.

PR #5697, opened 2026-09-14
feat(project-server): serve self-hosted apps on a single URL

adrians5j:
### What changed

`webiny watch` and `webiny serve` now run your whole self-hosted project on one URL instead of one port per app. Admin is served at the root and anything under `/api` goes to the API. The port comes from `PORT` when something sets it, which is what lets a tool like portless put the project on a domain such as `https://myadmin.localhost`; otherwise it takes the first free port from 3001. The api and admin apps move to ports nobody has to type. `--no-proxy` runs them separately as before.

This also fixes a real bug. The project templates pointed admin at `http://localhost:3002` while the API auto-advances off a busy 3002, so a second project running at the same time left admin talking to the first project's API. Relatedly, the admin bundle now resolves a relative API URL against the page origin, so one build works on localhost, on a portless domain, and behind a reverse proxy without being rebuilt.

One caveat for projects that already exist: the File Manager's file delivery URL is stored in settings and does not follow the new address, so images can 404 until it is updated on the File Manager settings screen. Tracked in #5703.

### Changelog

**Run your whole self-hosted project on a single URL**

Starting a self-hosted project used to mean running the API and Admin apps in separate terminals on separate ports, and two projects running at once could end up talking to each other's API. Now one command runs both and serves them on a single address, which also works with tools that give local apps a proper domain name.

### Squash Merge Commit

```
feat(project-server): serve self-hosted apps on a single URL (#5697)
feat(cli-server): run api and admin on one URL in watch and serve (#5697)
```


adrians5j:
Backlogged the two open questions from review as #5703: the api's public origin being baked at build time, which covers the persisted `srcPrefix` on pre-existing projects, the likely mixed-content break on uploads under portless, and `serve` needing a build whose API URL already works behind the proxy.

Not fixed here on purpose. The obvious fix (a per-request origin fallback) was tried in this PR and removed as dead code, and making it live would let a client header override a pinned deployment origin.

adrians5j:
Both slop cop findings check out as intentional. Details, since it asked for confirmation:

**1. `references.json` churn** — legitimate, and smaller than it looks. It is a single-line minified generated file, so any change rewrites the whole line. The only semantic difference against `next` is:

```
devDependencies | vitest@4.1.11 | + cli-server, project-server
```

which is exactly right: this PR adds vitest as a devDependency to those two packages so they can have tests. `dependencies`, `peerDependencies` and `resolutions` are byte-identical to `next`.

It was regenerated by `webiny sync-dependencies` during the two merges with `next` (it conflicted both times), not hand-edited, and running the generator twice leaves it unchanged.

**2. `self-hosted-auth/Extension.tsx`** — the guess in the report is correct. It had its own second copy of the rules for turning `WEBINY_ADMIN_API_URL` into a GraphQL endpoint. That copy happened to still work with the relative URL the proxy bakes in, but only by accident, and would have drifted from the real resolver. It now imports `resolveGraphqlUrl` from app-admin, which the package already depended on. Net -15/+5, no behaviour change; verified login still posts to `/api/graphql` and lands on the dashboard.

Came out of review thread on `resolveApiUrl.ts`, where the question was whether that file should be an abstraction.

PR #5707, opened 2026-09-15
feat(app-headless-cms): let a field be marked read-only

adrians5j:
Adds a **Read-only** switch to the CMS field editor. A read-only field still shows its value in the entry form, but editors can't change it.

This came out of a client asking whether a field could be marked read-only. Turned out most of the machinery was already there.

## How it's stored

No new model property. A read-only field is an ordinary field rule: a condition rule with a new `always` operator and the existing `disable` action.

```json
{ "type": "condition", "target": "", "operator": "always", "value": null, "action": "disable" }
```

`CmsModelField.rules` already round-trips through GraphQL in both directions, is zod-validated, and is already mapped onto the entry form's field builder in `applyFieldProps.ts`. So the only genuinely new piece is the `always` operator in `ConditionRuleEvaluator`, plus the switch that writes the rule. Nothing had to change in the model schema, the GraphQL types, or the entry form.

The side benefit is that adding API-side enforcement later needs no migration: models already carry the rule, and the server would just start reading it.

The switch sits on the **General** tab next to "Use as a list" and "Use predefined values", since it describes what kind of field this is rather than how it reacts to other fields. That does mean two groups write condition rules on the same submit, General first and Rules after, so `RulesGroup.mapFromForm` has to leave the read-only rule alone when it rewrites the rest. Without that guard, editing any condition would quietly clear a field's read-only flag. There's a test pinning exactly that interaction.

## Scope: admin UI only

Like every other field rule, this is evaluated in the admin app and nowhere else. The API does not look at `rules` in the entry write path, so a read-only field can still be written to over GraphQL. That matches how the Rules tab's existing `Hide` / `Disable` actions already behave, so read-only isn't making a promise the neighbouring features don't.

If we want it to be an actual guarantee, that's a separate piece of work: evaluate rules in the entry write path (next to `mapAndCleanUpdatedInputData`, so create/update/singleton all get it from one place), share the operator evaluation with the frontend rather than writing it twice, and decide what a write to a read-only field does. Dropping unchanged values and erroring only on real changes is probably right, otherwise every honest read-modify-write round trip breaks.

## Screenshots

Field editor, General tab:

Field editor General tab with the Read-only switch on

The resulting entry form. `Title` is editable, `External ID` is disabled:

Entry form with a disabled read-only field

Verified end to end against a local model. Toggling the switch and saving persists the rule, and reading the model back over GraphQL returns it unchanged:

```json
{ "fieldId": "externalId",
  "rules": [{ "type": "condition", "target": "", "operator": "always", "value": null, "action": "disable" }] }
```

## Pre-existing bug found nearby, not fixed here

The condition rules UI offers operators that the evaluator doesn't implement. `getOperatorOptions` serves `==`, `!=`, `contains`, `startsWith`, `endsWith`, `>`, `<`, `>=`, `<=`, while `ConditionRuleEvaluator` handles `eq`, `neq`, `isEmpty`, `isNotEmpty`, `isTruthy`, `isFalsy`, `matches`. Nothing translates between them, so a rule authored through the model editor falls through to `default: return false` unless it happens to be `isEmpty` or `isNotEmpty`.

Read-only works because `always` goes straight into the evaluator, but the conditions on the Rules tab mostly don't fire. Left alone because the fix has a decision attached (widen the evaluator, or change the UI options) and doesn't belong in this PR. Happy to open an issue.

## Tests

- `Rules.test.ts`: `always` matches without resolving a target; a field carrying the rule is disabled and stays disabled regardless of form state.
- `GeneralGroup.test.ts`: reading and writing the flag, leaving other groups' rules alone, and surviving a submit that also rewrites the condition rules.
- `RulesGroup.test.ts`: the read-only rule stays out of the condition list, isn't cleared when conditions are edited, and isn't re-added if submitted as one.

## Summary by CodeRabbit

* **New Features**
  * Added support for unconditional rules that always apply without requiring a target field.
  * Added read-only field configuration in the admin interface.
  * Read-only fields remain visible but cannot be edited, including after unrelated form changes.
  * Read-only settings are preserved when editing field rules and do not create duplicate entries.
  * API writes remain available for read-only fields.
* **Bug Fixes**
  * Corrected unconditional rule evaluation so it no longer depends on resolving a target field.


adrians5j:
Superseded by #5719, which takes the `disabled()` approach we settled on: a first-class `disabled` property on the field rather than a rule with an `always` operator. Simpler storage, and it made the model-editor switch nearly free.

PR #5726, opened 2026-09-16
fix(admin-ui): dropdown scrolling inside dialogs

adrians5j:
## The reported bug

In the ACO "Manage permissions" dialog, grabbing the team list's scrollbar to drag it closed the list.

The autocomplete closes on its input's `blur`, and pressing the mouse on a scrollbar moves focus off the input, so the list vanished the moment the drag started. Options in the list already guard against this with `onMouseDown` + `preventDefault()`; the scrollbar had no such guard, and while it was the operating system's own scrollbar there was no element to hang one on.

The scrollbar now suppresses the default focus change on mouse down. Dragging the thumb scrolls the list and leaves it open.

## Two more that turned up on the way

**The mouse wheel did nothing over any popover inside a dialog.** A modal `Dialog` locks scrolling through `react-remove-scroll`, which listens on `document` and calls `preventDefault()` on every `wheel` and `touchmove` that lands outside the dialog content. Popover content is portaled to `document.body`, so it counts as outside. `PopoverContent` now attaches a callback ref from a new `useEscapeScrollLock` hook, which stops those events on the popover before they reach the `document` listener. The lock still holds everywhere else.

The callback ref matters: a `useRef` + `useEffect` version runs before Radix mounts the content node, so the listeners never attach.

**The lists scrolled natively**, so they showed the operating system's scrollbar, thin while scrolling on macOS and then a wide bar with a track under the pointer. Everything else in the admin uses the `ScrollArea` thumb. The autocomplete list and the multi select dropdown now scroll inside a `ScrollArea`, which hides the native scrollbar and draws ours. `ScrollArea` gained a `viewportClassName` prop for it, since the height limit belongs on the scrolling element when the area grows with its content rather than filling a parent of known height. The thumb sat flush against the popover edge, closer than the native one it replaced, so the scrollbar is now 10px wide with 2px of padding around a 6px thumb.

The multi select builds its popover on the Radix primitive instead of `PopoverPrimitive.Content`, so it needed the scroll-lock escape hatch as well.

`DropdownMenu` and `Select` were never affected: Radix menus are modal and push their own lock, which supersedes the dialog's. Still native: the icon picker grid, which scrolls through `react-virtualized` and owns its scroll container.

## Verification

Driven in Storybook with real CDP mouse input, in a modal dialog:

| | before | after |
|---|---|---|
| drag the thumb | list closes on the first move | stays open, scrolls 0 → 232 |
| wheel over the autocomplete | scrollTop 0 → 0 | 0 → 352 (list max) |
| wheel over the multi select | 0 → 0 | 0 → 360 |
| wheel over the autocomplete, no dialog | 0 → 200 | 0 → 200 |

Drag and wheel checked on all three: multi autocomplete, single autocomplete, multi select. The popover holds exactly one scroller, its viewport reports `scrollbar-width: none`, and the 8px-track thumb renders. Also checked: over-scrolling the list does not leak into the dialog body or the page behind it, the dialog body still scrolls with the wheel, arrow-key navigation still scrolls the highlighted option into view, Escape still closes the popover, and short lists and the empty state still size the popover to their content.

An `InsideDialog` story in `Popover.stories.tsx` reproduces the case.

Workspace: wby-next6 · 🐛 dropdown scroll inside dialogs


adrians5j:
Split into four smaller PRs, one story each:

- #5767 keep dropdown lists open while dragging the scrollbar (the reported bug)
- #5768 scroll popovers with the wheel inside dialogs
- #5769 one scrollbar look across dropdowns (stacked on #5767)
- #5770 pick the content model group from a list again

Merged together, the four reproduce this branch's changes byte for byte.

